PR #452, opened 2014-12-21
Surfboard: same transportability rules as Valk

### LINKED ISSUE #363: Surfboard transportability rules
Valkyrie is currently limited by mass (350) and footprint (4).
Surfboard currently has only a footprint limit (3).

Proposition is to make Surfboard use the same rules as Valkyrie. Affected units:
Now surfable: Firewalker, Banisher.
No longer surfable: Ultimatum, Jack, Comms.

sprunk:
Affected units:
Now surfable: Firewalker, Banisher.
No longer surfable: Ultimatum, Jack, Comms.
Fixes #363

### DIFF
--- a/units/armtboat.lua
+++ b/units/armtboat.lua
@@ -4,7 +4,6 @@ unitDef = {
   description         = [[Transport Platform]],
   acceleration        = 0.102,
   brakeRate           = 0.115,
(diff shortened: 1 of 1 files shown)


GoogleFrog:
It seems to be important that surfboard can transport commanders.

PR #474, opened 2015-01-01
Stardust: avoidFeature = false

sprunk:
Prevents getting blocked by a tree and such


GoogleFrog:
Why not just fix the maps? Won't this lead to Stardust shooting wasting time shooting into wrecks?

sprunk:
Fixing maps won't necessarily work because it also happens for wreckage which is generated naturally. Map features that come in abundance (and thus obstruct fire the most) are usually trees which tend to be brittle. Stardust has a lot of DPS so it will cut through the features quickly enough and likely hurt the target in the process due to spread and AoE.

PR #479, opened 2015-01-04
zerg's proven OD algo

sprunk:
Changes OD algo to the one zerg has proven (has ideal solution already implemented).
Extreme OD is nerfed as a result and low OD slightly buffed (equal at 7 energy).
od
od2

### DIFF
--- a/LuaRules/Gadgets/unit_mex_overdrive.lua
+++ b/LuaRules/Gadgets/unit_mex_overdrive.lua
@@ -253,7 +253,7 @@ end
 -- local functions
 
 local function energyToExtraM(energy)  
-	return -1+sqrt(1+(energy*0.25))
+	return sqrt(energy)*0.25
 end
 
 -------------------------------------------------------------------------------------


Licho1:
Wont it give too big boost at very low OD inputs? Im ok for merging this, waiting for others

sprunk:
Max difference in mex production is +12% at 1 E per mex.

http://www.wolframalpha.com/input/?i=max%28%281%2Bsqrt%28x%29%2F4%29+%2F+sqrt%281%2Bx%2F4%29%29

Licho1:
If I remember correctly, there are also other equations elsewhere, inverse function etc please seek them out.
They also can be replicated in other tools (energy bars)

KingRaptor:
I'm fine with this.

GoogleFrog:
Sprung your graphs are very wrong.Your implementation should have 100% OD income at 4 energy and 200% at 16 energy.

sprunk:
Because that gives a relative change in total production instead of an absolute one. For example 100->125 and 300->325 are both +25 but one is +25% and the other is +8%, I think totals matter more.

GoogleFrog:
I think an absolute comparison is more useful because it tells you the difference in metal income for the same amount of energy. A 12% difference is small if the incomes are small.

The largest amount of extra metal you can get from a single mex is 0.13*income at 1.33 energy.
http://www.wolframalpha.com/input/?i=max%28sqrt%28x%29%2F4+-+%28sqrt%281%2Bx%2F4%29+-+1%29%29+x+%3D+0+to+7

If you put 32 energy into a mex then the new system gives you 0.58*income less metal and the difference continues to increase.

PR #480, opened 2015-01-04
New model for Archangel

KingRaptor:
Uses a kitbash of Beherith's T1 Core bot models
Much prettier than hunchback!

http://i.imgur.com/gtzkUX3.png


Pxtl:
Looks great, but I worry it could be mistaken for the Archer.

sprunk:
This is an Archer who changed his gear from Shoulderpads of Dirtbaggery +2 to Cape of Outlawing +3 :P (pretty, but too lookalike imo)

GoogleFrog:
I had to look closely to see how this is not Archer.

Licho1:
Yeah looks like archer :(  Why not use current archangel?

KingRaptor:
Hmm, good point. Less Archery version: http://i.imgur.com/nLp31Hb.png

PR #482, opened 2015-01-05
Merge Dev into Master

GoogleFrog:


GoogleFrog:
I will have to do this locally.

PR #489, opened 2015-01-06
Context Menu handles the new Gauss regen type

sprunk:
It is now read from a unitdef customparam.


sprunk:
"I don't initially put things in custom params because they are very annoying for developing anything." - what does that mean? Do custom params hinder further development or is it just the added one-time cost?

KingRaptor:
I don't know if this was already communicated elsewhere, but anyway:
Customparams aren't subject to /luarules reload, which is annoying if they need to be changed during testing.

PR #490, opened 2015-01-06
Shockley: same movement as other tacnukes

sprunk:
Shockley has longer rise time than all other tacnukes (5s, other tacnukes have 3s) and turns somewhat slower. This makes tacnukes sequentially launched at the same area arrive in different order. The PR equalises these.


GoogleFrog:
Shockley intentionally takes longer to hit and is better at peeking over cliffs. Also you did not change the widget.

PR #492, opened 2015-01-08
Feature fx

Anarchid:
I got too fed up with potentially huge features doing nothing when crushed, destroyed, blown up or reclaimed. This tries to make me less annoyed at this by hiding the disappearance between a cloud of dust.

The cloud scales with destroyed feature's radius and is always 10 particles.

There is no sound included because googlefrog hates sounds.


Anarchid:
this is what it looks like

PR #506, opened 2015-01-15
Claw remodel

Anarchid:
This is artillery mine that looks better than Starlight. It is much less huge in flight than current Claw, which is compensated by unpacking like an alien egg, complete with visible submunitions.

It's still too thick to actually fit in Wolverine's barrel, but at least now it's possible to imagine a barrel that would fit. Maybe this should/will be followed by thickening the Wolverine barrel.

I think it would be nice to make it non-upright but i don't know why current Claw is a building, and non-upright necessitates non-building afaik.

(kudos to Sprung for s3o conversion and script)


Anarchid:
preview

Licho1:
looks good!

PR #507, opened 2015-01-16
Small screen handling for Commander selection menu...

xponen:
Commander selection menu is quite big, and designed for 1024x786 resolution according to KingRaptor. 

I added some code to scale down the menu for screen size less than 1024x768. 

It benefit low resolution screen and also user with big screen but running the game in Windowed mode. 

Side commit:
I also tweaked Epicmenu and Eco Panel 2 (default widget) to support at least 640x480.


Licho1:
I dislike the gigantic list at all..
I think that we should limit number of options to 3-4 by default and instead give people option to add "preset" commander to list with default being list with presets..

PR #512, opened 2015-01-18
Revert "Initial queue turns off for spectators before game starts..."

xponen:
because spec bug was already fixed in prior commit (commit 7ff7ad0364270f2e694b87ac595657eb95a1797a ).

This reverts (the logic change of) commit 641620ad6849306b85e95c3ac4937507641c1675


xponen:
ops wrong branch

PR #520, opened 2015-01-20
Windmill patch

sprunk:
* removes wind modoption (broken)
* logs teams' wind economy in a team rules param
* fixes widgens spinning animation speed


GoogleFrog:
Why private and not ally?

sprunk:
Private is actually ally, see eg. the RoI params:
https://github.com/ZeroK-RTS/Zero-K/blob/develop/LuaRules/Gadgets/unit_mex_overdrive.lua#L263

KingRaptor:
Merged

PR #522, opened 2015-01-21
Block "/take" of inactive-team or AI

### LINKED ISSUE #457: /take - permanent ownership for afk-units disables rejoining players
this overrides the 2nd com or factory plop assignment, which is done to distribute the com/fac of afk players.

the difference is, it is permanent and the units are not given back if the person is (re-)joining later. So, if you are late you cannot participate the game. 

it is scriptable and already abused with calling cyclic in short times. example at http://zero-k.info/Battles/Detail/316898.

xponen:
Why blocking "/take" is a good idea:
  - https://github.com/ZeroK-RTS/Zero-K/issues/457 reported that "/take" prevent lagmonitor from returning unit to its owner. (fix #457 )
  - also, "/take" prevent some game mode like "player as chicken" from working properly (since the original design didn't want player to have all chicken)

Changes:
  - added "game_disable_take.lua"
  - tweaked some other gadget to make sure they use "GG.allowTransfer" to differentiate user's "/take" from gadget's unit transfer.

### DIFF
--- a/LuaRules/Gadgets/game_disable_take.lua
+++ b/LuaRules/Gadgets/game_disable_take.lua
@@ -1,0 +1,40 @@
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+if not gadgetHandler:IsSyncedCode() then
+	return
+end
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


xponen:
ops wrong branch

PR #530, opened 2015-01-22
Jsonify and fix modstats.

Anarchid:
- removed old teams reporting (springie exists)
- converted all reporting to JSON
- converted all recording to unitnames instead of unitDefID's to avoid double conversion
- added detection of reused drones like Firefly and Viper which can exist on multiple carrier defs (TODO: use `is_drone` customparam everywhere)
- fix GameOver not firing by increasing layer to 1 (TODO: figure out why 0 breaks it)
- enact facplop stats gathering
- `SPRINGIE:stats` header changed to `SPRINGIE_STATS:` for no real reason


LuaRules/Gadgets/mod_stats.lua
```diff
@@ -204,63 +216,34 @@ end
...
-		unitCounts[unitDefID] = tab
+		unitCounts[unitDefName] = tab
 	end
 	tab[2] = tab[2] + 1
 end
 
-function SendData(statsData) 
-	Spring.SendCommands("wbynum 255 SPRINGIE:stats,".. statsData)
+function SendData(name, data) 
+	Spring.SendCommands("wbynum 255 SPRINGIE_STATS:"..tostring(name).."="..encode(data));
```
Licho1: This will cause springie to ignore data .. 
SPRINGIE: meant that the message is for springie (and not chat)
and STATS meant just stats .. now its merged into one single prefix, not sure if its good

GoogleFrog:
Does that mean I don't merge? Changing the format for springie seems weird.

Anarchid:
Undid the springie prefix change.

LuaRules/Gadgets/mod_stats.lua
```diff
@@ -204,63 +216,34 @@ end
...
-		unitCounts[unitDefID] = tab
+		unitCounts[unitDefName] = tab
 	end
 	tab[2] = tab[2] + 1
 end
 
-function SendData(statsData) 
-	Spring.SendCommands("wbynum 255 SPRINGIE:stats,".. statsData)
+function SendData(name, data) 
+	Spring.SendCommands("wbynum 255 SPRINGIE:stats:"..tostring(name).."="..encode(data));
```
Licho1: I thought SPRINGIE: is ok but meh i will modify it myself later

PR #531, opened 2015-01-22
Featurepoof 3

Anarchid:
Third take on getting features to not disappear silently because that is just too jarring.

- smaller 
- starts at 0.75 opacity
- enabled


GoogleFrog:
Why not little shards of metal instead of a giant brown cloud?

Anarchid:
Brown dustcloud is a neutral and generic thing, metal basically means it's for metal things. 

Some features such as tree shouldn't have metal shards but i can make some different CEG's and autodetect based on feature type, and maybe even add a featuredef customparam to override such autodetect. 

One thing i'd like to do would be to detect reclaim from crushing/burning/blowing but there doesn't seem to be a reliable way.

Anarchid:
Withheld for repairs as described above.

PR #536, opened 2015-01-24
Revert "Jsonify and fix modstats."

GoogleFrog:
Reverts ZeroK-RTS/Zero-K#530


Anarchid:
Generally adding at least one sentence describing a revert's reason would avoid quite some unnecessary confusion.

PR #537, opened 2015-01-24
i-cons

sprunk:
cons get icons that reflect their factory


Skasi:
Back when I made them in 2011 people complained the icons looked like they were prisoners...
evennewericonz

Licho1:
Why is it "M" anyway? Who designed it that way? Me?
It could be W for worker instead.

KingRaptor:
It's an M because I based the original icon set off [this](http://en.wikipedia.org/wiki/NATO_Military_Symbols_for_Land_Based_Systems).

sprunk:
Skasi yours look like prisoners because they are off scale. The M is way too large compared to the backgrounds and is also a bit too high. Compare to mine

Issue #471, opened 2014-12-29
CBAI Issues


mtroyka:
I decided to actually take a look at the CBAI code and sort of figured out some of what is causing bad behavior.

1: Constructors going in circles.

The IsTargetReachable function produces a bunch of intermediate waypoints to the build site which very often end up in nonsensical locations (for example in the opposite direction). This leads to units moving between two waypoints without ever going to a job. I can't figure out how those waypoints get added to the order array, however, so I don't know how to safely remove them. I assume that telling the constructor to move directly to the build site would use normal unit pathing AI, and should work fine, but I don't know that to be a fact.

2: Constructors mobbing on jobs. For example, if you have 2 cons and queue a mex and a solar right next to each other, you still get both cons mobbing on the solar or mex instead of assigning 1 cons to each. 

The way jobs are assigned is kind of silly, but also kind of smart. Each reachable job is more or less given a priority relative to the selected worker and then the highest prio job is chosen. Workers already assigned to a given job are also accounted for, but the way priorities are determined is kind of derp. A more straightforward approach would be to give each job a weight equal to:

distance to job * (number of workers already assigned to that job + 1)

then choose the job with the lowest weight. I think the complexity of that would be about the same as it is now, too.

3: Constructors won't get assigned to new jobs once they're already assigned to a job, even if the newly added job is (much) closer.

Note that the general complexity of job assignment is num.workers * num.jobs, so you have to limit the rate at which constructors are reassigned so as to avoid pathological performance. I can see that there's already been an attempt to make some compensation for this, but it fails to account for common cases (ie queueing a job right next to a mob of workers that want to run across the map to some other job). Jobs need to be re-weighted more frequently, somehow or other.

4: Workers suiciding into enemy units/porc.

I haven't had any real trouble with this, but I did find out after looking at the code that it's supposed to automagically remove jobs that fall into enemy-controlled territory. However, I've never seen it actually do this, probably because either the radius is set too small for it, or the %enemy threshold is set too high. This one is harder to test, but easy to adjust. It might also be good to test that code in general and make sure it's actually working properly.

rlcevg:
> 1: Constructors going in circles.
Constructors running in circles even with direct build command (tested with C++ AI). Possibly an engine bug.

> 2: Constructors mobbing on jobs.
More mobbing = faster construction. Mobbing is good, but should be constrained (also) by something like target_cost/used_build_power ratio.

xponen:
[...]

PR #542, opened 2015-01-27
CBAI widget improvement, Issue #471

xponen:
Added feature:
1) reassign job between job every 4 sec
2) less mobing (more mobing for expensive building only)
3) keep queue of unfinished nanoframe (cons can return back to work if interrupted)


xponen:
ops wrong base

Issue #278, opened 2014-09-30
cant jump off surfboard


sprunk:
fix should be easy (steal force unload from air transports and use it before jump if jumper is surfing).
i might do it on the weekend.

xponen:
can I try? I have experience with Air Transport drop and jumpjet gadget.

I think its simply done by calling the following:
local env = Spring.UnitScript.GetScriptEnv(unitID)
Spring.UnitScript.CallAsUnit(unitID,env.DropUnit,cargoUnitID)

Going to try it first and possibly will inspection jumpjet gadget (maybe tweak some comment or something).

PR #545, opened 2015-01-28
Jumpjet gadgets: Add "jump-off-surfboard" feature, fix issue #278

xponen:
Summary of changes:
1) Implement/ Fix issue #278 
  - jumper can jump off surfboard using MoveCtrl-jump (and Impulse-jump)
  - reload bar will fill-up faster/slower/none depending on the unit that the jumper morphed into, during reload-cycle

-  Changes 1: https://github.com/xponen/Zero-K/blob/3c2dd29ce8e06127dad2de212979d2e1a56c55d6/LuaRules/Gadgets/unit_jumpjets.lua#L301-L308
-  Changes 2: https://github.com/xponen/Zero-K/blob/3c2dd29ce8e06127dad2de212979d2e1a56c55d6/LuaRules/Gadgets/unit_jumpjets.lua#L401-L410

2) tweaked Impulse-jump gadget:
  - jumper will jump faster/slower depending on the unit that the jumper morphed into, right before jumping.
  - cleaned Impulse-jump code by removing legacy cob support, reload-cycle are effected by stun & slow (Merge changes from MoveCtrl-jump)


xponen:
ops wrong branch again!

PR #547, opened 2015-01-29
fix #528

### LINKED ISSUE #528: slow bar only half fills
This is kind of a minor aesthetic issue, but the bar displaying slow dmg on units only half fills even when slow damage is maxed. I think it's probably fine to have it read "50%" in the text since that's what slow maxes at, but the purple bar should read full and not half full when slow is maxed (otherwise it looks like slow is -never- maxed).

sprunk:
Makes slowdamage healthbar filled on max slow

### DIFF
--- a/LuaUI/Widgets/unit_healthbars.lua
+++ b/LuaUI/Widgets/unit_healthbars.lua
@@ -830,7 +830,7 @@ do
 	  --// SLOW
       local slowState = GetUnitRulesParam(unitID,"slowState")
       if (slowState and (slowState>0)) then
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Why? 50% shows how slow a thing actually is.

sprunk:
Ticket #528 argues that the bar should be 100% filled when the slow is at cap and cannot be increased anymore, because being unable to fill a bar is unintuitive.

GoogleFrog:
I see, you did not change the text so it still communicates the maximun amount of slow.

PR #551, opened 2015-01-29
Context Menu: unit guide grouping

sprunk:
This is the same grouping as Initial Unit States:
Units get grouped by fac.
Buildings get grouped by Economy/Defence/Special.
Misc stuff (Comms, Chicken, etc) go under Misc.

Currently everything is grouped alphabetically and includes a ton of fake units and random people's comms.

### DIFF
--- a/LuaUI/Widgets/gui_contextmenu.lua
+++ b/LuaUI/Widgets/gui_contextmenu.lua
@@ -108,7 +108,7 @@ local colorCapture = {0.6, 1, 0.6, 1}
 
 local function MakeStatsWindow() end
 options_order = {'shortNotation'}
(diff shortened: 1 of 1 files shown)


KingRaptor:
Nebula, Kestrel, Eclipse etc. aren't included, should we put them in?

sprunk:
These are purposefully left out because they don't exist in normal games and would confuse newbs.

A category for campaign units that only gets shown in singleplayer would be a good idea but the availability of these is fairly random.

Another thing to do would be to improve the Commanders sub-menu (currently it includes tons of testing, campaign and CAI comms, and gets 30 more entries per person in multiplayer) but that's of low importance.

PR #552, opened 2015-01-29
Context Menu only shows DPS for continuous beamlasers

sprunk:


KingRaptor:
Should automatically detect (compare beamtime to reload time, or just reload time < x)

sprunk:
Not that simple.

Detecting low reloadtime won't work because Archangel has less reloadtime than a Banshee.
Low beamtime won't work either becase Flea has less beamtime than Banshee.

A simple beamtime to reloadtime comparison won't work because all combinations exist:
* Banshee has beamtime > reloadtime.
* LLT has beamtime == reloadtime
* Comm beamlaser has beamtime < reloadtime

Fixing the weapons can't be done because that's messing with gameplay values.
Some leeway for equality would probably work but that needs to be wary of edge cases.

PR #567, opened 2015-02-08
Increased reload time on airpads in attempt to nerf Raven

Orfelius:
Increased reload time on airpads from 15 cycles to 20 (that is 2.5 second) in attempt to lessen dps output of raven. This is a very minor change but I am afraid to hurt other bombers too much by it as well :\
While it wont hurt Wyvern and Thunderbird as much because these planes usually wait for their window of opportunity anyway and they hover around doing nothing after they spend a shot. Phoenix will suffer considerably and might require small buff?

### DIFF
--- a/LuaRules/Gadgets/unit_refuel_pad_handler.lua
+++ b/LuaRules/Gadgets/unit_refuel_pad_handler.lua
@@ -169,7 +169,7 @@ local function SitOnPad(unitID)
 			
 			landDuration = landDuration + 1
 			
-			if landDuration%15 == 0 then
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Oops I thought I got a comment. This is both:
 * A balance change.
 * A poor implementation.

Orfelius:
Sure at least thanks for response since you seem to ignore every single thing I say.
This sure is a good way to discourage new people trying to participate in the project I tell you that.
Oh and referring strictly to this PR: I think that this was not good change at all either as it harms every bomber in the factory.

PR #571, opened 2015-02-16
reenable take (until xponen does it properly)

### LINKED ISSUE #570: bugged coms when people resign during pregame phase
http://zero-k.info/Battles/Detail/330353

player nassier resigned during the startup phase after selecting his com, without picking a starting position. as a result, a noname-owned commander dropped from the sky that couldnt be /taken nor was it given by the afk monitor, despite blinking. 

http://zero-k.info/Battles/Detail/329616

in this other game with the same bug, alcur resigned during the startup and later rejoined, but did not get his com back.

sprunk:
fixes #570

### DIFF
--- a/LuaRules/Gadgets/game_disable_take.lua
+++ b/LuaRules/Gadgets/game_disable_take.lua
@@ -30,11 +30,11 @@ function gadget:AllowUnitTransfer(unitID, unitDefID, oldTeam, newTeam, capture)
 	end
 	local _,leaderID,isDead,isAI = spGetTeamInfo(oldTeam)
 	if isAI or isDead then --DISALLOW /take of AI or dead player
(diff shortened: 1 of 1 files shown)


xponen:
no no no, the problem is Lagmonitor broken (it forgot to check player resign before game start), not /take gadget broke

also, 
this is NON-ideal way to disable /take gadget.
Ideal way is to turn-off the gadget itself.

I can look at lagmonitor soon, or someone do.
But this is not ideal way to fix.

xponen:
I have to close this, I'm sorry.

Don't want other dev to accept this.

xponen:
Refering: https://github.com/ZeroK-RTS/Zero-K/issues/570

PR #572, opened 2015-02-17
Add blinking to slowed units

sprunk:
The same way EMPed and Disarmed units blink. Intensity is based on slow%.


sprunk:
#573 better attempt

PR #573, opened 2015-02-17
Add blinking to slowed units

sprunk:
The same way EMPed and Disarmed units blink.
Intensity is based on slow%.


GoogleFrog:
I disagree. Slowed units are visibly slower. The flashing would not accurately show the proportion. Slow and disarm flashings could both be relevant and somehow overlap. The PR has a capture reload change.

sprunk:
ok

PR #574, opened 2015-02-18
#571 again in the suggested way

sprunk:


sprunk:
eh let xponen handle that

PR #576, opened 2015-02-20
Disable game_disable_take.lua gadget for workaround random orphaned Com blinking problem

xponen:
orphaned com issue is reallly long issue, exist since ancient time, not sure can fix while people might experience random case of flashes of light.

People will usually /take this units. In mean time I investigate lagmonitor.

If its not really important issue then don't Pull this yet.
ref: https://github.com/ZeroK-RTS/Zero-K/issues/570


xponen:
new lagmonitor fix

PR #586, opened 2015-02-25
mex base share obeys communism/RoI modoption

sprunk:


xponen:
have you tested this? this is to disable refund, is that right?

also, this is commit-ing into Master not Develop

sprunk:
ok will re-submit to develop

PR #592, opened 2015-03-01
stockpile obeys disarm

sprunk:


GoogleFrog:
Screamer?

sprunk:
Stockpile: Screamer & Silencer.

GoogleFrog:
I don't believe you.

sprunk:
...???

GoogleFrog:
Does it interact with priority to communicate the change in drain? It does not look like it does.

sprunk:
https://github.com/sprunk/Zero-K/blob/patch-88/LuaRules/Gadgets/unit_stockpile.lua#L74
disables the drain when disarmed

https://github.com/sprunk/Zero-K/blob/patch-88/LuaRules/Gadgets/unit_stockpile.lua#L58 re-enables the drain when un-disarmed

PR #593, opened 2015-03-01
Capture scaling (3-2x) -> (1/x)

sprunk:
That is the same scaling as other status effects, ie. at eg. 25% hp the unit will be capped 4x as fast.
Currently it goes up to 3x at 0%, which:
* is not consistent with the other status effects
* relies on the 3 constant which is not really documented anywhere
* doesn't keep the time-spent-capping-to-received-value ratio constant; currently you get best efficiency for 75% hp units due to the equation
* previous also causes targeting issues (low health targets are preferred but actually have much worse cap efficiency)

Balance is kept intact for full health units; units at (0.5, 1) health would be captured slightly slower than currently and units at (0, 0.5) health would be capped faster.

### DIFF
--- a/LuaRules/Gadgets/91_unit_capture.lua
+++ b/LuaRules/Gadgets/91_unit_capture.lua
@@ -20,8 +20,6 @@ local RETAKING_DEGRADE_TIMER = 15
 local GENERAL_DEGRADE_TIMER = 5
 local DEGRADE_FACTOR = 0.04
 
-local DAMAGE_MULT = 3	-- n times faster when target is at 0% health
-
 include("LuaRules/Configs/customcmds.h.lua")
 local CMD_STOP = CMD.STOP
 local CMD_SELFD = CMD.SELFD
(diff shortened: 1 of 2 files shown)


Pxtl:
Iirc, wasn't this deliberate?  heavily-damaged units being hyper-easy to capture was problematic... You had to carefully beat the unit into an inch of its life for near-instant cap.

sprunk:
Heavily damaged units being hyper-easy to capture is okay because they are much less valuable.
In combat, a heavily damaged unit is about to die so it's not going to contribute much if capped. If someone manages to cap a near-dead unit and somehow save it then they probably deserve to keep it.
Capturing a unit safely (eg. a stunned and napped one) is usually quick enough for it not to matter much; note that the damage you do must subsequently be repaired for proper usage so even though you save some time you will still need to sacrifice time to repair them.

KingRaptor:
I think the main reason I originally implemented this was the silliness of instacapping nanoframes, which should probably get a "proper" fix anyhow.

sprunk:
One way to easily fix nanoframe blocking is them not causing the 12s cooldown. Downsides include a consistency decrease but capture is pretty wonky anyway.

xponen:
I have graphed the 2 capture multiplier here, to aid decision:
capture_multiplier

xponen:
just keep using existing scale, work fine

PR #594, opened 2015-03-01
allow air transports to unload comms into water

sprunk:


xponen:
for the record:
this commit make transport drop comm into water (usually it can't unload anything into water).

PR #595, opened 2015-03-01
Mexes obey RoI modoption + additional RoI settings

sprunk:
The patch makes Mexes obey RoI modoptions (currently they disregard communism and get refunded anyway).

There are now 3 RoI modoptions. Current default one is unchanged, everything is returned.
The new ones are OD Return (where only energy structures are refunded through OD), and Mex Return (where only mexes are returned through base income).


xponen:
I think its good, 
It disable mex payback in communism mode, but also added option to disable mex or energy payback selectively... (maybe unnecessary?).

Problem is to make the option sound good...
Maybe this will take too much time, I'll just Pull.

Issue #562, opened 2015-02-04
New "structure on terraform" interface fails with mexes


MLanghof:
Reproduction steps:
With the (always-on) mex-snapping active, manually select a mex from the build menu while holding alt. Click slightly away from a metal spot.

You can now choose the terraform height, but it will be centered on where you clicked, not on the mex spot, thus defeating the purpose.

It will also fail to queue the mex onto the mex spot after the terraform (even if the mex could be built there before and you clicked perfectly centered).

In short, the Mex Placement Handler and the new building-terraform interface don't cooperate.

xponen:
okay, I think I know how to fix this. 
(central_build widget knows how to cooperate with Mex Placement Handler, will copy this)

PR #601, opened 2015-03-05
fix Alt+mex terraform not working

xponen:
This address issue https://github.com/ZeroK-RTS/Zero-K/issues/562

Changes:
  - I moved "PlacingRectangle" code into "widget:CommandNotify()" (in gui_lasso_terraform.lua widget). This immediately fix https://github.com/ZeroK-RTS/Zero-K/issues/562

  - I made WG.Terraform_SetPlacingRectangle() to be defined at widget:Initialize() because defining WG. function outside widget:xxxxx() function will broke local copy of same widget due to conflict problem!**

  - also, I had to modify stuff in cawidget.lua to make  "CommandNotifyMex()" event work with multiple widgets (the event from cmd_mex_placement.lua being used by both terraform and CBAI widget to read mex placement).

  - additionally (minor) I tweaked CBAI widget to recognize the build-queue produced by "structure on terraform", so cons will assist the queued buildings.

I tested locally and fully satisfied with the result, 
**Unfortunately I couldn't use local gui_lasso_terraform.lua in ZK:stable MP because of the WG. conflict. However I'm 100% sure it worked.

### DIFF
--- a/LuaUI/Widgets/gui_lasso_terraform.lua
+++ b/LuaUI/Widgets/gui_lasso_terraform.lua
@@ -2,7 +2,7 @@ function widget:GetInfo()
 	return {
 		name      = "Lasso Terraform GUI",
 		desc      = "Interface for lasso terraform.",
-		author    = "Google Frog",
+		author    = "Google Frog", --also tweaked by xponen
 		version   = "v1",
 		date      = "Nov, 2009",
 		license   = "GNU GPL, v2 or later",
@@ -1536,6 +1536,7 @@ end
 
 function widget:Initialize()
 	widgetHandler:RegisterGlobal(self,"CommandNotifyMex", CommandNotifyMex) --an event which is called everytime "cmd_mex_placement.lua" widget handle a mex command. Note that "self" is added because this widget is tagged "handler=true"
+	WG.Terraform_SetPlacingRectangle = Terraform_SetPlacingRectangle --MUST Initialize like this! or else the very act of putting the file in Widget folder will add these function into WG.
 end
 
 --	receive broadcasted event from "cmd_mex_placement.lua"
@@ -1663,7 +1664,7 @@ function widget:CommandNotify(id,pos,options)
 	--]]
 end
 
-function WG.Terraform_SetPlacingRectangle(unitDefID)
+function Terraform_SetPlacingRectangle(unitDefID)
 	
 	-- Do no terraform with pregame placement.
 	if Spring.GetGameFrame() < 1 then
--- a/LuaUI/Widgets/gui_lasso_terraform.lua
+++ b/LuaUI/Widgets/gui_lasso_terraform.lua
@@ -791,23 +791,6 @@ local function calculateAreaPoints(mPoint, mPoints)
 	
 end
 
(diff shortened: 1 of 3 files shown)


GoogleFrog:
Too complicated. Why the CBAI and especially why the widget handler changes?

Licho1:
Imo it needs more rethink.
What i suggested is to make it work like this:
With structure to be placed selected, hold mouse to enter "height mode"
adjust height, when mouse is released, both terra and structure are queued.

This should listen to stanard modifiers (shift, space etc) for adding/inserting into queue.

This will also make it work with gesture menu and mexes.

If you provide simple API it could also be used by CBAI

xponen:
@GoogleFrog I fixed issue #562 by adding a global function in gui_lasso_terraform ("CommandNotifyMex") which is called from cmd_mex_placement. Unfortunately I can't add 2 global function with same name :( this is bad for CBAI widget 

So I modify the cawidget.lua to allow adding 2 or more global function with same name! 
It seem okay. I can definitely test in regular game and see if it is wrong.

xponen:
@Licho1 I only fix 1 case (terraform mex placement), so I wasn't touching the key shortcut.

But, since I moved "structure on terraform" code into engine's "widget:CommandNotify()" it now able to read SHIFT& ALT, so if @GoogleFrog  make it obey SHIFT & ALT then we can make like... a wall of spikes for example :P

CBAI only assist buildings, not terraform (CBAI is just constructor assisting other constructor and split queue)

Anarchid:
Mousedrag altude change would be interesting as an UI option. 
I still don't know how to use the current one, i tried a few times and always failed.

xponen:
@Licho1 and @Anarchid ,the suggestion to change key shortcut need to be in dedicated issue, this PR just fix ALT+ mex_placement_terraform,

although complicated, its just fixed that thing.

xponen:
@Licho1 , I could add the mex terraform shortcut to any build menu widget. will do in other PR

Anarchid:
> this PR just fix ALT+ mex_placement_terraform and will be closed upon PULL

Wouldn't the drag-terraform *circumvent* it and make that fix unnecessary?

xponen:
@Anarchid , you mean, don't need alt+structure_terraform at all because can do manually?
IMO, doing the alt+structure_terraform right have potential in future, its a step toward where you could terraform grid and stuff, although its still not able to do that, but this is 1 step closer.

anyway, somehow this version broke! I must have changed something before Push and broke it... :/

Anarchid:
> you mean, don't need alt+structure_terraform at all because can do manually?

No: i mean that a better UI could be had instead. However, i now realize that just dragging the build option wouldn't help because shift already uses it to draw lines.

xponen:
@Anarchid  the widget is terribly complex, too much global variable... only @GoogleFrog should reinvent the terraform User Interface.

Licho1:
Dragging helps its simple. Hold with mouse still (with lmb) when placing to
enter height mode (after say 0.5).
Its distinct from drag to draw line of structures.

2015-03-05 17:53 GMT+01:00 xponen <notifications@github.com>:

xponen:
merged for the fix only

PR #607, opened 2015-03-07
Crabe: don't curl when stunned/disarmed/reclaimed

sprunk:


xponen:
waitz! disarm is just weapon disabled

xponen:
I'm so silly, only found out the following problem after merge:
1) disarm shouldn't stop curl because its not stun,
2) there's problem with the animation: the crabe don't stop walking when emped, same is when being disarmed,
3) also (minor) it didn't actually stop the animation of curling if being emped/disarmed when it begun curling.

PR #608, opened 2015-03-08
Hide minimap in tab mode

### LINKED ISSUE #553: When zooming out with tab, minimap does not hide itself (nvidia gtx 970)
http://i.imgur.com/OW8MlyT.jpg

kaen:
Closes #553 

Not sure what mode == 7 referred to originally, but I tested this against all of the camera options. LMK if this is not the right way to solve the problem.

### DIFF
--- a/LuaUI/Widgets/gui_chili_minimap.lua
+++ b/LuaUI/Widgets/gui_chili_minimap.lua
@@ -650,16 +650,13 @@ end
 
  --// similar properties to "widget:Update(dt)" above but update less often.
 function widget:KeyRelease(key, mods, label, unicode)
-	if key == 0x009 then --// "0x009" is equal to "tab". Reference: uikeys.txt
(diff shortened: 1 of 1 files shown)


dvr333:
This should happen on any zoom-out. this solution is okay, but hacky, but the existing setup with COFC and transparency should be extended to apply to any camera (line 830 of gui_chili_minimap). That assumes that the maximum zoom distance of any camera is something Lua can know.

sprunk:
Tab is not guaranteed to be the hotkey for the "tab mode" so it might be a good idea to check for the constant instead of the key.

In future engines the constant will become 5 (not 7).

dvr333:
Mode as a check is far more complicated, if for nothing besides 91 compat. Name seems to be more consistent between engines.

kaen:
cool, thanks!

PR #609, opened 2015-03-08
Revert "Crabe: don't curl when stunned/disarmed/reclaimed"

xponen:
ops! disarm shouldn't stop curl because its not stun,

and, there's problem with the animation: the crabe don't stop walking when emped, same is when being disarmed, 

also (minor) it doesn't actually stop the animation of curling if being emped/disarmed after it begun curling.

### DIFF
--- a/scripts/armcrabe.lua
+++ b/scripts/armcrabe.lua
@@ -107,10 +107,6 @@ local function Curl()
 	if nocurl then return end
 	--Spring.Echo("Initiating curl")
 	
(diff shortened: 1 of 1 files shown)


Anarchid:
> waitz! disarm is just weapon disabled

No. Disarm also disables jump, cloak, shield generation, and basically all active special abilities. And yes it should also disable changing armor state because EMP already does that, at least in case of Gauss.

I'm afraid you're going to have to do a revert revert now.

Licho1:
Yeah disarm should block all abilities incl curl imo ...

PR #611, opened 2015-03-08
Crabe: don't curl when stunned (#607 without disarm and moonwalk)

sprunk:
#607 except it no longer moonwalks and disarm won't stop the curl.


xponen:
this work as told.

Issue #577, opened 2015-02-22
Game Details presets set BumpWaterRefraction to 2


dvr333:
The higher game details presets (Ultra, and maybe High) set BumpWaterRefraction to 2. This mode causes issues with commander names and can sometimes cause issues with underwater unit visibility as well.

The setting should probably never go above 1 in presets.

dvr333:
Wait, this should be in Infrastructure. Sorry

Licho1:
Nah it belongs here because its a game bug (the rendering)

dvr333:
I can remove the depth check on names (and icons) and that helps, it's probably not a bad idea either. However, the main issue is probably engine-side, as the water layer has a depth mask when  BumpWaterRefraction is set to 2 and doesn't when it is set to 1. This also affects facplop effects, and generally messes with everything that assumes water is considered transparent to depth, so that the depth value that matters is the ground below the water, not the water itself.

PR #613, opened 2015-03-08
Fix for underwater names and icons

dvr333:
Partial fix for #577, though the real fix is engine-side.

EDIT (2014/04/21): Fully fixed with the most recent commits. Engine side fixes only need to worry about unit build wireframe effects.


GoogleFrog:
Then you can see the icons behind terrain, which is bad.

dvr333:
I don't think this affects unit icons, just state icons, some of which you can already behind terrain.

GoogleFrog:
I meant the state icons. Currently they are blocked by terrain.

dvr333:
In that case, we're dealing with an engine bug (bumprefraction 2 makes water surface have depth value like terrain), and the lobby settings presets shouldn't set bumprefraction to 2.

I still think names should show through terrain.

xponen:
i think this conflict can be resolved by making the fix only activate for underwater unit and under specific settings.
we can read setting like this: Spring.GetConfigInt("bumpwatersettingname"),
and probably detect if its drawn underground/underwater if height is <0

dvr333:
There are other issues with bumprefraction set to 2 and depth testing, both in widgets and in general play, so the only real fix is to the engine, and that means we can't really use that setting in the lobby presets until then.

dvr333:
Finally found a solution that handles water properly. Still not sure how to fix outlines and halos, but everything else works properly.

PR #615, opened 2015-03-08
More complete minimap fade

dvr333:
Partial fix for #553, though a full fix requires a function that returns engine camera max zoom distance. NB: this value probably changes per map, so don't even start with suggesting hardcoding.


Jamanno:
Is it possible to have least a option to "not" hide minimap. I often zoom out and need minimap to get overview. Many other players think so too

dvr333:
Settings/HUD Panels/Minimap/Fade minimap on zoomout.

Jamanno:
Thats the way we go... I would demand minimap always visible and for those who do not want can check that option

dvr333:
Talk to Licho, he had the feature request, though I did make it initially (without compete hiding) when tightening the zoom bounds for COFC to keep the minimal from blocking the main map.

PR #616, opened 2015-03-08
fix nanos/cons/facs not having income

sprunk:


Licho1:
Wut removal of their income should happen (except for comm)
Dne 9. 3. 2015 2:19 "GoogleFrog" <notifications@github.com> napsal(a):

Anarchid:
Except this was not intentional income removal but a flawed refactor.

PR #625, opened 2015-03-11
differentiate razor pew

sprunk:
the patch makes razors use a new pew graphic to differentiate the aa-only red pew from the universal pew.

razor


Anarchid:
Shredder also exists!

sprunk:
both are affected

GoogleFrog:
I don't like the look of it for Razor. This looks more like a bolt of something instead of a pew pew laser. The sound, speed and length differentiation seems like changes are not required.

sprunk:
Would a bolt (the bitmap, not necessarily red and spammy) look good on anything else?

Anarchid:
IMO it looks good on spammy AA lazor and differentiation has utility.

sprunk:
Swift and Krow are both universal pew, not AA pew.

KingRaptor:
I like this but I think the bolts as shown look a bit long.

sprunk:
That is trivial to change; want a bunch of comparison pictures?

Pxtl:
Why not teal to match other AA lasers?  Keep a distinct shape so it doesn't look like the archangel weapon, but otherwise matching up with the jumpy and cloaky as would be good.

sprunk:
I kept red because the shape was differentiating enough and because a colour scheme does not really exist (eg. teal is not AA because LLT).

sprunk:
trimmed pew to 2/3 the size

sprunk:
lack of response; closing

PR #626, opened 2015-03-11
minor optimisations for stockpile

sprunk:


xponen:
I have tested. The change from UnitCreated -> UnitFinished doesn't seem to cause any difference, even with /give screamer .

xponen:
this pulled into master @_@

PR #627, opened 2015-03-11
X button in vote now closes window

Skasi:
The X button on the vote window now hides the current vote. New votes that are started will still be displayed again.

Previously it cause !endvote spam by new players who did not understand what the X button does and kept pressing it. Providing an !endvote button to people who mostly don't even have admin/boss rights, let alone know what !endvote does is not needed.

### DIFF
--- a/LuaUI/Widgets/gui_chili_vote.lua
+++ b/LuaUI/Widgets/gui_chili_vote.lua
@@ -35,6 +35,7 @@ local button_end, button_end_image
 --------------------------------------------------------------------------------
 local voteCount, voteMax = {}, {}
 local pollActive = false
+local hidePoll = false
 local votingForceStart = false
 
 local string_success = "END:SUCCESS"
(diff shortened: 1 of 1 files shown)


xponen:
Skasi, I think you need to make it re-appear when vote status changes. Like ZKL.

sprunk:
if somebody cares about the vote status then they simply won't close it

Skasi:
Exactly. If people press X on something then they want it gone. Simple as that.

Making the window reappear when someone votes would be as easy as only calling "RemoveWindow()" when the button is pressed in the original version of gui_chili_vote.lua at line 282, but that's not what I wanted.

xponen:
okay, I think maybe I would PR a minimize button.

Because I remember my first time I also wanted to close the window but not because I don't like it, but because it blocked my view.

xponen:
pls check if pulling into master or develop later.okay

PR #628, opened 2015-03-12
LUS notification gadget

sprunk:
This is #331 done the suggested way. Adds a gadget that other gadgets can use to communicate with LUS. Currently used by EMP and Disarm gadgets to tell LUS a unit got stunned by their respective status effect.


Anarchid:
Exposing the `callScript` function directly might (or might not) be better here because it will then not require all those wrapper functions to be added for other or future potential uses (such as carrier drones notifications, or maybe dyncomm morph notifications to show/hide specific pieces).

sprunk:
I understood GF's words in #331 comments as a request for the wrappers. Gadgets ask the notification gadget to tell LUS about stuff and don't care about the exact implementation details. If GF doesn't mind then a modification to expose callScript can be made.

PR #633, opened 2015-03-14
Fixed end-graph time formatting for games that take an hour or longer

dvr333:


xponen:
not sure what the fix look like (didn't try any hour long game), but seem like an aesthetic fix by Shadowfury333, so must be good.

dvr333:
Watch my latest cast. The bug was that if something took, for example, 1h 7min, it would show 1:67:00, when it should show 1:07:00.

PR #636, opened 2015-03-18
remove mantis_4244.lua: bug was fixed in latest engine

xponen:
https://springrts.com/mantis/view.php?id=4244

Have re-tested using Example Mod on Spring 98.0.1 423 and the small paralyzeTime no longer replace big paralyzeTime

Example Mod: https://code.google.com/p/zero-k/source/browse/trunk/mods/examplemod/?r=12381
http://zero-k.googlecode.com/svn/trunk/mods/examplemod


GoogleFrog:
The gadget is only enabled in 91.0 so this is not a large change. Removing the gadget does only these two things:
 * Breaks behaviour in 91.0.
 * Makes the repository slightly neater.

It looks like we will be able to stick with the dev engine (ie, break 91.0) but I would not like to break it piecemeal. It should be broken all at once, everywhere, in a branch which is then merged to master. We should have a branch that is left dormant for the last 91.0 compatible version. The last stable release that works for 91.0 should also be known and recorded.

PR #640, opened 2015-03-20
prevent obediation

sprunk:


Licho1:
Same check could be done in springie where similar checks already exists.

PR #644, opened 2015-03-21
Make mex spots actually team coloured for spectators, with player-specific option

dvr333:


GoogleFrog:
What do you mean? Do they match team color or allyTeam?

dvr333:
Player coloured so far, but I have a local untested change to make it allyteam coloured when this option is false (rather than based on who is selected).

sprunk:
Team color. This mostly matters for spectators who can change the [ally]team they're spectating which shifts the mex circle coloring but keeps teamcolours.

For example cyan player has cyan mexes and red has red mexes. But if the spec starts spectating the red player, the red player now has cyan mex circles and the cyan player now has red circles.

dvr333:
In more detail:

Current master: Teal for selected allyteam, red for others.

Current PR: the above when new option is false, player-coloured if true (teal player has teal mexes, green player has green mexes, yellow player has yellow mexes, etc)

Local untested change: like the current PR, but now when option is false it is allyteam coloured, rather than based on spectator selection.

xponen:
I noticed a bug, the new option isn't visible in menu!

turn out you didn't check this 
https://github.com/ZeroK-RTS/Zero-K/pull/644/files#diff-c432c444bd85b95b56f68ad39f4cbbc0R115
in your recent commit.

otherwise, after typo is fixed, it work as described.
I will pull after typo is fixed.

dvr333:
Oops, I thought I got everything. I'll fix it ASAP (which may be a while, I'm at work now and have plans in evening.

xponen:
okay thanks!

PR #648, opened 2015-03-25
COFC: fix jumpyness in new spring

xponen:
It fixed jumpiness in zoom & zoom out, and when follow cursor


Licho1:
To correctly interpolate angles, one should use quaternion representation and standard quternion slerp.

Anarchid:
Inb4 quat ipo in Lua.

GoogleFrog:
So is this supported by others?

PR #650, opened 2015-03-26
Widget translation support

N0U:
Example:
```
local tr
local hellWorld
function widget:Initialize()
	tr=WG.initializeTranslation(GetInfo().name,langCallback)
	hellWorld=tr("helloworld")
end

...
...

function foo()
	Spring.Echo(hellWorld)
end

...
...
function langCallback()
	hellWorld=tr("helloworld")
end
```


sprunk:
>tr=WG.initializeTranslation(GetInfo().name,WG.lang)

Why not skip the WG.lang argument and let this variable hold all the translations for given widget? This way a user changing language would be seamless: the tr() function would be the one to pick the correct translation based on current WG.lang

Anarchid:
I've been thinking that this could also look nicer in widget handler rather than having to modify every widget.

Unless anyone objects, i'll however merge this in a couple days.

N0U:
> Why not skip the WG.lang argument and let this variable hold all the translations for given widget? This way a user changing language would be seamless: the tr() function would be the one to pick the correct translation based on current WG.lang

I guess than this stuff should be embeded into epicwidget, so it would call `setLocale(WG.lang)` everytime it changes, or i18n should be changed. I think second way is better because I don't want to put all these stuff into one widget. 

> I've been thinking that this could also look nicer in widget handler rather than having to modify every widget.
Anyway, it is just two lines.

Anarchid:
There were no new objections in last 4 days.

PR #667, opened 2015-04-01
Added outlines to mex spots in economy view

dvr333:
For visibility. Number of segments reduced to try to minimize performance hit, each circle now has 21 segments, rather than 32 for the old single circle.


xponen:
this is really pretty!
even if a person still want yellow, this will look like nice upgrade.

PR #670, opened 2015-04-04
Partial fix for new COFC interpolation assumptions

dvr333:
COFC modified to use the state that was passed into the engine initially, rather than the myriad intermediate states used for SetCameraState to make the interpolation work properly.

Cleans up some issues with zooming and rotation.


xponen:
this fix MMB camera rotation?

dvr333:
No, there is still an odd jitter throughout when using ctrl+MMB or alt+MMB, at least when moving the mouse.

PR #672, opened 2015-04-08
Code polish for camera state get, should have been in #670

dvr333:


xponen:
I found an instance where GetTargetCameraState() could not be used, such as 
third person mode (because its different camera):
https://github.com/dvr333/Zero-K/blob/final-cofc-cleanup/LuaUI/Widgets/camera_cofc.lua#L1562
https://github.com/dvr333/Zero-K/blob/final-cofc-cleanup/LuaUI/Widgets/camera_cofc.lua#L2145

however its super low priority, because I want soo much for local implementation in COFC because it had low tweakability and always trailing on unit's butts.

PR #676, opened 2015-04-09
Overwhelming Numerical Advantage for FFAs

sprunk:
Allow Overwhelming Numerical Advantage in FFAs

### DIFF
--- a/LuaRules/Gadgets/game_over.lua
+++ b/LuaRules/Gadgets/game_over.lua
@@ -61,7 +61,7 @@ local chickenAllyTeamID
 local aliveCount = {}
 local aliveValue = {}
 local destroyedAlliances = {}
-local allianceToReveal
+local allianceToReveal = {}
 
 local finishedUnits = {}	-- this stores a list of all units that have ever been completed, so it can distinguish between incomplete and partly reclaimed units
 local toDestroy = {}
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I don't think this should be implemented because players with very little can build up while stronger players fight each other.

Licho1:
Value used should be different, like total metal spent so far. You can win without having a single mobile unit in FFA.

sprunk:
Overwhelming Numerical Advantage means there is only one "stronger player" though, and much stronger than everyone else. The value of a single player has to be a multiple of *everyone else's value combined*.

Basically the situation is *even more skewed towards the big guy* than in 2-alliance situation unless all the little guys perfectly cooperate in which case it's the same.

db81:
> players with very little can build up while stronger players fight each other

I think this can be solved with only enabling ONA once there's only a single player supreme over everyone else.

sprunk:
That's what the code does. The value multiplier compares each person's value with the sum of everyone else's.

db81:
Ah, sorry, didn't read before posting.

GoogleFrog:
I still think numerical advantage should not do anything in FFA. Consistency in information is important and there is more room for a swing (due to some sort of assignation).

sprunk:
Alright

PR #678, opened 2015-04-10
facmarker

sprunk:


xponen:
armcsa might benefit "delayedRefresh" tag, like spherepole because it re-add marker when unit decloaked. But not sure if it will be spammy, it don't remove old marker.

xponen:
seem compelling to merge because of discussion in Chat

PR #681, opened 2015-04-10
remove "command console activated"

sprunk:
super annoying and useless (gets played halfway through loading)

### DIFF
--- a/LuaUI/Widgets/init_startup_info_selector.lua
+++ b/LuaUI/Widgets/init_startup_info_selector.lua
@@ -268,7 +268,7 @@ function widget:Initialize()
 	or (not coop and Spring.GetTeamRulesParam(teamID, "commSpawned") == 1)	then 
 		noComm = true	-- will prevent window from auto-appearing; can still be brought up from the button
 	end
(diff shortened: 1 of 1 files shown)


Licho1:
Why? If anything it needs more sounds to nofity that game is about start etc. Exlain why. Unless there is a major license problem (its warzone voice so it should not be), keep it

sprunk:
It does not really notify that game is about to start because in Spring 98 pathing is done after widgets are loaded so it fires off when load is at ~60-70% which can mean few minutes of wait when calculating pathing for the first time.

It doesnt respect volume settings (it is played when loading and volume settings might arent loaded yet) and is not togglable.

Most people seem annoyed by it (maybe poll would give good data on this).

Anarchid:
It seems like the proper solution would be to figure out when to start it properly.

KingRaptor:
Move it to loadscreen?

Anarchid:
Does loadscreen know the last instant of loading? Then yes.

sprunk:
Moving to loadscreen sounds sensible and superior to removal.
Closing

PR #682, opened 2015-04-10
purge needless gadgetry

sprunk:


Licho1:
Lock units might get used again by planetwars. Zombies are still played
sometimes (koth not ) Chicken control can be kept as well. Imo no need to
delete features, even if obscure, if those are optional (modoption) and
does not incur maintenance cost. If its not broken, no need to delete.

2015-04-11 1:43 GMT+02:00 sprunk <notifications@github.com>:

sprunk:
"Lock units" is not unlocks, this is some broken thing (unlocks are game_perks.lua) so Planetwars can still use unlocks.

Didn't see any zombies played for the last few months (cant search replays by modoptions. Last "zombies" in room title was 7 months ago, even older for koth/lava/ctf).

If people really want to they can make a mutator (like zkdota).

Licho1:
What is the reason for removal request of working features with no known bugs or maintenance cost?

sprunk:
They belong in mutators, they introduce a ton of modoptions which make modoptions hard to navigate in, they are enabled by default and add random global vars, they give no benefit because everybody stops playing them 2 weeks after they're made, and they introduce maintenance cost because related code is also in "real" gadgets (like eg. start_unit_setup.lua) which makes them harder to work with.

KingRaptor:
Lock units is for players to lock units (say "Krow, nuke, Dirtbag") at game start.

Sprung why is send enemy death being removed?

sprunk:
Because it does nothing (it sends enemy deaths in LoS to unsynced, but that is already known to unsynced).

KingRaptor:
Hmm, widgets have access to enemy deaths in LoS in newer game/engine versions? IIRC they didn't before.

xponen:
Why do we have to deal with this thing now? 
can't we solve game breaking issue first?
Also,
1) is rude to contributor
2) loss of feature that nobody ever have time to reimplement ever again

If you delete anything, nobody will ever readd them again.

sprunk:
2) is why we use git.

xponen:
git don't reimplement deleted gadget, people do. Its a work. nobody have time to re-add it later.

sprunk:
thats the point

xponen:
you said 
1) its because it is difficult to navigate the ModOptions. This could also be ZKL/UI problem
2) difficult to maintain start_unit_setup.lua

but proposed fix is to
1) remove those modOptions
2) remove related gadgets.

can there be different solution than this?

also, why not let the mode available so people can play it.

sprunk:
>why not let the mode available so people can play it

Because people don't play it anyway.
But if they want to they can use maps (eg. if you want lava you have hotstepper) so there is no need to keep that inside ZK.

>can there be different solution than this?

The other solution is to do what Licho suggests and keep these files. This means you make working with the codebase harder for 0 benefit (which i think is worse than removing them, hence the pr).

xponen:
I think you should start with something like code refactor and comment.

sprunk:
how is that even related?

GoogleFrog:
I agree with the changes with two exceptions:
 * Lava rise should be retained but perhaps not as a modoption. A good game-side lava implementation with a map interface is a good thing to have because any updates to the behaviour or visuals would be centralized. There are quite a few old lava maps with bad looking lava which could be re-released with a config to use gameside lava.
 * Incorperate the removal of 91_ gadgets and put the change in a ZK-repo side branch so we know where to find it if the change is needed.

GoogleFrog:
Send Enemy Death is not 'entirely pointless' because widgets do not get a unitDestroyed callin for enemies, even when they are in LOS. But iirc the only widget that uses this information is a silly announcer widget so I do not think the gadget is worth the complexity and sendToUnsynced cost.

GoogleFrog:
I guess if anyone wants to keep some gadget (like for example, zombies) they could demonstrate that it actually works.

Licho1:
Zombies do work, we played it many times in recent month, it's quite fun. Had both team and FFA games with it enabled. Removing it removes value, adds nothing.

sprunk:
Ok, here's the plan then:
* keep lava as a modoption for now. Later change into a modside API for lava maps.
* keep zombies. They have the highest quality of the silly modes and are the only one played.
* remove send_enemy_death. Evidence suggests widgets get that info, see the gui_attrition_counter widget.
* remove CTF and KotH. Nobody plays them and they are better with dedicated maps.
* remove Ore mexes. Nobody plays them and they decrease maintainability of unit_mex_overdrive.
* remove Player Chickens. Player chickens are massively imbalanced and they add code to start_unit_setup.
* remove no airplop, lock units, no air, and no def. Nobody uses these and if people don't want to play with a unit then the unit should be made more fun instead.

sprunk:
I also want to remove the modoption choice between boost and facplop because boost is never used and removal will allow a large cleanup of start_unit_setup, is that ok?

db81:
I don't think mutators are a solution, as they need to be re-released for new zk versions and that's too much maintenance cost for a minor novelty feature.

I remember @Anarchid complaining that the method used in Attrition Counter is unreliable. If send_enemy_death is superior to that, it makes sense to keep and use it.

sprunk:
The unreliability is because things like arty can kill units outside of their LoS and it does not count. Except being notified when stuff dies outside LoS would be spec hax and the send_enemy_death does not (and should not) do that.

xponen:
why remove anything at all just to clean start_unit_setup?? 
this is just a crazy shortcut!

you should figure out how stuff connect, then ensure your new code support existing feature.

xponen:
>The unreliability is because things like arty can kill units outside of their LoS and it does not count. 
too easy to fix, just add Los check in the gadget.

db81:
I just played 5 games of playerchickens and was reminded how fun they are. It would be sad to see them go just because of 15 lines in start_unit_setup. I think it's one of the more fun novelty features and it *does* get played.

TomFyuri:
If you'd called me. I would've joined to play them as well.

sprunk:
This is going nowhere at this rate, closing

PR #683, opened 2015-04-11
Cleanup Unit Marker

sprunk:


xponen:
IMO delayed refresh & com upgrade message don't need to be deleted.

com upgrade message worked fine 
and delayed refresh avoid unnecessary Marker when scythe decloak rapidly

sprunk:
Not the proper place to add 1
2 is implemented, read the code

xponen:
2 is not implemented, see 
https://github.com/sprunk/Zero-K/blob/patch-115/LuaUI/Widgets/unit_marker.lua#L44

this added marker each time Scythe appear. In previous code it add marker if Scythe re-discovered after 5 second.

Its rude to not add you as author when you delete the original code.

Anarchid:
`git blame` exists, so author-attribution is probably superfluous.

sprunk:
Read the previous code to see what 2 really did (hint: it did the same thing the new one does.
Git can handle last update date and authorship.

xponen:
okay nvm, I'll add timer later, or you can add it if you want to help.

sprunk:
timer is not needed

xponen:
worse thing about this PR is:
widget barely have code from original author, but never want to add self to author name.

PR #687, opened 2015-04-13
fix areashield module

sprunk:
also remove unused stuff

### DIFF
--- a/gamedata/modularcomms/moduledefs.lua
+++ b/gamedata/modularcomms/moduledefs.lua
@@ -566,6 +566,7 @@ upgrades = {
 	},
 	module_personal_shield = {
 		name = "Personal Shield",
+		order = 1,
 		description = "Generates a small bubble shield",
 		func = function(unitDef)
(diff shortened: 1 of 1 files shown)


xponen:
to make certain where the removed stuff originate, I tracked it to the following commit:
https://github.com/ZeroK-RTS/Zero-K/commit/61e864836baa53d12353f94e4c666eb9862001b3

the code was slowly maintain until its removal by N0U (banana_ai , https://github.com/N0U) in:
https://github.com/ZeroK-RTS/Zero-K/commit/989b060987540d52cf3858e6b748101bb76efe48

however the removal missed moduledefs.lua & start_unit_setup.lua

In light of this discovery, I recommend to exclude the cleanup from this PR and do a separate quantized PR that cleanup zkdota only.

xponen:
I forgot to ask what this fix is for,,,

Anarchid:
http://zero-k.info/Forum/Thread/13053

I suspect it's the cleanup that did the breaking, though.

PR #691, opened 2015-04-15
Add a simple ingame guide

sprunk:
* accessible through F10 Menu -> Help -> Guide
* contains info on various concepts (eg. radar, wreckage, priorities) sorted into categories -- currently not much but it is done in a format that makes it easy for random non-coder people to contribute
* the unit list moved to the same submenu

In the future accessibility will be improved:
* a way for newbies to easily access that submenu directly (most likely a big [?] button somewhere visible)
* space+clicking various UI elements (eg. the firestate button) should bring up the relevant article
* a way to easily embed icons and link to other "articles" would help quality

### DIFF
--- a/LuaUI/Widgets/gui_contextmenu.lua
+++ b/LuaUI/Widgets/gui_contextmenu.lua
@@ -108,7 +108,7 @@ local colorCapture = {0.6, 1, 0.6, 1}
 
 local function MakeStatsWindow() end
 options_order = {'shortNotation'}
-options_path = 'Help/Unit Guide'
+options_path = 'Help/Guide/Unit List'
 options = {
 		
 	shortNotation = {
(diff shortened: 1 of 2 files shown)


Licho1:
Is this a fourth attempt at this? Why not reuse existing stuff.
There is 
1) tutorial mission
2) nubtron
3) the tips dispenser (the blue icon that talks)

sprunk:
No, this is not a "what to do" guide, this is a "what this is" wiki.

Tutorial mission is one-time so cannot provide any help during normal games.
Nubtron is a "what to do" gameplay guide for complete nubs who don't yet need info on advanced stuff features like priority.
Tips are not available on demand, and dispenser does not use the menu for browsing.

Ingame wiki articles can be made to appear when space+clicking UI buttons (eg. firestates or priority) to provide information on what these buttons are.
The wiki can also explain concepts such as radar, EMP or unit roles in more depth than tips or nubtron can.

Licho1:
Ok just worried about dilluting effort and not having anything polished.

PR #692, opened 2015-04-16
COFC zoom speed fix

dvr333:
Also fixes MMB rotation


xponen:
tested it yesterday, super extremely responsive!

PR #693, opened 2015-04-16
Smart ai for Slasher.

YurGaZK:
Removed Halberd from list of skirmed units since it's faster than Slasher.
Trying to make Slasher Smart AI less cowardly so you could actually send Slashers on fight command and expect them to advance.


GoogleFrog:
I do not agree with this. Tactical AI should just make Slashers cowardly without attempting to implement particularly smart behaviour. Also the change seems to cause Slashers to do nothing vs Leveler.

PR #695, opened 2015-04-17
make Leviathan untransportable

### LINKED ISSUE #694: Exploit: Leviathan on land
It is possible to put the Leviathan sub on land. One procedure for this would be: build it in water, raise the bottom under it to lift it out of the water and have it picked up by a transport and placed on land. This is not bad, but seemingly not anticipated by the game logic. The Leviathan stays immune to damage from land based units while on land. Players already exploit this to fire tactical nukes at enemy bases.

sprunk:
fixes #694

### DIFF
--- a/units/cornukesub.lua
+++ b/units/cornukesub.lua
@@ -15,6 +15,7 @@ unitDef = {
   canMove                = true,
   canPatrol              = true,
   canstop                = [[1]],
+  cantBeTransported      = true,
   category               = [[SUB FIREPROOF]],
   collisionVolumeOffsets = [[0 -5 0]],
   collisionVolumeScales  = [[30 25 110]],


saucyjack42:
How is that any different than using a transport?  It takes time, money,  and immobilizes the sub. You need a few workers to terra under the sub when it gets close to enemy land and that's it.

saucyjack42:
Is there a reason you can't use whatever code you use to determine what can hit amphibs and mod it to work on the leviathan?  And every other sub since we're on the subject.

Anarchid:
Setsfxoccupy solultion should be much better in the consistency department. Regular subs can still be transported and will still be unattackable on land just by making leviathan heavy.

sprunk:
Closing (not the proper way to fix the issue).

PR #697, opened 2015-04-17
Some fix to CBAI widget

xponen:
based on feedback in http://zero-k.info/Forum/Thread/13058
CBA's cons slightly prefer assist nanoframe > 50% complete
CBA's cons don't remove queue in enemy territory if user is watching
CBA's cons start build ASAP while user is queuing instead of waiting for user

some fix to cancel command behaviour.

Note: this widget allow user to place jobs and let cons decide which job it like.


xponen:
i just make PR because it look pretty

PR #699, opened 2015-04-18
Remove sub 95% damage resistance to ground weapons

sprunk:


GoogleFrog:
Subs are too close to the surface of the water. AoE would deal significant damage.

Anarchid:
Is the option where they lose armor when projected onto dry land by wyvern/terraform/valkyrie an acceptable alternative?

GoogleFrog:
Yes, explain how. UnitPreDamaged?

Anarchid:
I'd go with SetSfxOccupy instead.

sprunk:
Closing (will submit a proper solution shortly)

PR #707, opened 2015-04-19
Concussion Shell buff

Sgathec:
The commander weapon Concussion Shell now applies a lot more impulse to target units and damage has been reduced considerably. Consequently most damage is done by collisions except for in the case of very heavy units (tests revealed Reapers were only slightly affected by the impulse). Area of effect has been increased by 33% and reload time decreased by 25%. Numerous tests have been conducted in order to determine balanced values. Additionally, the weapon is now able to smooth ground. A new CEG has been devised to signify this. 

The weapon has also been converted into a MissileLauncher in order to reduce overshooting. It's still functionally and visually a cannon. A flame trail CEG was written for better looks. Range had to be reduced by 31% because the weapon outranged defenses too easily with the proper modules. Additionally, it has gained 50% more velocity because that looked more convincing. Testing showed that didn't have any negative side effects.


GoogleFrog:
Doesn't this make it situationally more powerful?

Sgathec:
Yes, I suppose hills make using it more worthwhile.

Skasi:
Soo more impulse means it's better against raiders and worse against structures/heavies. -25% velocity means its worse against raiders and better against structures/heavies, duh. Reminds me of old dodgeable raven projectiles: "Go micro this unit because I tell you to!"

Sgathec:
It was 100% more velocity at first. I reduced it by 25% afterwards.

Sgathec:
I have to invalidate certain tests due to wrong commander HP values. Please disregard the things I said about which units defeat commanders equipped with the Concussion Shell.

Sgathec:
The PR description has been updated to reflect the current state of the weapon. The earlier versions of the description should be disregarded.

GoogleFrog:
I don't think it needs a buff.

Sgathec:
How about I make this version of the weapon equal to the current Concussion Shell in power? My version is possibly cooler so players might use it more.

Issue #704, opened 2015-04-18
Upsize cursors to power of 2


GoogleFrog:
All cursor images need their size increased to their next largest power of 2. This is because some graphics cards cannot deal with anything else.

The upscale should simply add a transparent border to the cursors, do not rescale. Also pay attention to the hotspot of each cursor.

jk3064:
engine already handles that

xponen:
@GoogleFrog how do we know its the power-of-2 problem? 
I found some script for Window's Powershell and ImageMagick program that could possibly automate padding all files into a square of power-of-2, but how sure its needed? 

useful script:
http://viziblr.com/news/2012/1/11/recursive-batch-image-conversion-with-powershell-and-imagema.html
http://www.fmwconcepts.com/imagemagick/tidbits/image.php#power2_pad

sprunk:
assigning to self

GoogleFrog:
jK has does the engine handle it in 98.0.1-451? We have reports of some cursors being invisible and the image sizes was a possible problem.

jk3064:
hasn't changed, code is here:
https://github.com/spring/spring/blob/develop/rts/Game/UI/HwMouseCursor.cpp#L311

if at all you changed something, e.g. changed cursor images to something too large (some gpu drivers only support 48x48, some 96x96, and very rare do 128x128)

xponen:
I checked all the cursor icon, all now have 32x32 or 64x64 as nearest power-of-2 size

Original size:
non-power-of-2 size for optional cursor set:
17x21
21x21
26x26
20x20
39x39
42x42

for zk set:
24x24
27x27
40x40
48x36
48x48

xponen:
all cursor icon have been upsized to nearest pow-of-2

PR #714, opened 2015-04-21
Convert the rest of cursor to nearest pow-of-2

xponen:
Resize masse,erom,lathan, and ca_static cursor set to nearest power-of-2 size
Added automation script (for Windows) for future reference!

related to https://github.com/ZeroK-RTS/Zero-K/issues/704
so far there's no image greater than 64x64 and majority is 32x32


sprunk:
These new images do not obey hotspots properly.

xponen:
@sprunk what is the correct hotspot?
I will try to make the conversion properly.

sprunk:
Also, some of them had a lot of empty border space and could safely be downsized to 32 (instead of upsized to 64). Remember that cards where >32 is a problem exist.

xponen:
okay ty @sprunk  
I will find out how to script empty space trimmer

xponen:
tested each file for hotspot, transparency, aspect ratio after padding, now all okay.
although no resize to 32x32 because 32x32 or 64x64 seems to work for me.

PR #716, opened 2015-04-22
Added cruder backup outline effect for bump & dynamic underwater

dvr333:
Known issues: Masks XRayShader. Not sure how to force XRayShader to be executed after Outline. Alternative fix is to comment out rendering the units as normal above the outline lines, but that causes spurious internal lines underwater to show.


sprunk:
Layer can be used to control widget ordering.

dvr333:
Are lower layers in front of or behind other layers? 

Anyway, I did experiment a bit with layers, but it didn't seem to help. Part of that is that it don't want to put #613 in here accidentally, so testing will be tricky until that is merged.

dvr333:
In case this gets merged before I figure out how to avoid the aforementioned issues with XRay Shader, I've removed the lines that interfere with XRay Shader. This should have no widget interaction issues if merged, but there will be spurious internal outlines when underwater.

PR #718, opened 2015-04-23
Increased outline size for mex spots

dvr333:
* Also added outlines to minimap
* Increased size of mex spots slightly to match overdrive radius.

### DIFF
--- a/LuaUI/Widgets/cmd_mex_placement.lua
+++ b/LuaUI/Widgets/cmd_mex_placement.lua
@@ -598,14 +598,7 @@ function calcMainMexDrawList()
 			
 			glPushMatrix()
 			
-			glColor(0,0,0,0.7)
 			glDepthTest(false)
-			glLineWidth(spot.metal*3.3)
-			glDrawGroundCircle(x, 1, z, 50, 21)
-			glColor(mexColor)
-			glLineWidth(spot.metal*1.5)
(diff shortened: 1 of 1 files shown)


dvr333:
I'm curious how many people are unable to use shaders anymore? If everyone can, this code should probably be rewritten at some point to use them on a quad to make the circles. It would be a perfect circle, and it wouldn't look like a dotted line when thick and the camera is zoomed out.

Licho1:
It would also be nice to unify this with geospot marker and make it follow same convention/colors.

GoogleFrog:
Large mex visual:
 * Could overlap.
 * Obscures overdrive range. It is useful to see the exact range for linking.

If anything the visual should be a bit smaller to better match the model size.

GoogleFrog:
What was wrong with the old circle size and border thickness? The border looks weird as you zoom out and these mex spots are too small.

dvr333:
ikinz couldn't see it super well on Iceland, so I thickened it. However, because of how GL.LINE works, it looks like a dotted line when zoomed out, so I made the whole thing a bit thinner to make up for it, but then you suggested making it smaller to fit the mex model:
 
"If anything the visual should be a bit smaller to better match the model size."

I made it bigger because Sprung mentioned something about using it to show overdrive range for the mexes when placing power plants.

PR #726, opened 2015-04-27
Repair cursor: modify to fit into 32x32.

sprunk:
Replaces pulsating halo around the wrench (which didn't fit) with the wrench spinning.
image


GoogleFrog:
The outline of the spanner in your example gif looks pixelated. Does that occur ingame?

sprunk:
No, that's just due to GIF. The individual frames are smooth (eg. image )

PR #729, opened 2015-04-27
Submarine patch.

sprunk:
- Subs lose resistance to ground weapons when out of water.
- Subs are targetable by ground weapons when out of water.
- Subs no longer fireproof when out of water.
- Leviathan LUS.
- Updated Leviathan helptexts to reflect lack of antinuke.


GoogleFrog:
Do they lose resistance to AA weapons?
What if they are hit by a torpedo while above the water?

sprunk:
When on land, their armor class changes from SUBS to ELSE (which is everything but subs and planes). This means:
AA resistance is still there.
Torpedoes do the usual damage.

PR #741, opened 2015-05-01
Add global build command widget

mtroyka:
Please review/test

-Added global build command widget (and terraform support)
-Added callin code to gui_lasso_terraform so that other widgets can see the commands it gives and capture them

(note: it can't handle the build part of raise-and-build, but raise-and-build still works as long as it's given as a direct order)

### DIFF
--- a/LuaUI/Widgets/cmd_unit_mover.lua
+++ b/LuaUI/Widgets/cmd_unit_mover.lua
@@ -75,7 +75,7 @@ end
 function widget:UnitFinished(unitID, unitDefID, unitTeam)
 	if unitTeam == myTeamID then
 		local unitDef = UnitDefs[unitDefID]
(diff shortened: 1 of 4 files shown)


xponen:
I'll test gui_lasso_terraform changes

mtroyka:
I'm currently trying to see if I can't properly capture raise-and-build, but the way it's written in gui_lasso_terraform is difficult to deal with, and needs to have more compatibility code added. :|

GoogleFrog:
I don't see much purpose in including this but historically it is the sort of thing we do because we are nice.

mtroyka:
Alright, this should be ready to merge. There's nothing left to implement and I've tested it half to death.

GoogleFrog:
I'm not sure when or if I should merge this.

PR #743, opened 2015-05-02
Add global build command widget

mtroyka:
Ok I tried to fix compatibility with raise-and-build, but due to the unusual way that terraform works it turned out not to be practical. I reverted to a backup of gui_lasso_terraform and added some fixes so that the compatibility code now properly accounts for when groups of units are given a TF order. Everything should be in order now.


GoogleFrog:
Duplicate

GoogleFrog:
Duplicate

Issue #725, opened 2015-04-25
Missile range names overlap


dvr333:
The names of "Tacnuke and EMP" and "EMP AOE" overlap in the missile range display(missilesilo_range.lua). This should probably be fixed either by shifting the side one of them is on, or adding in a rotation element along the circle and setting up the elements along a curve.

xponen:
okay tested, it only happen when world is rotated exactly 90 degree,

dvr333:
Along which axis? Also, I'm pretty sure "world rotated precisely 90 degrees" implies being in an overview state, or similarly zoomed out, which is common enough that I'm still concerned. Regardless it's probably easy enough to polish, and polish is something we can always use more of.

xponen:
its along y-axis. 
I will try the following formula to make the text rotate with camera rotation 
```
if (cs.name == "free") then
  z = -(radius)*cos(cs.ry)
  x = radius*sin(cs.ry)
end
```

GoogleFrog:
I can only reproduce this by tilting the camera to look along the ground at an angle completely useless for playing. Provide a screenshot.

xponen:
won't happen to overhead cam because the cam can't rotate the world

xponen:
![silo range overlap text](https://cloud.githubusercontent.com/assets/2525365/7440882/76597d44-f100-11e4-8871-20047f02a1a1.jpg)

Anarchid:
Assuming this is fixed now.

PR #746, opened 2015-05-02
missilesilo_range: text follow COFC's cam rotation

xponen:
-Silo range text able to follow COFC rotation, fix https://github.com/ZeroK-RTS/Zero-K/issues/725
-Text size tweak. Is the pic below look good? (pls click to expand)
silo range zoom

For comparison, the original:
silo range ori

The aim is to help avoid squinting to see what the text at the outer circle wrote when at full zoom-out.
The catch is it isn't a uniform size...


dvr333:
Looks like a good fix.

xponen:
okay will test later, resting

Issue #720, opened 2015-04-24
Jump commander teleports over map twice


Skasi:
Upon landing after a jump order a jump commander teleported towards the other side of the map, getting stuck on a wind gen. After trying to move for a short moment it teleports again. **No Djinn was used** (because none were built).

Replay: http://zero-k.info/Battles/Detail/346786
Watch Coc_fr's commander slightly east of map center at 7:35 - 7:45 game time.

Attached is a screenshot I made shortly after it happened in a demo-replay. It includes a list of the commanders modules. I noticed that Coc_fr has been lagging almost the whole game (and games prior). Just pointing that out.

![teleporting commander bug](https://cloud.githubusercontent.com/assets/132906/7323552/a63aaf08-eaaf-11e4-9f7f-73f533347330.png)

xponen:
Debugged that jumpy com. Found a triggers: morph.
I made a video because its really useful https://youtu.be/QdXHMfr3FfA

Anarchid:
Reported fixed.

PR #749, opened 2015-05-04
Jump improvements

sprunk:
- jump cooldown now obeys status effects (previously healthbar did but actual cooldown did not).
- jump cooldown is remembered through morph.
- morph orders are no longer blocked from being queued mid-jump.
- morph will no longer finalize on a mid-jump unit (will instead wait at 100% for it to land). This increases maintainability of the gadget by simplifying logic (removes a lot of workarounds for mid-air replacement) and might increase safety against bugs that stem from weird interaction between morph and jump (as it's usually Recons undergoing morph who are involved).


xponen:
wow, really simple solution, just finalize morph at landing!

it might solve https://github.com/ZeroK-RTS/Zero-K/issues/720 but haven't confirm what cause it (it can be command not removed in time, for example).

+1 vote to merge after tested this PRy.

PR #750, opened 2015-05-04
replay_controls.lua: add easy to use fast-forward function to help debug stuff

xponen:
Really simple but neat feature, just click at any time to fast forward to, the game will be paused when you got back to it. 
replay bar update

### DIFF
--- a/LuaUI/Widgets/gui_replay_controls.lua
+++ b/LuaUI/Widgets/gui_replay_controls.lua
@@ -123,6 +123,10 @@ function CreateTheUI(showProgress)
 					if x>progress_speed.x and y>progress_speed.y
 					and x<progress_speed.x2 and  y<progress_speed.y2
 					then
+						progress_speed.color = {0.9,0.15,0.2,0.75}
+						progress_speed:SetValue(progress_speed.max)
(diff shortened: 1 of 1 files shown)


GoogleFrog:
What is the weird grey arrow?

Add a tooltip to the bar which tells you the time that will be jumped to. Make it clear that you cannot jump backwards.

Make an option to hide the bar because it is possibly a spoiler to know how long a game is.

sprunk:
The site already spoils duration.
Presumably, arrow is cursor.
Tooltip for the time that will be jumped to seems to exist (see 3rd and 4th picture).

xponen:
@GoogleFrog IMO is good enough, 

IMO it isn't much spoiler because
game minute won't show up if we didn't hover over the bar, and there's time in pre-game to drag the window offscreen.

if that don't work, then we can fix.

GoogleFrog:
You should never have to drag something offscreen. All this needs is a toggle in the menu to hide/unhide the bar.

xponen:
Ok
have made bar hidden (disabled) by default. Also work better than before.

To activate progress bar, double click it. Visual feedback is added to explain its clickable.

PR #757, opened 2015-05-08
Remaining fire damage for burning units is now readable by widgets

sprunk:
through the 'on_fire' unit rules param (previously this param was boolean whether the unit is burning).

Fire burning DPS is now readable through the 'unitsOnFire' game rules param (previously this param was boolean whether units can catch fire).

### DIFF
--- a/LuaRules/Gadgets/91_unit_attributes.lua
+++ b/LuaRules/Gadgets/91_unit_attributes.lua
@@ -357,7 +357,7 @@ local function UpdateUnitAttributes(unitID, frame)
 		end
 	end
 	
-	local cloakBlocked = (spGetUnitRulesParam(unitID,"on_fire") == 1) or (disarmed == 1)
+	local cloakBlocked = (spGetUnitRulesParam(unitID,"on_fire") > 0) or (disarmed == 1)
 	if cloakBlocked then
 		changedAtt = true
 		if not unitCannotCloak[unitID] then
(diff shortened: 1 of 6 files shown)


GoogleFrog:
Why?

GoogleFrog:
Why?

sprunk:
It is already partially available (you can poll for on_fire, and check which fire weapon hit that frame to get the duration and thus the damage). I don't think giving the exact value would allow for any spechax. On the other hand, being able to see the remaining duration and how healthy a unit is going to be after the fire is useful information. For example knowing which units are going to die allows the fire user to switch away from the doomed units or the burning user to choose which units to repair. The main use case I had in mind was to flash the burning part of the healthbar.

GoogleFrog:
You can't get the duration because it is non-deterministic.

sprunk:
That's why currently it's available only partially (you do know the min/max value for duration; though, the player cannot know when the duration gets reapplied). 

What is the concern? Is it bad (and if so: why) to make these nondeterministic rolls known when they happen instead of only being able to infer what they were after their effects fully resolve?

GoogleFrog:
Why should you know which units are going to survive?

maackey:
If you knew which units were going to be burned alive, they could have a special "I'm melting!" death animation.

sprunk:
Death animations already know that. The player doesn't.

There isn't any real reason why the player should know that but I don't think there is any reason why he shouldn't, either. Most other random processes seem to be unavailable to the player though.

sprunk:
Closing: the fire random rolls aren't supposed to be revealed.

PR #760, opened 2015-05-11
New Sea Update Alpha Version

RyMarq:
Current Issues:

Kraken requires wreckage model.

Kraken skirmishing AI needs to exist. Halted based on next issue.

Kraken range update in water non-functional.


Licho1:
Kraken!

sprunk:
Perhaps it would be a better idea to make the PR into the balance branch instead?

GoogleFrog:
I'm not merging this as is for reasons discussed. Try the underwater riot Scallop change.

PR #775, opened 2015-05-16
Remove send_enemy_death.lua

sprunk:
Redundant with the newest engine (widgets now have access to relevant events).


lhog:
Doesn't it break 91.0 compatibility? If so, is this so necessary to do on purpose?

sprunk:
No, the gadget has been disabled for a long time, 91 already had the info available.

lhog:
+1 for cleanup then

PR #777, opened 2015-05-17
Anti overkill prevention improvements

lhog:
1. Added Banisher
2. Made AOKP algorithm respect shielded units

### DIFF
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -74,7 +74,6 @@ local HandledUnitDefIDs = {
 	[UnitDefNames["corcrash"].id] = true,
 	[UnitDefNames["cormist"].id] = true,
 	[UnitDefNames["tawf114"].id] = true, --HT's banisher	
-	[UnitDefNames["shieldarty"].id] = true, --Shields' racketeer
 }
 
 include("LuaRules/Configs/customcmds.h.lua")
@@ -153,7 +152,19 @@ local function GetTargetShieldPower(unitID, targetID, timeout)
 	return totalShieldPower
 end
 
-local function OKP_CheckBlock_Regular(unitID, targetID, damage, timeout)
+function GG.OverkillPrevention_CheckBlock(unitID, targetID, damage, timeout, troubleVsFast)
+	if not units[unitID] then
+		return false
+	end
+
+	if spValidUnitID(unitID) and spValidUnitID(targetID) then
+		if troubleVsFast then
+			local unitDefID = Spring.GetUnitDefID(targetID)
+			if fastUnitDefs[unitDefID] then
+				damage = 0
+			end
+		end
+		
 		local incData = incomingDamage[targetID]
 		local addDamage = true
 		
@@ -177,7 +188,7 @@ local function OKP_CheckBlock_Regular(unitID, targetID, damage, timeout)
 						local teamID = spGetUnitTeam(unitID)
 						local unitDefID = CallAsTeam(teamID, spGetUnitDefID, targetID)
 						if unitDefID then
-							spSetUnitTarget(unitID, 0)
+							spSetUnitTarget(unitID,0)
 							return true
 						end
 					end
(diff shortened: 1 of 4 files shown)


GoogleFrog:
Remember, OverkillPrevention_CheckBlock can be called a lot. Every unit in a battle could potentially have its shot blocked every frame. This means you should avoid slow calls and cache where you can.

Do not do string operations within that code. Pass a bunch of booleans to the function or, if you must, use constants.

Stuff like Spring.GetUnitsInCylinder is slow. Also shields should not be handled for reasons unrelated to performance. Units are able to move in and out of shields within the time it would take for a guided missile to hit. The only problem with the current version of overkill prevention is when you are fighting a much smaller army of shielded units. In this case there is barely a problem but if you want to improve the behaviour make overkill prevention check the shield strength of the target. That will ensure that enough missile are fired.

lhog:
```
Remember, OverkillPrevention_CheckBlock can be called a lot. Every unit in a battle could potentially have its shot blocked every frame. This means you should avoid slow calls and cache where you can.
Do not do string operations within that code. Pass a bunch of booleans to the function or, if you must, use constants.
```
Agree, I'm likely to do different OverkillPrevention_CheckBlock flavors for Regular and Disarming damage types as the current signature doesn't fit what I need anyway. So disregard the current code in that part.

```
Stuff like Spring.GetUnitsInCylinder is slow. Also shields should not be handled for reasons unrelated to performance. Units are able to move in and out of shields within the time it would take for a guided missile to hit. The only problem with the current version of overkill prevention is when you are fighting a much smaller army of shielded units. In this case there is barely a problem but if you want to improve the behaviour make overkill prevention check the shield strength of the target. That will ensure that enough missile are fired.
```
* Stuff like Spring.GetUnitsInCylinder is slow. <-- Agree, it could be slow. Any idea how to make large scale stress testing?
* Units are able to move in and out of shields within the time it would take for a guided missile to hit. <-- this one is actually addressed by the code.
* The only problem with the current version of overkill prevention is when you are fighting a much smaller army of shielded units. In this case there is barely a problem but if you want to improve the behaviour make overkill prevention check the shield strength of the target. That will ensure that enough missile are fired. <-- In the current implementation OKP eligible units shoot target like if there were no shield(s) present. This leads to the problem that high alpha units like scalpels or banishers shoot missiles sequentially instead of parallel. My implementation solves this issue completely. Units shoot whatever number of missiles to drain shields and kill the target. I've tested it locally and observe perfect match.

lhog:
Should be more or less ready for performance and bug testing.

I've decided that racketeer OKP needs another branch, since disarming damage addition into the current framework has proven to be more complex than I expected

GoogleFrog:
*In the current implementation OKP eligible units shoot target like if there were no shield(s) present. This leads to the problem that high alpha units like scalpels or banishers shoot missiles sequentially instead of parallel. My implementation solves this issue completely. Units shoot whatever number of missiles to drain shields and kill the target. I've tested it locally and observe perfect match.*

They only shoot sequentially when there are few targets compared to the number of Scalpels. This is not really a problem but, as I said, you could fix it by taking the personal shield value of the target into account.

I have never seen overkill prevention cause a problem against shielded armies. This is because there are many targets in the army so the Scalpels will always have something to fire at. A player would have to be quite observant to notice an edge case and if they do so then they should be competent enough to decide to turn off overkill prevention for the battle.

My solution to Spring.GetUnitsInCylinder is to not do anything which requires it.

lhog:
I did benchmark myself:

100 core_spectre team 0
100 hoverassault team 0
200 nsaclash team 1

http://i.snag.gy/1VJtJ.jpg

Scalpels put on holdfire.
0. Wait till shields are fully charged. CPU load ~14-15%
1. Scalpels put on holdfire.
2. Disabled OKP for scalpels. Put fire at will on for scalpels. Made 3 measurements
a. CPU load jumps up to 37%
b. CPU load jumps up to 34%
b. CPU load jumps up to 31%
3. Pressed stop, put on holdfire for scalpels.
4. Wait till shields are fully charged. 
5. Enabled OKP (mine version) for scalpels. Put fire at will on. Made 3 measurements
a. CPU load jumps up to 31%
b. CPU load jumps up to 39%
b. CPU load jumps up to 33%

CPU stays 20-24% loaded during shields recharge/charge donations.
Relevant function is triggered 400 times in a very short period of time.

From that I can draw conclusion that the improvement doesn't cause extra significant CPU load. In fact, other gadgets like shields charge donations (presumably) cause magnitude bigger CPU load.

Feel free to propose better way of benchmarking.

GoogleFrog:
I am not sure how to do a good profile of the performance difference. The worrying part of the gadget is a GG function called from within the LUS of the affected units. I do not think that area is tracked by the gadget profiler. Your data just looks like noise.

You have yet to convince me that it is worthwhile to take all nearby shields into account.

lhog:
Data is indeed not accurate, but you and I have no better way to measure it. The measurement I've done is quite large-scale, and the toll is around 1-3% CPU time. Shouldn't be a cornerstone.

Improvement automate OKP behavior when shields are present and guarantee precision (no underkill) as long as projectiles are precise.

Without taking shields into account it would be almost impossible to make good racketeer OKP. Since there is no other way to tell if next racketeer projectile is going to disarm target.

If you are fine with current derpy way racketeers auto-target things, so be it. Accept my bugfixes and remove GetTargetShieldPower and references to it.

GoogleFrog:
The data we have here is not good. These callins are probably not covered by the current gadget profiler. I probably could develop a way to measure the impact but that would cost work. However, in my experience GetUnitsInCylinder is a slow function so it is safe to be wary of it.

The PR has a new problem pointed out by Sprung. A Flea (or really any small unit) near a bunch of shields can cause a lot of units to waste their shots because they think it is about to run inside a shield. Can you give an example where ignoring non-self shields has caused a problem in a real game? A heavy unit with trailing shields sounds like a likely case but nobody has complained. Perhaps if people attempt to exploit OKP in this way then their opponents should notice and disable it for the encounter.

If ignoring shields proves to be a problem perhaps there is some other way to fix it. For example you could poll every large shield every second and mark the nearby units as being under a shield. OKP would then become increasingly lenient for units under shields. This solution only has one area check per second per large shield which is an acceptable solution.

Also I should note that your implementation of shield avoiding seems to have a few problems. Firstly incoming damage is the damage of the burst, not individual projectiles. So a Scalpel will fire at a 500 health shield and expect both projectiles to penetrate when in fact only one will get through. Secondly you either assume all shields are penetrated or none of the shields are penetrated. This may cause a target to be shot at many more times than required.

*If you are fine with current derpy way racketeers auto-target things, so be it. Accept my bugfixes and remove GetTargetShieldPower and references to it.*
False dichotomy, Racketeer can have decent OKP without considering shields. Also you can change your PR after creating it so the relevant bits should be removed here.

lhog:
I disagree with everything you wrote, but see no point in further discussion.
You can merge now.

GoogleFrog:
You're not going to fix the more common edge cases by making the gadget take shields into account? Why would you not add Racketeer OKP?

GoogleFrog:
You could use my suggested area shield heuristic to fix the large unit + many area shields case.

lhog:
GetUnitsInCylinder is not that slow nowadays. Whole code was noticeably slow in 91.0, but not that bad in 98+. Privately I tried to put pacing/caching code in place instead of GetUnitsInCylinder and the code has become only slower.
```
For example you could poll every large shield every second and mark the nearby units as being under a shield
```
This is going to be worse than the implementation I had. It would require GetUnitsIn... calls every second for every shield, while
1) RoF for most of the OKP-eligable weapons is more than 1 second 
2) Alternative (mine) doesn't require polling of every shield but only shields in the relatively small area

Sprung's exploit was more a theoretically possible thing. Obviously it didn't have 100% right resolution, but I could have mitigated it by introducing few heuristics, taking into account target metal cost, overkill factor, target unit position or something else.

The bug you mentioned was known for me, but fixup required introduction of additional parameter (# of projectiles per shot) to the function call. I didn't do it for reason in the expectation of denial of my "shielded" proposal.

I'm not going to add Racketeer OKP. As I wrote before few times it requires taking shields into account. You said my dichotomy was false. Ok, go ahead and let's see what kind of OKP implementation you can come up with without taking shields into account. I'd be more than happy to test it and point you possible bugs/exploit, same way you did a favor to me.

GoogleFrog:
"1) RoF for most of the OKP-eligable weapons is more than 1 second "

If a weapon is loaded but the shot is being blocked the check is made once a frame. It would be interesting to see whether GetUnitsInCylinder is significantly faster now. If it is fast then these checks could be fine.

lhog:
So far my attempts to make former code faster ended up so it became only slower. I may try the approach you've proposed, if/when I have passion. I think it would be done in different branch, so current one is ready for merge.

PR #783, opened 2015-05-21
Fix quick Raven dives

sprunk:
Raven cannot actually dive in water.

### DIFF
--- a/scripts/corshad.lua
+++ b/scripts/corshad.lua
@@ -91,7 +91,7 @@ local function SpeedControl()
 	SetSignalMask(SIG_SPEED_CONTROL)
 	while true do
 		local x,y,z = spGetUnitPosition(unitID)
-		local terrain = math.max(spGetGroundHeight(x,z), 0)
+		local terrain = max(spGetGroundHeight(x,z), 0) -- not amphibious, treat water as ground
 		local speedMult = minSpeedMult + (1-minSpeedMult)*max(0, min(1, (y - terrain-50)/(fullHeight-60)))
 		Spring.SetUnitRulesParam(unitID, "selfMoveSpeedChange", speedMult)
 		GG.UpdateUnitAttributes(unitID)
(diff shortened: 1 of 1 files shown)


Skasi:
To explain the issue: Currently Raven diving to bomb (under)water targets do not suffer from the speed penalty if the water is deep enough.

xponen:
this commit remove 1 behaviour. The equation ( https://github.com/ZeroK-RTS/Zero-K/pull/783/files#diff-b2e9386a6b58fee2e3448bc193cb6bdfR95 ) did lots of things at once.

Specifically the one Skasi mentioned:
corshad special behaviour

Skasi:
Oh right maybe I should actually read through code before commenting. Anyway, here's the line this is about:

    local speedMult = minSpeedMult + (1-minSpeedMult)*max(0, min(1, (y - terrain-50)/(fullHeight-60))) 

Unless I'm drunk a terrain height of -200 could lead to problems in the min() part:

    min(1, (y - (-200) -50)/(fullHeight-60))) 

The double - turns it into y + 200 - 50, assuming y (current bomber flight height) is 22 that's 22 + 200 - 50 = 172, insert into code:

    min(1, (172)/(fullHeight-60))) 

fullHeight is probably somewhere around 200, so that leads to:

    min(1, (172)/(140)))

Because 172 is > 140, min() picks 1, insert 1 into above code:

    local speedMult = minSpeedMult + (1-minSpeedMult)*max(0, 1) 
->
    local speedMult = minSpeedMult + (1-minSpeedMult)*1
->
    local speedMult = minSpeedMult + (1-minSpeedMult)
->
    local speedMult = 1

And this might be why Raven currently doesn't lose speed. (unless I made a mistake)

xponen:
Is Raven not loosing speed desired or not? who made those equation?

Skasi:
To clarify, right now Raven does not appear to lose speed when diving against water-targets (unless water is shallow). After the commit it will (unless I'm wrong) lose speed.

I think it **is meant to** lose speed in this situation, that's probably what people expect when bombing ships and hovers on sea.

Issue #544, opened 2015-01-28
Improve start position shuffle.


GoogleFrog:
Start position shuffle should shuffle start positions before the game starts. People should be placing their command in the start box they will actually start in.

sprunk:
won't happen, licho said it's perfection not worth the cost

Licho1:
If you can do it without breaking other stuff, do it. But any additional cost is up to you (PW, QM, whatever this breaks, because I don't know).
Testing environment is your friend

Licho1:
In other words, springie needs some control over boxes and it needs ability to disable shuffle and set them as before.

sprunk:
The infra-side implementation was supposed to just change how startboxes are put into spring start-script (eventually: into modoption instead of the engine startbox fields -- the one i added didn't remove engine boxes because thats not yet ready game-side).

Springie would still have full control over how the boxes are set up and control how the shuffle behaves through another modoption, so nothing is affected.

Licho1:
In that case you probably made erorr in format string did you test on local or test springie?
Deploying blind changes in key infra is really bad idea.

sprunk:
It compiled so i assumed everything worked

Licho1:
that equals to blind test...

2015-05-20 10:10 GMT+02:00 sprunk <notifications@github.com>:

xponen:
@sprunk I discover this can be done without gadget changes. 

We shuffle the startbox during startscript creation, suddenly the startbox ingame is randomized and didn't match the ZKL screen.

sprunk:
No, that wouldn't do anything. You would still know where everybody starts except you'd know that from Spring.GetAllyTeamStartBox instead of the lobby.

xponen:
@sprunk we can attach some tag to the startscript that tell ZK to hide player position by overriding Spring.GetAllyTeamStartBox() like this: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Utilities/weapon_index_compat_91.lua

sprunk:
Perhaps that would work. Want to implement that?

xponen:
Turn out there was much help! the override for Unsynced was separate from Synced, like here:
https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/main.lua

Is check first whether user can override this file to undo the override,

xponen:
@sprunk yes, I want to try. will get back to you if it is a fail

sprunk:
[...]

PR #784, opened 2015-05-22
Startbox reimplementation (backwards compatible)

sprunk:
* removes current shuffle and replaces with startbox shuffle, with 3 options (off / shuffle / all-shuffle)
* normal shuffle shuffles among boxes that would be used, all-shuffle among all boxes
* backwards compatible (no behavior change until Springie support exists)


xponen:
what is this, 
what I mention in https://github.com/ZeroK-RTS/Zero-K/issues/544 don't need any changes at all in gadget nor widget.

sprunk:
except 1 override

and except changing widgets to use something else than the call-out because you still need to know the original boxes
and except implementing the shuffle in the infra anyway
and except not being able to get extended with more power (like personal boxes, or arbitrary/disjoint shapes) if we wanted to
and except this is already implemented

xponen:
Widget & gadget override were separate, I don't need to change gadget because I'll just override widget calls.

xponen:
you really want to implement split starbox?

sprunk:
i could if we wanted that (and most likely will at some point because it allows some interesting map designs); the point is, it is actually possible to modify my implementation to allow that while engine boxes will always be a single rectangle

GoogleFrog:
Why do you need a Springie change. Lua can already read the start boxes so just use those instead of your modoption.

sprunk:
Start boxes as currently provided by the engine cannot be overridden. Engine also clamps all start positions to them. If your box gets shuffled somewhere else you won't be able to place because engine will still clamp your statr position to the original box.

This means engine boxes need to go and be made map-wide. That in turn means the info is no longer available and has to be provided through a modoption. The engine API allows us to define our own boxes for the picker through AllowStartPosition but that still doesn't prevent clamp which makes the modoption needed. The modoption is very cheap infra-side because all it does is copy the startbox data.

There are two alternatives:

1) It would be possible to discard the engine position picker and reimplement the whole picking mechanism in Lua but I think it's better to use the provided engine API.

2) xponen's version would have shuffling moved fully to infra and the game-side API modified to prevent reading them. It isn't really any cheaper, doesn't allow extension and causes fragmentation between infra and game. There is also no working version so far.

GoogleFrog:
This sounds reasonable. Just another check, with your version here is it possible to make a raw startscript which implements each settings? With xponen's version it sounds like that is impossible.

Another thing to check; are the enemy start boxes drawn correctly? By this I mean that if set to "off" or "shuffle" then you should only draw the boxes which are known to be occupied. Of course if set to "shuffle-all" all boxes should be drawn. This reminds people of the possible start locations.

This looks hard to test locally so I have to ask; have you tested this?

sprunk:
The start-script difference is that instead of the engine startbox fields, the modoption with the boxes listed exists. If this does not exist everything uses legacy engine boxes.

So far all boxes are drawn (because that's how it works with current boxes). I will update it to only use the known-occupied boxes.

I have tested it (including CAI and bots), and minor issues exist:
* non-CAI bots are forced to the middle of the startbox. The picker accepts a playerID and bots don't have those so they have to use the default. I don't think any bot actually does any custom placing though and CAI is unaffected and starts correctly (including multiple CAIs spreading evenly over their box).
* clicking outside the box currently places the start at the edge of the box. After the patch invalid start positions are dropped instead of being clamped. I think it's reasonable as its side effect is preventing accidential clicks that put people in weird places.

sprunk:
Now only occupied boxes are drawn (of course unless all-shuffle is used).

Further testing with weird box setups has shown that pathological cases might be handled poorly:
* if there are more teams than boxes, all-shuffle always shuffles them among the teams with lower IDs.
* the code assumes Springie box numbering (1-based) so any custom 0-based autohosts/lobbies will get wrong boxes.
* if a box is missing (eg. box 5 exists but 4 doesn't) wrong boxes might get shuffled.
Are these worth fixing?

xponen:
this, IMO is a hax. Shuffling startbox in Springie use significantly less code than this (as I already demonstrated).

The only reason to want this PR is if octagonal startboxes or splited startbox is required (as a feature), but other than that its just wasting CPU cycle.

GoogleFrog:
Sprung those issues can be fixed by saying that the required-but-missing boxes span the entire map. If there are more teams than boxes then fill in the remaining with entire map boxes.

I think it is bad to support 0-based lists because standards that allow for too many things are easy to misunderstand/misuse. Regardless, many things are 1-indexed in lua.

sprunk:
Springie's 1-based ordering is actually wrong because the engine is 0-based. Anything that implements the proper ordering would then get messed up boxes. Should the support get reverted? It's a 1-liner.

If there are more teams than boxes but no team is skipped then everything works correctly (lacking box == global box).

Spring messes the allyteam numbers up if there are gaps teams with neither a box nor people so wrong boxes get used (it never allows gaps so it shifts the subsequent teams). A box with no people is fine, as are people without a box (they get a global box).
It should be possible to detect that situation but the question is whether such situations actually exist. I can't think of any example where team N is supposed to exist (with either a box or people) but N-1 is supposed to be completely skipped (neither box nor people).
Of course it would be good if this was detected and dealt with but it's a pathological case that will only appear if someone on a custom host doesn't know what they're doing and never for proper games.

GoogleFrog:
There is no 'actually wrong'. If Springie sends boxes starting with 1 then that is the standard. I suggest that it be impossible to implement anything which is inconsistent with that standard.

Springie should define a springBox for every allyTeam which is able to exist. Does this fix the box skipping problem?

It should be possible for players in a custom host to play a game between allyTeams starting in box 3 and box 7 on an 8 way FFA map. Do not treat it as a rare pathological case. Do you know the design of the custom battle room and what sort of things would be expected?

sprunk:
>Springie should define a springBox for every allyTeam which is able to exist. Does this fix the box skipping problem?

Yep.

>It should be possible for players in a custom host to play a game between allyTeams starting in box 3 and box 7 on an 8 way FFA map.

It's okay to use teams 3 and 7 as long as boxes 1-8 exist. There would have to be a missing box, eg. it would be a problem if eg. box 5 didn't exist but boxes 6 7 8 did. Assuming the map is 8-way all 8 boxes should be set.

As far as I know the new design always gives people correctly set boxes so the pathological case is not supposed to happen.

sprunk:
I updated the infra-side PR to fix the missing boxes issue (by filling the gaps with fake teams). Everything should work correctly now.

xponen:
this won't work

sprunk:
Obviously it won't work if the infra support PR got rejected

PR #794, opened 2015-05-24
More interesting award names

sprunk:
Because calling the capture award "Capture Award" is the epitome of boring.


KingRaptor:
"Apocalypse Bringer" is more... appropriate? Thematic? Something like that - but it also loses the alliteration.

GoogleFrog:
"Dominator" or "Puppetmaster"

"Master Necromancer"

sprunk:
Dominator is a bit broad and so far all the names are multiple words.
Master Necromancer sounds alright but that would be the 3rd thing called "master" already (after puppet and commander).

GoogleFrog:
"Arch Licho"?

Orfelius:
more like "Arch Shaman" :P

KingRaptor:
BTW, is there any particular reason it's "Master of Puppets" instead of just *HoN announcer voice* Puppet Master ?

sprunk:
Mostly because metallica but I don't mind Puppetmaster.

PR #796, opened 2015-05-24
Fix #557 (laggers hogging metal)

### LINKED ISSUE #557: Players that have surrendered still receive metal from their team's mexes
e.g.: http://zero-k.info/Battles/Detail/328767

When SpaceAids surrenders after losing his base, THOL's metal income is cut in half until SpaceAid's metal storage of 500 is full. 
This effectively takes 500 metal from a team for each player that has dropped out on that team.

Possible fix: Set the maximum metal storage to 0 for dead players, thus stopping metal leech and giving back any of the 500 metal currently stored.

sprunk:


GoogleFrog:
Where do you reset the share level to what it was?

sprunk:
There is no GetTeamShareLevel so I didn't think that was possible, but it turns out some other call-out provides that. Fixed.

GoogleFrog:
Was this tested?

sprunk:
No (because lagging and the AFK monitor requires mutiplayer to work)

GoogleFrog:
You can test components manually.

For example make a team toggle whether it is afk whenever a shield that it owns is hit. Then make a game with a few AIs on a team and try toggling their afk state to test the changes.

sprunk:
Tested and fixed some more outstanding issues with the PR and the gadget:
* fixed anti-hogging not happening if the lagger had 0 units
* fixed the afk time threshold being larger than nominal
* fixed the case where if the 2 best players on the team are afk their units would bounce between them instead of going to the 3rd guy

PR #798, opened 2015-05-26
Remove innate con metal incomes

sprunk:


GoogleFrog:
I see no point to this. You just slightly change game pacing by making everything effectively more expensive. I would not mess with the pacing at this point.
1. This is not what people find confusing about the economy. At that level you barely notice and it just acts as something which makes the economy go more smoothly.
2. I don't mind if we have a 12 minute return metal maker.

sprunk:
I think the pacing change would be barely tangible. It would approximately reverse the slowupdate rate change and that one didn't seem to have much impact at all.

Economy confusion is probably not a large issue but it would be better if units didn't randomly produce minuscule amounts of resources nevertheless.

A part of the metalmaking problem is that this income is wholly private unlike other sources. Perhaps that's a social issue though.

Anarchid:
I'm pretty sure innate non-commander unit income accounts for less than 1/16th of one's economy until caretaker spam occurs.

Orfelius:
It is a minimal change to pacing while it also removes resource production inconsistency. This is one of the things that people find confusing in the economy. The other 2 things being lack of information of what exactly spends or requires energy and the different mex values per maps (such as 1.4 or 2.2).

CarabusX:
I am against this.
It would remove the FFA comeback mechanism, which is there since OTA, and is pretty important part of FFA gameplay. If you played more FFAs you would agree with it.

The only real problem I see with innate incomes is the new caretaker spam meta. But keep in mind it is efficient only in large teamgames, and can be fixed in other ways without hurting FFA. For example by making innate incomes be subject to communism. This way making caretaker wouldn't give more personal benefits than making a fusion or singu, so investing into OD would again become the main way of ecoing in these games.

As for resource production inconsistency, I would call it variation. If every map had same amount of mexes with same metal output it would be just boring and repetive.

sprunk:
FFA rebuild is still possible without incomes.
* reclaim. If you died then there should be a ton of wreckage lying around.
* reserve. If you see an opportunity to rebuild before dying then you can set reserve to make sure you have enough to rebuild.

Admittedly I don't play FFA so I don't know how often that happens and how impactful it is but I think these two mechanisms already provide ample opportunity for comeback. You won't be able sit in the corner and wait 40 minutes to afford a fac but I don't really think that's an issue.

Anarchid:
In practice the "comebacks" people do from a single con (or Athena) are typically about as extensive as "i just got a factory back and it got starlol'd".

It's still fun to do this, but i don't think removing con income will hurt it that bad; reclaim still exists, etc.

Orfelius:
Oh look I am going to rebuild so easily with this 0.4 energy! Surely my 2 cons will be able to win back me a game in this one corner of the map. FFA comebacks don't happen at the stage where con income matters. Because every other enemy already has at least 30 times better eco so there is almost no way that one can get back into the game and win it. Thus I don't think that FFA would be hurt at all actually.

Rafal variation is one thing and consistency is another. One still would have variation in mexes if they would be 2m, 1m, 1.5m, 3m, 4m and sometimes 5m. On most maps it is at least stable income per every mex type like there are 3.2 everywhere on Red Comet. But some maps like Tombstone Desert have slightly different mex incomes 2.13, 2.35, 2.24 etc. This is just annoying "variation".

Licho1:
You could also have bas per player income of 1/1 for example to allow FFA
rebound.

2015-06-09 8:30 GMT+02:00 Orfelius <notifications@github.com>:

sprunk:
Passive income is bad because it has absolutely no interaction. It's not obvious it exists and the enemy cannot do anything about it. It would also introduce a new mechanic while not serving any purpose because FFA rebuild is still possible regardless.

CarabusX:
About FFA rebuild:
- Reclaim: 
Many maps have no energy features so energy income is a must.
As for metal you frequently have to build radar first to see if the wreck covered area is free to take, so you need some metal to start.
- Reserve:
With the new resource bars it can't be set easily, only in settings. Yes you can't click on resource bars anymore, you have to go settings, enable reverve and set the reserved amount.
But reserve is still not something you can reliably count on, because resources in reserve can be spent too, by high-priority constructions, and you are likely to be rushing something on high-priority where your base is dying.

Attempts to make comeback happen quite frequently in FFAs with several good players on maps that have good spots for them. They most often fail if the player is discovered quickly, but sometimes they are as meaningful as deciding the outcome if the game is long enough.
Typical comeback doesn't mean sitting in corner building fac for 40 mins. Usually a player loses main base, but has several athenas or cons (usually rectors) and some other, possibly powerful units left. He needs to get some energy production going, spam some radar to find some metal spots or wreck fields to take, then take them, rebuild fac and he is back to the game, assuming other players are busy with each other for while. But the start needs resources and +1.0M/E from several cons is infinitely more than +0.0M/E. What is the value of metal spot if you don't have metal to make mex?

CarabusX:
As for passive income I too think it is bad for reasons Sprung mentioned.

Another idea would be to make innate cons income dimishing.
Pros:
- works kinda like passive income, there is soft limit of how much you can get.
- there is interaction, you can kill last cons to remove the income. Without cons there would be no income but it is also not needed anymore...
- FFA rebuilding is not hurt much. Con income matters the most when you try to get your first windgens and mexes, and it would be preserved.

Cons:
- adds complex logic, not tranparent to players. Could be so confusing, that I would rather stay with current innate incomes mechanism.

sprunk:
You can always reclaim live units if you need metal, or use them to scout/escort the reclaiming con if you cant afford a radar. Energy is easy to secure (you can hide a contingency windgen beforehand or something) so I don't think it is a problem.
 
Reserve can be set directly using ctrl+click. If you use high prio then you should be aware that rushing some last ditch effort to save the base will drain your rebuild reserve and will have to decide whether to take that risk and go va banque, or to delay the thing a bit but be guaranteed the ability to rebuild.

Basically if you do some preparation you won't really be affected. The situation that gets prevented is making a comeback after you screwed completely.

Licho1:
I vote to keep them in some form for now, though they are ugly and would be
non-existent if this had been designed from scratch, the awesome FFA
comebacks are worth the minor annoyance of non standard income.


2015-06-10 4:20 GMT+02:00 sprunk <notifications@github.com>:

let4be:
Bad idea, cons, factories, caretakers should give minor income in my opinion
This makes some interesting strategies possible, don't make one another boring stadard rts out of zk

sprunk:
Nobody cares about greenfields and Comm keeps its income

let4be:
I do care., if you don't it doesn't mean nobody else does :)

Anarchid:
Nometal maps are not 'supported' by ZK. You argument is void.

let4be:
But I remember such games were played, some map with ore-mex when people had to reclaim ore that slowly moved to center of the map
Ore was limited and the only income was possible - came from cons/caretakers ;) which was quite interesting

Anarchid:
"Some people did it" is not evidence for "it is supported".  If it were, then you could argue that ZK needs flying Warlords because some maps have that.

KlonZK:
distributing con incomes over team sounds reasonable, some people like fireman have already trolled with huge caretaker farms in the last couple of days. when asked why he doesn#t build more efficient energy instead, the answer was "because it dont want to share income".

Anarchid:
There was an awesome comeback in http://zero-k.info/Battles/Detail/362409, and there was something related to con income. Namely:
1) Multirotor made a nanofarm which provided him more income than what Yogzothoth had while controlling half of the map and having yellow grid.

2) Rafal didn't need con income to make a comeback and win.

GoogleFrog:
I am just happy to wait and see where people take nanofarms. No need to remove the option at this point imo.

sprunk:
Is providing more income than half a large map's worth of yellow grid not taking it too far already? How far would it have to go?

From the first comment:
>1. This is not what people find confusing about the economy. At that level you barely notice it

People *do* find it confusing when you take it to the extensive farming scale. For example see http://zero-k.info/Battles/Detail/363402 - multiple people complain about a nanofarm and kick votes are even started "because useless" even though the farm actually provides the owner with three times more income than other people receive through OD. The exponential nature of this income means the player is encouraged to just keep cycling income back into more nanos. Nanofarms got some popularity in teamgames recently and usually receive negative response from the team.

KlonZK:
shared con income would fix this tho. dont see a reason to remove it completely as OD has better returns in 90% of real games, just the fact that con income is personal incentivizes people to prefer them over OD.

sprunk:
I think removal is better than making it shared because:
* removal is simplier and much cleaner. Sharing would need some new gadget to track cons and redistribute their income which eats performance and adds code to maintain for no gain.
* sharing does nothing about nanofarming in FFA. One could argue that it's alright to have nanofarms there but I think nanofarming goes against both the eco design that OD follows (is not tied to territory and is exponential) and intuition (can have large impact through a very obscure side feature of a unit).
* the income itself does not have a design or purpose and is kept as legacy. Apart from farming, it is currently supposed to allow FFA comebacks from a position where you utterly screw up. I argue that it is good to punish complete failures as it is possible to make a comeback without con income. It also makes everything effectively a little bit cheaper but again I'd argue that the change in pacing would be minimal and barely tangible (like the 16/15 change which it approximately reverses).

Anarchid:
Some more data http://zero-k.info/Battles/Detail/363620

CarabusX:
After putting more thought into it, I am no longer strictly against the removing con metal incomes, as long as their Energy income stays. Most FFA comebacks I remember, didn't really need con metal income, but many needed their energy income to make first few windgens. Metal can be obtained other ways, like already suggested, but in most maps energy cannot. Leaving their energy income doesn't pose a problem, as energy is cheap to spam anyway.
As for change in general pacing and incomes, perhaps a buff to OD could counter it? (and imho would be desirable anyway with how bad OD is now)

Licho1:
That sounds more reasonable.

let4be:
Are you going to kill the whole meta of comnap + capture, comnap + kill + resurrect, or just ressing coms and using them as mexes?
This is fun!

Anarchid:
Don't mix up commanders and constructors please.

I very much like the idea of removing the metal fraction while keeping energy fraction of innate incomes on non-commander units.

sprunk:
>Most FFA comebacks I remember, didn't really need con metal income, but many needed their energy income to make first few windgens

Why not make emergency windgen (with the surviving cons, in a safe place) when you notice yourself dying but still have energy? It takes 5-7s depending on con.

I will modify this PR to only include metal removal because at least there is consensus.

sprunk:
Since it seems the surviving con is Athena most of the time: how about giving it some energy generation, say +1 (and increase Cloak cost by the same amount), but remove it from other cons?

As the unique multirole spec-ops thing, I think it can probably afford to have a small energy income. If GoogleFrog's idea of rez removal passes through, it also becomes a compensation buff.

PR #801, opened 2015-05-26
fix #800

### LINKED ISSUE #800: CTF and KoTH doesn't work anymore
Fails to spawn the CTF map points and the KoTH spawn zone (I dont know the details on the issue).

sprunk:


Orfelius:
What can I say? Looks legit to me!

sprunk:
It prevents confusion because there is actually no working koth/ctf. If somebody wants to have koth/ctf they are free to reimplement but there shouldn't be an option for something that doesn't even work.

Orfelius:
Nono I don't disagree its just quite... I dunno. Interesting to see?

sprunk:
Game-side my implementation is used but infra-side xponen's version got merged.
Both implementations require the counterpart to also get merged to be useful.
Only one version of the shuffer is needed so one of them needs to be chosen and the other reverted.

PR #803, opened 2015-05-27
Widgets call UnitDestroyed for enemies that die in LoS

sprunk:
Similar to what Send Enemy Death was supposed to do, but working and done properly (calls UnitDestroyed directly without any input needed whereas SED required widgets to explicitly subscribe a separate function to the API and employed some weird redirection through synced).

Updated widgets that assumed the contrary not to break.
Updated Attrition Counter widget to use the new detection method, which also enables it not to count as losses units that didn't actually die (eg. morphs or factory cancels).

### DIFF
--- a/LuaRules/Gadgets/api_widget_events.lua
+++ b/LuaRules/Gadgets/api_widget_events.lua
@@ -1,0 +1,26 @@
+if gadgetHandler:IsSyncedCode() then return end
+
+function gadget:GetInfo() return {
+	name      = "Widget Events",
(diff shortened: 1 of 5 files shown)


xponen:
have you check all widget?

sprunk:
Yes: the ones that would have a behaviour change got the additional checks (some of them looked like they were supposed to also work on enemies but I chose the safe route). Two widgets now behave differently: Music Player (it can now play the war music when enemies die in addition to your own stuff dying) and Attrition Counter (doesn't count fake deaths as deaths).

PR #804, opened 2015-05-28
Add support for "unknown" Elo rating

sprunk:
To differentiate people undergoing placement matches and actual noobs.


Licho1:
There is elo weight that corresponds to how known player's elo is.
How is this change meant to work?

sprunk:
See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/635

PR #812, opened 2015-06-02
Impulse jump improvement

xponen:
Rewrite of impulse jump rule toward broader user interactivity. Allow:
  1) easier skuttle launch. Newton interaction now quantifiable and repeatable (aka: a better ramp).
  2) more damage when sumo squash enemy structure. Collision is detected properly (that trigger explosive anim) & lesser bounce

Minor tweak
  1) add teleport exception to factory_hax gadget
  2) reduce tree bounciness (for skidding unit only) using fall_damage gadget

### DIFF
--- a/LuaRules/Configs/jump_defs.lua
+++ b/LuaRules/Configs/jump_defs.lua
@@ -14,6 +14,9 @@ for id, ud in pairs (UnitDefs) do
 			rotateMidAir = (tonumber(cp.jump_rotate_midair) == 1),
 			cannotJumpMidair = (tonumber(cp.jump_from_midair) == 0),
 			JumpSpreadException = (tonumber(cp.jump_spread_exception) == 1),
+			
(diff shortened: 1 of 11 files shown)


xponen:
Hopefully this will be the default Jump in future, 

It give jumper units more power. 
and more interactivity.

GoogleFrog:
Do not accept your own PRs before anyone can respond.

Do not add minor tweaks alongside a large PR. Make PRs for the tweaks in separate PRs and justify them there.

I thought we had already discussed the problems with impulse jump. It is not cool physics, it is annoying physics. Jumpjets have to be reliable for people to trust them. It is too much 'physics randomness' for a unit to be nudged slightly by a leveler shot and end up on impassible terrain.

There are much more useful things to spend time on before this change to jumpjets is considered. My remaining problem with normal jumpjets is that jumpers cannot jump to land on top of structures.

PR #827, opened 2015-06-09
Remove modoption bloat

sprunk:
* Comm Ends and Com Control (unfair due to unlocks, soft comm ends already exists earlygame, pointless lategame because you can cloak/bury the com)
* No Air and No Defenses (never used, there's already a modoption that does the same except better by allowing to specify arbitrary set of disabled units)
* Share Control (never used; enforced the default, ie. block sharing to enemies)
* No Wreckage (super silly (both gameplay and visually), never used)
* Fac Cost Multiplier (never used, arbitrary)
* Special Decloak (never used)
* Wreckage Value Multiplier (broken, never used)

Also removed Scan Sweep ability implementation (was supposed to be a cool Planet Wars ability but was never used as such)


GoogleFrog:
I would not remove Comm Ends because it is a classic option that people might want to try. The old widgets dedicated to Comm Ends could be removed though.

Also you could retain effects/scan.lua because someone might use it for something else.

sprunk:
Scan effect looks very poor and is basically just 1 groundflash.
I'll bring back Com Ends.

Also consider what to do about other modoptions that dont see use:
* Water level (maps already tend to have wet/dry versions and modoption sucks at conveying its existence through lobby)
* Weapon cratering multiplier
* Pathfinder
* Impulse jump
* Terraform cost mult

Licho1:
I have seen no air used in team game.s

Licho1:
Same for cratering, terrform and water level.
Had seen games altering that

GoogleFrog:
If you are removing modoption which aren't used then you would remove all of them. So instead you have to remove modoptions which change the game too much or are unreasonable to maintain.

Water level seems like a reasonable option. Same with Cratering and terraform cost mult. The Pathfinder is a useful modoption for testing.

sprunk:
Some modoptions are used regularly on the official hosts (like zombies or player chickens which seem to appear every now and then) and some are popular among complete noobs in private games (like metalmult). These can stay.

On the other hand nobody ever changes cratering.

I don't see why deleting stuff that is never used would be a problem even if it meant removing most of the modoptions. Their existence makes navigating the modoption menu difficult and I think it's important that people who want to change some settings be presented with the meaningful ones instead of being forced to wade through loads of crap that serves no real purpose.

Licho1:
I played many cratering disabled ganes..
Modoptions are not meant to be easilly navigable..its like power user adv.
settings, you want options there not accesibility.
Regarding maintenance, unless its costly to maintain now, keep it.
Dne 9. 6. 2015 20:36 napsal uživatel "sprunk" <notifications@github.com>:

sprunk:
Cratering-disabled is not cratering-multiplier, it's a separate thing

sprunk:
Brought back Comm Ends.

GoogleFrog:
I would retain defeatmode. It is at least useful for development and sandboxes. Missions may want to use it too.

sprunk:
Defeatmode retained.

PR #828, opened 2015-06-10
Rework of GF Sea Rework of my Rework

RyMarq:
Hunter made fragile with 360 range.
Makes a ship that is actually fragile. Decreases need for high numbers of hunters to increase ship composition variety.

Snake changes averaged to be less extreme.
Range diminished from base and can take 1 more urchin hit, but stays closer to base.

Typhoon changes normalized.
-Maintains margin of speed of Hunter, but did not need extra buff.

Scallop DPS halved with small range increase.


GoogleFrog:
I tested some things about this change.

The speed difference between Hunter and Typhoon is too small for Typhoon to catch and beat Hunter. 

10 Scallops (2800) lose to 40 Ducks (3200) with the Ducks slightly spread but still approaching from one direction. The riot density required to be effective here is too large.

GoogleFrog:
Can't be merged anymore probably because I made similar Hunter changes.

RyMarq:
Typhoon can catch hunter in this for the same reason scallop can sometimes hit the hunter. Because of relative turn rates.

When that micro occurs, the typhoon is able to close upon and murder the hunter. I find it odd that you claim the hunter can skirmish the typhoon but cannot skirmish the scallop when the typhoon is radically faster and longer ranged. 

Scallop range increased to decreased chance of harming itself/allies when it attacks with its considerable AoE.

Likewise, I find it a bit odd that scallops lose to ducks that significantly, with even slight micro, but the number can be tweaked if they are that off. Sadly, I can only really micro one side of this equation in personal testing. The Scallop 2-shots a duck, in 1.8 seconds after engagement. The scallop gets its second shot off right after the duck gets its first shot. The ducks should take heavy casualties in the first wave of shots, then trade evenly with their remainder for metal, then die off to the remaining scallops.

I don't think its unusual that a riot trades not-fantasticly against a relatively well micro'd raider, especially one as well suited against Riots as the Duck.

Am I that far off here? I put some thought into these numbers, and initial tests seemed to show results consistent with my estimates.

GoogleFrog:
Typhoon also has a large turn rate so it is much harder to react and form up while chasing Hunter. Some of the Typhoons can shoot while chasing Hunter but not enough to win while all of the the Hunter get to focus on one Typhoon.

Scallops can position themselves well and make it very painful for a Hunter to end up within Scallop range. Hunters will often end up in range due to their large turning circle.

GoogleFrog:
Merge issues and current sea seems to work.

PR #831, opened 2015-06-13
Okp imp v2

lhog:
This is a working implementation of overkill prevention for disarming (racketeer only for now) and regular weapons (same as before). It's based on the common code.

It's possible to add shields into account (that's why I needed stepped iteration over frames). Please give your feedback to the existing part before I start adding shields.

P.S. Also preferably use 769 engine, where @ashdnazg rectified issue, that units standing still hadn't used ZK targeting priorities.


GoogleFrog:
Split out the unit_target_priority changes because they are unrelated and require their own discussion.

The paralysis and disarm mechanics function with respect to health and max health in exactly the same way. There is no difference in the resulting mechanics.

Could you describe the changes to the OKP gadget? What improvements are there? What have you fixed?

lhog:
Unit_target_priority changes are in fact related to the whole thing. I found that targeting priorities for partially disarmed targets were suboptimal, although valid. Not every disarmed target has the same priority as it was put in the code. Another question whether or not my implementation makes it better. I haven't yet felt it, so it indeed may need separate discussion. Ideally I'd like to have separate functions handling priorities for regular, disarming, slow and EMP damage. The current implementation, I think, is quite bloated. 

Changes allow for disarm damage handling as I wrote before. You can in fact observe almost working OKP for racketeer. It also works with regular damages with same result as before. The need for radical changes in code was due to the future shields interaction. Damage to the shields will be applied each frame as a part of simulation.  Previously code operated only with cumulative damage,  which was fine as long as it was the only damage type. Now with two and potentially 4 damage types it's no longer possible. In example order at which regular and disarming projectiles hit targets will cause different results.

GoogleFrog:
The content of your changes to the two gadgets is unrelated though. They have no dependence on each other. The other one needs discussing and I would like to keep individual tickets focused on one thing.

Many of the added LOC have no purpose for the PR. There are two large commented out functions and a function which is never used. It would be much nicer for the diff and repository if you had a cleaner PR. I do not know enough about git to tell you how to do that.

I would like to know what exactly has been improved and why there are things such as new data structures? What do they do? Not just what do they eventually achieve.

lhog:
As for 'why' I think we discussed that #zkdev.

I can revert unit_target_priority changes. It's not critical.

Gadget was WIP, however I actually like your idea to clean it up and leave it with no new functionality for now. Racketeer OKP is part of current functionality though, so upon following cleanup this could be merged.

GoogleFrog:
Make a new PR with the priority changes, don't just leave them.

lhog:
Yesterday I had an interesting and eye-opening conversation with @ashdnazg on how things work. Apparantly they do in a slightly different way than I thought.

I was under impression that BlockShot=true causes unit to go for next target in priority order specified by gadget:AllowWeaponTarget beforehand. Apparently I was wrong. BlockShot doesn't cause anything more than blocking shot. In order to change target, new and better priority need to be specified in gadget:AllowWeaponTarget for a desired target and yet to be discovered actions to be made in OKP code to make unit "forget" the current target (as spSetUnitTarget(unitID, 0) may not be enough and CMD.STOP issue was suggested in some cases).

Though the code works pretty well for moving racketeers right now, there is a fundamental flaw of bad/slow feedback between blocking shots from OKP and updating targeting priorities.

Having that said:
1. I'm going into science mode temporary to find out how things work exactly and therefore update framework (likely slightly).
2. Code in unit_target_priority.lua no longer has minor meaning it has rather major meaning and therefore likely to be an integral part of the current branch. These two files are like yin and yang: can't be separated.

lhog:
I have found that (current) uniform function for setting targeting priorities is not good enough in some cases.

Say for two targets within range of few racketeers, setting defPrio = defPrio + 25 for disarmed units, make them all target nearest one if both targets are currently disarmed. Although right metric would probably be to prioritize the target which is sooner to lose disarm state or something slightly more sophisticated. For such units the shot is being fired to the right target too late: when disarm is already gone (thus no defPrio = defPrio + 25). I'll try to modularize targeting priorities (in separate branch) in accordance to damage types, if you don't mind.

lhog:
Closing due to lack of response or acceptance.

GoogleFrog:
I thought you were still working on the new stuff that you mentioned in your last few comments.

lhog:
Targeting priority thing has proven to be very complex, to say the least. So rethinking this may require separate branch, as you proposed initally. I'm having intents to ask engine devs to augment targeting with new callin in a way so that target units are sent in a bulk hash array --> unitID-priority. I have arguments why it's a better way compared to current one, however lets have a separate discussion on that.

As for current branch I think you can merge it as it's (give me a day or two  to test how it works though). Next thing I'm going to do, in case you merge, is to add shields (probably in a way you proposed previously) and handle targets under repair properly.

ashdnazg:
You don't really need such a system, always store the last winner, and if you want to pick a new one, just give it (lastpriority - 1).

I have a suspicion that any implementation using the bulk method can be transformed into priority mode.

lhog:
@ashdnazg , oh since you around, this thread might be appropriate for discussion. So let me state what diffuculties I stumbled upon.

The amount of possible unit state is huge: disarmed, emped, under capture, slow, unit is considered doomed, static vs moving, built or under construction, health, cost, carrying shields or not, etc. Basically what current call-in asks is to state the numeric priority for a single target without much context. And by context I mean what other targetting alternatives are. Having such information it's much easier to sort targets out based on multiple levels of "if then" clauses and therefore return one and only (best if you will) target for a unit & (unit weapon). With the present method of operations one needs to collapse all that huge state into a single variable. Take a look into that file: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_target_priority.lua It's full of empiric constants and related assumptions. It's probably doing very good now, but it's very hard to expand to support something else. New additions are required to operate within its own numeric space and don't overlap with anything else.

ashdnazg:
I'm entirely aware, I wrote the previous comment taking what you said in mind.
The approach doesn't have to be expanding the current system.
it can also be:
1) give the first target a priority of 1000 and remember it.
2) compare the second target to the first, if it wins give it a priority of 999 and store it, otherwise give it 2000, and keep storing the first.
3) etc.

lhog:
This may work. It's not going to give same level of debug easiness and clarity like the approach I'm proposing though.

Are you that strongly against doing it right (should be faster as a additional merit)? I'm sure it could be done without breaking compatibility with present mode of operation.

ashdnazg:
It won't be faster, and it will require rewriting parts of the engine.
So no, I'm not going to implement it myself.

I actually think the priority system is very clean and easy to support and debug, it's your lua implementation which is giving you a hard time, so what you really need to do is devise a new system that works and that can take into account many variables in a smart way. This has nothing to do with how the engine provides you the targets.

lhog:
Fair enough. It makes sense to try the approach you have proposed.

GoogleFrog:
I was going to comment on whitespace and variable names again but instead I made a PR https://github.com/lhog/Zero-K/pull/1

lhog:
Thank you. I kind of missed that comment regarding whitespaces. I will try to stick to that style going forward.

PR #842, opened 2015-06-14
Fix #811: wrong Ctrl+C behaviour.

### LINKED ISSUE #811: Getting commander from resigned player breaks commander find
After you get commander from resigned player neither clicking on commander icon nor ctrl+c find your original commander.

This is a new bug.

sprunk:
Actually a workaround because the cause seems to be in engine (it didn't happen in 91). Currently the linked function gets called multiple times per Ctrl+C which can cause some commanders to be always skipped. The workaround prevents it from being called more than once per keypress.

### DIFF
--- a/LuaUI/Widgets/gui_chili_core_selector.lua
+++ b/LuaUI/Widgets/gui_chili_core_selector.lua
@@ -709,9 +709,15 @@ local function StripNanos()
 end
 ]]--
 
+-- this prevents SelectComm from firing off multiple times per key press.
+local alreadyFired = false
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Are you sure being called multiple times is not some weird behaviour within chili. epic menu or the widget handler?

sprunk:
Widget handler calls the action handler and we use the basecontent action handler.
In 91 it didn't seem to happen.

Weird behaviour within Chili could be a cause though. I'll investigate a bit more.

sprunk:
Closing: see #843

PR #843, opened 2015-06-14
Alternate fix to #811

sprunk:
It appears Epic Menu screws things up by trying to register an existing action.


GoogleFrog:
So selectcom registered as an action and then epic menu also registers that action? Weird.

PR #848, opened 2015-06-14
Remove a quit command send from autoquit, seems to fix GetConfigData() callin not being called

dvr333:


GoogleFrog:
Is it just relevant for this case? quit is used alongside quitforce in many situations. Ideally someone would decide which one to use and then use it consistently.

GoogleFrog:
It looks like the quit command has been removed from the engine. Thus this change is good but should have no effect on anything.

PR #859, opened 2015-06-18
Queued Solars shown properly.

sprunk:
Before and after:
solar_orientation


Orfelius:
I have to question this. Wile I like that the positioning has been changed but on the other hand solars start up closed not open thus they should also have closed unit shadows.

sprunk:
Solar build icon also shows the open position and that's what live Solars default to so I think it's more intuitive to use it.

PR #865, opened 2015-06-22
Move com module metal costs game-side

sprunk:
Gameplay values belong game-side. In the database they cannot be easily looked up, modified, or known by widgets.

The patch also makes modules that do not meet prerequisites (e.g. having Shotgun Autoflechette without a Shotgun installed) not cost metal.

Local testing was done on values copied from a real springie comm.


KingRaptor:
Do you have any particular plans for keeping this synced with the DB?

sprunk:
No, but the DB would only be used for telling people the cost values when composing a comm. These can be maintained manually (they already are, anyway) as they aren't as important. Ideally the site would read that from lua (I'll look at the unit guide generator).

Licho1:
You need to replace full functionality -> sync from game to DB. Then and only then, would having them ingame be preferable.

Issue #868, opened 2015-06-22
What is going on with the conditional transportability gadget


GoogleFrog:
The gadget has a recursion error https://springrts.com/mantis/view.php?id=4833

There is something at the bottom talking about a new version for later engines. This should be implemented if it is real.

https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_conditional_transportability.lua

sprunk:
This didn't happen when testing so presumably newer engines changed something or some corner case happened.

The gadget is only for allowing empty Surfboards to be transported so it's not a big deal to disable it for the time being.

The later engines implementation was in case AllowTransport call-in started existing in engine. It hasn't so far so the simple version can't be used but it shouldn't be hard to make it happen.

GoogleFrog:
So to clarify: are the suggestive comments at the bottom for a real engine version which we are not using yet? The surfboard thing is rather unimportant so I will just disable the gadget and remove surfboard transportability.

sprunk:
The call-in in the suggestive comment is planned only, so no real engine version has that yet.

PR #869, opened 2015-06-22
Quickfix #868

sprunk:
Disables Surfboard conditional transportability until more investigation is done.


GoogleFrog:
Oops, already made this change.

PR #870, opened 2015-06-23
Fix script division by 0

ashdnazg:
All other speedmult calculations work that way


ashdnazg:
No, wait, it doesn't make sense mathematically.
Also isn't reloadtime handled in unit_attributes.lua?

PR #871, opened 2015-06-23
Fix division by 0 in dante's script

ashdnazg:
These lines don't do anything anyway, since everything is overridden by unit_attributes.lua


GoogleFrog:
They are a poor attempt to make the missile salvo weapon block the normal missile weapon. The implementation is broken but cannot simply be removed.

PR #873, opened 2015-06-24
fix div0

sprunk:


GoogleFrog:
Although, it is probably better to just remove this bar.

PR #879, opened 2015-06-26
yolo gaetors

sprunk:
Scorcher engagement range 270 -> 64.
Heatray has more damage up close so it is reasonable for Scorchers to try to maximize it.


sprunk:
On further testing it seems that SetMaxRange also prevents automated target acquisition beyond that range so this method is undesirable.

Perhaps it could be given a fake main weapon with low range instead, or the tactical AI widget could make them move closer even against things they do not jink (they already attempt to move very close against the things they jink).

PR #881, opened 2015-06-27
fix #880

### LINKED ISSUE #880: resign during commander morph may leave com unowned and unusable
if the morph was started before the resign, and did not finish before the com was given to the new owner, and was not canceled by the new owner. that is, the new owner of the com will own a perfectly usable com until the morph is finished, at which point the com will change owner to "noname" and cannot be used anymore. 

similar to this one: https://github.com/ZeroK-RTS/Zero-K/issues/570

steps to reproduce:
- play a team game
- start a com morph
- resign after starting the morph
- receiving player has to not cancel the morph

http://zero-k.info/Battles/Detail/360556

few minutes into the game, fireman resigns and his com in the northeastern corner gets given to me. i own the com for a brief period, when the morph is done and it changes ownerships to noname.

the morph gadget seems to set the teamID when the morph starts and then not change it anymore unless the unit gets capped during morph. so presumably just setting it at the end of the process will do away with this issue.

sprunk:
* units that change owner during morph now spawn in the correct team
* priority is reassigned


GoogleFrog:
Reassigning priority should also be done for stockpile.

PR #882, opened 2015-06-30
Autorepair System that works in combat

Sgathec:
Quoting GoogleFrog after a bit of discussion about bringing back the old autorepair: "I feel like giving up and letting you owl it with a level 3 requirement".

The amount this commander module repairs remains the same. The database needs to be accessed in order to set the level requirement. At first I thought a cost increase would have made sense too because the module used to be rather powerful but on second thought the level requirement might actually suffice.


GoogleFrog:
How about, as the nerf, a reduction in HP/s instead of a cost increase?

Sgathec:
I thought the level increase would be enough because there are other modules that need at least level 3.

Sgathec:
Would you consider 25% less HP/s and 20% less cost adequate?

sprunk:
How about 12.5 regen? This way it maxes out at a nice round 100.

Sgathec:
That could be done if GoogleFrog deems that low a value fitting. I'm not sure why he still hasn't responded.

PR #887, opened 2015-07-02
Remove target move error across the board

sprunk:
* Projectile weapons can get lower velocity to make them worse vs fast stuff in a physicsy way (if they aren't bad enough already).
* Hitscan weapons can be widgetised to emulate 0 TME (by targeting ground behind target), missing looks like the unit is arbitrarily and deliberately being stupid (because it is).

Zeus and Panther are probably the most affected ones and would likely need some monitoring because this does mess balance up a bit.

### DIFF
--- a/gamedata/weapondefs_post.lua
+++ b/gamedata/weapondefs_post.lua
@@ -264,9 +264,10 @@ end
     weaponDef.noselfdamage = (weaponDef.noselfdamage ~= false)
  end
  
--- remove experience bonuses
+-- remove experience bonuses, target move errors
 for _, weaponDef in pairs(WeaponDefs) do
 	weaponDef.ownerExpAccWeight = 0
+	weaponDef.targetmoveerror = 0
 end
  
 --------------------------------------------------------------------------------


Licho1:
imo it has its uses exactly for stuff like zeus .. approximate is that unit has slow tracking/turret which can also be done with script.
It is visible and can be kept imo.

Voting against unless there are some significant arguments for removal.

sprunk:
Arguments for removal:
* for hitscan weapons it can be emulated with a widget (by targeting ground behind victim), which would then have to be included in ZK for equality anyway, so we might as well remove the missing.
* it is arbitrary and not physical. More physical options exist (eg. slower turret turn rate like you said, or slower projectiles for non-hitscan weapons).
* the unit deliberately misses which should be considered a unit AI bug.

GoogleFrog:
Slower turret turn rate is not a replacement because re-aiming is infrequent and lightning cannot miss once it is fired.

I dislike doing this in _posts because it is unclear what has changed and later devs may think it is still in place. Removel from unitdefs can be done by removing "\s*targetMoveError[^,]*," with the particular way that np++ implements regex.

I will implement the unitdef removal with some sprayangle to simulate the target move error for lightning. Zeus, Scorpion, Bantha wit h 0.3 had and Panther had 0.1.

PR #890, opened 2015-07-03
targets set manually using the set target command do not clear when using stop command

KlonZK:
attack commands will now always set a (weak) target, which will be cleared through stop and fight commands. manually set targets will be overriden by attack, but persist in the latter case.

note that this only occurs when attacking units. attacks on ground will be ignored by the widget and thus wont set a target, nor will they remove any existing target (this needs some consideration, units show strange behaviour when they have both a target and an active attack order somewhere else in range).

### DIFF
--- a/LuaUI/Widgets/cmd_keep_target.lua
+++ b/LuaUI/Widgets/cmd_keep_target.lua
@@ -2,8 +2,8 @@ function widget:GetInfo()
   return {
     name      = "Keep Target",
     desc      = "Simple and slowest usage of target on the move",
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I realized that this widget does not take shift queued commands into account and this potentially makes your implicit target tracking complicated. i also have doubts on the usefulness of the option as its only advantage over no target removal is that some targets are removed automatically. If you are using set target you should be able to do this manually. You did not implement options so I decided to implement them myself.

PR #892, opened 2015-07-03
Prevent drones from being assisted by cons

sprunk:


sprunk:
Apparently this won't work because drones are not real buildees but gadget-made.
Perhaps blocking CMD.REPAIR orders on drone nanoframes would work though.

PR #893, opened 2015-07-04
keep target widget update

KlonZK:
- moved the whole thing to UnitCommand() callin. this was necessary for area/line commands to be 
  properly registered, as they dont appear in CommandNotifiy(). this has some side effects:
   - units get adressed directly and need no longer be looked up in the units list
   - it appears to be no longer possible to check for internal commands as the UnitCommand() function receives no usable value for this
   - line orders now correctly clear targets

- added CMD.DGUN and CMD.AREA_ATTACK to commands which set a target, added CMD.PATROL
  and CMD_UNIT_SET_TARGET_CIRCLE to the commands that remove a target. 

- the widget does respect command queues as good as possible, there are some rules on how the shift key and previously queued commands will be taken into account. targets will only be set if the very first command in the queue is an attack command, otherwise shift key will generally preserve targets. 

it would also be possible to insert set/remove commands into the que such that all attack commands would always set an appropriate target when executed, regardless of their position in the queue. this would however require a bit of synched code(possibly in the unit_target_on_the_move gadget) to catch the CommandFallback function, as target commands normally get handled before they enter the queue. (insertion skips that)

### DIFF
--- a/LuaUI/Widgets/cmd_keep_target.lua
+++ b/LuaUI/Widgets/cmd_keep_target.lua
@@ -2,8 +2,8 @@ function widget:GetInfo()
   return {
     name      = "Keep Target",
     desc      = "Simple and slowest usage of target on the move",
-    author    = "Google Frog",
-    date      = "29 Sep 2011",
+    author    = "Google Frog, Klon",
+    date      = "3 Jul 2015",
     license   = "GNU GPL, v2 or later",
     layer     = 0,
     enabled   = true  --  loaded by default?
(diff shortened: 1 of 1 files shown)


GoogleFrog:
You've done something which makes this unmergable. Also it is a waste of time to add the target when the attack command is given as then the gadget has to do extra work and an extra line is drawn. Avoid it if you can.

KlonZK:
well you just have to merge the older(closed) commit i think, if thats possible? anyhow ill give this another look first.

GoogleFrog:
There was a merge button. It was grey.

Issue #905, opened 2015-07-04
AI start position broken


rlcevg:
ZK denies attempts of ai to choose its position (via Game::SendStartPosition) and throws them all in one spot (middle of startbox).
[start_boxes.lua](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/start_boxes.lua) to blame?

sprunk:
Yes, see https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/start_boxes.lua#L96

The AllowStartPosition call-in only receives a player ID to identify a team and gives all bots ID == 255 so it is impossible to discern which team the bot is from (and thus which box it belongs to).

As an alternative, bots could be allowed to place anywhere. The main downside is that bots would be allowed to cheat this way. Bots already can cheat so I guess that's alright, and the change would also benefit ones that play fair.

rlcevg:
Ok, thanks.
I think workaround #906 is the best atm.

Probably proper engine-side solution would be to pass teamID in every lua-callin (breaks backward compatibility), or refactor AI so it will have playerID (amount of work and new bugs unidentified).

rlcevg:
#906 awaits its hero who'll lift his mighty finger without a shadow of regret and push the damn merge button!
Or is there another opinion on the topic?

Anarchid:
I think this can be closed now :)

rlcevg:
Tested single-player - works. Thanks.

sprunk:
I'll expose the boxes to GameRulesParams and hopefully add an IsPosInBox call-in.

GoogleFrog:
Does this now work since the start box changes?

rlcevg:
I have no idea how to test it atm. Tried on
BETA TEST / Iridium / 100.0
Terra  1
Zero-K test-10483-95040dd
But had issues with UI (no UI on start, but only after /luaui reload) and 9/10 Circuits were dead (probably new-modular-commander related)

rlcevg:
Works for me

PR #906, opened 2015-07-05
Startboxes: let AI pick startpos

sprunk:
See #905

External AI can now pick a startpos.
AI that doesn't pick any or tries to pick an invalid one will still get the middle-of-box default.

### DIFF
--- a/LuaRules/Gadgets/start_unit_setup.lua
+++ b/LuaRules/Gadgets/start_unit_setup.lua
@@ -354,7 +354,7 @@ local function SpawnStartUnit(teamID, playerID, isAI, bonusSpawn, notAtTheStartO
 	local x,y,z
 	local startPosition = luaSetStartPositions[teamID]
 	if not startPosition then
-		if (startboxString and not Spring.GetTeamRulesParam(teamID, "valid_startpos")) or (not startboxString and notAtTheStartOfTheGame and (Game.startPosType == 2)) then
+		if (startboxString and not (Spring.GetTeamRulesParam(teamID, "valid_startpos") or isAI)) or (not startboxString and notAtTheStartOfTheGame and (Game.startPosType == 2)) then
 			x,y,z = getMiddleOfStartBox(teamID)
 		else
 			x,y,z = Spring.GetTeamStartPosition(teamID)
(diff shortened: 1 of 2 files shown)


Anarchid:
> The obvious scope of picking inside own box is not possible because AllowStartPosition does not discern bots so it would not know which box to use.

It can learn to do that, however:

1) obtain playerID from `AllowStartPosition` callin
2) obtain teamID from playerID via `Spring.GetPlayerInfo `
3) obtain isAI from teamID via `Spring.GetTeamInfo`
4) ???
5) Glorious victory awaits.

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -354,7 +354,7 @@ local function SpawnStartUnit(teamID, playerID, isAI, bonusSpawn, notAtTheStartO
 	local x,y,z
 	local startPosition = luaSetStartPositions[teamID]
 	if not startPosition then
-		if (startboxString and not Spring.GetTeamRulesParam(teamID, "valid_startpos")) or (not startboxString and notAtTheStartOfTheGame and (Game.startPosType == 2)) then
+		if (startboxString and not (Spring.GetTeamRulesParam(teamID, "valid_startpos") or isAI)) or (not startboxString and notAtTheStartOfTheGame and (Game.startPosType == 2)) then
 			x,y,z = getMiddleOfStartBox(teamID)
 		else
 			x,y,z = Spring.GetTeamStartPosition(teamID)
```
rlcevg: Clamping position
```
-- FIXME: AI's teamID in AllowStartPosition
local boxID = Spring.GetTeamRulesParam(teamID, "start_box_id")
if boxID then
	local box = startboxConfig[boxID]
	local bx = x / Game.mapSizeX
	local bz = z / Game.mapSizeZ
	local valid = (bx > box[1]) and (bz > box[2]) and (bx < box[3]) and (bz < box[4])
	if not valid then
		x,y,z = getMiddleOfStartBox(teamID)
	end
end
```
right after 360th line should make start for NullAI/RAI/zkmbai/zkgbai more appropriate. (Yet a bit wrong if startbox contains map center, [engine](https://github.com/spring/spring/blob/1c0cf0a2d024c453a97c1f6d70f89ed6740721c8/rts/Game/Game.cpp#L1519) to blame)

sprunk: Good idea.

PR #907, opened 2015-07-05
replaced cmd_keep_target with gadget implementation

KlonZK:
now respects queues and will prevent wonky concurrent attack/target unit behaviour in most cases. 

attacks will now always remove existing target links, and if the target of the attack is a unit, set a new one (we have to give at least a cancel target order anyway, we might as well just give a set order. it does draw the additional line tho). 

fight, stop, partrol, guard commands will remove those targets, but will not remove manually set targets. otherwise automatic und manual target setting has the same effects. 

other commands will always preserve targets of either kind.


LuaRules/Gadgets/cmd_auto_target.lua
```diff
@@ -0,0 +1,114 @@
...
+		end
+	end
+	
+	return true	
+end
+
+function gadget:UnitCommand(unitID, unitDefID, teamID, id, options, params, tag)	
+	local shifted = math.bit_and(options,OPTIONS_SHIFT) == OPTIONS_SHIFT
+	-- if cmd queue is empty, treat shifted commands as non-shifted (or else <GiveOrderToUnit() recursion is not permitted> may occur)	
+	if shifted then
```
KlonZK: this may have been due to bad return value in CommandFallback(), which is fixed in last commit.

LuaRules/Gadgets/cmd_auto_target.lua
```diff
@@ -0,0 +1,114 @@
...
+			unitHasManualTarget[unitID] = nil		
+		end
+	end
+	
+	return true	
+end
+
+function gadget:UnitCommand(unitID, unitDefID, teamID, id, options, params, tag)	
+	local shifted = math.bit_and(options,OPTIONS_SHIFT) == OPTIONS_SHIFT
+	-- if cmd queue is empty, treat shifted commands as non-shifted (or else <GiveOrderToUnit() recursion is not permitted> may occur)
```
KlonZK: this may have been due to bad return value in CommandFallback(), which is fixed in last commit.

GoogleFrog:
Why is this a gadget? I dislike this because gadgets are not user configurable and they should (almost always) not be directly configurable. It also feels like a small improvement to the target priority gadget would let you do this in a widget.

KlonZK:
the part where it works with queues is pretty much impossible to do without synced, also fully unsyched version would probably get called on bogus commands (empty area attacks, ...) and have some overhead.

it could be split tho and the part that needs no synch could be in a widget but data about existing targets needs to be shared, unless we just let the hard/soft target logic out and use config options again. id prefer to just keep this gadget because it needs no config.

anyhow just keep this open if you dont want to puil it and ill make changes next week if necessary.

KlonZK:
3 options to go from here:

- keep the gadget
- keep widget (cant work with queues

KlonZK:
basically there are 3 options:

- keep this
- keep the widget, it wont work with queues
- put the unsyched part into a widget and the rest into the target gadget. its just a few lines and likely doesnt affect anything outside the scope of this

PR #908, opened 2015-07-05
Okp imp v3

lhog:
Rough version of how OKP, that takes shields into account. will look like.
Still requires general clean-up and probably some testing (although basic one is done and results are satisfying), should be fully functional though.

Few questions:
1) How to perform tests best?
2) Do you still want to take into account that target might be repaired during missiles flight time or is recent slasher's OKP hax enough? (will be a separate branch if former)
3) I believe there could be few other units that can benefit from OKP:
3.1) I was initially thinking about artillery shooting at discovered static units, but realized that artillery is slightly inaccurate in that case, so it could be not applicable  
3.2) Merl seems to be accurate. Makes sense to OKPy?
3.3) How about assault units like reaper, goliath, ravager, thug, zeus(maybe) against statics?
3.4) Any others?
4) For whole thing to operate more smoothly, targeting gadget needs to be rethought. For instance targeting could flag a target with about to be expired disarm timer as a high priority for disarming weaponry. I have had it working it in my private branch, but the approach I took wouldn't scale to the whole state of the game, so I'm thinking to give it a try to rewrite it in a way, suggested by @ashdnazg . What do you think?


LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -285,6 +404,38 @@ function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOpt
...
+		
+		for shID, _ in pairs(shields) do --this will iterate through shield unitIDs
+			local xSh, _, zSh = spGetUnitPosition(shID)
+			
+			local shieldWeaponDef = shields[shID].shieldWeaponDefID --get shID's biggest shield		
+			local unitShieldRadius = WeaponDefs[shieldWeaponDef].shieldRadius --get its radius
+			
+			local unitsAround = spGetUnitsInCylinder(xSh, zSh, unitShieldRadius + addSearchRadius)
+			
+			for _, uId in pairs(unitsAround) do
```
GoogleFrog: Don't use pairs. unitsAround is a list indexed from 1 to #unitsAround

lhog: ipairs or 1, #n?

Anarchid: Suggested reading: https://springrts.com/wiki/Lua_Performance

GoogleFrog:
1) I do not know. You would need to look into profiling script.BlockShot because that is where a lot of the work in the gadget occurs.

2) If the hax is enough to fix regeneration cases then I would stick to the hax. I have not noticed any more cases which occur in practice. Unless some cases are discovered I would not worry about regen.

3) It depends on how likely it is that the shot will be intercepted. Impaler and Rogue arc so are likely to hit turrets whereas Rocko can easily hit something which is in the way. Perhaps assaults could have OKP with reduced reported damage, this would try to account for the times in which the shot is intercepted mid-flight.

4) Not sure, I would work on this first.

If you want some premature optimization look at reducing the number of tables you create in script.BlockShot. Also the pairs with relevantShields could probably be avoided by creating and using it as a list.

I doubt I will merged within the next two weeks as I am away soon and busy preparing. I don't want to add anything major for the week of reduced support.

lhog:
1) Ok, let's see how it works.
2) I meant I could track regen quite seamlessly within framework. It won't come for free, so let's stay where we are for now
3) OKP-imp-v4 already has Rogue, Thug, Impaler and Halberd OKP for static targets. Only thing that stops me from adding Reaper, Ravager, Goliath is lack of lua unitscripts. Levelers could benefit from ordinal OKP, however it's bos/cob - scripted as well. Also I'll take a look if any of ships may leverage OKP (Reef and Crusader come to mind).
4) Yeah it's middle range target. Though it was something I wanted to touch after OKP is more or less leveraged.

Few optimizations could be done probably, but first it's good to have some data on how heavy the whole thing is.

I'm not in hurry regarding the merge, in fact I'll be out next two weeks as well.

lhog:
Fighting with CMD.FIGHT targeting is hopeless. Engine automatically chooses nearest target, regardless of targeting priorities or hints I'm giving from gadget.
https://springrts.com/mantis/view.php?id=4916 needs to be done.

Same is happening with "Dont fire at radar" gadget/behavior, so that despite all the hints from gadget, a unit on FIGHT, having "shoot at radar dots" set to false, will still target radar dot if it's a euclidean closest target.

For now, racketeers on fight, that don't move, and situated closest to the same non-moving target will have their shots blocked, despite alternative targets could be available. Hopefully it's not a very common in-game situation.

GoogleFrog:
Tracking weapons cannot miss due to radar wobble so the last change should not be in place.

lhog:
Tracking weapons can miss due to radar wobble. The only thing that usually prevents them from doing so, is homing property of the projectiles. Racketeer's weapon type however is not good at tracking on radar and that was the main reason to limit area where OKP applies its power.

It used to be the case that *only* identified (once seen) targets were handled by OKP and I don't see why it was a good idea. Target could be seen few minutes ago in game by completely different ally units and still OKP had perfect information about it. Sounds like a cheating and unfair advantage to me. Approach proposed in 356a064 is better, so that OKP work only with information directly available to player.

If you think that idea has drawbacks or pitfalls, I'll be glad to hear them out and then either reconsider or propose quick fixes.

GoogleFrog:
OKP should come into effect against identified radar dots to prevent a Chainsaw wasting shots on a dot which is known to be a Gnat. It only takes 2 shots to kill a Gnat so should not shoot a third.

"Target could be seen few minutes ago in game by completely different ally units and still OKP had perfect information about it." - This is not true. A different allyTeam spotting a unit does not cause it to be identified to your allyTeam. Unless you meant different team, in which case it is irrelevant. OKP should not care about team at all.

It sounds like the valid issue here is that the health, armour etc.. of the radar dot is known. This can be solved by assuming that the unit is healthy with no status effect (and armoured if applicable).

MissileLauncher weapons track real unit location so they should have OKP whenever applicable.

StarburstLauncher weapons track radar dot location so should perhaps not have OKP on radar dots. They should possibly even have a radar dot fire toggle. Although if a StarburstLauncher is going to miss an unidenfied radar dot then extra missiles will not be helpful, they are all going to miss. So StarburstLauncher should have OKP on identified radar dots as well.

lhog:
```OKP should come into effect against identified radar dots to prevent a Chainsaw wasting shots on a dot which is known to be a Gnat. It only takes 2 shots to kill a Gnat so should not shoot a third.```
Agree, makes sense. I'll add per attacker UnitDef characteristic. Something like blockOnDot.

```"Target could be seen few minutes ago in game by completely different ally units and still OKP had perfect information about it." - This is not true. A different allyTeam spotting a unit does not cause it to be identified to your allyTeam. Unless you meant different team, in which case it is irrelevant. OKP should not care about team at all.```
I meant your flea could have seen that gnat 10 minutes ago for the sake of argument with half health, still OKP had perfect information about its state at the time of shooting. This concern is however in conflict with previous point as we are dealing with continuous game flow (1st case) vs discrete game flow (this case). One solution could be to implement "last seen on LoS" timer on targets, however I want to avoid over-complicating already complex gadget.

```It sounds like the valid issue here is that the health, armour etc.. of the radar dot is known. This can be solved by assuming that the unit is healthy with no status effect (and armoured if applicable).```
LoS approach had nice side property of solving this issue elegantly, if I'm to partially revert that, I'm not sure I should bother with assumptions above.

```MissileLauncher weapons track real unit location so they should have OKP whenever applicable.```
blockOnDot will be set to true

```StarburstLauncher weapons track radar dot location so should perhaps not have OKP on radar dots.```
blockOnDot will be set to false. Although I'm going to perform a test how starburst based weapons work with identified dots, that are buildings. So exception might be needed.

```Although if a StarburstLauncher is going to miss an unidenfied radar dot then extra missiles will not be helpful, they are all going to miss.```
Actually miss rate on radar dots is usually not 100%. As radar dots drift, it may/or may not hit the target. Thought I'm not sure why do you mention "unidentified radar dots", as those were never handled by OKP.

```So StarburstLauncher should have OKP on identified radar dots as well.```
I honestly don't understand how you came to that conclusion. Please rephrase/elaborate.

GoogleFrog:
LOS does not solve the issue elegantly. I suggested that the attacker should know the max health of an identified radar dot. Your solution seems to be to make some weapons use the current behaviour and some weapons not used OKP on dots at all.

StarburstLauncher should have OKP on identified radar dots because two shots fired at the same time will either both hit or both miss. Additionally the current StarburstLaunchers have high flight time so units are able to enter LOS after the shot has fired. If OKP assumes full health for radar dots there will still be plenty of shots going for large targets.

lhog:
Ok, you seem to dislike LoS idea. I don't really agree with that, but it's a minor thing to waste my and your time on, so I'm going to revert 356a064.

GoogleFrog:
I'll just fix those bits myself

GoogleFrog:
Oops, I thought I had pushed mine. I'll merge with the UI.

lhog:
Thank you. I left a comment regarding your changes in the relevant commit.

As I was mentioning before there is another follow-up OKP-imp-v4 branch that adds OKP to several precise assault units (targeting statics). Also here is my OKP targets wishlist:
*Buoy - static
*Panther - all, but fast
*Leveler - all
*Ravager - static
*Reaper - static
*Goliath - static
*Claymore - TBD
*Raven - static or consider dive altitude vs target speed/jump
*Hermit - static
*Moderator - all, but fast?
*Skeeter - all, but fast?
*Crusader - static
*Enforcer - all, but fast (and medium possibly)

Makes sense?

PR #910, opened 2015-07-07
Add coin awards for minor community events.

sprunk:


Licho1:
For consistency, player representation should be identical in game and on site. Those changes should be mirrored.

Anarchid:
Site actually mirrors award images from game repository since very long time ago (used to be just links into google code repository, now is a teamcity copy task).

Anarchid:
Correction to above: actually, site didn't mirror awards images *then*. It will, now.

PR #934, opened 2015-07-20
Target on the move & auto target

KlonZK:
target on the move:
- moved execution of commands to CommandFallback
- target commands can now be qued like any other command
- setting target while an attack command is active should now override the attack command
- reports to UnitCmdDone now. to make this work for target commands that are given while another command is active (but without shift, ie. the normal usage case) are inserted into first que position via CMD.INSERT. this makes the normal usage of set target more expensive

auto target widget:
- always cancels target on fight, patrol
- cancels target on stop, guard only if the target wasnt set by the user 
- will always set target when an attack is given, and thus override existing target. 

right now the only thing that uses the UnitCmdDone callin for target commands is the widget, so i am unsure what the value of this feature is and i dont know how big the performance impact of that insert operation is.

other than that, there seems to be a bug that i need to find before you can pull this, if you intend to do so.


GoogleFrog:
There seems to be a merge and I am not so sure that the whole change is required. CMD.INSERT for the normal use case sounds expensive. Queuing set target commands would need to work with the command queue visualization.

If you want queuable set target commands why not intercept non-shifty commands in AllowCommand and handle the queued ones in CommandFallback?

PR #937, opened 2015-07-22
Cleaned up index mistakes in gui_lasso_terraform.lua, added terraTag …

lhog:
…and few supporting attributes. terraTag is an optional terraform action identifier. It could be set and stored on a widget side and then retrieved/compared upon receipt of UnitFinished(of terraunit) or custom callin called TerraformUnitCreated. The use of this attribute (or strictly speaking UnitRulesParam) will allow for better interaction of terraform with follow-up construction events.

The problem with previous terraform implementation was that resulting terraunit didn't give any hint of what kind of work it was. It was not possible to know the shape, the height, the type or the location of the actual terraform construction volume. Now it'll be possible to match work issued through widget with custom unit parameters (UnitRulesParam) set by gadget through use of the shared terraTag value. The proposed implementation adds WG.Terraform_GetNextTag(), that is a simple value autoincrement function. Other widgets might either use it or supply own values.

The use case would be to:
1. obtain terraTag on widget side,
2. send it as a part of CMD_TERRAFORM_INTERNAL parameters,
3. use it as a hash key for storing any relevant information widget author would like to store (e.g. terraform type, height, shape, coordinates, followup construction, etc), 
4. next use either UnitFinished(check fo unitDefID==terraunit) or custom callin called TerraformUnitCreated, read/match terraTag;
5. retrieve info stored in the mentioned hash, do some follow-up action.

The use of terraTag is optional, so it won't break any 3rd party widgets.
Additional attributes "terraTagSegment" and "terraTagSegmentsCount" are set to values >1, if a terraform action, identified by terraTag, requires creation of more than one terraunits. Respectively if a widget author foresees such occasion, they should check if all terraunits have been created, finished, progressed to particular stage, or whatever they would like to check.

### DIFF
--- a/LuaRules/Gadgets/unit_terraform.lua
+++ b/LuaRules/Gadgets/unit_terraform.lua
@@ -16,7 +16,7 @@ end
 
 -------------------------------------------------------------------------------------
 -------------------------------------------------------------------------------------
-if (gadgetHandler:IsSyncedCode()) then
+if (not gadgetHandler:IsSyncedCode()) then return end
 -------------------------------------------------------------------------------------
 -------------------------------------------------------------------------------------
 
@@ -185,6 +185,8 @@ local currentCheckFrame 	= 0
 local terraformUnit 		= {}
 local terraformUnitTable 	= {}
 local terraformUnitCount 	= 0
+
+local terraTagInfo 			= nil
 
 local terraformOrder		= {}
 local terraformOrders 		= 0
@@ -396,11 +398,11 @@ local function getPointInsideMap(x,z)
 	return x, z
 end
 
-local function setupTerraTag(unitID, terraTag, i, count)
-	if terraTag then					
+local function setupTerraTag(unitID, terraTag, segment, segmentsCount)
+	if terraTag then
 		spSetUnitRulesParam(unitID, "terraTag", terraTag)
-		spSetUnitRulesParam(unitID, "terraTagSegment", i)
-		spSetUnitRulesParam(unitID, "terraTagSegmentsCount", count)
+		spSetUnitRulesParam(unitID, "terraTagSegment", segment)
+		spSetUnitRulesParam(unitID, "terraTagSegmentsCount", segmentsCount)
 	end
 end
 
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Looks good but where were the index mistakes?

setupTerraunit had the terraTag parameter added to it but the parameter is not used.

I dislike the inclusion of TerraformUnitCreatedWrap because it is unused and probably unnecessary. Surely you could just use UnitCreated.

lhog:
1. There were redundant few rendundant i = i + 2 in lasso
2. My bad. At one point I was thinking to place a tag inside the setupTerraunit, but I ended up placing it in the separate function. Will be fixed
3. My first impression would be bad as well, but there is reason. UnitCreated is fires as early as Spring.CreateUnit, so there is no way I can add terraTag and others custom params prior to the UnitCreated call. Thus TerraformUnitCreated is a workaround to this issue. TerraformUnitCreatedWrap is needed cause I couldn't find any way to setup a call to some widget callin from synced part of gadget. I'm willing to redo that part if you have better ideas on UnitCreated or synced gadget --> widget::something issues.

GoogleFrog:
3. Why not store the created terraform in the widget and read the custom params one frame later? I still do not see what this is for.

lhog:
For a single terraform task you can do it, like you do in lasso... widget. Your implementation there has some non-intuitive code, including taking a player's delay value (ping) into equation. Next a GameFrame needs to be involved, which is not an easiest call-in to deal with (see Svatopluk's code, where he performs query every frame without followup callin removal). The way I see it, the proposed event-based method is more convenient and much easier to use.

For multiple terraform tasks issued same frame (tasks on pause or area tasks, like area-mex), the followup is trickier, you will never know what terraUnit has been created one frame later or whenever, because the only meaningful attribute it has is unitID. I fail to see how it could be resolved, without having a parameter, shared between widget and gadget, thus terraTag.

GoogleFrog:
terraTag seems good. The two problems are:
 * terraTag is not used in the code.
 * UnitCreated should be able to detect terraunits instead of the new script action. An offset of a single frame should be sufficient to get the unitRulesParams.

PR #940, opened 2015-07-26
Lua unitscripts for Leveler and Ravager (mostly copied from banisher) + small bugfix

lhog:
For review and not necessarily for merge.

### DIFF
--- a/scripts/corlevlr.lua
+++ b/scripts/corlevlr.lua
@@ -1,0 +1,258 @@
+include "constants.lua"
+include "RockPiece.lua"
+include "pieceControl.lua"
+
+--------------------------------------------------------------------------------
+-- pieces
+--------------------------------------------------------------------------------
+
+local base, body, turret, sleeve, barrel, firepoint,
+			rwheel1, rwheel2, rwheel3,
+			lwheel1, lwheel2, lwheel3,
+			gs1r, gs2r, gs3r,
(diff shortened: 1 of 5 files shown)


GoogleFrog:
The wheels clip through their wheel housings which is a problem. Also when the units fire the wheels do not show any suspension.

lhog:
Regarding 1st point. Is it for all units or some specific ones? Can you do a screenshot?
2) If you/anyone knows how to fix both, you/that person better do that. I'm not very good with unitscripts.

GoogleFrog:
The second issue is probably caused by the rock script not changing the location of the wheel contact pieces.

lhog:
I've done some experimenting few days ago and found out that, Rocking base piece of model causes change of wheels' vertical position at least. And this is what suspension script is looking for.

Also the issue is probably caused by difference in explicit Rock scripts parameters such as force and speed and "magic constants" that are used to calculate ztilt and xtilt. I removed x/z tilt handling few days ago and got sane results. When I have time I'll try to put them back but make them aligned so one effect does not contradict with other.

GoogleFrog:
Look at the piece hierarchy. It may need changing if the wrong piece is being rocked or you require a stable piece.

GoogleFrog:
I don't think this is going anywhere.

PR #951, opened 2015-08-03
improved Attrition Counter

KlonZK:
- works in team games (counts attrition rate of team vs team)
- tracks info about individual players' losses
- new layout

http://imgur.com/IIZEnDH


GoogleFrog:
The settings are wrong. This looks like a HUD Panel to me, not a Misc.

Why is updateFrequency in settings?

In GameFrame you can update with a certain frequency with the check updateFrequency %n == 0.

Why are allyTeams and teams metatables?

KlonZK:
updatefrequency is in settings because it could conceivably arise the situation where someone wants the update to kick more or less often. it is in misc because its not a default hud panel? idk.

why should the tables not be meta? its just a convenient solution to initialize new entries. it shouldnt have any side effects?

Anarchid:
I guess it should self-disable in FFA after all, since i'm having great fun spectating a "Jasper vs OldGhostStalker" which in reality is four-way FFA, and neither of those are winning

PR #957, opened 2015-08-08
Fix wreck levels for Domi-based facs and Impaler.

sprunk:
Convert these facs' scripts from Cob to Lua.
Fixed Amph Fac wreckage not being blocking.


GoogleFrog:
Did you check that using explodables like that looks good? The Mex change had a similar explodables table which caused the explosion to look bad and I had to fix it.

Did you check the amph factory is not blocking for a reason? Can ships path over it when it is underwater?

sprunk:
Explodes look good.

I have tested feature blocking and they only block ship pathing if they peek over the surface. Submerged features won't block ship pathing.

Blocking wrecks do block construction even on the surface above them. However this is consistent with the live unit (it also blocks construction on the surface above, even if submerged deep enough not to be a problem otherwise) and with wreckage of most other water buildings (for example fusions, but also other factories: Gunship and Air).

Admittedly, Shipyard and Hoverfac wrecks are not blocking, but one is not buildable on land and the other is pretty much flat though so they aren't as bad. Amphfac is taller and bulkier so units and buildings clipping through it look very buggy.

GoogleFrog:
Many ship, hovercraft and amph units have non-blocking wreckages. This is probably from a time where underwater wreckage blocked surface movement. So if underwater wreckages do not block surface movement then most of those wreckages should become blocking.

Dagger and Duck should not block because there seems to be a loose rule that small units do not have blocking wreckages.

PR #958, opened 2015-08-08
Fix UnitDestroyed being called twice when speccing

sprunk:


KlonZK:
the error used to occur when not in specFullView. i think this will still happen with 

(not (spec and specFullView) and spGetUnitLosState(unitID, spGetMyAllyTeamID()).los)
? 

(if in spec with selected player, this resolves to true)

PR #964, opened 2015-08-10
Revert "Revert "Okp imp v3""

GoogleFrog:


GoogleFrog:
The repository regarding this gadget is all messed up. I thought I had pushed my small change and then it was dealt with in the fork.

lhog:
I'm not sure what happened to my PR of OKP-imp-v3. It says branch was merged although,  what is in the master now is not the latest code from that branch nor correct one. It's rather a modified OKP-imp-v2.

GoogleFrog:
I merged it too early. So you need to somehow open the same PR again.

lhog:
I don't know what you made exactly, but now github thinks that changes of OKP-imp-v3 have already been incorporated into the master branch of code. Creation of the new branches doesn't help either. It's likely to search for commit IDs and it finds them all, although they were all reverted later (which apparently is not taken into account).

The only way I can think of to fix this is to edit a project history and take wrong commits out. Though git experts (I'm not nearly one of them) don't recommend doing that for public reps.

GoogleFrog:
I think I scared lhog away so nothing is going to be done here. I still have not seen shields be a problem for OKP.

PR #968, opened 2015-08-11
The Autorepair System commander level requirement has been changed to…

Sgathec:
… two to match the value in the database.


sprunk:
This file has no effect. It has tons of other wrong values but they aren't actually used anywhere.

Sgathec:
Alright.

PR #970, opened 2015-08-12
added UnitDestroyedByTeam() callin

KlonZK:
reveals teamID of attacker to player(if within los) and to specs(always).
unlike UnitDestroyed(), specs will always receive full information no matter what mode they are in.

i want to add radar support for both this and the original callin. meaning that if an enemy unit that was previously identified
a) dies within radar los, you get information about that kill. 
b) kills one of your units, you get the attacker id

and c) if an unidentified enemy unit dies in radar los, you get informed about that something died, but not what.

a and c however reveal a bit of information that the player doesnt necessarily know: if the unit actually died or just disappeared from the radar for some other reason. 

regarding the widgethandler change, it seems necessary to do that or register the callin in some other way. without this the engine will complain about unlinked callin and not call it at all.
conceivably the whole thing could use some other way of communication if changes to the handler are not wanted.

### DIFF
--- a/LuaRules/Gadgets/api_widget_events.lua
+++ b/LuaRules/Gadgets/api_widget_events.lua
@@ -3,7 +3,7 @@ if gadgetHandler:IsSyncedCode() then return end
 function gadget:GetInfo() return {
 	name      = "Widget Events",
 	desc      = "Tells widgets about events they can know about",
-	author    = "Sprung",
+	author    = "Sprung, Klon",
 	date      = "2015-05-27",
 	license   = "PD",
 	layer     = 0,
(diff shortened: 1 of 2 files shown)


KlonZK:
what also could be conidered is to add attacking unit id and def id, if that information would be avaibale to the player. possibly some other widgets could use it.

GoogleFrog:
Apparently this does not affect UnitDestroyed. It adds UnitDestroyedByTeam which has unitTeam and attackerUnitTeam. It always triggers for spectators and only triggers for players when both units are in LOS.

GoogleFrog:
I dislike a), b) and c) because radar dots can disappear for many reasons and projectiles from outside LOS could have come from anywhere.

KlonZK:
regarding b), you only get team info anyway, so projectiles coming from things that you have on radar likely can be tracked to their owner by the power of player observation. tho there some ffa cases imaginable where this wouldnt be possible.

PR #976, opened 2015-08-13
fix #975: draw morphed slave's leash to master

### LINKED ISSUE #975: Morphing captured units removes the line indicating connection with Domi
Captured units after morph now correctly maintain their capture-lineage since it was fixed in https://github.com/ZeroK-RTS/Zero-K/commit/c4b1d3035a83dea7425916fefc64d961aacf2ad0 and https://github.com/ZeroK-RTS/Zero-K/commit/0f055bb3d917ee3fedb3e4f4a35ab45e291cc015, but the white line indicating connection with controlling Dominatrix is no longer drawn afterwards.

sprunk:


sprunk:
got a better solution (will implement tomorrow)

PR #985, opened 2015-08-15
buff welder dps by 50%

KlonZK:
to make it win vs a single scorcher. 

this may help tank lab a bit to get through the early game without having to rely on panthers or static defenses as much, both of which are expensive and of limited effectivity - and slow down expansion even more.


sprunk:
The PR actually contains a 50% buff.

KlonZK:
that is true, i lied. its 50%.

PR #986, opened 2015-08-16
Area mex: queue mexes at real height

sprunk:
Ctrl+Shift (building surround) works at the requested height while the rendering was done at ground height which led to discrepancy where hovering over the queued mex did nothing but a random point below worked.


GoogleFrog:
Who would have thought passing the y parameter in Spring would ever do anything.

PR #987, opened 2015-08-16
Ctrl+Repair now does pure repair (no construction)

sprunk:
Will only repair live units. Does not cause construction assist.

Spring supplies this functionality natively through space (ie. META key) but Insert overrides that without providing an alternative.
This patch replaces native Ctrl behaviour with the native META behaviour for repair.
Reclaim is handled by Insert the same way (engine META behaviour is now achieved through Ctrl, engine Ctrl not accessible).
Native behaviour of Ctrl+Repair causes repair not to ignore units currently being reclaimed which I don't think is of much use.
For comparison, native Ctrl+Reclaim was used to reclaim "own" wreckage like Dragon Teeth.


GoogleFrog:
So in short:
 * Area Repair with Ctrl on LMB release "Don't ignore units being actively reclaimed" -> "Only repair live units"?

sprunk:
Yes.

PR #988, opened 2015-08-16
fix #980

### LINKED ISSUE #980: Morphing comms doesn't keep their Priority, Misc Priority and Retreat states
Instead every one of them behaves differently:
- Priority is reset to whatever is set in Initial States settings for commander
- Misc Priority is reset to Normal (cant be set in Initial States settings)
- Retreat is set to No Retreat (ignoring what is set in Initial States settings)

Fire State, Move State and Repeat are kept between morphs, so I think others should be too.
Initial Sates should be only taken in account for level 0 comm, and morphiing should keep whatever state the unit has actually set (as with Fire State, Move State and Repeat already).

sprunk:
Morph keeps custom states (priorities and retreat)

Untested so far (will do tomorrow most likely)


GoogleFrog:
I'm going to wait for testing.

sprunk:
Tested and fixed

PR #990, opened 2015-08-16
Added on/off button for Aegis and Aspis

lhog:
Is there a reason on/off button has been not available to area shield units?

http://zero-k.info/Forum/Thread/19856


GoogleFrog:
This needs to be done as part of the attributes system which I am currently writing.

PR #993, opened 2015-08-17
Clippy comments updated slightly

Skasi:
I tried to make them sound less like an annoying smartass.


Skasi:
Still not perfect, but I think clippy should be disabled by default anyway until all the conditions are fool-proof. There should not be any situation where hints can potentially be wrong.

Examples:
Second factory warning in first few minutes: What if people plopped a land factory on a sea map, or a vehicle factory in a crater?
Excess warnings don't consider allies and don't check for actual excess.
Superweapons are viable 2 minutes into the game on speedmetal.
Defense spam is actually op.

PR #1003, opened 2015-08-19
Sniper: replace 10-elmo jammer with stealth

sprunk:


GoogleFrog:
Weird. Remove the stealth entirely. Cloak automatically applies stealth (some widget) and cloaked units are not supposed to be stealthed when not cloaked.

sprunk:
see #1006

Licho1:
Yeah remove the stealth flag. Issue is that stealth = true => perma stealth. We don't want that...
We only want cloak which also auto applies stealth in gadget

PR #1009, opened 2015-08-20
prevent eardrum mutilation

sprunk:


Licho1:
wont happen

let4be:
Why? If feature is required by 10% minority it should be disabled by default imo

PR #1010, opened 2015-08-21
remove "i choose you pikachu" spam

sprunk:


Licho1:
I like comedic aspect of this. Voting against merge..
Don't want ZK to become sterile, serious environment...

GoogleFrog:
I should go to chat. Just not TTS.

sprunk:
What's the point of spamming the chat with info nobody cares about?

Licho1:
Nobody cares? If nobody cares why people spend time/energy inventing original/fun names?
You should speak for yourself . .you don't like them, others apparently do..  Personally I like to listen to fun names...

GoogleFrog:
I care if I am on a team or spectating someone with comm names I know. It can be useful and is not harmful when just sitting in the chat.

Licho1:
Ok i will block it from TTS

PR #1017, opened 2015-08-23
[untested] fix #1016

### LINKED ISSUE #1016: Disabled units can be built using Initial Queue
Units disabled using "setoption disabledunits=unitname" can be queued in pregame using Initial Queue, and they get built without problems.

To test use "setoption disabledunits=armsolar". Before placing start position use command menu to queue a few Solars with shift. Notice the Solar button is not grayed. Now start the game and watch the comm building Solars... 
Bring command menu again - notice the button is now correctly grayed and no more Solars can be placed.

sprunk:


CarabusX:
Would be nice to have the units grayed, not completely removed.

sprunk:
Graying out engine-side is current behaviour (which fails) so the ways to do that are either:
* using this PR and spawning fake unclickable gray buttons for removed units in the menu
* keeping engine gray-out and intercepting locked build orders in AllowOrder; as a Lua solution this could lead to potential unforeseen hax (like currently).

PR #1021, opened 2015-08-25
Ease local icon modding [don't merge]

sprunk:
Each constructor now has a separate icon category (they all still look the same, the M).

This is to allow local modifications to affect each one individually (which is needed for chassis-based icons: see #537).


GoogleFrog:
I dislike how the backup constructor differentiation icons were deleted.

sprunk:
"Backup"? But sure, they can be restored.

sprunk:
Readded the chassis-showing icons.

I have discovered something that might be a better solution though so don't merge yet.

PR #1026, opened 2015-08-26
Revert "Change defaults for minimap LMB behaviour and Command Panel u…

Licho1:
Reverts ZeroK-RTS/Zero-K#1023

### DIFF
--- a/LuaUI/Widgets/gui_chili_integral_menu.lua
+++ b/LuaUI/Widgets/gui_chili_integral_menu.lua
@@ -112,7 +112,7 @@ options = {
 	unitshotkeyrequiremeta = {
 		name = 'Units tab hotkeys require Meta',
 		type = 'bool',
-		value = false,
+		value = true,
 	},
 	unitshotkeyaltaswell = {
 		name = 'Units tab can use Alt as Meta',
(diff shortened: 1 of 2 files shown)


Licho1:
It's maybe "genre standard" but so are other countless stupid things, like locked cameras.
Middle mouse scroll makes perfect sense, its self consistent with main map navigation and allows complex selections and commands to be issued through minimap easilly.

sprunk:
Is issuing commands through minimap worth it? In my experience minimap is too small (= imprecise) to give commands reliably and it's very easy to zoom-out instead.

Genre standards are quite important for ease of use. Minimap commands can still be re-enabled by pros even if they are no longer default.

dvr333:
Dammit Licho, do you want new players to play this or not? Everyone coming into this game now (since I'm guessing the Spring well has dried up) expects map movement via minimap click. I'm not sure selecting units is even practical to do precisely at speed without them being on the main screen, not to mention that most mice with MMB as a clickable scroll wheel make LMB much less physically strenuous to press and hold than MMB.

In general, it only makes sense to buck genre conventions and standards if what we are offering is (such as with camera zoom) a superset of those standards. Granted, for anyone not coming from StarCraft the camera system is already conforming to their expectations.

sprunk:
Remember the default is for newbies who don't need advanced usage (and minimap commands are advanced, it needs a lot of skill to make them more precise than our powerful zoom-out) but who are likely to be familiar with RTS standards.

The change would only affect defaults. There is already a toggle so veterans (or experimenting newbies) could still re-enable the setting if they wanted to.

Licho1:
We are discussing it internally, decision pending.

PR #1029, opened 2015-08-26
fix #1027 (rezzable outlaw debris)

### LINKED ISSUE #1027: Outlaw Debri Rezzable
This is probably due to the offmap Outlaw hax. This hax may no longer be necessary because I made an engine change (which got into 100.0) which gave projectiles either allyTeam or teams. They no longer have to rely on their ownerID to get their allyTeam for damage check purposes.

http://snag.gy/60tKy.jpg

sprunk:


GoogleFrog:
Needs a name.

PR #1033, opened 2015-09-01
Revert "Revert "Change defaults for minimap LMB behaviour and Command…

sprunk:
Reverts ZeroK-RTS/Zero-K#1026


Licho1:
2 out of 4 core devs don't want the change. It's not accepted

Anarchid:
Wait who's the second? Saktoth?

PR #1036, opened 2015-09-04
Fix #1020 (gunship brakes obey attributes).

### LINKED ISSUE #1020: Gunship Brake Rate should be Attributed
Currently it is missing. See
 https://github.com/SpringCabal/AlienResources/commit/1df43e8e6976b57a3d0cfc63ecb86547ee468294

sprunk:


GoogleFrog:
Are you sure accRate = 0 is safe? I would lower bound maxAccelerationRate by 0.001 so units are able to stop if fully move disabled.

sprunk:
Bounding for this value already exists:
https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_attributes.lua#L288

GoogleFrog:
Crap, this broke the speed to acceleration inheritance.

sprunk:
? Both values, max speed and acceleration are calculated pretty much the same, with the exception of the selfFactors which are fully Lua. The only use where only one of them was used but not the other was transport slowdown but this patch makes that use case apply both forms of reduction.

PR #1038, opened 2015-09-06
Shrink + move loading bar

KingRaptor:
This makes sure it doesn't overlap with the load screen tips.

Preview: http://i.imgur.com/HDTfst1.jpg

### DIFF
--- a/LuaIntro/Addons/main.lua
+++ b/LuaIntro/Addons/main.lua
@@ -13,6 +13,7 @@ if addon.InGetInfo then
 end
 
 ------------------------------------------
+local BAR_SCALING = 0.85
 
 local lastLoadMessage = ""
 local lastProgress = {0, 0}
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Are you sure that doesn't overlap any loadscreens? The half-hearted offset looks poor, I would instead fit the whole bar between the left side of the screen and "Spring 100.0".

KingRaptor:
Oops, it does overlap the movestate one (doesn't touch the text, but does cast a shadow on it) and the command queue one (does touch the text). Well, easy enough to shrink further (just change one variable).

It'd need to be pretty small to fit between the left side of the screen and the Spring version number text though...

KingRaptor:
Here's one that almost just fits in the gap. It's a bit annoying because the current scaling method causes text aliasing issues.

http://i.imgur.com/c1FoL6N.jpg

Anarchid:
A cunning alternative would be to draw it vertically along the left edge.

GoogleFrog:
Looks reasonable to me. It should just move slightly left so the bar looks well placed between the screen edge and Spring words.

PR #1039, opened 2015-09-06
Zenith tooltip: at 500, say it's maxed

sprunk:


ashdnazg:
Maybe "Controls n/500 Meteors", so you know the limit comes in advance?

(even though it's still not entirely clear they start falling in your own base)

Licho1:
+1 to @ashdnazg

sprunk:
done

PR #1040, opened 2015-09-06
"Exit to Desktop" does not cause resign

sprunk:
See http://zero-k.info/Forum/Thread/20014


Licho1:
Imo menu should be specific about this option "resign and exit to desktop"
or something like that

sprunk:
Sometimes people want to exit without resign, for example because something is bugged and needs restart.

mojjj:
or ragequit -.-

Licho1:
I guess that if auto AFK kicks in instantly after exit this is fine..

GoogleFrog:
"I guess that if auto AFK kicks in instantly after exit this is fine.."

This was done because auto AFK no longer kills teams.

Licho1:
I mean if AFK transfer happens instantly, which it does so all is fine.

PR #1049, opened 2015-09-14
Fix attrition

Anarchid:
Rewrites attrition counter team detection wholesale, uses modified specpanel naming scheme with additional top player detection; now works for ai vs player, ai vs ai, player vs player and varying team sizes.

The team description generator function should probably be moved to an utility include. The way i've hacked it should keep it more or less compatible with specpanel aside from storing player subtables rather than only player names.

### DIFF
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -434,7 +434,7 @@ function widget:Initialize()
 		label_self.y = label_self.y+18;
 	end
 	
-	if not allyTeams[myAllyTeam].name:find("\n") then
+	if not allyTeams[enemyAllyTeam].name:find("\n") then
 		label_other.y = label_other.y+18;
 	end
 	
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -202,78 +379,65 @@ function widget:Initialize()
...
+	if not allyTeams[myAllyTeam].name:find("\n") then
+		label_self.y = label_self.y+18;
+	end
 	
-	--Echo("Friend team is: "..allyTeams[myAllyTeam].name..'with '..allyTeams[myAllyTeam].numPlayers..' players')
-	--Echo("Enemy team team is: "..allyTeams[enemyAllyTeam].name..'with '..allyTeams[enemyAllyTeam].numPlayers..' players')
-			
-	if spectating then
-		Echo('<Attrition Counter>:Running as Spectator'); -- this doesnt seem to do anything
+	if not allyTeams[myAllyTeam].name:find("\n") then
```
FuckYouMeatLovers: if not allyTeams[enemyAllyTeam].name:find("\n") then

FuckYouMeatLovers: if not allyTeams[enemyAllyTeam].name:find("\n") then

Anarchid: Thanks for noticing!

PR #1053, opened 2015-09-17
Revert "Mangualfire can custom formationed." to fix #1052 (dgun vs air)

### LINKED ISSUE #1052: Manualfire command clicks through units
It is not possible to target any units with manualfire (e.g. dguns) if they have any ground behind them, because the command instead gets targeted on the ground under the cursor.

Targeting units against the sky works. 

**The problem disappears if LuaUI is disabled**. `/luarules disable`, on the other hand, has no effect.

There were commits related to click-through on some UI elements in the past months. They should be investigated.

Nice picture to make a dull report less boring: a Dante successfully releasing a napalm salvo against a Brawler directly against it on condition that order is given with the Brawler rendered against the sky.
nabaldgun

Anarchid:
Since commit 25185fb19d97853e4d44733837f03a8518048751, it is impossible to target any manualfire commands against air units if the screen ray clips any ground (#1052).

 This reverts the commit.

### DIFF
--- a/LuaUI/Widgets/cmd_customformations2.lua
+++ b/LuaUI/Widgets/cmd_customformations2.lua
@@ -78,7 +78,6 @@ local formationCmds = {
 	[CMD.MOVE] = true,
 	[CMD.FIGHT] = true,
 	[CMD.ATTACK] = true,
-	[CMD.MANUALFIRE] = true,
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Bad fix.

PR #1056, opened 2015-09-18
Firewalker "Artillery/Skirmish" -> "Artillery/Riot"

Skasi:
Small adjustment to Firewalker's short description. It's some sort of Artillery/Skirmish/Riot, but I've decided to leave out the "skirmish" part because most artillery can be used as a skirmisher anyway and "Artillery/Skirmish/Riot" sounds silly. Also because Firewalker is (in my limited experience) definitely not a "give a fight command to and forget about it" unit like many skirmishers can be (unless you don't mind Firewalker shooting at your own army/base or not shooting at all because of "Hold fire").

On the other hand, the "Riot" part is very important. Firewalker excels at taking out large groups of cheap units/buildings, that's probably its biggest strength. It not having "Riot" in its description would be a big mistake.

### DIFF
--- a/units/firewalker.lua
+++ b/units/firewalker.lua
@@ -18,10 +18,10 @@ unitDef = {
   corpse                 = [[DEAD]],
 
   customParams           = {
-    description_de = [[Roboter mit Feuerunterst?zung (Artillerie/Skrimish)]],
+    description_de = [[Roboter mit Feuerunterstützung (Artillerie/Riot)]],
     description_pl = [[Robot wsparcia artyleryjskiego]],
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Firewalker is easily killed with raiders. Just spread out, wait for it to fire and then close in. It is not like the other units called "riot".

GoogleFrog:
I've added "Saturation" to the description. This adjective means that unit deals damage over a particularly large area. This can make it good vs clumps of light units but it does not necessarily mean it can fight them on its own. Tremor, Catapult and Phoenix have "Saturation" as well.

Anarchid:
Tremor is not very good at saturation, though.

PR #1061, opened 2015-09-22
Heatray sound change

Blackdutchie:
- Adds a new heatray sound file
- Updates sounds.lua to contain a new sound for heatrays, modulating pitch and volume slightly for each instance. This works to reduce the "horrible repetitive racket" effect.
  - Editing the pitchmod line to be a slightly higher number will reduce the racket effect further, but will     at some point introduce a "sci-fi laser" effect where it starts to sound quite silly.
- Edits Scorcher, Dante, DDM, and Commander Heatray to use this new sound.


Blackdutchie:
Example video. First the slashers demonstrate the original sound, then the scorchers demonstrate the new sound.
Sound quality is worse than I'd like, but gives a reasonable idea of what it sounds like.

https://www.youtube.com/watch?v=ytZb0lEgPPg

Anarchid:
It would be awesome to get of the eye-shattering heatray finally.

The old file is forever stored in git history, so there's no need in keeping it except for comparison.

GoogleFrog:
The new sound is too indistinct for me, just sounds like noisy steam.

sprunk:
It sounds like a constant stream of barrels rolling on a metal floor a few rooms across.

sprunk:
There seems to be a consensus that this is not better than current.

PR #1065, opened 2015-09-25
Add modifier keys to Area-mex command

sprunk:
* Ctrl: adds Solars around the mexes (same as ctrl+shift click)
* Alt: adds LLT

### DIFF
--- a/LuaUI/Widgets/cmd_mex_placement.lua
+++ b/LuaUI/Widgets/cmd_mex_placement.lua
@@ -328,7 +328,7 @@ function widget:CommandNotify(cmdID, params, options)
 			local mex = WG.metalSpots[i]
 			--if (mex.x > xmin) and (mex.x < xmax) and (mex.z > zmin) and (mex.z < zmax) then -- square area, should be faster
 			if (Distance(cx,cz,mex.x,mex.z) < cr*cr) then -- circle area, slower
-				commands[#commands+1] = {x = mex.x, z = mex.z, d = Distance(aveX,aveZ,mex.x,mex.z), mex = (not spotData[i])}
+				commands[#commands+1] = {x = mex.x, z = mex.z, d = Distance(aveX,aveZ,mex.x,mex.z)}
 			end
 		end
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I am not sure what Space is for. Wouldn't you just queue mexes and then later do a modifier to add solars to them? Space should be reserved for command insert.

sprunk:
The two approaches have a difference:
* queuing mexes and solars at once leaves you building mex-solars - mex-solars, each spot getting the full package instantly.
* queuing mexes, then solars without mexes, leaves you building all the mexes naked, then after that coming back and adding all solars.

But Space being reserved for Insert is more important after all and it's still possible to add solars to naked mexes without it (just do it after mexes are ready). Will remove the Space mode.

sprunk:
Removed space, updated first post.

lhog:
I don't think it's a good idea that you got rid of "handledExternally" check and following code.
I use to for areamex terraform.

Licho1:
Sounds rather too special to be default widget. 
Solars or LLT around mexes is highly situational and both can be achieved using alt space command insert after mexes are queued.

I don't see solars or LLT being a standard. Sometimes you want solars around some mexes, sometimes you want winds and you definitely want to control exact LLT placement.

Its ok change but should not be default on, it's not something all players do/have to do  (solars around mexes) and keybinds can interfere with other keybinds...

Alt and space certainly do a thing with command queues and should not be touched.

sprunk:
Modifier keys exist for other commands already so they won't interfere with keybinds. They are usually very advanced usage but it has never been a real problem even though they are all default on. For example if you hold Ctrl when giving area-reclaim you will area-reclaim your own live units.

I don't think alt does anything with building queues but for safety I'll remove this for now and only leave ctrl for solars.

I agree that solars might not be *the* standard but they are the closest thing there is to standard. Late-game you will almost always want energy around mexes for overdrive and the widget solves the problem of adding reliable E with minimal effort. It won't be used all the time and won't always be optimal but it doesnt have to be - normal areamex also isn't perfect (for example it doesn't prioritize ubermex) but it's useful enough.

GoogleFrog:
I think a more complete solution would be a building template system which can be area-mexed to put the template around each mex.

Licho1:
That would superior yeah.. i could imagine complex structures with terraform, which might become usable if some templates save from microing

sprunk:
The complete solution with templates has some problems to figure out (like how to define facing/rotation which is important, or how does the user actually define a template), areamex solars skip all that to provide a very simple solution (just hold a modifier key) for a common problem (getting a basic grid for all mexes).

It would be good to see how much people would actually use a real template system and what use cases they would find. Personally I can't think of many things you would want to have near each and every mex. Basic energy and maybe LLT are fine, but anything more than that is too expensive to build at each mex.

Licho1:
Oh I meant general templating, not for area mex, sure for area mex its overkill

PR #1071, opened 2015-09-30
Add notification for double comms

sprunk:
Adds an infobox that pops up pre-game if someone has double comm.


GoogleFrog:
This should probably be made to not mess with other docked windows. Also putting the notifier in the middle of the screen looks annoying. I would rather have a fully opaque "game rules" (or similar) window off to the side of the screen to display this type of information. It would also replace the current modoption notifier window.

Licho1:
Even better solution would be real time display, because this number can change (with AFK)
You can display it on playerlist maybe?

PR #1073, opened 2015-10-01
Reorder factories for better hotkey mnemonics.

sprunk:
After the patch, more factories will have their hotkey based on their first letter.

Previously:
[T] Gunship
[A] Spider
[S] Jumpy
[D] Tank
[F] Amph
[G] Ships

After patch:
[T]ank
[A]mph
[S]pider
[G]unship
[D] Jump
[F] Ship

Rest unchanged:
[Q] Cloaky
[W] Shield
[E] Vehicle
[R] Hover
[Y] Plane
[H] Strider

To also consider is [H]over but that requires moving the striders away from the last spot (which is meaningful somewhat). Perhaps they could get the last spot on the first row but that would mean planes get R which would be quite odd.


GoogleFrog:
The aim was not to group them mnemonically, it was to group by type. I don't think mnemonic has significant advantages. The current layout clusters:
 * Bot pathing.
 * All-terrain and jumpy (cliff avoidance).
 * Flying factories.
 * Seaborne factories.
 * Vehicle pathing.
I feel it is messy to mix up these clusters.

sprunk:
I think things having their "natural" hotkey (eg. T for Tank) is very elegant but on the other hand I agree that grouping helping newbies find the stuff they want is important.

Both systems have their advantages so perhaps it would be good if there was a toggle? That still leaves the question of what the best default would be though - newbies should learn the factory names about the time they learn their movetypes so better mnemonics might also help them to an extent.

One of the problems here is that the hotkeys are tied to layout (eg. leftmost thing of the 1st row has to be Q) so we cannot combine the natural hotkeys with group-based layout.

GoogleFrog:
I think such a toggle would be very rarely used. I also think it is a bad idea to mix grid and name-derived hotkeys. It would also be really annoying to current people if we changed their structure positions and painful to keep compatibility between two systems.

sprunk:
Fair enough.

sprunk:
Current above, new below (as proposed + plane/strider swapped) :

groups

The mnemonic-based layout would still keep the clumps. Moving Amph under A means the bots clump gets its outlier back to the clump, but sea loses clumpiness. I think that's a net gain: sea is a rare subdomain and the icon backgrounds already show seafaring.

I also consider Amph a good beginner fac. It's the most reliable fac in that it has the most versatile pathing so a newb would almost never be forced out of comfort zone due to things like hills or water. At the same time it doesn't have any complicated mechanics and I think is the simplest of the seafaring facs. With this in mind moving Amph to the left is also a slight improvement since that's where the beginner facs are.

So the clumps are not necessarily worse off and depending on how you look at it it might even be a slight improvement. This is before the benefits of mnemonics.

GoogleFrog:
I think its a matter of constraints which can somewhat be inferred from the factory build list layout. Here are some, I think they are in order:
 * Strider on H. This is because H is the key for weird utility units that don't fit with the roles of the other things in the tab.
 * Plane and Gunships adjacent.
 * Sea factories adjacent.
 * Ease/necessity such that the top row contains the most straightforward and 'vanilla' options with difficulty increasing from left to right.
 * Vehicle and Tank presented as a light and heavy version, Hovercraft nearby.
 * All terrain adjacent.
 * Bots adjacent.

With these constraints the order could be: 
 *Cloak, Shield, Veh, Hover, Gunship, Ship
 * Amph, Spider, Jump, Tank, Plane, Strider

I don't see how it is much better than the current system though. I like how the current system has a 2x2 block of hilly/cliff units. Moving Spider and Jump across by one slot removes that simple barrier between vehicles and non-vehicles.

In the end, I don't see any advantage of the modifications significant enough to outweigh the current playerbase-derived cost of moving everything around.

PR #1074, opened 2015-10-01
Make Solar buildable in knee-high water (depth 6)

sprunk:
Wind does not yet become tidal at this depth so there is no reliable basic E source.


GoogleFrog:
This would break consistency, afaik no other structure which neither floats or can be built fully underwater can be built in shallow water. I do not know of any maps in which this inefficient zone of -6 to 0 elmos is a problem. It could even be considered a feature.

Anarchid:
On IoG it's not possible to build tidals or solars in the shallows. 

It is impossible to build tidals because the depth is insufficient for windgens to morph into tidals.
It is impossible to build solars because they don't float.

Maybe a more radical version of this would allow all buildings to be built at depth 6, unless they require a minimum depth.

sprunk:
If GF agrees I can prepare a patch to make all ground buildings able to live in shallows.

GoogleFrog:
I suspect waves lapping over buildings would look silly. Why do you feel structures need to be buildable at this height range?

Anarchid:
The height range is very interesting for mappers because it is the minimum depth required for ships while only applying tolerable (though still pretty huge) speed malus to ground units.

Mexes placed on such terrain, however, cannot be overdriven with solars *or* tidals, removing a choice.

sprunk:
Inconsistency and looking silly with regard to placement already have a precedent. Windgen was recently made buildable absolutely everywhere which is different to all other buildings and looks stupid when on cliff side:
cliff_windgen

GoogleFrog:
I did not do the windmill slope tolerance change.

sprunk:
So, would making every land building able to live in shallows okay?
What should be done with windgen?

PR #1093, opened 2015-10-14
Win counter cleaned up to use gameover winners parameter

dvr333:


Licho1:
Why does it read from local lua table? Server could send you actual historic counts...

dvr333:
How does the server know to reset, such as in the second grand finals series? Who would have control over manual resets of win counts, or of correcting mistakes should they come up?

Licho1:
It could only detect that player A and B play just with each other (and not with someone else).
If either A or B plays with someone else -> reset.
I think it would work for tourneys right?

dvr333:
Ah, sorry, I should have mentioned that auto-reset code already exists, for the cases where the players don't all have the same teammates as the previous game. The reset issue is that those cases are the only things the win counter can know to auto-reset on, but cases related to tournament rules are not known. 

The main case is going from grand finals 1 to grand finals 2, that needs a reset, but the win counter sees the same players on the same teams, and can't know if the win counter needs resetting (and before you suggest a "first to X wins" configuration variable, that's more possibility for human error and thus more reason to require a manual fallback). The other cases have to do with players messing up how they leave games (before you suggest we control how players leave games, that won't work and would piss people off) or other human error, or possibly a mistake in code, but in a tournament that all needs a manual fallback.

Licho1:
Yeah but server does see it, because it sees all games, so it can reset it reliably.. thats why i mentioned that..

dvr333:
That's irrelevant, since the logic for resets works both locally and server-side. The issue is with the cases where the expected number of wins won't match up to the listed number, either because a tournament rule invalidated a win, technical problems resulted in a false win, or because we moved to a new tournament stage, but with the same players playing (usually between the first and second Grand Finals of a double elimination tournament).

None of these cases are covered by the algorithm, and only the last one has a chance of being covered by "first to X wins" configuration, which in turn leads to an inevitable misconfiguration and either a premature reset (ex. configured to first to 2/best of 3, but stage is first to 3/best of 5) or a missed reset. As such, a manual fallback is necessary, and so the only relevant question is who has control over that (since it would be chaos if everyone could mess with it, and it isn't really a question of a vote). I guess it could be a room boss thing, though I don't know how abusable that is

PR #1094, opened 2015-10-14
Area-mex: change solars modifier key to Alt

sprunk:
This is because of a proposal to make the Ctrl modifier to instead work like it does for Attack, ie. split up the cons among different mexes instead of all getting the same path.


Licho1:
Isnt alt for queue insert? - Should be instead of space..

sprunk:
Space is queue insert.

Alt is used as modifier for commands. For example with attack it means you can draw a line and with repair it means the command is persistent.

Licho1:
Yes but alt should be used for command insert, because its factory insert.
Its just legacy gap in standardization that should be solved eventually..

PR #1101, opened 2015-10-18
Add localisation to a bunch of small widgets.

sprunk:
Comes with Polish translations.

Affected are:

* healthbar labels
* Morph complete
* X is under attack
* Build ETA


Licho1:
Btw I have given up localization of website (due to massive text walls and maintenance cost) but I guess for unit info and some widgets its doable..

Anarchid:
Now that i've gotten quite handy with teamcity, i think i'll look into actually integrating transifex... eventually. >.<

PR #1102, opened 2015-10-18
Add an API for maps to define arbitrary startboxes

sprunk:
Fully backwards-compatible (ie. maps that don't define their own startboxes see no change).

Maps can include a file which describes the boxes (similar to the mex layout file). Being Lua code these can read the game setup to allow any arbitrary setups based on what the teams are (for example Throne could detect whether the FFA is 7-way or 5-way and expose either the first or the second floor to start on).

The boxes are defined through sets of triangles which allows any arbitrary starting area shapes (including concave and disjoint areas).

The mapper also defines a set of start positions for CAI and lagger purposes. Currently laggers drop in the middle of the box and CAI distributes itself evenly over it. With the new API a set of startpoints is used for that which makes sure laggers and CAI are distributed to sensible places (eg. near mex clusters and not on a cliffside).

### DIFF
--- a/LuaRules/Gadgets/ai_CAI.lua
+++ b/LuaRules/Gadgets/ai_CAI.lua
@@ -3849,14 +3849,29 @@ end
 
 local function setAllyteamStartLocations(allyTeam)
 	if Game.startPosType == 2 then -- Apparently this is 'choose ingame'
-		
+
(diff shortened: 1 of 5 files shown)


Anarchid:
You forgot the screenshot!

sprunk:
I'll make a wiki page on usage and with examples soon.

PR #1103, opened 2015-10-19
An implementation of wading wakes.

Anarchid:
This provides an include function with some ugly globals, which should be easily applicable to most LUS in ZK. 

It uses a poll to determine when unit is between `0` and `-unitHeight` altitude to spawn wake sfx of required kind. The poll is deactivated when unit is above `0` by use of `SetSfxOccupy`. Seeing how no unit currently uses `SetSfxOccupy`, the include takes full control of that callin.

It is applied to Grizzly, Angler, and Djinn here. For some reason, Djinn does not terminate the wake spamming on StopMoving while others do.

The end goal would be to apply this to every non-ship non-hover non-flying mobile unit in ZK.

The questions:
- is the ugliness really worth the performance savings of not running the poll above zero?
- is there a way to reduce ugliness without massive copypasta?
- why does Djinn refuse to terminate wake spam? I've checked the signals, and they match; there's no cross-contamination between individual unitscripts and Djinn script doesn't seem to use any globals that the include relies on.

### DIFF
--- a/scripts/amphtele.lua
+++ b/scripts/amphtele.lua
@@ -17,6 +17,11 @@ local SIG_WALK = 1
 local SIG_DEPLOY = 2
 local SIG_BEACON = 2
 
+SIG_WADE = 4
+WADE_SFX = SFXTYPE_WAKE1
+WADE_PIECE = {rfoot, lfoot}
+include "wade.lua"
+
 --------------------------------------------------------------------------------------
 --------------------------------------------------------------------------------------
 -- Create beacon animation and delay
@@ -258,9 +263,15 @@ function script.StartMoving()
 	deployed = false
 	GG.tele_undeployTeleport(unitID)
 	StartThread(Walk)
+	isMoving = true
+	if inWater then 
+		StartThread(Wade);
+	end
 end
 
 function script.StopMoving()
+	Signal(SIG_WADE)
+	isMoving = false
 	Signal(SIG_WALK)
 	StartThread(Stopping)
 end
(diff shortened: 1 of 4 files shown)


scripts/amphtele.lua
```diff
@@ -17,6 +17,11 @@ local SIG_WALK = 1
 local SIG_DEPLOY = 2
 local SIG_BEACON = 2
 
+SIG_WADE = 4
```
sprunk: Djinn doesn't stop because this is 4 instead of 64.

scripts/wade.lua
```diff
@@ -0,0 +1,35 @@
...
+
+function Wade()
+	local spGetUnitPosition = Spring.GetUnitPosition
+	local maxWadeDepth = -Spring.GetUnitHeight(unitID)
+	
+	if not (WADE_PIECE and WADE_PIECE[1]) then 
+		return 
+	end
+	
+	SetSignalMask(SIG_WAKE);
```
sprunk: This is wa*K*e instead of wa*D*e. Probably the engine only allows wakes to live on the surface though anyway.

Anarchid: Oh, good catch. So, somehow Sig 64 kills the thread in case of grizz and angler purely by coincidence, while djinn manages to expose it (probably because too few signals).

GoogleFrog:
Replace the global variables with setters. The include would have global functions WadeEnterWater and WadeLeaveWater.

Replace the globals set at the start of each individual LUS with a function WadeInitialize. Different files should only communicate via functions. Global variables which are required for extra files to work are too spaghetti.

GoogleFrog:
Also look into script signals. I hear there is a different signal mask system available to lua. Perhaps the include itself could contain a unique signal.

Anarchid:
Reimplemented as a gadget in less loc and without requiring conversion of COB and modification of every unitscript.

PR #1105, opened 2015-10-22
Startbox API improvements

sprunk:
* ease area definition: allow any convex polygons (was only triangles previously)
* add mod-side override
* add GetTeamCount() which is callable from the config and returns the real number of allyteams (Spring.GetAllyTeamList not usable because of empty teams and Gaia)
* mark the recommended startpoints during the pre-game
* allow the usage of math.random in the config
* expose the manual startpoints to WG
* improve performance and reduce code duplication

Also comes with some ready configs which showcase various things one can do using the power of the API:
* Speedmetal. The simplest config, just two rectangles.
* DSD. Still just 2 areas and no magic but the shapes are quite detailed. The typical use case.
* Throne. Shows how a map can pick between different sets of boxes based on circumstances.
* Isle of Grief. Shows how boxes don't need to be hardcoded - generates one box by mirroring the other.
* Speedball 10x10. Shows boxes which are generated completely algorithmically (slices the map like a pizza based on how many teams there are).


GoogleFrog:
The commits range over 5 days. Does that mean you do not want it to be pulled yet?

sprunk:
It's pullable.

Licho1:
Lua config files like these, with active lua code seem to be too hard to edit..
My guess is that it won't be used much due to lack of map profiles.
Maybe we could instead do proper web gui for drawing arbitrary shapes on site and store polygons in server infrastructure? 
All that would be needed then is a way for springie to pass those shapes to game.

PR #1108, opened 2015-10-23
Sharpshooter → Spectre

sprunk:


Licho1:
Bleh, dont, we call it sniper not "spectre"
Sharpshooter is awkward, sniper is not.. and otherwise whats the reason for rename? Just because 5 year old OTANames wiki article got ressed by import we don't have to go crazy on renames...

Anarchid:
> Just because 5 year old OTANames wiki article got ressed by import we don't have to go crazy on renames...

Maybe it shouldn't show wiki as fresh forum posts then, so that they wouldn't invite rezzing :)

PR #1119, opened 2015-11-08
DDM: require power for heatray

sprunk:
* is quant's (grid reliance is a weakness - disabled heatray allows to capitalize on it eg. by enabling raider usage)
* consistency: unpowered is now always harmless


Licho1:
All power reliance is atm meaningless, because it is possible to reserve power..
Now if it wasn't possible ..

sprunk:
But DDM power is grid, you can't reserve grid

Licho1:
Grid linkage does make sense other forms of energy dependence mostly not. They would make sense if energy was unreservable and it was not possible to disable units.

PR #1130, opened 2015-11-14
Wait pauses non-free Stockpile

sprunk:


Licho1:
I was hoping we would remove wait completely..
Its mostly for people to shoot themselves.

Issue #1067, opened 2015-09-26
Minimap appears too dark on many maps


Licho1:
Default minimap is barely visible and not useful for navigation.
Stuff outside los should be more visible.
http://i.imgur.com/5ChsgoA.jpg

Map itself should also be brighter.
MAp is Fetid Marsh 05

Licho1:
It's actually two problems
1) fog brightness default value is too low
2) minimap does not render sea properly

GoogleFrog:
Whether the fog brightness, minimap brightness and minimap sea is set correctly is largely up to the map. We should pick maps which do it correctly.

PR #1131, opened 2015-11-15
Increase Fog of War default brightness: 0.4 -> 0.5, reduce air LOS brightness

dvr333:
* Addresses issue #1067


GoogleFrog:
It is map dependent so this does not fix the issue.

dvr333:
0.4 is too low on every map I tested. Maybe 0.62 is too high on some maps, but 0.4 is definitely too low.

GoogleFrog:
I use 0.4 and it looks fine on many maps. For example Titan Duel is fine. It is important to make the fog dark enough to make air LOS distinguishable.

dvr333:
That's why I had it set to 0.4 initially, since air LOS was 1.5x as wide as ground LOS. With air LOS now equally wide (a great aesthetic change, btw. Not sure why, but it makes me more eager to play ZK), the main contrast is between ground LOS and no LOS, which can be really stark at 0.4.

How do you find.0.5 on Titan Duel?

GoogleFrog:
Air LOS is a fixed circle around all units, it is not affected by terrain. So air LOS is visible when you have ordinary LOS blocked by terrain. If, to take the extreme example, we drew air LOS at the same brightness of LOS it would mislead people because they would not be shown that terrain blocks vision. So fog has to be dark to make air LOS sufficiently darker than full vision.

Is there a problem with starkness of LOS boundry? I have not been thinking about starkness as a problem. The only problem I know of is whether the terrain is visible though fog or on the minimap.

dvr333:
The tricky part is that the LOS boundary should be stark, but in a lot of maps the FoW area seems darker than it needs to be for sufficient starkness, like Isle of Grief, Barren, or Zed. Other than the colour of the terrain I'm not sure what is map dependent about the look of maps in FoW. It's more noticeable on the minimap than the main map.

Licho1:
I think we should not display air los at all.
It exists only for two purposes: air to air visibility - people would assume that air high up sees farther naturally
Exploit fix - to reveal sunken turrets and firing arty.

Both of those are natural and become obvious/visible, so there is no need to indicate air los everywhere.

Showing it means only more confusion.

dvr333:
Okay, testing Titan Duel I see what you mean, but I'm kindof in agreement with Licho. I'm suspecting that air LOS could be 25% or even less of the total LOS brightness and still be quite visibly different from fog even at fog brightness of 0.55 or 0.6 on Titan Duel.

GoogleFrog:
What is 80/20? Have you removed air LOS in one of your images?

My aim has been the following: Players should know what is covered by LOS and what is not covered by LOS. In particular they should be able to distinguish things covered by LOS and things covered only by air LOS. Because LOS is not air LOS.

The for (at least most of) this ticket is that you cannot increase fog brightness because that would increase air LOS brightness and then people would not be able to tell air LOS from LOS. If you can change the shader to "80/20" then that issue is resolved.

Not drawing air LOS at all (that is, drawing it like fog) would solve the issue too but a small distinction between air LOS is probably worthwhile. It would be more for discoverability rather than ingame usefulness. Players would eventually notice air LOS and realize that it is a mechanic.

dvr333:
When I say 80/20 a mean that air LOS contributes 20% to the total LOS view. The images in the album show different proportions of air and ground LOS in the LOS display shader colour mix function, as mentioned in their titles. In other words, 80/20 means the air LOS display is 1/5th as bright as the full LOS display, rather than default of being half as bright.

In one of the images air LOS is removed, just to see how that would look. I prefer having faint air LOS as mentioned before, so if that works with a higher fog brightness for you (more like 0.5 or so than 0.62) then I'll update the PR with that.

Licho1:
I'm now 100% convinced there is no point in displaying this obscurity. Players dont know about it and its largely inconsequential, so no point smacking it into their faces this way.

PR #1132, opened 2015-11-15
All COFC camera actions respect smooth mesh scrolling option

dvr333:


GoogleFrog:
What do you mean all camera actions? Is this just COFC?

dvr333:
Yeah, sorry, All COFC camera actions

PR #1133, opened 2015-11-15
Fix ground collision damage

sprunk:
This reverts 602ca7182171926f8b6aaabf235980e6a4e39b30

The normal should not be reversed. This causes complete damage mitigation if the unit is colliding exactly along the normal. For a practical example see http://zero-k.info/Battles/Detail/384893

Explanation: let's say the normal is 0 / 1 / 0. Unit velocity is 0 / X / 0. ZK uses 0.5 multiplier for the tangent damage.
So the situation that should happen, the normal component is X * 1 == X, and the tangent component is X-X == 0. The damage is X + 0.5 * 0 == X.
Now with the normal reversed the normal component becomes X*(-1) == -X. The tangent thus becomes X - (-X) = 2X. The damage is (-X) + 0.5 * (2X) == 0.


GoogleFrog:
The purpose of the change was to fix the normal direction. I've made a change which fixes damage without re-breaking direction.

PR #1153, opened 2015-12-07
Improvements to Unit AI

mtroyka:
Major Changes:
-Scorchers dive at their targets on fight move and drive in small circles around the target once at close range.

-Air units no longer nochase versus solar panels and razors. This was especially bad with gunships, which would completely ignore them even if they were the only targets around.

-Swifts no longer nochase versus ground units. May be reverted if people hate it enough.

Minor Changes:
-Arty now disprefers shooting at raiders.
-Spectre, Penetrator and Racketeer now disprefer shooting at cheap porc. (in addition to raiders)
-Rackets disprefer shooting at unarmed stuff.
-Raiders prefer to shoot at mexes above other things.
-Added solars as a universal bad target in target priorities.

Note: I did not get great results with target priorities. Target prio only handles weapon targeting and not unit targeting/chase behavior, which leads to inconsistent results. It also has some nonsensical preferences, for example having rackets target unarmed things rather than overkill something armed but already disarmed. Clearly rackets should have a preference for what armed targets they should overkill rather than disarming something that it's more or less useless to disarm. I'll work on that when I have time to tear it apart and figure out how to make all that work properly.


timo:
i wonder if tactical AI could make a difference between closed solars and opened solars, so that opened solars are prefered over closed ones?

mtroyka:
That wouldn't make sense, since solars close immediately after taking any damage.

timo:
My thought was that a group of glaives would first close all solars next to them in any order and then shoot the solars until they are destroyed (or when they open up again in the mean time). this would maximize damage to the enemy's energy economy without costing too much time (assuming glaives wouldn't run off to find new open solars when all nearby ones are closed)

Anarchid:
High-alpha stuff can kill solars which are not closed sometimes.

mtroyka:
After watching this for a while I'm actually pretty happy with how it works. Rackets really don't attack unarmed stuff unless there's nothing else around, snipers and penetrators are noticeably better about not wasting shots on stupid stuff, raiders are noticeably smarter about targeting mexes and raiding is less frustrating overall.

On the down side hammers haven't really improved much, although it's still better than it was, and target priorities don't work quite as well with gunships since they don't use turreted weapons, but besides that lots of units are smarter. It's good enough that I don't think it really warrants more effort, anyway. :P

GoogleFrog:
Firstly, this is not all tactical AI. I would call all these changes unit AI changes. Some of the changes are tactical AI changes.

Good (probably, untested):
- Scorcher dive.
-Spectre, Penetrator and Racketeer now disprefer shooting at cheap porc. (in addition to raiders)
- Rackets disprefer shooting at unarmed stuff.
- Puppy behaviour.

Clarification:
- What is an "air unit" for the purposes of chasing Solars and Razors? It is reasonable for gunships and bombers to chase Razors but not Solars, consider splitting the category.
- How does arty disprefer shooting at raiders? I thought they already had this through target priorities.

Discussion:
- It is probably bad for raiders to prioritize mexes. There are definately cases where a raider should kill the nearby combat-enemies and then have free reign to kill the economy. Reasonably consistent automatic targeting is desirable, it makes the game easier to understand. Users can choose to focus on mexes with the set target command and it is much easier to set target a single static mex than a group of enemy raiders.
- What is the purpose of making Solar special in target priorities? What does the current HP/Cost equation modified by unarmed-ness not achieve? I don't want to have the priority gadget turn into a mess of ad hoc rules.

Bad:
- Swifts need to not chase ground units. Otherwise fighter patrols can be baited by ground units. It is rarely a good idea to patrols with the aim to hit ground. Fight could be useful for hitting ground but I think it is reasonable to require area attack.
- Roaches and Ticks should have their tactical AI behaviour. Remember, tactical AI does not aim to make the units better, it aims to give more control options. Fight is not always optimal. It is hard for an AI to know when to jink Roaches and Ticks because those units have very low LOS. Humans are good at knowing when to jink and it is good to give them a choice of speed vs dodging. Manual dodging can be hard and Fight has no other purpose for these units so Fight may as well automate dodging.
- The tactical AI changes are just whitespace. It is better for the history to remove the changes.

mtroyka:
>>What is an "air unit" for the purposes of chasing Solars and Razors? It is reasonable for gunships and bombers to chase Razors but not Solars, consider splitting the category.

That's possible, but part of the problem was that banshees and brawlers sitting right next to undefended solars would not attack them. It doesn't seem to favor attacking solars inordinately without nochase for solars either, at least for gunships. As far as bombers are concerned none of them are used without micromanagement anyway and since they bomb anything they fly over otherwise (for example wyverns on fight move attack drones indiscriminately in spite of target priorities) the AI changes wouldn't make much difference. If there's a specific problem then it can be dealt with easily enough, but I haven't seen any yet.

>>How does arty disprefer shooting at raiders? I thought they already had this through target priorities.

The only units that were previously set to disfavor attacking raiders were penetrator, rogue and annihilator. No other arty units were listed in target priority defs. None of them used nochase for raiders either (nor was there such a category), which I added.

>>What is the purpose of making Solar special in target priorities? What does the current HP/Cost equation modified by unarmed-ness not achieve? I don't want to have the priority gadget turn into a mess of ad hoc rules.

Mostly that was to balance the changes to air. The other reason was simply to set up a more ordinal preference for targets that would be respected by units that use high ratio overrides for other things. It seems to work well enough for gunships, and for ground units I haven't seen any difference in behaviour at all.

>>It is probably bad for raiders to prioritize mexes. There are definately cases where a raider should kill the nearby combat-enemies and then have free reign to kill the economy. Reasonably consistent automatic targeting is desirable, it makes the game easier to understand. Users can choose to focus on mexes with the set target command and it is much easier to set target a single static mex than a group of enemy raiders.

I suggest watching what it does. Raiders still seem to prefer armed targets even with the changes, but in edge cases where it's either dead mex or wasted raider it does seem to help. The difference overall is subtle. Also, set target is extremely clumsy to use since it has no key binding. I don't know if that's configurable but with the defaults it's not a reasonable option when the time to micromanage raiders is limited and the time to select priority target is long.

>>Swifts need to not chase ground units.
I'll revert that then.

>>Roaches and Ticks should have their tactical AI behaviour.
Fair enough.

>>The tactical AI changes are just whitespace. It is better for the history to remove the changes.
True, I'll do that as well.

GoogleFrog:
I tested Scorchers and they need longer order distance to make them move at full speed as they circle their target.

It is better for Banshees to sit idly next to a Solar than for a Fight command across a base to cause them to waste time on Solars. If they chase Solars then fight commands cause them to target Solars if they are the first thing encountered.

Set target should have a binding and there is a reasonable place to configure bindings in the ingame menu. Set target is also automatically set with an attack command. The priority change to mex causes units to consistently arget mex over any other target. They target unarmed things before they target armed things. This is inconsistent and bad behaviour.

mtroyka:
>>I tested Scorchers and they need longer order distance to make them move at full speed as they circle their target.

The reason I set strafe order length so short was to keep them at the range where they do max damage. I tried increasing it, but even a small increase makes them much worse and also more likely to get stuck on random things around their target. Speed isn't really the most important factor when circling a target anyway, and I'm pretty sure that scorchers slow down when turning regardless of what you do.

>>It is better for Banshees to sit idly next to a Solar than for a Fight command across a base to cause them to waste time on Solars. If they chase Solars then fight commands cause them to target Solars if they are the first thing encountered.

Added nochase for solars for air as a new category. I kind of see that behaviour as being bad either way, but I guess this would be the lesser evil.

>>Set target should have a binding and there is a reasonable place to configure bindings in the ingame menu. Set target is also automatically set with an attack command. The priority change to mex causes units to consistently arget mex over any other target. They target unarmed things before they target armed things. This is inconsistent and bad behaviour.

Set target has never and still does not get applied when attacking with right click, unless you specially set it up that way somehow. Also, I've never considered any part of the game's configurations to be reasonable. For the moment I commented out the part that set mex priority for raiders, but personally I think there should be a specific "loot and pillage" option or something to make it possible to decide what you think your raiders' priorities should be.

mtroyka:
I just tested and no, attack never, ever sets priority target unless you replace your "a" (and/or alt-a) keybinding with priority target instead of attack. There is no way that I can tell to replace attack with priority target as the default behaviour when right clicking enemy units, either. I confirmed this because when you set priority target vs ground the unit continues to attack through move orders, but when using attack the attack order gets cancelled immediately if any other order is given without shift. Stop orders do cancel priority target though, which I think is how it should be (that cancel priority target button is stupid).

KlonZK:
i dont like raiders picking mexes over other targets, if that means they will happily lose battles to kill/shoot a mex (that they would otherwise kill anyway).

if a player decides he wants his glaives to kill a mex instead of attacking something that is shooting back, he should attack or use set target on the mex. it is much easier to set target on a mex than to set target on a small, mobile target such as another raider.

in addition, this target preference may lead raiders that are in combat with enemy raiders to switch their target midfight, and then die?

mtroyka:
If you could use right-click to set target I think that would be much better. I removed the target priority for raiders anyway. For AI having raiders prioritize mexes is definitely superior though. Set target is also stupid for AI due to problems with order spam/bandwidth and determining whether a unit has orders or not, but that's a different issue.

GoogleFrog:
Attack always sets target priority in my testing. You must have broken your UI or changed the setting.

The way it works is that if you give a unit a non-shift move order and its current order is a user-give attack command then a priority order is given to the target of the attack command. The attacking unit turns into a moving unit with a priority target order.

This is equivalent and less spammy than giving target priority orders at the same time as attack orders and canceling priority for non-move orders. The aim is to make the target of attack commands persist through move orders. If players want more control they can look up the raw command and hotkeys.

KlonZK:
what if no attack or set target orders were given at all? your raider meets an enemy raider, then proceeds to run by a mex while exchanging fire with the enemy raider?

PR #1154, opened 2015-12-10
Make XP a better gauge of cost made

sprunk:
* remove last hit xp bonus.
* rescale xp so that the number says exactly how many times cost was made (eg. made cost twice => 2.0 XP).
* extend the rank icons widget a bit (max shown rank 4 -> 5) and make coloring consistent (recolor star from silver to gold like the chevrons)
* Decorated Veteran award requirements: 15 times cost made in damage and last hits -> 3 times cost made in damage with a minimum of 1000m


GoogleFrog:
What are the new cost thresholds for the ranks widget? They should be changed simultaneously to reflect a useful progression which distinguishes most units (it is useless if high ranks are very rare or too common).

sprunk:
The rank widget shows:
* nothing if cost made is [0; 1) times
* 1 chevron for [1; 2) times cost made
* analogically up to 4 chevrons at [4; 5) times cost made
* a gold star at [5; ∞) cost made

PR #1163, opened 2015-12-16
Add cylinderTargeting = 1 to Flail

Skasi:
Most AA units have cylinderTargeting set to 1.
Exceptions: Swift has it set to 6, Defender to 5 and Hacksaw to 3.

I assume someone just accidentally forgot to add cylinderTargeting for Flail.

I also sneaked in a little helptext change. For some reason github marked the helptext line green even before I changed anything so while reading through it I decided to change it a bit. I hope the fact that the line was green even without any changes doesn't break anything.

### DIFF
--- a/units/hoveraa.lua
+++ b/units/hoveraa.lua
@@ -25,7 +25,7 @@ unitDef = {
     description_de = [[Fulgabwehrgleiter]],
     description_pl = [[Poduszkowiec przeciwlotniczy]],
     helptext       = [[The Flail launches a single large, short-medium range SAM that does heavy damage.]],
-	helptext_de    = [[Der Flail verschieﬂt ein einzige, groﬂe SAM auf mittlerer Distanz, die wirklich groﬂen Schaden anrichtet.]],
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Convince me that this isn't cargo cult coding.

Skasi:
Why would I? I can't know how AA is meant to work in ZK because like most things in ZK it's not documented anywhere. That's the same reason I didn't bother to make a PR for Snakes which can't hit amphibs in deep water.

sprunk:
Relevant source:
https://github.com/spring/spring/blob/develop/rts/Sim/Weapons/Weapon.cpp#L929

For cylinderTargeting == 0 the result is a cone whose tilt angle is based on heightMod with the top vertex being (range/heightMod) elmos above the unit and the radius being range at the unit level (if heightMod is 0 then it's an infinite cylinder).

For cylinderTargeting ~= 0 the result is a cylinder with radius range and radius*cylinderTargeting height.

The defaults are 0.2 for heightMod and 0 for cyliderTargeting:
https://github.com/spring/spring/blob/develop/rts/Sim/Weapons/WeaponDef.cpp#L88

Currently Flail is a cone with 4000 height. This patch would change it to a cylinder with 800 height.
flail

PR #1175, opened 2015-12-30
Added hover effect to unit_shapes

dvr333:


GoogleFrog:
Why? In any case it should be toggleable through its epicmenu options.

dvr333:
Toggle-ability is planned, as is ally-coloured ally selection circles (definitely an option). Main reason is to have instant feedback when it's possible to click on a thing. Otherwise I find the UI feels insufficiently connected to the game world.

PR #1179, opened 2016-01-01
Startbox API for Skirmish AIs

sprunk:
Add a call-out that lets them ask if a given position is valid and expose the startbox polygon lists to GameRulesParams.

### DIFF
--- a/LuaRules/Gadgets/start_boxes.lua
+++ b/LuaRules/Gadgets/start_boxes.lua
@@ -1,5 +1,3 @@
-if (not gadgetHandler:IsSyncedCode()) then return end
-
 function gadget:GetInfo() return {
 	name     = "Startbox handler",
 	desc     = "Handles startboxes",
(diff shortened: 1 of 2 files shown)


LuaRules/Gadgets/start_boxes.lua
```diff
@@ -143,3 +155,29 @@ function gadget:AllowStartPosition(x, y, z, playerID, readyState)
 		return false
 	end
 end
+
+else -- Unsynced ------------------------------------------------------------------------------------------------
+
+startboxConfig, manualStartposConfig = ParseBoxes()
+
+function gadget:RecvSkirmishAIMessage(teamID, dataStr)
```
rlcevg: For some reason this function is never called (and probably not registerd in [RecvSkirmishAIMessageList](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/gadgets.lua#L1163)).
When gadget:RecvSkirmishAIMessage is in synced part - it works, just like [ai_commander:](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/start_unit_setup.lua#L548) or [sethaven](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/cmd_retreat.lua#L397) that is also inside synced code.
[Lua<->AI communication thread](https://springrts.com/phpbb/viewtopic.php?t=26298) says that it should be unsynced only, so i have no idea what's going on (maybe some lua error/crash before RecvSkirmishAIMessage is registered as i used git-latest-commit 04c738e?).

PR #1195, opened 2016-01-09
Pylon sight 273 → 512 to match grid linking radius

sprunk:


GoogleFrog:
I never knew that this was a rule. I don't think it would be helpful for the UI because any constructed Pylon will put out a circle in relevant situations. Is that the purpose? What is the purpose of the change?

The rule that I apply to economy structures is that they are basically blind.

sprunk:
The grid is (often perceived as) a territory thing so I thought it would make sense to have sight over it. 

The sight values for buildings are seemingly arbitrary so being tied to something might be slightly helpful too. Though the proper solution to this is probably to show a sight circle for buildings that provide it (for example Radar Tower has huge sight but a lot of people never notice due to the radar coverage).

GoogleFrog:
The sight ranges aren't arbitrary. Buildings with no ranged abilities are usually almost blind. Buildings with a weapon have sight at 110% their weapon range. Radar tower is explicitly supposed to help with sight and I think the current LOS view is sufficient to discover this fact (it is barely vital knowledge

Almost every structure in the economy and factory tabs have 273 sight distance. The exceptions are Strider Hub and Caretaker because they have a ranged ability which might need to aim. Pylon has range but it doesn't actively do anything with it. So it may as well match the 273 of all other economy buildings.

sprunk:
Fair enough

Issue #1199, opened 2016-01-12
Some updates for 101.0


ashdnazg:
* Currently bombers have a check for GetUnitFuel in script.BlockShot. since ZK doesn't use fuel and it was also removed from the engine, the check can probably be removed.
* instead of calling SetProjectilePos(-100000, -1000000, -100000) call DeleteProjectile(projID)

Anarchid:
Possibly relevant wrt 101: http://zero-k.info/Forum/Thread/20934

ashdnazg:
ooh right! LOS_MULT  needss to be 1 IIRC

PR #1214, opened 2016-01-27
Various changes for 101.0

ashdnazg:
also see #1199, 
**This is not intended for merge wholesale, pick your favourite fixes when you want to support 101.0**
Included are:
- [x] Fix for gadget/widget handlers to support new unitdestroyed event
- [x]  Fix for custom lua shaders to correctly show teamColor
- [x]  Bombers fix
- [x]  ud.mid* => ud.model.mid*
- [x]  SetProjectilePos(-100000, -1000000, -100000) => DeleteProjectile(projID)
- [x]  fix LOS_MULT to be 1

### DIFF
--- a/LuaRules/Gadgets/CustomUnitShaders/UnitMaterials/Shaders/default.lua
+++ b/LuaRules/Gadgets/CustomUnitShaders/UnitMaterials/Shaders/default.lua
@@ -21,7 +21,6 @@ return {
   #endif
 
     varying vec3 cameraDir;
-    varying vec3 teamColor;
     //varying float fogFactor;
 
   #ifdef use_normalmapping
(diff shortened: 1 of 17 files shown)


Anarchid:
So much whitespace >.>

ashdnazg:
I've set my editor to clean trailing whitespace when I save files :S

GoogleFrog:
Too many whitespace changes. Make a separate commit for the whitespace changes if you need to.

The changes are not backwards compatible (as far as I can tell, with whitespace it is hard to see). Changes should be backwards compatible. For example Spring.DestroyProjectile could be defined in the gadget handler if the function is absent. The definition would use the current position hax. Another example is with aircraft, aircraft could run a CheckAmmo function from bombers.lua which contains the backwards compatibility switch.

Issue #1254, opened 2016-02-07
Guardian and Strike commander sizes


GoogleFrog:
Guardian base model needs to be 5% larger. Strike base model needs to be 10% larger.

- [x] Guardian
- [x] Strike

Changing the size of the base model implies that the other models need changing. The upgrade scales are as follows (consistently across commanders).
1 - 100%
2 - 110%
3 - 120%
4 - 125%
5 - 130%

PR #1260, opened 2016-02-08
Guardian model sizes

sprunk:
Half of #1254


GoogleFrog:
You also need to change the size of the other Guardian models. There is a fixed size relationship between the base model and the higher level models which I wrote in the ticket.

sprunk:
Patched the other models. The heights are 54, 60, 65, 68, 71.

PR #1261, opened 2016-02-08
Hacksaw damage 1050 → 1200

sprunk:
Shots to kill:
 * Rapier 2 → 1
 * Vindicator 2 → 1
 * BD 4 → 3
 * Wyvern 3 → 2


GoogleFrog:
If you go to 1200 you two shot a Wyvern and get a round number. Perhaps this is desirable?

sprunk:
The tiny coverage and baitability make it unlikely to work so making it good when it actually does work sounds good.

Licho1:
its op

sprunk:
Currently or after this buff?

Anarchid:
440 metal to stop a single Wyvern in radius less than that of LLT, if the enemy fails to bait it? Fine, i'll use Newtons instead, they can shoot ground, too! 

Such OP!

GoogleFrog:
I think it is a good change to try. Rapier and Wyvern are in the meta and the Hacksaw is supposed to create a small bubble of air fear.

PR #1265, opened 2016-02-09
quick fix for weird TV effect until engine bug is fixed

dvr333:


dvr333:
It looks like engine 100.0.1 (101RC) can't handle Spring.SetCameraState() without clobbering screen-space drawing.

dvr333:
Engine bug mantised at https://springrts.com/mantis/view.php?id=5075

dvr333:
Apparently the engine bug is fixed, so if the TV problem is gone without needing this PR then don't bother merging.

GoogleFrog:
We'll assume engine fix fixed it.

PR #1271, opened 2016-02-13
Changed command fx display defaults to provide maximum feedback

dvr333:


GoogleFrog:
As I recall this was disabled because the custom formation dots drew very inefficiently. I noticed the framerate drop when I tested the feature after upgrading the custom formations widget.

dvr333:
Hrm, I do see it is using immediate mode. I'll switch it to display lists at some point and add that to the PR.

PR #1276, opened 2016-02-15
Unbind the useless engine map info screen.

sprunk:


Licho1:
It can be useful for game speed or map name. I dont think its completely useles now. Game speed should be displayed elsewhere.

GoogleFrog:
It also displays game and engine version.

sprunk:
The hotkey is never mentioned anywhere (so if someone wants to know gamespeed he wouldn't use it because he won't know about it) but is easy to press accidentally. It doesn't go away when you press Esc, covers half the screen and is ugly.

Game speed could be displayed alongside the clock. So instead of the clock saying [12:34] it would be [12:34 (x1.23)] if speed is not 1.

Showing the versions is a debug thing. It's ok to have a default binding but it should be hard to accidentially click.

GoogleFrog:
ctrl+i?

PR #1280, opened 2016-02-17
Trainer default setup changes

sprunk:
For pre-made setups there are the site comms.
The original reason for trainers having modules was because the modules were otherwise locked which is no longer the case.
The defaults were crappy noobtraps anyway and a blank setup is more flexible/useful overall.


GoogleFrog:
The original reason for trainers having modules was so new players could just click through the morph screen without needing to stop and spent brain power on selecting a module. A slot saying "No Module" demands more of a response from players than a slot with a  module in it. If the defaults are crappy then fix them.

(Unless by original trainers you mean the pre-dyncomm ones. I ignored their setup completely and don't really consider them a predecessor)

sprunk:
A different approach.

 * radar at level 1. Newbies often don't know about radar so it's good to have it builtin. At the same time it's also the only cost-efficient module so in competitive play forgetting to cancel it won't be as disastrous as say a range module.
 * HMG at level 1 (where applicable). If someone is morphing mindlessly then they might as well get the best weapon instead of the budget ones.
 * hp + regen at level 2. Makes it harder for a newbie to die. In theory shield and cloak also do that but they do not provide sustenance (regeneration of real health). They're also expensive so a nub would likely stall before getting them.
 * shield/cloak moved from level 2 to 3
 * Guardian gets a shotgun at level 3 instead of riot (to synergise with the disruptor conversion it later gets at 5).
* level 4/5 unchanged

PR #1284, opened 2016-02-20
SetTarget hotkeyed for SetTargetCircle on T

dvr333:
- Clear target on Ctrl+T, Resurrect moved to Ctrl+Shift+T

Resurrect is rarely used, SetTarget can be used all the time, and the default hotkey ('t') for it is for SetTarget, not SetTargetCircle, exposing that terrible square for area targeting.


GoogleFrog:
Why is there "None" set as a hotkey? Also I am not so sure about unbinding resurrect.

dvr333:
"None" was what my zk_keys.lua had after unbinding the hotkey for SetTarget, and Resurrect isn't unbound, just not bound to a plain 'T'.

sprunk:
Resurrect being Ctrl+something seems a bit odd in that Ctrl+something is usually selection. Are there any other Ctrl+something order hotkeys?

Then there's the fact that settarget and resurrect are pretty much unrelated to each other so being under the same key and a different modifier is also weird.

dvr333:
Resurrect doesn't have to be on 'T' at all, I just didn't spend any time thinking about that, but I needed to put it somewhere, and I figured it might as well stay close to where it historically was.

sprunk:
This doesn't seem to have happened.

PR #1289, opened 2016-02-22
Nanoframes suffer the in-combat buildspeed penalty

sprunk:
x0.25 buildpower when under fire

### DIFF
--- a/LuaRules/Gadgets/unit_repair_speed.lua
+++ b/LuaRules/Gadgets/unit_repair_speed.lua
@@ -79,7 +79,7 @@ function gadget:UnitDamaged(unitID, unitDefID, unitTeam, fullDamage, paralyzer, 
 			bt = bt,
 			frames = REPAIR_PENALTY_TIME,
 		}
-		if select(5,spGetUnitHealth(unitID)) == 1 then
+		if select(5,spGetUnitHealth(unitID)) == 1 or not buildTimeChangeNeeded[unitDefID] then
 			spSetUnitCosts(unitID, {buildTime = bt*REPAIR_PENALTY})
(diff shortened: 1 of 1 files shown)


GoogleFrog:
As far as I am concerned the current system is a design decision. Building is intentionally distinct from repair.

Repair can be done while a unit is firing/moving/whatever and does not cost the full energy cost of the unit. Something about full units makes them cheaper to repair (makes sense) but harder to repair while in combat. The purpose is to stop wars being halted by background constructors but to still enable repair to be fast enough to be useful.

Construction drains metal so should not be messed with by something as common as taking damage. People would not be able to predict their drains. Construction is often about "time to completion" whereas combat repair is mostly concerned with the fact that the unit is gaining health. The speed penalty is also too long for construction, it would be weird to be able to send things in to poke constructions to deal buildtime damage.

Perhaps the short version is that I feel health and build-completeness are different parameters and that damage should not affect your rate of build-completeness.

Anarchid:
Neon: visuals for combat and normal repair should be subtly different.

Skasi:
> it would be weird to be able to send things in to poke constructions to deal buildtime damage

Why would an effective counter to porcpush be weird?

> Perhaps the short version is that I feel health and build-completeness are different parameters and that damage should not affect your rate of build-completeness.

Would implementing this be an option? I mean, separating HP-gain from build-gain.

Adding to Anarchid's neonstorming: Build progress costs 100% of the metal and 50% of the energy, repair "progress" costs 50% of the energy.

GoogleFrog:
I think there are good reasons for not doing this. I think the change increases complication and it is not otherwise required.

PR #1296, opened 2016-02-26
Fixed #1288 defaults issue

### LINKED ISSUE #1288: Default config is to only show selected unit commands on shift
Should be selected visible, all on shift.

dvr333:
* Also apparently fixed a hidden issue with setting the command visilibity behaviour properly at LuaUI load


dvr333:
It guarantees the OnChanged function gets called at widget start. Otherwise the command mode stuff isn't properly set, even if the setting is correct. This seems to be the case regardless of defaults, but the problem may be only with engine 101+.

sprunk:
I mean why "do OnChange end" instead of just "OnChange"

dvr333:
Just "OnChange doesn't work, AFAICT

PR #1300, opened 2016-02-27
Revert "Changed command fx display defaults to provide maximum feedback"

sprunk:
Reverts ZeroK-RTS/Zero-K#1271

Causes crash on ATI:
[04:02]  KingRaptor OH COME ON
[04:03]  KingRaptor Shadowfury333: line move with multiple units causes CTD, stacktrace attributes to GFX driver


sprunk:
http://pastebin.com/qw6NQc6U

Issue #1287, opened 2016-02-21
Idle indicator does not show single idle unit


Licho1:
http://i.imgur.com/7wQz9EQ.jpg

PR #1312, opened 2016-03-01
Fixed #769 and #1287

### LINKED ISSUE #769: Quick Selection Bar doesn't show idle workers (properly)
After the worker is built by a factory (athena and athena made conjurer is shown properly) it is not detected as idle up until it builds something. The bug is present in all kinds of factories.

N0U:
So far it works, but I think it must be investigated further. 
So far I found that con (And I think any unit) get CMD.WAIT command after moving. However, con doesn't get CMD.WAIT after finishing building. Maybe it is another widget messes with wait command?


GoogleFrog:
Write what the issues are when you make a fix commit.

PR #1319, opened 2016-03-02
Added double wait check

N0U:
Now widgets thinks that WAIT WAIT queue is the same as an empty queue.
Cleaned up my debug echos


LuaUI/Widgets/gui_chili_core_selector.lua
```diff
@@ -841,12 +840,20 @@ function widget:UnitIdle(unitID, unitDefID, unitTeam)
...
-		--Spring.Echo("Con "..unitID.." is idle")
 		idleCons[unitID] = true
 		wantUpdateCons = true
 	end
 end
 
+-- Check current cmdID and the queue 
+local function isDoubleWait(unitID, cmdID)
+	if cmdID==CMD.WAIT then
+		local cmds=Spring.GetUnitCommands(unitID)
```
GoogleFrog: Don't call this with only one argument. First call it to check how many commands the unit has and then call it to check the ID of the first and second command.

Also, don't use GetUnitCommands at all. Use GetCommandQueue instead because that is what is used thoughout the rest of ZK.

N0U: Why? This queue should have only one command, because the new added command isn't in queue yet, but given as an argument.

Licho1:
Wont it interfee with transport AI?
It uses waits for embark/disembark.

N0U:
I think It shouldn't. This widget just mark cons with WAITWAIT commands as idle.

PR #1323, opened 2016-03-04
Fixed wrong weapon ranges for dyncom

N0U:


LuaUI/Widgets/gui_defenseRange.lua
```diff
@@ -910,6 +910,56 @@ local function DrawRanges()
 	end
 end
 
+local function GetWeaponRanges(unitID,unitDefID)
+	if Spring.GetUnitRulesParam(unitID, "comm_level") then
+
```
sprunk: Huh? This is empty and the else is also empty but has a return. Looks like you missed some code.

N0U: Ops. Firstly I wanted to used this function, but than I had another idea and I just forgot about this.

PR #1336, opened 2016-03-14
Precision bomber selector

CrazyEddieTK:
This PR adds functionality to the Core Selector widget to allow for easier selection and management of precision bombers (Ravens). It creates a new action bindable via Epic Menu: Game -> Selection Hotkeys -> Individual precision bombers, default unbound.

The hotkey selects "ready" Ravens: armed, not being refueled or repaired, and not already tasked with an attack order in their command queue. The hotkey selects one Raven at a time, allowing you to give it a specific attack order and then quickly select another one that's ready to attack. In this manner you can quickly assign individual Ravens to individual targets, ensuring that targets will not be overkilled nor underkilled and that high-priority targets will be hit first.

The hotkey has some intelligence. If you press it once, it will select a ready untasked Raven. If you press it again, it will continue to add ready untasked Ravens to the selection. This allows you to, for example, press the hotkey five times to select a group of five ready untasked Ravens, then click on a factory to hit it with five bombs. So not only can you quickly assign individual Ravens to individual targets, you can quickly assign _the right number_ of Ravens to each individual target.

The Ravens selected will be those that are closest to the mouse pointer.

If you already have a selection containing various types of units, pressing the hotkey will filter out everything except ready untasked Ravens. So you can select all the ready Ravens flying around some area by box-dragging across the area and then pressing the hotkey, or you can select all ready Ravens everywhere by pressing ctrl+A and the hotkey.

This PR does not add any icons to the Core Selector display. It only creates the bindable action.

### DIFF
--- a/LuaUI/Widgets/gui_chili_core_selector.lua
+++ b/LuaUI/Widgets/gui_chili_core_selector.lua
@@ -180,7 +180,7 @@ options = {
 		name = 'Select Precision Bomber',
 		action = 'selectprecbomber',
 		path = 'Game/Selection Hotkeys',
-		-- dontRegisterAction = true,
+		dontRegisterAction = true,
 	},
 }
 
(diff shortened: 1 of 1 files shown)


sprunk:
Why count moving Ravens as idle?

CrazyEddieTK:
Not idle, but rather available to launch an attack. If they've been given a move order (or a series of move orders) but no subsequent attack, the assumption here is that you'd be willing to pull them away from their destination and give them a target instead. Whereas if they're currently moving but have an attack queued up after their move, they're considered already tasked and won't be selected by this hotkey.

It might be the case that you _don't_ want to pull them away from their destination on the east in order to attack something on the west. The hotkey's intelligence helps out somewhat here, as when you hover over the targets on the west, it will select the _closest_ Raven first. So if you have an east flock and a west flock, it will automatically use the west flock to attack the west targets and allow the east flock to continue its move order undisturbed. But for this to work you have to remember how many Ravens are in the west flock and not use the hotkey to task more than that number on the west targets.

But I suspect that's an unusual scenario, or perhaps one in which you should use a different selection-and-targeting method than this hotkey.

I think a more common case is that you'll select a flock, give them a move order across the map towards the targets, clear the selection, then start using the hotkey to assign individual Ravens to individual targets while the entire flock is still in motion. This will work well, as each target will get whichever Ravens are closest, and the flock will move together while you allocate targets instead of heading off by ones and twos.

PR #1338, opened 2016-03-14
Gadget version of wading wakes.

Anarchid:
This reimplements the wading wakes proposal as a gadget, which should be much cleaner and less error-prone. Wade-admissible units are autodetected. Wakes are spawned every seventh frame for every fourth unit, so each unit currently generates ~one wake per second.

The part i dislike about this is relying on piece zero (which is nevertheless guaranteed to be present and likely to be the unit's core). Units could specify which of their pieces should emit wakes, but really i hope to see engine 102 include a way to SpawnCEG wakes without having to go down to EmtiSFX level. 

The staggering behaviour could be made dynamic so fewer units spawn wakes per cycle.

This replaces and supercedes #1103.

Eyecandy:

workspace 1_060
workspace 1_058


GoogleFrog:
Why don't you store a list of units and then iterate with an increment of 4 each time. This method is messy.

PR #1339, opened 2016-03-15
Warrior → Ghoul; Tick → Wisp

sprunk:
Warrior is a somewhat boring OTA name which doesn't fit either of the Cloakybot themes. The unit has recently received fairly large changes so it seems like the best moment to do a rename.

Tick is there mostly because the other rename creates a good excuse to do more at the same time and the unit is fairly non-crucial. The name is not OTA but almost: it seems to have been included in most popular unit packs. It also does not belong in either theme.

Both proposed names fit within the evil magic theme (like Gremlin, Spectre) and are really good at describing the units:
Ghoul - slowly walks with arms stretched forward and mauls everything that gets close.
Wisp - a volatile flash of light out of nowhere, which doesn't harm people directly but still leads to their demise.


CrazyEddieTK:
The cloakybot "evil magic" theme is terrible, and these names are terrible.

Cloakybot has some of the best names in the game. They are interesting and distinctive. The idea of making all names in every factory fit some kind of theme dilutes the very idea of having themed names, dilutes the interestingness of the names in each theme, increases the learning curve for newbies because all the names are meaningless synonyms, and robs units of individuality.

I urge all the devs to stop this misguided headlong rush into madness.

knorker:
There is nothing "evil magic themed" about dual machine guns.

dvr333:
While I kinda like the themed name idea, the options don't work:

http://classic.battle.net/war3/undead/units/ghoul.shtml is a ghoul
http://classic.battle.net/war3/nightelf/units/wisp.shtml is a wisp

Since WarCraft 3 is getting a bit of recent publicity due to a patch announcement, I'm pretty sure that's the first thing people are going to think when they hear "ghoul" or "wisp". I do think Warrior is a bit generic, though I like Tick because they are sneaky and disabling units is a neat reference to Lyme disease. 

I think the bigger issue with themed names is "why?" Now, from an out-of-fiction dev perspective and to have a sense of factory name consistency it makes sense to theme names, but from a player perspective it's just a bunch of random words. There's no underlying theme or flavour behind the names (an issue shared with OTA), which means anything more elaborate and specific than one short word name is so out of place as to not fit at all, but just having random one-word names makes it harder to create a sense of identity as a game.

I realize that ZK's units are primarily OTA units remodeled, renamed, and regrouped, so it might come off as pretentious to do this, but having some kind of in-fiction justification for the names and themed names would help sell the theming. Maybe each factory was once a product of a specific human faction or corporation, and the robots just managed to get access to all the blueprints by the time of the game, so these random names are actually model/brand names, which tend to be a bit random, cf. car model names.

ParzivalX:
I agree that cloaky's name were possibly a bit misguided (and rushed) In the early development stage. OTA models were probably the primary basis for ZK's units so I'm not really suprised by that, but it is time to start to pull away from OTA refrences.

Since hover fac needs renamed as well why don't we rename cloaky units to names of weapons (I.e. Glaive, Hatchet, Axe, Mace, Longbow (Sniper), Grenade (Rename for tick), etc.)?

dvr333:
My point about OTA models and concepts is that nearly every Spring engine game uses them as a basis, which is why I'd be worried adding ZK-specific fiction justifying ZK's use of those unit concepts would be seen as pretentious.

The evil magic theme works to a point, but probably only inasmuch as the names sound like something you could see for a piece of military equipment. "Spectre" sounds like some kind of SpecOps thing, "Ghoul" does not.

CrazyEddieTK:
The entire idea of name themes is terrible.

See further comments in [the forum post](http://zero-k.info/Forum/Thread/22180).

springraaar:
Not sure about the idea, but "warrior" and "tick" are already fitting names for the units, much better than "ghoul" and "wisp".

Anarchid:
A lot of this reads like "gremlin is a bad name for jethro because i'm used to jethro".

Ghoul is awesome.

knorker:
Often names are different but not better. Things that get broken by name changes:

- forum posts
- videos with commentary
- the brains of people
- tutorials / missions (yes, very few exist but some mention unit names to player)
- manual texts or helpful forum postings about how to play

Imo bad to break so much for a minimal (if any) improvement.

sprunk:
I can fix missions, manuals and even forum posts.

dvr333:
Jethro->Gremlin went from one ill-fitting name to another, and given that Gremlins are sneaky it's a bit less ill-fitting.

Warrior is going from fitting to the ill-fitting Ghoul. If you must stick with the theme, then something like Shambler or Executioner or some demon name associated with mass destruction would work better. Same with Tick, even if we don't consider WarCraft 3, (Will o' the) Wisps are still a specific thing, and Ticks don't look like one at all, and remodeling it to fit would likely not work either given the grounded sci-fi theme. I'd almost rather go with Wight, but that's only because of the Myth reference, which fits the Roach better anyway.

Anarchid:
Warrior is not `fitting`. It is `overfitting` because it matches every armed unit in ZK.

dvr333:
Yeah, that's true, but Ghoul doesn't seem to fit at all. It doesn't even sound like a brand name/military designation.

Anarchid:
> t doesn't even sound like a brand name/military designation

Doesn't it just fit alphabetically right between SA-8 "Gecko" and SA-16 "Gimlet", and thematically between SA-14 "Gremlin" and SA-20 "Gargoyle"? :P

sprunk:
Warrior "fits" the same way calling something "Unit" would fit. Warrior is just a fighting person, every combat kbot could be called that. For example Zeus could be called Warrior just as well. It's nondescriptive.

In most depictions Ghouls fight like ZK Warriors do (ie. they maul stuff in close combat) so I claim it fits perfectly and describes the unit traits. Explain why you think Ghoul doesn't fit.

Orfelius:
To be honest this sort of discusions only remind me that ZK can still be considered as OTA mod and even the devs treat it as such in many cases, clinging to the past and unable to see the old things go.

dvr333:
Hey, I don't much like "Warrior" either, I just don't see "Ghoul" as the best replacement term. I can sortof see the argument for it being a replacement term, but I think it could be improved.

CrazyEddieTK:
Anarchid: overfitting is _not_ a problem. Names are not descriptions or definitions; they are handles.

ParzivalX:
I do feel like the unit name issue is a little over stressed though. Shouldn't we be focusing on balancing out the existing units instead of just renaming them?

sprunk:
Eddie: Overfitting is not a problem for you because you know the game well. For someone who is learning the game it's better if names can be easily connected to the unit.

Shadowfury: if you think it can be improved then figure out a better name than Ghoul. Can't be Wight because that's a synonym to already-existing Spectre. Shambler is pretty much a synonym to Ghoul except longer and less common so while also valid Ghoul is superior.

Parzival: you're free to go and contribute to balance. This issue is not taking any dev time.

CrazyEddieTK:
_Overfitting is not a problem for you because you know the game well._ - No, it's not a problem, period.

_For someone who is learning the game it's better if names can be easily connected to the unit._ - No, it's not, for reasons I covered on the forum.

PR #1344, opened 2016-03-20
Rewrite of Tiltzoom correction

dvr333:
* Tiltzoom now works properly on every frame.
* Overview also fixed to not just tilt wildly.
* Zoom out to center option works, but it's a bit jittery.
  I can't figure out the problem and it's relatively minor
* Drift to center now an option, used to be part of tiltzoom.
  When on, will slowly move the object under cursor to center when zooming in to cursor
* COFC helper functions will now properly not modify input camera state

### DIFF
--- a/LuaUI/Widgets/COFCtools/Interpolate.lua
+++ b/LuaUI/Widgets/COFCtools/Interpolate.lua
@@ -7,9 +7,15 @@ local spDiffTimers		= Spring.DiffTimers
 local mathPi 				= math.pi
 local pow = math.pow
 
+lockMode = {
+	xy = "xy",
+	free = "free",
+}
+
 local beginCam = {px=nil,py=0,pz=0,rx=0,ry=0,rz=0,fov=0,time=0}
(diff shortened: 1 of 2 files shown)


LuaUI/Widgets/COFCtools/Interpolate.lua
```diff
@@ -7,9 +7,15 @@ local spDiffTimers		= Spring.DiffTimers
 local mathPi 				= math.pi
 local pow = math.pow
 
+lockMode = {
```
sprunk: What is this for? This table assigns each key the same value as the key and never modifies it. Why not just use "xy" instead of lockMode.xy and "free" instead of lockMode.free?

dvr333: It's the closest I can get to an enum in Lua. It's better than just a raw string because the bounds are more clearly laid out without having to dig into the code, and if you enter one that doesn't exist it'll just be nil, rather than something almost right, though you could just reference the table creating it for the correct options.

PR #1363, opened 2016-04-10
Reef: remove anti. Protector: buildable on water.

sprunk:


Anarchid:
Does the model have an underside?

sprunk:
No.

GoogleFrog:
Why do this? I am ok with fleets not fearing nukes. Also Protector needs modifications to not look out of place on the water.

Anarchid:
Protector scouting, sniping, placement and defense are interesting, "fleets are immune to nukes" less so, imo.

I can do model modifications to make it be less out of place.

sprunk:
Reasons:
 * having anti built into an already-good combat unit just removes the choice of making anti.
 * as a unit, Reef is cloakable. This ruins scouting.
 * open sea tends to become a long porcy standoff. Nuke is supposed to deal with that kind of situation.

Licho1:
Dunno im ok with removing anti from protector if something else gets it.
But imo for sea it should be a mobile unit - perhaps specialized anti nuke cruiser.

Anarchid:
Mobile unit will mean it retains most of the objections that prompted this PR:
- cloakability defeating scouting 
- ability to avoid being tacnuked after being scouted by relocating

Also since the inception of this PR, antis have been changed to work on missile flying above their interception radius. That means a mobile anti has become much better: as you conquer territory, you move onwards with it, protecting an ever increasing sector of the map.

sprunk:
The MM map pool has no open sea.

DeinFreund:
nu

Anarchid:
Brutal.

mtroyka:
that's a stupid reason to close this. It should have been pulled long ago, and is just as bad if not worse of an issue than it was when this was first proposed. I remember seeing an FFA where shaman sat in the corner with a cloaked reef and prolonged the game vs 4hundred for like an hour for no reason. Shaman had already lost beyond recovery but thanks to the stupid terrain on isla bonita (and terraform) berthas couldn't hit and thanks to the cloaked reefs neither could silencer.

Also the map pool can and will change in the future, so that's not even a valid excuse.

Anarchid:
I have reasons to suspect the map pool is not ever going to feature maps with open sea combat.

mtroyka:
>> I have reasons to suspect the map pool is not ever going to feature maps with open sea combat.
Is that my cue to fork ZK and start a new project where bs like that actually gets fixed within geological timescales?

Anarchid:
> Is that my cue to fork ZK and start a new project where bs like that actually gets fixed within geological timescales?

I don't know. I just will strongly oppose adding sea maps until sea stops irking me. Which is a good reason for me to suspect that sea maps will not be ending up in matchmaker.

ArchShaman:
> I remember seeing an FFA where shaman sat in the corner with a cloaked reef and prolonged the game vs 4hundred for like an hour for no reason. Shaman had already lost beyond recovery but thanks to the stupid terrain on isla bonita (and terraform) berthas couldn't hit and thanks to the cloaked reefs neither could silencer.

That was a glorious day.

Anarchid:
> Is that my cue to fork ZK and start a new project where bs like that actually gets fixed within geological timescales?

Doesn't effective removal of sea fix it?

GoogleFrog:
Sea seems ok from the games I've seen recently, although there were few of them. I have not heard any decent suggestions for rework and nobody is stepping fowards to implement something so I don't think it's getting changed at the moment.

The PR was unmergable, which is a good reason for closing it. I never looked at the PR since it was opened because it doesn't fix the uglyness of the floating antinuke model. It could be better to have no floating antinuke, just expect people to use islands or terraform. Another approach could be to significantly reduce Reef interception range.

GoogleFrog:
> Is that my cue to fork ZK and start a new project where bs like that actually gets fixed within geological timescales?

This issue itself is really unimportant compared to other required work. You're welcome to open a PR about "bs issues like this".

DeinFreund:
Sea is back, reef lost anti, we have no anti for sea.

sprunk:
Terraform a platform.

DeinFreund:
This has a completely arbitrary cost which is easily more than the nuke itself. Is this really what you want sea antinuke to be?

ArchShaman:
I have to agree with  @DeinFreund. It would be more interesting if it was sea placeable. This means it doesn't get an amphib/sub immunity side arm.

sprunk:
My vision of water is to be a border thing (see eg. Bonita or Adansonia) or small inland (eg. Onyx Cauldron, Charlie in the Hills, Lonely Oasis). So:
 * sea won't usually get enough density to warrant nuke
 * if it does, you can just place anti on nearby land
 * if you still can't, terraform a platform.

In this regard no floating adds choice. You can either put some effort to secure nearby land, or pay extra for the terra platform. I don't think it's going to cost nearly as much as a silo.

Sub immunity if unaffected by floating because floating does not remove the ability to live on a platform.

PR #1371, opened 2016-04-21
Silo_valley_v02 metalmap

Skasi:
Metal spots as suggested in http://zero-k.info/Maps/Detail/8122

map defaults were 1.0200001001358 for small and 4.0800004005432 for large metal spots according to dbg_mex_to_infolog.lua


CrazyEddieTK:
Should this file end in .lua ?

sprunk:
Yes, should be lua.

PR #1373, opened 2016-04-22
Radio hotkey partial cleanup

dvr333:


dvr333:
* Spectator State intentionally left hotkeyable, it's super useful when casting, though I suppose a more direct way of choosing to show the vision of either a specific player or all players at once would obviate the whole widget.
* Selection displays for allies left hotkeyable. I'm torn on this, but it seems like a thing that people would use for quickly being able to see what allies are doing or not. Probably obviated in most cases by using shift to see all commands/selections, but shift is pretty overloaded for temporary information at this point.
* TTS left hotkeyable. Too much drama around that to not let it be turned off in a hurry.
* Camera tracking options left hotkeyable. They seem like things you'd want to flick on and off quickly and surreptitiously when following units around for videos and such.
* Spectator Panel subpanel toggles left hotkeyable, since the idea is ultimately to have them work like StarCraft 2 where you switch out subpanels to show different parts of the game state.
* Options in gui_hud_presets.lua left hotkeyable since they use a totally different interface for their options and I don't want to mess with it in a large PR like this one
* Hardware Cursor option left hotkeyable, since it may be needed in an emergency, though I'm not sure I want to keep it this way.
* Defense ranges left hotkeyable, since it's useful for casting, to control screen clutter.
* Retreat zone options left hotkeyable, same reasoning as ally selections.

dvr333:
If being able to revert individual sections isn't a priority, let me know and I'll re-PR as one big commit.

CrazyEddieTK:
woo

quite the busy beaver

good stuff

sprunk:
>Hardware Cursor option left hotkeyable, since it may be needed in an emergency, though I'm not sure I want to keep it this way.

If someone has cursor problems commonly enough to hotkey, they should just keep this on.

>TTS left hotkeyable. Too much drama around that to not let it be turned off in a hurry.

Ditto. Either you like it or not.

dvr333:
Okay. My concern with TTS is that it sounded like it could be maliciously enabled remotely with the right chat messages.

sprunk:
Because it can. But that eventually gets fixed and a hotkey won't help you anyway.

Issue #1201, opened 2016-01-15
Wakes visible for cloaked ships.


GoogleFrog:
This occurs at least with Halberd. A check for cloaking should be inserted into the relevant scripts.

Anarchid:
Should cloaked units also:
- not emit smoke (i think you can catch scyhhes now with this)
- not emit wakes

?

GoogleFrog:
Yes. Cloaked units should not emit things which can seen merely by viewing the position. LUS fixes the smoke emitting problem so I am fairly sure you cannot see Scythes. See https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/constants.lua#L78

GoogleFrog:
Todo: Ships and Dagger.

GoogleFrog:
https://github.com/ZeroK-RTS/Zero-K/commit/0166ac6ca07ec935530f6e5fb813dc689b54b1f9

sprunk:
Related: #1375 - has pretty much the same solution (convert ships to LUS and make wakes conditional).

Anarchid:
Not all ships require converting to LUS, so some are fixable as-is.

PR #1381, opened 2016-05-05
probably fix reef, typhoon, skeeter cloakwakes

Anarchid:
#1201


GoogleFrog:
Tested?

Anarchid:
No, hence PR. Will test today i think.

GoogleFrog:
Is this going to be tested?

Anarchid:
Tested, works.

PR #1386, opened 2016-05-08
Heatray targeting

sprunk:
Heatray uses distance for target priority (ignores everything else)

### DIFF
--- a/LuaRules/Gadgets/unit_target_priority.lua
+++ b/LuaRules/Gadgets/unit_target_priority.lua
@@ -95,11 +95,6 @@ function gadget:AllowWeaponTarget(unitID, targetID, attackerWeaponNum, attackerW
 	if not enemyUnitDef then
 		return true, 5
 	end
-	
-	--// For heatrays, ignore other priority and just attack the closest target.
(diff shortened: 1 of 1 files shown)


mtroyka:
should get rid of the 'divide by enemy weapon range' clause. It could cause screwy results, and what you really want is just prio = range.

GoogleFrog:
He is not dividing by enemy weapon range. He is dividing by own weapon range. In this form thedivision should still be removed because scaling all returned values should have no effect on weapon behaviour. The order is the only thing that matters.

However, I think the change is wrong. This could make a Scorcher attack a nearby Solar instead of a slightly further away enemy raider. I can think of many problem cases like this. I would like the following change instead:
 * Remove reference to proximityPriority, replace with customParams.dyndamageexp (as with the current PR).
 * Increase the maximum proximity multiplier from 0.1 to 10.
My reasoning for this is that most combat units have HP/Cost in the range of 1 to 9. This proximity multiplier makes it likely that a heatray will target close combat units over far combat units. However, in most cases it will prevent heatrays from targeting unarmed targets as they have a flat bonus of 15 to their priority.

PR #1399, opened 2016-05-16
Fix #1396 bug with wolverine claws not spawned

### LINKED ISSUE #1396: Wolverine Claws are not spawned if Wolverine dies
This should now be fixable with ProjectileDestroyed and Spring.GetProjectileTeamID.

N0U:


LuaRules/Gadgets/unit_explosion_spawner.lua
```diff
@@ -48,22 +48,26 @@ function gadget:Explosion_GetWantedWeaponDef()
 	return wantedList
 end
 
-function gadget:Explosion(w, x, y, z, owner)
```
ashdnazg: why use ProjectileDestroyed and not Explosion?

N0U: Explosion doesn't give projectile id and when wolverine die, owner id become invalid, so there is no way to find team id. Before this script just ignored any projectiles with a dead owner.

PR #1406, opened 2016-05-21
Cherry-picked chili features from upstream and Chonsole input method

gajop:
Probably requires 101+ for some minor gamerules/teamrules changes, but backward compatibility could be added if ZK gets stuck to 100.


gajop:
The last commit should add the required backward compatibility.

PR #1412, opened 2016-05-26
add lordmuffe construction gunship

Anarchid:
This has:
- model, texture, texture2
- kinda working script
- jets
- unit for testing 

This does not have:
- wreck
- animations
- working nanos
- textures in dds format 
- normals (i generated a normal map but it didn't seem to do anything - probably should have named it wrong, or converted to dds)

workspace 1_073
workspace 1_072


GoogleFrog:
Why are the images only 1/4 used?

GoogleFrog:
Why is the script full of errors...

Anarchid:
> Why are the images only 1/4 used?
I found it like that. I presume that either LordMuffe was going to make an atlas, or one of the pieces of geometry had a wrong piece. By the time i noticed, i already had gone through reassembling the geometries into objects, so it was kind of too late to rescale UV.

PR #1419, opened 2016-06-01
Allow AIs to give units to allies

DeinFreund:
Does this break something else?

Currently it's impossible for AIs to give units to teammates, which is a feature I would like to use for my AI in COOP games.

### DIFF
--- a/LuaRules/Gadgets/game_disable_take.lua
+++ b/LuaRules/Gadgets/game_disable_take.lua
@@ -21,6 +21,56 @@ end
 local spIsCheatingEnabled = Spring.IsCheatingEnabled
 local spGetPlayerInfo     = Spring.GetPlayerInfo
 local spGetTeamInfo       = Spring.GetTeamInfo
+local spGetUnitTeam       = Spring.GetUnitTeam
+local spTransferUnit      = Spring.TransferUnit
+local spSendLuaRulesMsg   = Spring.SendLuaRulesMsg
+
+
+function gadget:RecvSkirmishAIMessage(aiTeam, dataStr)
+	-- usage: callback.getLua().callRules("ai_give_unit|" + u.getUnitId() + "|" + team, -1);
+	
+	local dataSplit = {}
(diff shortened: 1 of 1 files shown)


GoogleFrog:
The comment you removed implies that players with AIs on their team would be able to take units from the AI if this is pulled.

DeinFreund:
How could this be avoided while still allowing giving units?

Anarchid:
The entirety of this gadget was committed in 2015. I don't think CAI was /takeable before 2015?

DeinFreund:
This doesn't work as it will allow /take of CAIs units. Possible alternatives would be having the AI call LuaRules or only allowing AI<->AI sharing. Any other ways I could do it?

DeinFreund:
Now AIs can give units via a LuaRules message. This circumvents issues with /take.

DeinFreund:
Are there any objections/Am I missing something?

rlcevg:
Bump! I'd also like to see unit sharing (though didn't test PR).

Less flexible alternative is to disable "disable_take" gadget when there are no CAIs.
Even less flexible alt is to wait until [FR](https://springrts.com/mantis/view.php?id=5259) fulfilled and switch to new engine immidiatly (i'm so naive).

mtroyka:
This seems like something that should be part of engine rather than zk.

GoogleFrog:
Doesn't RecvSkirmishAIMessage cause desync?

DeinFreund:
That's why it's getting forward to LuaRules Message, like other AI messages. (Com selection already uses the same system)

GoogleFrog:
It's still worrying. Doesn't the engine now support take blocking?

DeinFreund:
Alternative:
1) disable/remove disable_take 
2) add 'allowTake=false' into 'system' section of gamedata/modrules.lua

both should work without desyncs, but this one still needs a PR

DeinFreund:
Obsolete now

PR #1420, opened 2016-06-02
Changed name

ParzivalX:
Changed name from Infiltrator to Black Widow to fit spiders theme.

### DIFF
--- a/units/armspy.lua
+++ b/units/armspy.lua
@@ -1,6 +1,6 @@
 unitDef = {
   unitname              = [[armspy]],
-  name                  = [[Infiltrator]],
+  name                  = [[Black Widow]],
   description           = [[Cloaked Scout/Anti-Heavy]],
   acceleration          = 0.3,
   activateWhenBuilt     = true,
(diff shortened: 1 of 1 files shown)


GoogleFrog:
We seem to have an implicit policy against multi-word names. This was demonstrated by "Razor's Kiss" -> "Razor". I think it is a decent policy because it makes it easier to talk about units.

However, Widow by itself could be a decent name.

ParzivalX:
Somehow Sneaky Pete never got changed, maybe exception to rule? 

I'll see if I can modify pull request to change to widow.

GoogleFrog:
Sneaky Pete is legacy. Also, changing names is more complicated now as there are probably some translation files which need changing.

CrazyEddieTK:
Sneaky Pete is an awesome name, which compensates for it being somewhat cumbersome. Black Widow is not only cumbersome but also pedestrian.

I hate to say it, but Widow is not bad. Much better than Infiltrator, which is terrible. The only reservation I have with Widow is that people will keep thinking "Shouldn't it have been 'Black Widow'?" Yes, Widow by itself is a type of spider, but Black Widow is so prominent it people's minds that Widow by itself will seem wrong.

How about Fang? It goes along somewhat with Venom, and both units paralyze. The name is short, brutish, and ugly, i.e. not very mellifluous. I'm usually against names like that, but perhaps it's appropriate for a unit which isn't especially common and only comes out when something very nasty is about to happen.

Anarchid:
Evo has a Widow; noone complains. Neither are there "has to be black" sentiments that i know of WRT Starcraft's Widow mine.

Funfact: Redback is an Australian widow spider.

sprunk:
I always liked Mantis because it patiently waits and then lashes out.

ParzivalX:
But mantis isn't a spider.

sprunk:
Neither is Flea nor Crab(e).

ParzivalX:
Lol, the arthropod factory :D

CrazyEddieTK:
Mantis is good.

GoogleFrog:
PRs can be edited. I think everyone liked Widow. Part of the edit should affect the translation files.

GoogleFrog:
I doubt the PR is going to be updated. Window is a good name though.

Anarchid:
> Window

!

PR #1424, opened 2016-06-03
Starlight remodel

Anarchid:
Replaces Starlight with a superior model. Satellite physically crashes into ground when the unit is destroyed, and can be resurrected under very narrow circumstances.

Satellite is named Owl. It is completely useless.

Stuff left to do:
- `push the button`. I guess it's time to `push the button`. World, my finger is on the button `push the button`.


GoogleFrog:
Burrowing is normal. It is the downbeam cratering affecting the terrain. The cratering seems to be more than usual though.

Anarchid:
102 seems to be adopted, so i'm pushing the button `push the button`.

ashdnazg:
woah

Issue #1438, opened 2016-06-22
pressing flag icon in chat (near label text) does not update some camera variables


FuckYouCannonFodderLovers:
How to reproduce: left-mouse-click flag in chat, then press ctrl+arrow. Rotating will be performed around position of camera that was before clicking on a flag.

Camera - COFC.

sprunk:
https://github.com/ZeroK-RTS/Zero-K/pull/1439/commits/7d3efb3e5ebc68924ba630bdc337fbfb2e78b991

PR #1439, opened 2016-06-23
Fixed camera target setting not handling COFC properly

dvr333:
Fix for #1438


dvr333:
NB: Mission Objective Widget and Deploy Draw gadget not tested (not sure how to test them), Central Build AI tested partially, but I'm not sure how to trigger its SetCameraTarget call. The pattern is fairly straightforward, though, so it should work.

PR #1445, opened 2016-06-28
Add widget which shows decloack range

N0U:
Can this widget be official?


dvr333:
Neat, but I'd recommend mirroring the colours used when you shift+hover over a unit. I'm pretty sure de-cloak radius is blue there.

PR #1447, opened 2016-07-03
Add option for showing/hiding CPU and Ping status

dahn-d:
Add otion to menu and put a bunch of if's here and there. Also make a
small change into width calculator to account possible Ping&CPU hiding.

P.S. Who ever needs these columns btw? o_0

### DIFF
--- a/LuaUI/Widgets/gui_chili_deluxeplayerlist.lua
+++ b/LuaUI/Widgets/gui_chili_deluxeplayerlist.lua
@@ -3,7 +3,7 @@
 
 function widget:GetInfo()
   return {
-    name      = "Chili Deluxe Player List - Alpha 2.02",
+    name      = "Chili Deluxe Player List - Alpha 2.03",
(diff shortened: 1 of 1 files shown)


db81:
>Who ever needs these columns btw? o_0

They're good for telling when people are lagging or catching up, which is difficult to do otherwise.

LuaUI/Widgets/gui_chili_deluxeplayerlist.lua
```diff
@@ -3,7 +3,7 @@
 
 function widget:GetInfo()
   return {
-    name      = "Chili Deluxe Player List - Alpha 2.02",
+    name      = "Chili Deluxe Player List - Alpha 2.03",
```
sprunk: Don't change the name. This resets settings.

dahn-d:
Imo placing the widget version in the name is the worst idea ever.

GoogleFrog:
Yes, widget version in name is the worst idea. Never change the name.

PR #1453, opened 2016-07-08
Remove version number from Decloak Range widget

dahn-d:
Since the name change is resetting user settings, remove the version
form the name. Do it fast so people is not yet used to this widget and
there are not many settings yet.

Also change "Decloak Area" to "Decloak Range" to be consistent with
filename and naming used in Settings.


dahn-d:
I think it would be better close it and make another pr merged with color setting option that I've made in another branch.

PR #1454, opened 2016-07-09
Unit AI Update

mtroyka:
This update includes a bunch of unit AI stuff that I had been working on before dynacoms that never got merged, plus some bugfixes and other things.

EDIT: updated to reflect changes.

Major changes:
-Unidentified radar dots are now treated as lower priority by default, which is more conservative and should stop rackets from preferring to shoot at unidentified radar dots vs known armed targets.

-Added radar wobble penalty for some units that should have had it but did not (incl hammer and reef's cruise missile)

-Felons and other mid-range skirms now skirmish vs things which explode violently when killed, such as fusions, factories, etc, to prevent them from suiciding into the explosion.

-Fixed an issue that I accidentally introduced in a previous update which causes snipers to skirmish at their sight range instead of their max range.

-Fixed scorcher dive being broken.

-Wolverines now skirmish vs all ground units and buildings including unarmed. This is to stop them from wandering into undefended enemy territory and also keeps them from standing on their own mines (which may damage them due to aoe). This also fixes an issue where wolverines would not skirmish properly vs stingers which sometimes led to suicide.

-Added tactical ai for some units that did not have it (impaler, pillager, grizzly).

-Anti-heavies now have consistently higher priority vs heavy units and a consistent penalty vs light units. Bantha was also included since it otherwise tends to waste its tachyon shots vs small stuff.

-Increased the range penalty rate for heatrays somewhat, which should improve their targeting consistency.


GoogleFrog:
I have not had a full look, I should get some time for that next week.

Unidentified radar dots should not be unarmed because I feel it is better to shoot at an unidentified thing than to shoot at something known to be unarmed. The change could be to move unidentified radar dots between armed and unarmed. This would be around priority 12.

Radar wobble penalty should be per-weapon because the effect varys depending on weapon type. Homing missiles do not care about radar wobble at all and neither do inaccurate weapons. Long reload precise weapons are particularly bad against radar dots.

How is Scorcher dive fixed? My issue with it is that it removes your ability to control units. Sometimes it is really good for Scorchers to stay at max range.

Why does anti-air get a bonus against heavy gunships? Against gunships it is generally going to be best to fire at the target which will deal the most cost in damage. If the aim is to fix Trident then some handling for AA should be introduced, not a bonus to fire at heavy things.

mtroyka:
>>Unidentified radar dots should not be unarmed because I feel it is better to shoot at an unidentified thing than to shoot at something known to be unarmed. The change could be to move unidentified radar dots between armed and unarmed. This would be around priority 12.

Fair enough, except it should be 25 since dirtboxes are ~20 and unarmed is +35.

>>Radar wobble penalty should be per-weapon because the effect varys depending on weapon type. Homing missiles do not care about radar wobble at all and neither do inaccurate weapons. Long reload precise weapons are particularly bad against radar dots.

What I had in mind when I did that was probably unrealistic. :P I'll fix that.

>>How is Scorcher dive fixed? My issue with it is that it removes your ability to control units. Sometimes it is really good for Scorchers to stay at max range.

Are there any other units besides glaives and jacks that scorchers should skirmish against? I think proper skirmishing would be infinitely better than not diving vs things that they ought to.

>>Why does anti-air get a bonus against heavy gunships? Against gunships it is generally going to be best to fire at the target which will deal the most cost in damage. If the aim is to fix Trident then some handling for AA should be introduced, not a bonus to fire at heavy things.

The question is more why should bombers get higher priority than, say brawler or krow? If anything the brawler or krow are slower and more likely to be killable before they can run away.

GoogleFrog:
Bombers have bonus priority because they are in range for a very short amount of time and it is most useful to kill them as they approach. It is more important to target fast things with burst damage than it is to target slow things which deal sustained damage.

Scorchers should skirmish many units depending on the situation. Invariably you want to stay at max range against a stunned unit when moving in range would put you in range of turrets. There are many situations which fall somewhere less than this extreme. Ticks and Roaches should be avoided. Raiders should be kept at max range depending on positioning.

I dislike Scorchers closing range because it presents inconsistent behaviour and the behaviour is not always good. Players need to choose between staying at max range or diving and they need to have a UI which allows them to choose. It is easy to manually dive but it is hard to manually tell your units to stay at max range. Therefore it is better for there to be a helpder for max range than it is for diving. The end goal of unit AI is not to make units smart. It is to make it easy to tell your units what to do.

Anarchid:
> Invariably you want to stay at max range against a stunned unit when moving in range would put you in range of turrets. There are many situations which fall somewhere less than this extreme. Ticks and Roaches should be avoided. Raiders should be kept at max range depending on positioning.

I think there are fewer such situations than where it's worth to dive, and when they arise, the player can control them manually. IIRC UnitAI is not about being optimal in all situations and context-aware, or you'd have glaives path around Stardust threatfields.

GoogleFrog:
Are you going to make further modifications? If you do not then I will eventually pull this and make additional modifications.

mtroyka:
>>Bombers have bonus priority because they are in range for a very short amount of time and it is most useful to kill them as they approach. It is more important to target fast things with burst damage than it is to target slow things which deal sustained damage.

reverted this

>>Scorchers should skirmish many units depending on the situation. Invariably you want to stay at max range against a stunned unit when moving in range would put you in range of turrets. There are many situations which fall somewhere less than this extreme. Ticks and Roaches should be avoided. Raiders should be kept at max range depending on positioning.

>>I dislike Scorchers closing range because it presents inconsistent behaviour and the behaviour is not always good. Players need to choose between staying at max range or diving and they need to have a UI which allows them to choose. It is easy to manually dive but it is hard to manually tell your units to stay at max range. Therefore it is better for there to be a helpder for max range than it is for diving. The end goal of unit AI is not to make units smart. It is to make it easy to tell your units what to do.

I added a proper skirmish category for scorchers including most of veryShortRangeSkirmieeArray and some of shortRangeSkirmieeArray, which I've confirmed is working properly. That's about the best that I can do with tactical ai.

Furthermore there IS a UI to tell your units to stay at max range by default. Simply disable their tactical ai. There's a button for that (the lightbulb button..) and a setting in new unit states if you want to apply it to all of your scorchers by default. If you don't want them to chase things into enemy porc you can also set them to hold position, so I don't see the problem.

Anarchid:
Hubs and Caretakers are perceived by the engine as mobile builders with zero speed.

Builders seem to have had preferential targeting in 91.0 already. Engine really wants them to die.

GoogleFrog:
> Furthermore there IS a UI to tell your units to stay at max range by default. Simply disable their tactical ai. There's a button for that (the lightbulb button..) and a setting in new unit states if you want to apply it to all of your scorchers by default. If you don't want them to chase things into enemy porc you can also set them to hold position, so I don't see the problem.

The button takes a long time to press. I much prefer it when you are able to decide between behaviours just by giving either fight or move commands. I'll look at these Scorcher categories.

GoogleFrog:
Are you sure about that distAdd change? I set it to a bonus of at most 10 because the bonus for being unarmed is 15. The intention is to have Scorchers fire at armed targets even if they are further away.

mtroyka:
The bonus for unarmed is +35: https://github.com/ZeroK-RTS/Zero-K/blob/51e092793bbc6e92133aeb8b00cbba6fcc7f886f/LuaRules/Configs/target_priority_defs.lua#L238
+15 is only for bad targets.

afaik most armed units have a base priority between 1 and 5, with the notable exception of dirtbag which has a base priority of ~20. That means that a dirtbox at max range will be somewhat worse prio than an unarmed thing at point blank range, but any other armed unit will always be a better target than unarmed. I used +20 to make range the dominating factor so that they shoot at whatever is closest most of the time.

I haven't seen any issues with things shooting at unarmed when they shouldn't be, which apparently was caused by the gadget failing to load rather than any particular settings that I had used.

GoogleFrog:
Ah ok. With +35 the change makes more sense.

PR #1465, opened 2016-07-20
Autogenerate sprungboxes for featured maps

Anarchid:
This applies sprungboxes to all currently featured maps which had no sprungbox previously by converting Springie startbox commands in database to startpolygons in the following way:

1) Obtain Springie command lists for all featured maps:
```
SELECT TOP 1000 [ResourceID]
      ,[InternalName]
      ,[MapWidth]
      ,[MapHeight]
      ,'"'+[MapSpringieCommands]+'"' as Commands
  FROM [zero-k].[dbo].[Resources]
  WHERE TypeID = 0 AND NOT FeaturedOrder IS NULL
```

2) Export results of above as [CSV](http://hastebin.com/raw/ihakigurox). 

3) Parse and convert using [this script](https://gist.github.com/Anarchid/509028e94e840c78d08d639ebeaa8474).

PR because i have not tested this yet.


LuaRules/Configs/StartBoxes/duck.lua
```diff
@@ -0,0 +1,2 @@
+return {
```
Anarchid: Why is this even featured.

PR #1484, opened 2016-08-08
Allow wind swaying for all plant-like features ever made for spring

ForbodingAngel:
As well as all features that exist within spring features.

### DIFF
--- a/ModelMaterials/4_features_trees.lua
+++ b/ModelMaterials/4_features_trees.lua
@@ -102,12 +102,25 @@ local materials = {
 
 local featureMaterials = {}
 local featureNameStubs = {
-	-- all of the 0ad, beherith and artturi features start with these.
-	{str = "ad0_", prefix = true}, 
-	{str = "btree", prefix = true}, 
-	{str = "art", prefix = true}, 
-	-- Other trees will probably contain "tree" as a substring.
+	-- Support all features ever made for spring
 	{str = "tree", prefix = false}, 
+	{str = "bush", prefix = false},
+	{str = "vegetation", prefix = false},
+	{str = "baobob", prefix = false},
+	{str = "aleppo", prefix = false},
+	{str = "pine", prefix = false},
+	{str = "senegal", prefix = false},
+	{str = "palm", prefix = false},
+	{str = "shrub", prefix = false},
+	{str = "bloodthorn", prefix = false},
+	{str = "birch", prefix = false},
+	{str = "maple", prefix = false},
+	{str = "oak", prefix = false},
+	{str = "fern", prefix = false},
+	{str = "grass", prefix = false},
+	{str = "weed", prefix = false},
+	{str = "plant", prefix = false},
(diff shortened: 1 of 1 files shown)


Anarchid:
This already allows too much features: for example, streetlights on Intersections should definitely **not** sway eldritchly with the wind.

ForbodingAngel:
WTF? Why would streetlights on intersection sway?

`[streetlamp_1]
{
	world=allworld;
	description=Street Lamp;
	object=streetlamp_1.s3o;
	footprintx=1;
	footprintz=1;
	height=36;
	blocking=1;
	hitdensity=5;
	metal=5;
	energy=200;
	damage=40;
	flammable=0;
	reclaimable=1;
	autoreclaimable=1;
	featurereclamate=smudge01;
	seqnamereclamate=tree1reclamate;
	upright=1;
}`

This is the lone featuredef in Intersection v3. Please tell me at what point the name "streetlamp_1" contains any of the following words:

- tree
- bush
- vegetation
- baobob
- aleppo
- pine
- senegal
- palm
- shrub
- bloodthorn
- birch
- maple
- oak
- fern
- grass
- weed
- plant
- palmetto

As you can see, none of those words exist in the name "streetlamp_1", therefore it would not sway.

Anarchid:
The commit and PR are badly worded, but Intersection streetlights swaying is not something i **read** but rather, something i **observed** last time when i played Intersection in ZK.

ForbodingAngel:
Hmm, fair point. If this did category parsing 
```
	category			= "urban",
```
in features rather than name parsing, everything would work better. All the features in SF have categories assigned to them.

Anarchid:
That woild deal with streetlights, but i still fear for cloakbots. Do only Gaia features get sway?

ForbodingAngel:
Well, my thinking is that if it looks in the def for category and filters out anything that isn't in the category  of vegetation (also check for vegitation because some people can't spell :-)), then the problem solves itself, does it not?

Anarchid:
That seems safer in the sense that it extra unlikely to allow non-vegetative things to sway. You could also, at that point, remove the name matching wholesale.

Unsure if there exist "vegetation" category things which shouldn't sway, or vice versa.

Anarchid:
It seems that `category = "vegetation"` things exist as well - beherith's rocks are all "vegetation". So you definitely can't do simple category whitelisting, name matching has to stay...

GoogleFrog:
ZK-side features should be easily excludable. We can add customParams or, if customParams don't exist, put particularly blacklisted substrings in their name.

ForbodingAngel:
The rocks have that category? well fudge me. Ok I'll fix it right now. SF wasn't supposed to be a one person enterprise, but sadly that's the way it turned out. Thanks for letting me know. Feel free to post issues on the tracker on the SF repo if you find other stuff.

sprunk:
>some people can't spell :-)

>baobob

ForbodingAngel:
Touche :-)

Issue #1464, opened 2016-07-19
Rejoining as a spectator to a not-yet-started game zoomed to start position


GoogleFrog:
I rejoined a game as a spectator and the game had not yet started. My camera was zoomed to focus on a start location. Spectators should not get this zoom.

dvr333:
Odd. I thought https://github.com/ZeroK-RTS/Zero-K/commit/6e830bc76dfb2669163d30f83bc7e1eb0b2ceb70 fixed it.

mtroyka:
I've been seeing this issue as well. It zooms to a startbox at the beginning, but when you try to move the camera at all it snaps back to the middle of the map.

PR #1491, opened 2016-08-17
Possible fix for #1464

dvr333:


GoogleFrog:
This cannot fix #1464 because I do not use COFC. However if it has another purpose it can be pulled.

dvr333:
Oh, I see. Well, it fixes a related bug regarding the handoff from TA camera to COFC at game start, and guarantees that COFC will start zoomed out looking at the map as players would expect.

PR #1500, opened 2016-09-01
Unit AI Fix

mtroyka:
This fixes certain issues where arty (and particularly snipers) would sometimes walk into enemy units.


sprunk:
How does this work together with Impaler chasing? Won't this make an Impaler try to chase mobile units?

mtroyka:
it's not just about chasing, but also about skirmishing. Units will apparently only skirmish against things that they don't have set to nochase, so basically it allows them to avoid raiders and things. Impaler weapon targeting is independent of chase, so it makes no difference in terms of what impaler will shoot at.

sprunk:
Yes, skirming and chasing are controlled by the same tag. Won't fixing skirming break chasing?

mtroyka:
How so? It generally skirms vs the nearest enemy, so it still won't follow raiders into enemy porc or anything like that.

GoogleFrog:
@mtroyka is correct. NoChase units are potentially ignored by unitAI, preventing artillery from running away from small units. 

Impaler chasing small units is dealt with by the gadget. Automatically assigned attack commands (maneuver target acquisition and fight) are detected by the tactical AI gadget and cause units to skirm the nearest enemy, not the target of the attack command. If you want your units to chase a particular unit into a dangerous area you need to give the attack command manually.

PR #1504, opened 2016-09-07
Allow AIs to check for valid enemy startpos

mtroyka:
Since startboxes are no longer exposed to AIs through rules params they are no longer able to infer enemy start positions. To get around this I added an "ai_is_valid_enemy_startpos:" ai callin so that they can  do the same thing via rules calls.

Note: I haven't tested this because there's no skirmish mode.. additionally it's possible that Spring.GetAllyTeamList() and Spring.GetTeamList() do not return an ordinal list as documented but instead use the ally/team ids as keys, which would require changing the code.

### DIFF
--- a/LuaRules/Gadgets/start_boxes.lua
+++ b/LuaRules/Gadgets/start_boxes.lua
@@ -283,7 +283,14 @@ end
 
 function gadget:RecvSkirmishAIMessage(teamID, dataStr)
 	local command = "ai_is_valid_startpos:"
-	if not dataStr:find(command,1,true) then return end
+	local command2 = "ai_is_valid_enemy_startpos:"
+	if not dataStr:find(command,1,true) and not dataStr:find(command2,1,true) then return end
+	
+	if dataStr:find(command2,1,true) then
+		command = command2
+	end
+
+	local boxID = Spring.GetTeamRulesParam(teamID, "start_box_id")
 
 	local xz = dataStr:sub(command:len()+1)
 	local slash = xz:find("/",1,true)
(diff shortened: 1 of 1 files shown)


Licho1:
Hmm so we should add boxes to startscript again?

mtroyka:
Boxes previously cluttered up gamerulesparams like nobody's business, so I'd have to lean toward "no". It's also more complicated to have every ai parse the boxes instead of just using rules calls and letting lua do it.

For reference as far as I know only ZKGBAI parsed the startboxes directly anyway, circuit just uses the ai_is_valid_startpos call, but circuit doesn't check enemy positions either. KGB only tests metal spots so the number of tests generally isn't huge.

Anarchid:
Another option would be to add a CallRules for AI's to query boxes without cluttering up GRP.

LuaRules/Gadgets/start_boxes.lua
```diff
@@ -293,10 +300,33 @@ function gadget:RecvSkirmishAIMessage(teamID, dataStr)
...
-		return "0"
+		-- for checking enemy startpos
+		local enemyboxes = {}
+		local allyteamid = Spring.GetAllyTeamID(teamID)
+		local allyteams = Spring.GetAllyTeamList()
+		
+		for _,value in pairs(allyteams) do
+			if value ~= allyteamid then
+				local enemyteams = Spring.GetTeamList(value)
+				enemyboxes[Spring.GetTeamRulesParam(enemyteams[1], "start_box_id")] = true
```
sprunk: This reveals the true box after shuffling.

PR #1506, opened 2016-09-08
Area-reclaim behaviour changes:

sprunk:
* metal-less features (mostly trees) are no longer reclaimed by default.
   Many maps have abundant trees on which cons trying to reclaim wreckage
   waste a lot of time.

 * CTRL can now be used to force reclaim of trees.

 * CTRL is no longer force reclaim of own units.
   This functionality is bugged (cons in the area will reclaim each other),
   very rarely useful and detrimental if used accidentally. One can still
   reclaim units one by one.

### DIFF
--- a/LuaUI/Widgets/cmd_commandinsert.lua
+++ b/LuaUI/Widgets/cmd_commandinsert.lua
@@ -83,10 +83,10 @@ end
 
 local function ProcessCommand(id, params, options)
   local alt,ctrl,meta,shift = Spring.GetModKeyState() --must use this because "options" table turn into different format when right+click. Similar problem with different trigger see: https://code.google.com/p/zero-k/issues/detail?id=1824 (options in online game coded different than in local game)
-  if (ctrl) and not (meta) and (id==CMD.RECLAIM or id==CMD.REPAIR) then --Reclaim + CTRL
+  if (ctrl) and not (meta) and id==CMD.REPAIR then
 	local opt = 0
 	if options.alt or alt then opt = opt + CMD.OPT_ALT end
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Perhaps use CTRL to activate the area unit reclaim widget. The widget would queue reclaim orders on all units matching the unit type at the center of the command. It was disabled by default because it had no modifiers and people complained about accidentally hovering over their own units and reclaiming them when issuing an area command.

sprunk:
Any feedback on the other PR contents?

So ctrl-drag-reclaim would be a bit like drag-guard, where it does a different thing depending on whether a unit is in the center (G: circle orbit / R: reclaim unit type) or not (add individual guards / reclaim trees)?

GoogleFrog:
I think it is fine. I never knew that CTRL force reclaimed your own units. Failing to reclaim trees with area reclaim will be an annoyance for people who don't look up the modifiers (basically everyone) but then I think it is worth the extra control provided.

Ctrl + reclaim on units would reclaim units of a type on the centre of the command. Otherwise it would reclaim trees.

GoogleFrog:
Sorry, I thought it was fine but CTRL makes it very hard to reclaim trees as metal is prioritized over trees in every case.

Also, right clicking and dragging on a tree should reclaim trees.

sprunk:
As far as I can see the engine does the metal>energy prioritisation in the area-command so a tree-only reclaim would need to convert it directly into a series of individual reclaim commands. It's not terribly bad but it makes for inconsistent UI (energy reclaim would be shown as a series of individual reclaims but metal would show as the circle). On the other hand it would never be incorrect because new trees cannot normally appear.

PR #1507, opened 2016-09-08
Fix issues with ai callins to the startbox gadget.

mtroyka:
It turns out there is no such function as Spring.GetAllyTeamID(), and additionally there was a nil error when it tried to get the startbox for gaia assuming it to be an enemy team.


Anarchid:
I assume it's tested this time :P

PR #1508, opened 2016-09-08
In-game ignore

ArchShaman:
Adds an ignore feature in-game.


LuaUI/cawidgets.lua
```diff
@@ -15,6 +15,7 @@
 --------------------------------------------------------------------------------
 
 -- stable release?
+local ignorelist = {count = 0,ignorees ={}} -- Ignore workaround for WG table.
```
sprunk: Why the separate count if you're indexing by name?

PR #1514, opened 2016-09-14
Area-reclaim changes:

sprunk:
- Ctrl + Area Reclaim to reclaim own units now only works when the center of the area is a unit (like Circle Orbit Guard). Only reclaims that unit type. This converts the area command to a series of individual reclaims (so won't update when units come and leave the area).
- CTRL + Area Reclaim, when hovered over a feature with metal = 0 (usually a tree), will reclaim metal = 0 features only. Also a series of individual reclaims except this time it has no UI consequences because trees cannot move [citation needed].
- Ctrl + Area Reclaim not hovered over a tree or live unit will eat all features unconditionally. This means that in addition to metal, it will reclaim wrecks currently being resurrected and trees.
- Regular Area Reclaim (without CTRL) will no longer reclaim metal = 0 features.

### DIFF
--- a/LuaUI/Widgets/cmd_commandinsert.lua
+++ b/LuaUI/Widgets/cmd_commandinsert.lua
@@ -7,8 +7,7 @@ function widget:GetInfo()
     name = "CommandInsert",
     desc = "[v" .. version .. "] Allow you to add command into existing queue. Based on FrontInsert by jK"..
 			"\n• SPACEBAR+SHIFT insert command to arbitrary places in queue."..
-			"\n• SPACEBAR insert command in front of queue."..
-			"\n• CTRL+Reclaim force reclaim your own unit in an area.",
+			"\n• SPACEBAR insert command in front of queue.",
 	author = "dizekat",
     date = "Jan,2008", --16 October 2013
     license = "GNU GPL, v2 or later",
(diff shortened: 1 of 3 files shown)


GoogleFrog:
I dislike the Ctrl+Reclaim on trees because trees can be destroyed. This can make the order replacement version inefficient. It is also ugly on the UI.

sprunk:
How is it inefficient? If a tree is removed, so is the order.

sprunk:
My main concern is that by default, area-reclaim will pick whatever is closest. This makes trees bad because now cons waste time reclaiming useless forests instead of wrecks. 

The solution I used is to remove trees from being chosen by default.

But people sometimes do want to reclaim a forest. So there has to be some mode to enable that. At first I tried to use the builtin engine way to force reclaim. However features that are not reclaimed by default are de-prioritised under the force-reclaim mode so trees would be hard to reclaim. So I had to use manual replacement with individual orders.

Licho1:
It could just be smart about it .. stalling metal (Default state) prioritize metal, not enough energy? Go after tree.

Anarchid:
Smarts are bad when they add opaqueness.

sprunk:
Being smart about it requires rewriting the command in Lua.

Also, usually I want to prioritize the things I want regardless of what my storage is.
 * perhaps I have a lot of E but I still prioritize reclaim E because I have no solars and it's a felon rush.
 * even I have lots of metal I usually still want to reclaim wrecks to deny from enemy.

sprunk:
#1649

PR #1524, opened 2016-09-18
WIP: Loadscreen. Shows the map and player roster.

sprunk:
Examples:

Teams or 1v1:
teams_loadscreen

FFA:
ffa_loadscreen

Team FFA:
teamffa_loadscreen


sprunk:
The Spring license is not removable and there is no guarantee on how large it actually is horizontally (vertically it's 10% of the screen but screen resolutions differ). This means the bottom area is mostly unusable so I relegated the progress bar there.

For 2-team setups the map is at the center and the people are listed on the sides. Else, the map is at the edge and people are listed on one side groupped by teams. If the list is too long it gets split into two columns.

mtroyka:
Teams and team FFA need team titles to make the teams more obviously teams. Otherwise it looks awesome.

I dunno about the lack of tips. There's obviously nowhere to put them, but on the other hand showing them before the game (when you can't actually try anything) isn't necessarily the best either so.. I guess who cares?

Licho1:
I would like this to be a small overlay in some corner and not a fullscreen replacement.
Our loading screens are pretty and tips might be more useful to newbies than team layout or minimap (which becomes visible in just a few seconds after load).

Licho1:
Also you could render boxes on minimap

GoogleFrog:
I like this technically, the layout is a bit ugly. Our current small loading bar is better than a large one with no margins. The loadscreen must have a nice background. I'd suggest trying to place the loading information somewhere smaller in a way that is compatible with the current loadscreen (as tips are useful).

Alternately you could make a standardized area for tips and include some generic backgrounds.

sprunk:
The small loading bar is nicer but it leaves the license text area (bottom 10% vertically) devoid of anything useful. Vertical space is comparatively precious.

I could add a line for tips above the progress bar. It won't really fit anywhere else - on small resolutions the screen will be pretty cramped with a large FFA - but again it's vertical space. I feel like the minimap would feel way too small when shrunk, especially if both nice-loadbar and tip-area were done (since it has to keep ratio so X size would reduce as well) but it can be given a try.

Licho1:
Well the license is useless and it's space could be used, but loading screenhot and big tips are well made imo and not worth replacing by dynamic minimaps. It's just not as visually pleasing.

GoogleFrog:
The current tips are well made but I would like many more of them and nobody is going to make something which takes more effort than taking a screenshot and editing a text file. Here are my priorities of display for a loadscreen:
 * Loading bar.
 * Nice screenshot.
 * More tips than we have now.
 * Current tips.
 * Minimap.
 * Playerlist.
 * Map name.
 * Game version.
 * Spring version.
 * Licenses.

Licho1:
Tbh I woudlnt bother people with license or spring version or even game version .. map name .. maybe but they are long and odd, not short names from other games.

GoogleFrog:
I wouldn't bother with them either. Notice that they are the bottom of my priority list.

Anarchid:
License text is mandatory.

Licho1:
Mandatory?? On the loading screen?? With size 36 font?? Where in the license it says that license on loading screen is mandatory.

Anarchid:
Chobby had exactly this discussion and there was much digging about GPL. 

TL:DR Spring's GPLv2 requires the license text.

Anarchid:
See https://github.com/spring/spring/pull/254

sprunk:
Looking at the priority list I think the only thing I can do somewhat competently is map name.

Minimap and Playerlist would mostly obscure Nice Screenshot if they're large and I don't think they'd look any good if not covering most of the screen. Minimap in the corner (like in the UI) sounds alright in theory but if you have something that looks like the UI it gets interpreted as such which'd cause a lot of confusion.

Licho1:
@Anarchid linked spring ticket does not indicate game needs license text on every loading screen, quite the contrary.. 
Im fine with showing license it could even be on some chobby (launcher) loading screen, but wasting it "in-game" is imo very ugly and not-needed.

sprunk:
http://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

Point 2c:
>If the modified program normally reads commands interactively when run
Spring is interactive, so the next parts hold.

>you must cause it, when started running for such interactive use in the most ordinary way, to print or display an announcement including an appropriate copyright notice and a notice that there is no warranty (or else, saying that you provide a warranty) and that users may redistribute the program under these conditions, and telling the user how to view a copy of this License
So it has to print the license stuff.

>(Exception: if the Program itself is interactive but does not normally print such an announcement, your work based on the Program is not required to print an announcement.)
Exception does not apply because Spring normally prints the announcement.

Commits to Spring are works derived from Spring. So a patch that removes the license cannot be done except by the author, or if the license is changed (which itself can only be done by author).

Spring has no single author. You'd have to ask everybody who contributed to Spring to allow removing the license text (or to change license). Existing engine devs don't allow that.

Licho1:
This is perfectly fine, so if its shown on the "program start" from user's perspective - when chobby launcher checks for updates, it meets this condition and we don't need it on every load screen.

sprunk:
Under existing loadscreens there is nowhere to put the map name without obstructing the ZK logo or brotips.

Licho1:
Even if its really tiny 64x64 pix minimap + few player name highlights?

Anarchid:
It should be entirely possible to manufacture new screenshots, we have both the sources and the person who made them. Bonus point is that new screenshots may look better due to dnts and ssdl

The-Yak:
With latest maintenance it is not possible to make $maptexture calls from luaintro anymore.

sprunk:
sad!

PR #1535, opened 2016-09-25
absolutely nicer boxes

DeinFreund:
boxes and startpos for abs v2


sprunk:
Looks decent but:
absolutely
 * the blue areas could be added (why omit?)
 * the red areas should be removed (no normal exit and no mexes)
 * the yellow area could be removed because it's ground level and removing red would else create a hole

DeinFreund:
I left yellow in because it's a common startpos. Cutting out polygons is annoying to do..

Don't mind about the blue ones, but it's only 200 elmos to walk and no spot I see usually occupied. Feel free to change boxes or keep boxes as they are, I just want to have the startpos for AIs..

PR #1544, opened 2016-10-05
Unit AI: Impaler, Reef, Detri

mtroyka:
-OKP vs statics now applies to mobile units that are not moving.
-Impaler, Reef, Detri rockets now disfavor targeting moving units.
-Reef and detri rockets gain OKP vs statics.
-Increased skirmish leeway for impaler so they get closer to targets (and get more targets to choose from).

### DIFF
--- a/LuaRules/Configs/tactical_ai_defs.lua
+++ b/LuaRules/Configs/tactical_ai_defs.lua
@@ -914,7 +914,7 @@ local behaviourConfig = {
 		skirmRadar = true, 
 		swarms = {}, 
 		flees = {},
-		skirmLeeway = 150, 
+		skirmLeeway = 600, 
 	},
 	
 	["cormart"] = {
@@ -922,7 +922,7 @@ local behaviourConfig = {
 		skirmRadar = true,
 		swarms = {}, 
 		flees = {},
-		skirmLeeway = 100, 
+		skirmLeeway = 400, 
 	},
 	
 	["armraven"] = {
--- a/LuaRules/Configs/target_priority_defs.lua
+++ b/LuaRules/Configs/target_priority_defs.lua
@@ -84,6 +84,12 @@ local unitIsFighterOrDrone = {
 	[UnitDefNames["carrydrone"].id] = true,
 }
 
+-- Wolverine mines are stupid targets.
+local unitIsClaw = {
+	[UnitDefNames["wolverine_mine"].id] = true,
+}
+
+
 -- swifts should prefer to target air over ground
 local unitIsBadAgainstGround = {
 	[UnitDefNames["fighter"].id] = true,
@@ -171,9 +177,16 @@ for wdid = 1, #WeaponDefs do
 	end
 end
 
+local starburstWeaponDefs = {
+	[WeaponDefNames["armmerl_cortruck_rocket"].id] = true,
+	[WeaponDefNames["reef_armmship_rocket"].id] = true,
+	[WeaponDefNames["armorco_orcone_rocket"].id] = true,
+}
+
 local radarWobblePenalty = {
-	[WeaponDefNames["armmerl_cortruck_rocket"].id] = 5,
-	[WeaponDefNames["reef_armmship_rocket"].id] = 5,
+	[WeaponDefNames["armmerl_cortruck_rocket"].id] = 100,
+	[WeaponDefNames["reef_armmship_rocket"].id] = 100,
+	[WeaponDefNames["armorco_orcone_rocket"].id] = 5,
 	[WeaponDefNames["armsnipe_shockrifle"].id] = 5,
 	[WeaponDefNames["armanni_ata"].id] = 5,
 	[WeaponDefNames["armmanni_ata"].id] = 5,
@@ -244,6 +257,8 @@ for uid = 1, #UnitDefs do
 	for wid = 1, #WeaponDefs do
 		if unitIsUnarmed[uid] then
 			targetTable[uid][wid] = unitHealthRatio[uid] + 35
+		elseif unitIsClaw[uid] then
+			targetTable[uid][wid] = unitHealthRatio[uid] + 1000
 		elseif (unitIsFighterOrDrone[uid])
 			or (weaponBadCats[wid].fastStuff and unitIsTooFastToHit[uid])
 			or (weaponBadCats[wid].fixedwing and unitIsFixedwing[uid])
@@ -269,4 +284,4 @@ for uid = 1, #UnitDefs do
 	end
 end
 
-return targetTable, captureWeaponDefs, gravityWeaponDefs, proximityWeaponDefs, radarWobblePenalty, transportMult
+return targetTable, captureWeaponDefs, gravityWeaponDefs, proximityWeaponDefs, starburstWeaponDefs, radarWobblePenalty, transportMult
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -77,6 +77,8 @@ local HandledUnitDefIDs = {
 	-- Static only OKP below
 	[UnitDefNames["amphfloater"].id] = true,
 	[UnitDefNames["armmerl"].id] = true,
+	[UnitDefNames["reef"].id] = true,
+	[UnitDefNames["armorco"].id] = true,
 	[UnitDefNames["corstorm"].id] = true,
 	[UnitDefNames["corthud"].id] = true,
 	[UnitDefNames["spiderassault"].id] = true,
@@ -122,7 +124,7 @@ function GG.OverkillPrevention_IsDisarmExpected(targetID)
 	return false
 end
 
-local function IsUnitIdentifiedStructure(identified, unitID)
+local function IsUnitIdentifiedStructure(identified, inLOS, unitID)
 	if not identified then
 		return false
 	end
@@ -130,7 +132,21 @@ local function IsUnitIdentifiedStructure(identified, unitID)
 	if not (unitDefID and UnitDefs[unitDefID]) then
 		return false
 	end
-	return not Spring.Utilities.getMovetype(UnitDefs[unitDefID])
+	
+	if (Spring.Utilities.getMovetype(UnitDefs[unitDefID])) then
+		--for units
+		if (inLOS) then
+			local vx, vy, vz, vl = Spring.GetUnitVelocity(unitID)
+			local isNotMoving = (vx == 0 and vy == 0 and vz == 0 and vl == 0)
+			if isNotMoving then
+				return true
+			end
+		end
+		return false
+	else
+		--for buildings
+		return true
+	end
 end
 
 --[[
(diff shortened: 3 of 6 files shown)


GoogleFrog:
I dislike the tactical AI def changes. The basic purpose of the fight command is to provide predictable behaviour which works in many cases. The predictability is important so that players can give a fight command and know how their unit is going to behave. Adding a special case incurs a cost because it makes it harder to remember how your units will behave. Special cases should also not remove desired behaviour.

The easy-to-remember behaviour for long ranged units is that fight will make them stick to their maximum range. I don't want to add exceptions because sometimes it is better to move further forwards to fire at more important targets. If two behaviours are beneficial across equally many cases then fight should implement the one which is harder to implement manually. It is easier to make your artillery focus fire (one attack command) than it is to make it keep maximum range (constant attention and move commands). The increased skirmLeeway introduces failure cases, Pillager can be approached and killed by a Penetrator without trying to move away. Pillager will respond more slowly to approaching enemies which are trying to rush it.

Furthermore, changing skirmLeeway will not actually make Impaler and Pillager move closer to their targets. The "stop when in range" behaviour is an engine-inbuilt feature of Fight. All the skirmLeeway will do is to make them more tolerant of enemies which move towards them. I am not sure though as I have not tested it.

GoogleFrog:
I'm not sure about the change to disfavour targeting currently mobile units. However, my view here is mostly based on performance and code complexity. I think I would have added a static penalty to fire at mobile units directly in the defs.

I am beginning to think that dynamic mobile detection is good because it links into something else I wanted to do. Currently paralyzed units are given low priority, this makes sense because they can't do any damage and are often able to be permastunned. However, this does poorly in the example of  Venom + Hermit against raiders. The optimal way for Venom + Hermit to fight raiders might be for Hermits to prefer shooting at the paralyzed units because they are much more likely to hit them. Killing the paralyzed units lets the Venom switch target. I was thinking of some sort of fast unit target inversion cludge but your speed tracking may enable this to be done more cleanly. The IsTooFastToHit table may even be removable.

Every weapon could have a speed threshold. If a unit is currently moving faster than this speed then the unit is given a lower target priority. This priority penalty may scale with how much the unit exceeds the threshold. In general any function could be used but I think linear is sufficient. A fixed size factor might also be desirable because many weapons are much better at hitting large units.

A speed and size multiplier would be more flexible than an inversion cludge. Hermits may end up shooting at Bandits over Glaives because they are more likely to hit Glaives. Scalpels could fire at the Paralyzed aircraft instead of the one they have barely any chance of hitting.

If this change is meant to override the paralysis penatly, in certain cases, then paralysis probably needs some changes. Currently stunned or inbuild is lumped into the same case nanoframes should keep most of their current priority penalty. It may be important to shoot at a near-complete nanoframe, however, mobile units should be fired at before most nanoframes. I don't want people baiting Ravagers with Defender nanoframes instead of the Ravagers shooting Glaives.

GoogleFrog:
"OKP vs statics now applies to mobile units that are not moving." is bad, particularly for mixed armies. Imagine a Rogue and a Scalpel coming across a static Glaive. The Rogue fires first so the Scalpel holds its fire, thinking the Glaive is dead. The Scalpel will not fire until the Rogue rocket times out.

Is that the old problem? Assume that we fix the problem by having Glaive movement cause rockets flying towards it to time out, allowing all units to shoot at it. There are still poor cases:
 * If the Glaive moves out of the way of the rocket then this movement will cause the Scalpel to fire. The time which the Glaive waited to move is effectively added to the Scalpel's reload time. It would have been better to shoot the Glaive immediately.
 * A worse target than the Glaive may enter the Scalpel's range. This causes the Scalpel to fire, allowing the Glaive to escape (or possibly kill the Scalpel).
 * The Scalpel may move such that the Glaive is out of its range, letting the Glaive escape.

The only good outcome occurs if the rocket hits the Glaive (which basically never happens) or if something more valuable than the Glaive enters the Scalpel's range (such prediction, in general, is beyond the scope of this gadget). This OKP does not even work for Impaler, replace Impaler with Rogue and the situation is the same.

This issue arises from the assumed equality of all (damaging) projectiles in OKP. If a unit has enough incoming damage to kill it then it is assumed that the unit is dead. The Rogue is lying to the OKP gadget when it says that the rocket it fired is highly likely to hit the Glaive. Solving this problem probably involves multiple tiers of reliability, a Rogue may prevent other Rogues from firing but do nothing to Scalpels. However, Rogues not firing may still be bad and I don't want to increase the complexity of the gadget for this marginal benefit.

Rockos and many plasma assaults should not have mobile OKP because spamming projectiles at a lone target often decreases the survival chances of the target. Rocko has such a low reload time that it should fire whenever there is a chance of its rocket being useful. Stray rockets can hit nearby units which have not yet entered range or give the target more difficulty dodging.

Perhaps mobile OKP is only useful in edge cases, such as Impaler. This feels like a cop-out though as Impaler should simply be heavily de-prioritized from firing at mobiles in the first place.

I hold OKP to a higher standard than target priorities because target priorities fail gracefully whereas OKP can easily create broken and exploitable behaviour.

mtroyka:
>>Furthermore, changing skirmLeeway will not actually make Impaler and Pillager move closer to their targets. The "stop when in range" behaviour is an engine-inbuilt feature of Fight. All the skirmLeeway will do is to make them more tolerant of enemies which move towards them. I am not sure though as I have not tested it.
..
>>Pillager will respond more slowly to approaching enemies which are trying to rush it.

I was more or less aware of that. I considered implementing a dive but decided against it partly because lazy and partly because I didn't want them rushing towards enemies. I think the case for impalers isn't that strong, since even with the increased skirmish leeway they still have plenty of breathing space vs most enemies, and otherwise impalers can end up backing up far enough to be useless even when they have plenty of porc coverage. For pillagers the case vs penetrator is kind of bad so I'll just return that to 50. For impalers I'll decrease it a bit so that they properly skirmish vs penetrators at least.

>>vl is the length of the velocity vector. You don't need to use vx, vy and vz for what you are doing.

I knew that :P 

>>I am beginning to think that dynamic mobile detection is good because it links into something else I wanted to do. Currently paralyzed units are given low priority, this makes sense because they can't do any damage and are often able to be permastunned. However, this does poorly in the example of Venom + Hermit against raiders. The optimal way for Venom + Hermit to fight raiders might be for Hermits to prefer shooting at the paralyzed units because they are much more likely to hit them. Killing the paralyzed units lets the Venom switch target. I was thinking of some sort of fast unit target inversion cludge but your speed tracking may enable this to be done more cleanly. The IsTooFastToHit table may even be removable.

>>Every weapon could have a speed threshold. If a unit is currently moving faster than this speed then the unit is given a lower target priority. This priority penalty may scale with how much the unit exceeds the threshold. In general any function could be used but I think linear is sufficient. A fixed size factor might also be desirable because many weapons are much better at hitting large units.

>>A speed and size multiplier would be more flexible than an inversion cludge. Hermits may end up shooting at Bandits over Glaives because they are more likely to hit Glaives. Scalpels could fire at the Paralyzed aircraft instead of the one they have barely any chance of hitting.

That could work as well. I did not see any issues with performance but one issue that I did see was that it's sometimes bad about seeing moving units as being non-moving which I suspect is due to momentary pauses in the unit's motion. Tracking units that have recently moved and their max speed would probably work but introduces complications wrt los.

>>If this change is meant to override the paralysis penatly, in certain cases, then paralysis probably needs some changes. Currently stunned or inbuild is lumped into the same case nanoframes should keep most of their current priority penalty. It may be important to shoot at a near-complete nanoframe, however, mobile units should be fired at before most nanoframes. I don't want people baiting Ravagers with Defender nanoframes instead of the Ravagers shooting Glaives.

Paralyze was not something I had considered with this, but is an interesting case. Obviously it's probably better for units with slow projectile velocity weps to target paralyzed stuff that can't dodge compared to things that are moving since it's basically a free shot and now okp will be applied sensibly as well so nothing is wasted. Nanoframes should probably get a separate penalty based on 1-%completed so that units/buildings close to completion have a small penalty and new frames get a higher penalty.

That should be easy enough to implement, my question is should the paralyze penalty be kept at all or just be replaced by a nanoframe-only penalty?

mtroyka:
>>"OKP vs statics now applies to mobile units that are not moving." is bad, particularly for mixed armies. Imagine a Rogue and a Scalpel coming across a static Glaive. The Rogue fires first so the Scalpel holds its fire, thinking the Glaive is dead. The Scalpel will not fire until the Rogue rocket times out.

>>Is that the old problem? Assume that we fix the problem by having Glaive movement cause rockets flying towards it to time out, allowing all units to shoot at it. There are still poor cases:

>>If the Glaive moves out of the way of the rocket then this movement will cause the Scalpel to fire. The time which the Glaive waited to move is effectively added to the Scalpel's reload time. It would have been better to shoot the Glaive immediately.
A worse target than the Glaive may enter the Scalpel's range. This causes the Scalpel to fire, allowing the Glaive to escape (or possibly kill the Scalpel).
The Scalpel may move such that the Glaive is out of its range, letting the Glaive escape.

OKP inherently assumes that wasted shots are expensive for the unit using it, whether because the shots miss or because they overkill. Scalpel is more of an outlier case for OKP since its projectiles are homing. Also abusing those cases would be really difficult since the required conditions (small numbers of unsupported skirms vs small numbers of fast units that come into range and suddenly stop moving, or which aren't moving when skirms approach them and the skirms don't immediately retreat upon scouting them -and- the units don't start skirming/swarming when shot at) are not that common nor easy to control. Even widgetizing that usefully would be very difficult and it'd interfere with normal tactical ai.

>>Rockos and many plasma assaults should not have mobile OKP because spamming projectiles at a lone target often decreases the survival chances of the target. Rocko has such a low reload time that it should fire whenever there is a chance of its rocket being useful. Stray rockets can hit nearby units which have not yet entered range or give the target more difficulty dodging.

You're forgetting something in this case, which is that target leading can only be applied if a unit is already moving (and okp only applies if the unit isn't moving or I suppose if the projectiles are homing). If the target doesn't move then the shots are basically guaranteed to hit, but if the target does move then they all have a high likelihood of missing uniformly. On the other hand if the okp-inhibited units wait to fire then they can have a higher chance to hit by applying target leading, or may find better targets to shoot at intentionally. The random case is random and beyond the scope of unit ai.

>>The only good outcome occurs if the rocket hits the Glaive (which basically never happens) or if something more valuable than the Glaive enters the Scalpel's range (such prediction, in general, is beyond the scope of this gadget).

I thought it was the whole point of the gadget. "Don't waste shots because there may be better targets in range now or better targets may come into range soon", no?

>>Perhaps mobile OKP is only useful in edge cases, such as Impaler. This feels like a cop-out though as Impaler should simply be heavily de-prioritized from firing at mobiles in the first place.

That was originally my intention, but observing impaler kills informed me otherwise. There are lots of cases where impalers can get free kills on mobile units, such as constructors that have stopped moving in order to build stuff, idle/paralyzed units (including coms and striders), sitting crabes, etc.

>>This OKP does not even work for Impaler, replace Impaler with Rogue and the situation is the same.

My assumptions specifically for units like rogue and impaler were that if the unit is standing still then all shots hit, and otherwise they all miss. For impaler that's an absolute truth, and although for rogue it's much less so my argument about target leading still applies.

Actually it doesn't work that well for impaler in practice, but mostly because of issues with starburst launcher type weapons vs radar dots. It would probably be better to have them spread out their shots vs radar dots to amortise wobble misses and the cost of one-shot kills vs unidentified dots, but I haven't figured out how to go about that sensibly so radar dots currently have a lower prio than moving units do.

Also it would probably be a good idea to have radar wobble penalty be modified or eliminated depending on the amount of radar wobble reduction that the player has, but I don't know how to check that.

mtroyka:
I'm closing this for now, it needs more work.

PR #1549, opened 2016-10-08
Unit AI: Add velocity based target priorities

mtroyka:
…for units which have slow-moving projectiles.

-Got rid of the 'too fast to hit' category, but moved those units into 'cheap' so that antiheavies will still avoid them.
-Radar wobble penalty now respects wobble reduction.
-Speed-based priority penalties also apply to radar dots, using conservative speed information when possible.

### DIFF
--- a/LuaRules/Gadgets/unit_target_priority.lua
+++ b/LuaRules/Gadgets/unit_target_priority.lua
@@ -31,7 +31,7 @@ local spGetAllUnits = Spring.GetAllUnits
 local spGetUnitRulesParam = Spring.GetUnitRulesParam
 local spGetUnitSeparation = Spring.GetUnitSeparation
 
-local targetTable, captureWeaponDefs, gravityWeaponDefs, proximityWeaponDefs, radarWobblePenalty, transportMult = 
+local targetTable, captureWeaponDefs, gravityWeaponDefs, proximityWeaponDefs, lowVelWeaponDefs, radarWobblePenalty, transportMult = 
 	include("LuaRules/Configs/target_priority_defs.lua")
 
 -- Low return number = more worthwhile target
@@ -57,6 +57,9 @@ local remVisible = {}
 
 -- Remebered mass of the target, negative if it is immune to impulse (nanoframes)
 local remScaledMass = {}
+
+-- The number of radar wobble reductions that apply to each ally team.
+local remHalfWobble = {}
 
 --// Fairly unchanging values
 local remAllyTeam = {}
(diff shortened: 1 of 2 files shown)


LuaRules/Gadgets/unit_target_priority.lua
```diff
@@ -240,14 +290,35 @@ end
...
+	
+	local _,_,nanoframe = spGetUnitIsStunned(unitID) -- determine if it's still under construction or if it was instantly created
+	if not nanoframe then
+		if UnitDefs[unitDefID].targfac then
+			remHalfWobble[allyTeam] = (remHalfWobble[allyTeam] or 0) + 1
+		end
+	end
+end
+
+function gadget:UnitFinished(unitID, unitDefID)
```
sprunk: This will count multiple times if someone partially reclaims a radar and then brings it back. You'll need to subtract when reclaim happens.

LuaRules/Gadgets/unit_target_priority.lua
```diff
@@ -240,14 +290,35 @@ end
...
+	end
 end
 
 function gadget:UnitDestroyed(unitID, unitDefID)
 	remUnitDefID[unitID] = nil
-	remAllyTeam[unitID] = nil
 	remStatic[unitID] = nil
+	
+	local allyTeam = remAllyTeam[unitID]
+	if UnitDefs[unitDefID].targfac then
```
sprunk: This does not check for nanoframe so killing a half-built radar will make it go negative.

LuaRules/Gadgets/unit_target_priority.lua
```diff
@@ -240,14 +290,35 @@ end
...
+		remHalfWobble[allyTeam] = remHalfWobble[allyTeam] - 1
+	end
+	
+	remAllyTeam[unitID] = nil
 end
 
 function gadget:UnitGiven(unitID, unitDefID, teamID, oldTeamID)
```
sprunk: The patch does not consider transfers at all. Transfer can happen between enemy allyTeams in the form of Domi. In the future it might also happen between ceasefired FFA teams.

mtroyka:
Urg. Now I remember why using callins for that was a crappy solution. Unfortunately since spring's events are not deferred there's also no guarantee that the order they're called in will be consistent, either, so probably the best way to deal with that is to keep a list of all extant adv radars and check for completeness/ownership periodically.

GoogleFrog:
Check whether this properly detects MoveCtrl as moving. We wouldn't want jumpjets and aircraft landing on pads to cause many weapons to preferentially target that unit.

GoogleFrog:
I think the speed cutoff needs a linear component, otherwise units may change targets too often.

LuaRules/Gadgets/unit_target_priority.lua
```diff
@@ -129,24 +145,51 @@ function gadget:AllowWeaponTarget(unitID, targetID, attackerWeaponNum, attackerW
...
-		remStunnedOrOverkill[targetID] = ((stunnedOrInbuild or overkill or disarmExpected) and 1) or 0
+		remStunnedOrOverkill[targetID] = ((overkill or disarmed or disarmExpected) and 1) or 0
 	end
 
 	if remStunnedOrOverkill[targetID] == 1 then
 		defPrio = defPrio + 20
 	end
 	
+	-- de-prioritize nanoframes relative to how close they are to being finished.
+	local local _,_,nanoframe = spGetUnitIsStunned(unitID)
```
GoogleFrog: I am unhappy that this gadget was written without testing.

PR #1558, opened 2016-10-13
Adv Radar, Sonar weight reductions. Update adv radar desc/tooltip.

mtroyka:
Adv Radar:
Cost 500->250
Based on pure coverage area adv radar is only worth ~200 metal. 250 makes it a bit worse than basic radar both in terms of cost per area covered (using pi*r^2 for calculation) and cost per hp, which is fair for being able to see further into the enemy's base.
-Updated description and tooltip to reflect the removal of half-wobble. (needs translations)

Sonar Station:
Cost 450->125
HP 750->250
-Reduce sonar weight to make water more tolerable to play.

EDIT: adjusted costs based on a 55 metal basic radar rather than 75 (which is for field radar, derp).


Anarchid:
Being able to see into enemy base usually requires a Swift, a Vulture or a cloaked scout.

Cloaked scouts require energy, and Swift/Vulture are liable to be expended.

How about keeping the same halved metal cost, but making the E drain more noticeable?

GoogleFrog:
Closing, as discussed.

Adv and basic radar have their costs set based on the gameplay we wish to create, not their relative usefulness. It is probably the case that basic radar is more powerful than 55. It is cheap to allow for early raider game spotting and counterplay. Adv radar should be compared to other forms of base penetration instead of compared to basic radar.

I don't want small random walks around sea balance without a better overall design.

PR #1559, opened 2016-10-14
WIP: replace annoying laser meltage groundflash with heatray-like airflash

Anarchid:
In current stable ZK, a Lotus or a Bandit shooting at a solar will generate a flat ground flash under the projectile's impact point (shifted a significant distance vertically!).

This is obsolete and unnecessary because we now have actual flashes from dynamic lighting (even though they could extend to explosions as well as to projectiles).

Also i hate groundflashes.

The meltage effect, however, is pretty, especially when it doesn't snap  to ground. This attempts to reproduce the effect using heatcloud particles. This extends to all or nearly all lasers and lasercannons in ZK.

This is a WIP thing. Stuff left to do:
- Color variations instead of uniform orange
- Find unaffected lasers, and affect them.

### DIFF
--- a/effects/beam.lua
+++ b/effects/beam.lua
@@ -472,16 +472,16 @@ local cegs = {
 
   ["beamweapon_hit_red"] = {
     usedefaultexplosions = false,
-    groundflash = {
-      circlealpha        = 1,
-      circlegrowth       = 0.125,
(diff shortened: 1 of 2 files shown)


Anarchid:
This is going nowhere.

PR #1560, opened 2016-10-15
don't access model.height/radius on startup

ashdnazg:
prevents lots of models from loading and saves precious time


abma:
backlink: https://springrts.com/mantis/view.php?id=5370

PR #1561, opened 2016-10-17
Comshare mode

ArchShaman:
- Includes entry points for invites/invite handling and team mixing logic.
- Cheats allows one to do "invalid" things (like merging with AIs)

### DIFF
--- a/LuaRules/Gadgets/game_share_mode.lua
+++ b/LuaRules/Gadgets/game_share_mode.lua
@@ -57,6 +57,10 @@ local function GetTeamID(playerid)
 	return select(4,Spring.GetPlayerInfo(playerid))
 end
 
+local function GetTeamLeader(teamid)
+	return select(2,Spring.GetTeamInfo(teamid))
+end
+
 local function IsTeamLeader(playerid)
 	local teamid = GetTeamID(playerid)
 	local teamleaderid = select(2,Spring.GetTeamInfo(teamid))
@@ -117,7 +121,7 @@ if (gadgetHandler:IsSyncedCode()) then
 
 	local function UnmergePlayer(player) -- Takes playerid, not teamid!!!
 		local name,_ = Spring.GetPlayerInfo(player)
-		if originalplayers[player] then
+		if originalplayers[player] and config.unmerging then
 			Spring.Echo("game_message: Unmerging player " .. name)
 			if originalplayers[player] then
 				GG.ResetIncome(originalplayers[player]) -- Reset team income/storage.
@@ -185,46 +189,42 @@ if (gadgetHandler:IsSyncedCode()) then
 		playerlist,playerlist2 = nil
 	end
 	
-	local function InvitePlayer(player,target,ismergereq)
+	local function SendInvite(player,target,targetid) -- targetplayer is which player is the merger
 		if Spring.GetGameFrame() > config.mintime then
-			Spring.Echo("config.mintime passes")
-			if player ~= target then
-				local _,_,targetspec,_ = Spring.GetPlayerInfo(target)
-				if targetspec then SendToUnsync("errors",player,"You can't merge with specs!") return end
-				local targetteam = GetTeamID(target)
-				local _,_,dead,ai,_ = Spring.GetTeamInfo(targetteam)
-				local _,_,spec,_ = Spring.GetPlayerInfo(player)
-				if spec then
-					SendToUnsynced("errors",player,"You are spectating!")
-					return
-				end
-				if not dead and not ai then
-					Spring.Echo("alive passed")
-					if Invites[player] then
-						Invites[player][target] = {};Invites[player][target]["timeleft"] = 45;Invites[player][target]["ismergereq"] = ismergereq
-					else
-						Invites[player] = {};Invites[player][target] = {};Invites[player][target]["timeleft"] = 45;Invites[player][target]["ismergereq"] = ismergereq
-					end
-					SendToUnsynced("newinvite",target,player,ismergereq)
-					Spring.Echo("newinvite sent to " .. target .. "(" .. player,tostring(ismergereq) .. ")")
-				elseif playerspec then
-					SendToUnsynced("errors",player,"Target squad is dead!")
-				else
-					SendToUnsynced("errors",player,"Target is an AI player!")
-				end
-			else
-				SendToUnsynced("errors",player,"You can't merge with yourself!")
+			local targetspec = select(3,Spring.GetPlayerInfo(target))
+			local _,_,dead,ai,_ = Spring.GetTeamInfo(GetTeamID(targetid))
+			if player == target then
+				SendToUnsync("errors",player,"You can't merge with yourself!")
+				return
 			end
-		else
-			SendToUnsynced("errors",player,"You can't merge this early!")
+			if IsTeamLeader(player) and targetid ~= player then
+				SendToUnsync("errors",player,"You can't invite players when you aren't the team leader!")
+				return
+			end
+			if targetspec then
+				SendToUnsync("errors",player,"You can't merge with specs!")
+				return
+			end
+			if targetid == player then
+				local teamid = GetTeamID(target)
+				target = GetTeamLeader(teamid)
+			end
+			if not dead and not ai then
+				if Invites[player] == nil then
+					Invites[player] = {}
+				end
+				Invites[player][target] = {timeleft = 45,controller = targetid}
+				SendToUnsync("addinvite",player,target,targetid)
+			end
 		end
 	end
 	
+	
 	local function AcceptInvite(player,target)
 		Spring.Echo("verifying invite")
-		if Invites[player][target] then
+		if Invites[player][target] and Invites[target][player] then
 			Spring.Echo("invite verified")
-			if Invites[player][target]["ismergereq"] then -- player->target
+			if Invites[player][target]["controller"] ~= player then
 				MergePlayer(target,GetTeamID(player))
 			else -- target->player
 				MergeTeams(GetTeamID(target),GetTeamID(player))
@@ -295,23 +295,12 @@ if (gadgetHandler:IsSyncedCode()) then
 			if proccmd[2] and string.find(proccmd[2],"invite") then
 				if proccmd[3] then
 					proccmd[3] = string.gsub(proccmd[3],"%D","")
-					if proccmd[3] ~= "" then
-						InvitePlayer(playerid,tonumber(proccmd[3]),false) -- playerid <- proccmd[3] merger
-						Spring.Echo("inviteplayer")
+					if proccmd[3] ~= "" and proccmd[4] then
+						SendInvite(playerid,tonumber(proccmd[3]),tonumber(proccmd[4])) -- #4 should be the controller id.
 						return
 					end
 				else
 					SendToUnsynced("errors",playerid,"Error: Invalid invite!")
-					return
-				end
-			elseif proccmd[2] and string.find(proccmd[2],"mergereq") then -- playerid -> proccmd[3] merger
-				proccmd[3] = string.gsub(proccmd[3],"%D","")
-				if proccmd[3] ~= "" then
-					Spring.Echo("mergereq")
-					InvitePlayer(tonumber(proccmd[3]),playerid,true)
-					return
-				else
-					SendToUnsynced("errors",playerid,"Error: Missing mergeid")
 					return
 				end
 			elseif proccmd[2] and string.find(proccmd[2],"accept") then
@@ -329,12 +318,6 @@ if (gadgetHandler:IsSyncedCode()) then
 				else
 					SendToUnsynced("errors",playerid,"You aren't on any squad!")
 					return
-				end
-			elseif proccmd[2] and string.find(proccmd[2],"restart") then -- Lua was restart! Sender requests invites again.
-				if Invites[playerid] then
-					for k,v in pairs(Invites[playerid]) do
-						SendToUnsynced("widgetstuff",playerid,k,v["timeleft"],v["ismergereq"])
-					end
 				end
 			elseif proccmd[2] and string.find(proccmd[2],"decline") then
 				if proccmd[3] then
@@ -376,6 +359,7 @@ if (gadgetHandler:IsSyncedCode()) then
 	end
 	
 else -- unsynced stuff
+	local unsyncedinvitetable = {}
 	
 	local function Errors(_,playerid,msg)
 		if Spring.GetMyPlayerID() == playerid then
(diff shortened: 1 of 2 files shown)


ArchShaman:
TO DO:
- Figure out which widget should be modded to support the invite/merge process.
- Fix resource allocation
- Modoptions
- Support for pre-made parties (a la chobby/etc?)

mtroyka:
>Automatically merges similar AI's (Not tested with CSI [Likely unneeded]/Circuit AI)
No. CSI and Circuit both have internal methods of merging. If you just mash AIs together they may end up giving conflicting/redundant orders to the same units, which wastes tons of processing time and bandwidth for no reason and could also cause severely broken behavior.

>Cheats allows one to do "invalid" things (like merging with enemy team or AIs)

Well merging with enemy AIs is "invalid", but the AI interface has a specific interface for dealing with the actions of merged players. I'm not aware of any AI that actually makes use of that, though. Mostly the AIs will just trample over whatever orders the user gives its units.

GoogleFrog:
> Modoptions

Make a modoption to control how smart/allowed this behaviour is. Here are some settings
 * Disable merge - Merging is disabled.
 * Dynamic merge (default) - Merging has a request/refuse interface.
 * Force merge - All allyTeams are merged from the start of the game. No breaking the merge (effectively commshare mode).

Disable merge might be useful for certain situations. Force merge could be something to apply for clanwars or tournaments for convenience.

> Needs testers (I can't test this on my own)

You can run chobby twice in windowed-game mode with two or more accounts. Start a test room and set game to zk:dev to run your local $VERSION as the multiplayer game.

> Automatically merges similar AI's (Not tested with CSI [Likely unneeded]/Circuit AI)

Don't do this. AIs manage themselves. AIs on a team should not even be merged with the force merge modoption.

> Figure out which widget should be modded to support the invite/merge process.

I think it is something to put on the playerlist. See the ceasefire checkboxes. Alternatively, we could add a Sharing menu to the main game menu and put merging there. This sharing menu could also allow for resource and unit sharing because the playerlist buttons can be hard to understand. This sharing menu would be a new widget which is merely shown via a button on the main menu as I don't want to touch epic menu.

For now just create a window which allows for merge but also for unit and resource sharing. Later we can implement showing/hiding via epic menu and hotkeys.

> Fix resource allocation

Tell me what sort of GG function you want and I'll put it in overdrive. Merging with enemy teams makes no sense in overdrive and shouldn't be attempted to be supported.

ArchShaman:
>Tell me what sort of GG function you want and I'll put it in overdrive. Merging with enemy teams makes no sense in overdrive and shouldn't be attempted to be supported.

Okay I'll update the logic to always disallow enemy merging. 

As far as OD function I'd like merged players to have a greater-than-normal share to encourage the use of the feature. Perhaps the merged team should get 1 + (1/2 * mergedplayers) shares of the team's income? Such that a merged team of 3 gets 2/6 of the resources of a 6 player team? Maybe allow a modoption to set this with a default of either 1/2 share or 1? As far as I'm concerned, this is a "your-call" area, as you're the main balancer.

> No. CSI and Circuit both have internal methods of merging. If you just mash AIs together they may end up giving conflicting/redundant orders to the same units, which wastes tons of processing time and bandwidth for no reason and could also cause severely broken behavior.

It doesn't set the team id as the same for all of them, it simply gives all units belonging to all ai players that are the same type to the lowest id.

Anarchid:
>  it simply gives all units belonging to all ai players that are the same type to the lowest id.
The AI's should do this themselves. Maybe the author wants them to play team one by one, or the AI wants to be run on the most powerful machine available.

GoogleFrog:
So what GG functions do you want? You should define some dummy GG functions in the commshare gadget and describe their arguments, return values and intended purpose.

ArchShaman:
GG.ReassignIncome(teamid,targetteam)
- Tells OD gadget to funnel income from teamid to targetteam.
- Have modoption to decide how much of the income to give up to the targetteam (remaining gets divided evenly amongst non-empty teams) values: 0.0 - 1.0.

GG.ResetIncome(teamid)
- Tells OD gadget to give back teamid's share.

Would these work?

ArchShaman:
sharemodemergedincome is the modoption for how much per cent a merged player gives to the party. I have defaulted this to 0.75 (3/4). So each party gets 1 + (merged players * 0.75) shares of the team. So a party of 3 gets 1 + 1.5 = 2.5 shares and the rest of the team splits 0.5 shares. This can be lowered at will. I'm thinking that a value between 0.33 and 0.75 will incentize merging, while preventing "party is op"

There's also a min merge time (in seconds) as well, so that you can't "rush" with a huge portion of the team's income. The default should be set for this.

sprunk:
Why mess with the income at all? X people merge in a team that originally had Y people -> the merged team gets X/Y while others get 1/Y, period.

GoogleFrog:
I agree with @sprunk, why are you creating a complicated income sharing system for this? If two players are controlling a team with N allies then they should receive a 2/N share of the income. I'm thinking of implementing the income sharing as a function: Teams -> Teams which just remaps teamIDs in the final share-giving part of overdrive. Very little of the overdrive gadget would require changing.

Licho1:
2 - commshared should simply be shared as a single team, with weight 2 .. 
Meaning that all overdrive rules apply, except all incomes are weighted 2x in the division.

ArchShaman:
Alright. so mergeincome(teamid,target) and unmergeincome(teamid)? I'll remove all storage modifications (except the +500 base during merger).

GoogleFrog:
I think you should remove the storage modification as overdrive is effectively a full resource system manager at this point. Overdrive should be the one to change storage as I don't want any weird effect such as dividing by zero.

Don't make more than one modoption and keep the option simple. The first design rule with modoptions is to realize that basically nobody touches modoptions. The defaults had better be good and rampart configuration won't be used.

Also, stop thinking about game balance. Team merging is orthogonal to balance as far as I am concerned. I have been ignoring the presence of teams within allyTeams for years because a sufficiently organized group of players can approximate a single team. Duplicating factions on a single allyTeam was one of my justifications for the switch to 1faction being an empty change in terms of factional diversity. If merge removes the raiding phase then there was never such a thing as a raiding phase.

I want force merge to only ever have one team per allyTeam, from the start of the game. Any solution which involves transferring spawned commanders is hacky. Selecting your commander after the game starts needs to spawn your commander correctly, even if the team already has commanders. Making commander selection and spawning work with enforced sharing from the start is one problem that needs solving. Start point selection may require a rewrite.

Licho1:
It looks fairly complicated, experience shows that complicated stuff isnt used.
Simple  on/off modoptions. 
If it's on, let people invite and accept invitations using playerlist.
Idea is, that if it works, modoption can be always on .
If it's not always on its probably waste of work (judging from other options we have).

ArchShaman:
Perhaps the clan option can be integreted into the invite only version? Such that there's an option in a widget somewhere to autoinvite people?

DeinFreund:
Or use the infra and let clan leaders decide whether their underlings get automerged.

GoogleFrog:
I like the widget autoinvite idea. This means there should be a single modoption with these options:
 * Force merge.
 * Invite mode
 * Off

Invite mode is the default. I can imagine people wanting to set Force or Off (possibly tournaments).

GoogleFrog:
If I touch the gadget I will first replace all the double spaces with tabs. So you might want to do that now and save me the hassle.

ArchShaman:
Okay sorry. It was created before the ignore widget. I'll fix it now.

Licho1:
"Or use the infra and let clan leaders decide whether their underlings get automerged"
Hivemind !! Thats actually simple to do but it would be lol..

Join my clan underlings, i will mind control your units!

Anarchid:
Hivemind is a bad name because it's an existing widget that allows you to borrow units without asking from people who have whitelisted you.

ArchShaman:
How about anti-grief? Should I remove it or have it be some sort of widget-setting thing? It blocks self-d for nonkamikaze units in merged teams (so one doesn't just join a team and self-d everything). How do we track griefing cases for admin action? Should we just simply put a warning label that says 'use at your own risk' or add some sort of monitor that tells us who's self-ding what?

ArchShaman:
I mean for forced mode.

sprunk:
No touching gameplay.

GoogleFrog:
Forced mode is simply a modoption and it is about as abusable as some other modoptions. If someone is enticing people into a forced share host then they are easy to track and punish. Don't try to solve any social issues with this feature.

ArchShaman:
Next thing to do: Share menu?

Would it be best to remake the share menu (the engine default menu that shows up when you press h) in chili? One could simply unbind it to "disable" it right? Or what's the plan here? Any suggestions on where to give player the access to invite/accept invite? Where should the settings to have an "autoinvite this player" or "autoaccept this player" be?

GoogleFrog:
Yes, share menu. We should replace the engine share menu by unbinding the key. The design can be improved so don't faithfully remake the menu. The menu should do the things that the player list can do but with larger, harder to miss, icons. That would be:
 * Share resources.
 * Share selected units.
 * Ceasefire voting (optional at the moment as its underused).

The menu should also allow you to send share requests to other people. This would just be done with a tickbox next to their name. The UI should be entirely doable with a tickbox next to everyone on you allyTeam. I think its reasonable to say that a player joins a team if they have ticked all players on the teams boxes and the players on the team have ticked the box for that player. In this way the commsharing display picks out cliques. Any concept of team leader can be ignored from the players point of view.

Licho1:
Ceasefire is underused precisely because there is no proper gui. If it was working and clearly indicated/very visible that you are being offered ceasefire, it would be used much more.

ArchShaman:
I don't even think enemy team/player should even be visible under the new share dialogue box. After all in 90% of use cases, it's just needless clutter. For ceasefire, it should be per ally team IMO. And perhaps you could expose that the player is getting a ceasefire using game_message? If one really needs enemy teams, it should be available through a separate button below it (eg: enemy teams). After all, in the default game state, you can't do anything with enemy teams, so why have them visible in the share menu?

I thought ceasefire was underused because it's presently broken / the mechanics don't properly work?

ArchShaman:
Proposal # 1:

rough draft

Something like this would work?

GoogleFrog:
Ok I am convinced. Leave the ceasefires off the sharing panel. As far as I know ceasefires do work and are per allyTeam with voting if the allyTeam has multiple players.

I think the UI just needs to be a 5xN grid where N is the number of other players on your team. The columns of the grid are:
 * A tickbox to set commshare.
 * A button which gives the selected units to that player.
 * A button which gives that player 100 metal.
 * A button which gives that player 100 energy.
 * The name of the player.

The details of who is subordinate to who seem over complicated. If players need to know who has some sort of power then they are not in a situation where they should be commsharing. Simply have a single checkbox for each player which signifies "I want to share with this player". If a group of players all have each other ticked then put them all on the same team.

The 100 metal and energy button could be configurable with bars as shown in your mockup.

DeinFreund:
Unless you changed something during the last month, ceasefire doesn't work using deluxe player list.

Licho1:
The proposal panel looks way too complicated.
You are overcomplicating it.
We dont need extra "share" window.
Playerlist alreayd has/had buttons to share energy/metal and units.

What we need is "full" playerlist, because its hard to cram stuff into small area.
Make a full player list, which is displayed when you hold some key (TAB in traditional games).

Stuff that full player list might have includes
- economy details (energy/metal store)
- unit details (metal in units/structures/defs)
- button to localize player (his camera/mouse)
- button to send metal/energy/units
- button to mute/unmute player
- button to invite/accept ceasefire
- button to invite/accept commshare
- button to add to friends

Standard (non-full-screen non-interactive playerlist could be hidden or really simple)

ArchShaman:
Tab has a problem: It's assigned to strategic map / full zoom out. With the share window proposal, you get rid of the fugly engine share menu and get a better looking chili-based derivative which we can then make a better design of. Mind you, this was the first design draft, not the final proposal.

See the problem with the current share menu is:
* Enemies (which by default setting, you cannot interact with so this is a noob trap)
* Weird slider things
* Slow as f... to use. (This is why adv player list has a button to share units)
* You have to use it in order to share m/e, which then you bring up this fugly menu that looks out of place in our chili interface.
* A tickbox (lol) to share units.

Moving the commshare stuff to the new share dialogue box has an additional benefit: we don't force players to use yet-another-playerlist if they want to use the commshare feature.

An idea has approached: Why not keep both propositions? Make the share menu better like this:
rough draft

Then move the "social" functions to the playerlist?

Licho1:
Current playerlist is scheduled for replacement. Share dialog is scheduled for removal.
Imo its best to combine both as its basically same thing - whenever you want to share, or ceasefire, or commshare, its basically on playerlist..

You need to locate player and click him. It's natural extension.
Tab is the map now, it can be changed or playerlist binding can be changed, of course we wil lmake sure it does not collide..

Having some weird storage bars in sharing makes no sense.
You should be able to send stuff even if you dont have it too!
For example send 500m means you send your storage + 500m from your income over time. 

Having bars in share screen and having to spend only this storage is some weird old BAism that makes absolutely no sense.

There is even a consideration to remove energy storage completely from the game..

ArchShaman:
> The details of who is subordinate to who seem over complicated. If players need to know who has some sort of power then they are not in a situation where they should be commsharing. Simply have a single checkbox for each player which signifies "I want to share with this player". If a group of players all have each other ticked then put them all on the same team.

It's not about who has some sort of power, it's who you can merge with. These people would simply appear tabbed below the leader. This way you know it's a team consisting of DeinFreund, Licho and GoogleFrog instead of 3 seperate teams of DeinFreund, Licho, and GoogleFrog. For instance if I am a newbie and I think to myself if I click the unitshare button next to GoogleFrog, I should expect that the units go to GoogleFrog. To my surprise when I click it it says "I gave 3 units to Licho" -- wtf?! How am I supposed to know why my units went to Licho instead of GoogleFrog? The proposal would be to remove the boxes next to the players who are already merged. This way you know you can't interact with them.

> Having bars in share screen and having to spend only this storage is some weird old BAism that makes absolutely no sense.
In order to "send" resources that you don't have from income would require changes to overdrive/income scheme. For instance, how would you know how much of your income is still going to GoogleFrog? What if I send 600m I don't have to both GoogleFrog and Licho? How does that work?

How would you propose an input system to work as easily/quickly as storage bars without being storage bars?

> There is even a consideration to remove energy storage completely from the game..
Doesn't work with OD. Setting estore to 0 REALLY breaks stuff. Big time. Inability to build stuff, -100000m/s income display, etc.

DeinFreund:
Having a player list like it's common in other games would make sense: http://i.imgur.com/YDEGV4n.jpg
(A nearly full screen overlay that is opened by holding a key)

Then you can add one button to share units, and one for metal and energy. For metal and energy I'd recommend displaying a little vertical "storage bar": the further you move your mouse up, the more resources you share, by releasing the mouse button you confirm your selection. It's just click, drag up, release, so it should be really easy and quick to do.

I would prefer not having any actions that require more than one click in that overlay, to close it you just release the tab/whatever key. (No extra buttons to close needed)


Maybe along these lines? (Space left for things i forgot)

sprunk:
> For instance if I am a newbie and I think to myself if I click the unitshare button next to GoogleFrog, I should expect that the units go to GoogleFrog. To my surprise when I click it it says "I gave 3 units to Licho" -- wtf?! How am I supposed to know why my units went to Licho instead of GoogleFrog?

The widget saying "I gave units to X" needs to be modified to support comshare.

GoogleFrog:
> Having some weird storage bars in sharing makes no sense.
You should be able to send stuff even if you dont have it too!
For example send 500m means you send your storage + 500m from your income over time. 

Too much feature creep. Sharing is barely used as is because overdrive does basically all that is required. There is no need to have sharing over time.

> How would you propose an input system to work as easily/quickly as storage bars without being storage bars?

The system I proposed earlier is much simpler. Click the button to share up to 100 metal or energy. Click it a few times and all your resources will be shared. Really simple.

I agree with this:

https://camo.githubusercontent.com/aa3b4a7c889c2c76c02dddb3911ce5b4f644bcf2/68747470733a2f2f692e696d6775722e636f6d2f5a6b4c6d4158742e706e67

Put the commshare request on what is effectively a larger and more detailed Delux Player List which is toggled with some button. This is only a common feature for FPS games but we should be able to do it here. I think tab for overview is a useless hotkey because I have a mouse wheel, however, we should ask around to see who likes it. I also think tab is a common hotkey for strategic zoom games so we may need to keep it to avoid angering people.

Don't take any of the code from any of the existing player lists. Both Deluxe Player List and Crude Player List are very badly structured. Also I think that tab player list is too big for this PR. For this PR you should merely add a commshare tickbox to the current two playerlists. They have code for provisional tickbox addition so it should be simple.

GoogleFrog:
>> There is even a consideration to remove energy storage completely from the game..

> Doesn't work with OD. Setting estore to 0 REALLY breaks stuff. Big time. Inability to build stuff, -100000m/s income display, etc.

You're being annoying here (and with the concern about the unit sharing message). Obviously Licho meant coding a system to workaround those problems with naively setting a number to 0. The implementation of other gadgets is not set in stone.

ArchShaman:
Imperfect information exists.
Leaving it at this.

GoogleFrog:
Is all the remaining required work in luaUI? If so I would like to pull this and write the widgets to use the interface later.

sprunk:
This does not touch overdrive so presumably not.

GoogleFrog:
Hmm you're right. I still want some empty GG functions added so that I know exactly what overdrive is meant to be doing. Are GG.ResetIncome and GG.RedirectPlayerIncome the functions? I assumed they were defined elsewhere because otherwise the current form of the gadget would crash.

Issue #1551, opened 2016-10-10
Reimplement isTargetingUpgrade


GoogleFrog:
The isTargetingUpgrade seems to be broken. My guess is that nobody cares about it and it would be removed from the engine if anyone noticed that it was broken and cared. Brokenness:
 * Stunning or reverse building TargetingUpgrade does not remove the bonus.
 * A single TargetingUpgrade structure removes almost all radar wobble.
 * TargetingUpgrade probably fails to respond to energy stalls.

Reimplementation can be done with Spring.SetRadarErrorParams.

GoogleFrog:
Another thing to consider is whether this even adds anything to the game or is consistent. The only other global behaviour in ZK is related to the economy and the economy is consistently global. The targeting upgrade is akin to a tech requirement or upgrade structure.The effect of the targeting upgrade is to remove the vision requirement for precision artillery to hit units at range. The vision requirement makes for a more interesting game and the cost to remove it is rather cheap. Adv radar has a niche without it.

Radar wobble reduction might be interesting if it were limited to the area visible to the adv radar. However I don't think the engine would support this.

Nobody currently uses targeting upgrade so removing it now would not affect many systems. It may just be something waiting to cause problems.

DeinFreund:
Oh, this is a feature. Thanks for keeping us informed.

PR #1563, opened 2016-10-18
reflect #1551 (update arad description)

DeinFreund:


mtroyka:
I think you may have to change some files in LuaUI/Configs/lang as well. Also translations? I have no idea how translations work though.

PR #1568, opened 2016-10-31
Add smart bombers widget.

mtroyka:
The widget itself is very simple and has no settings. It detect what orders your bombers (raven, phoenix, tbird, wyvern) are currently following and if they're following fight or patrol orders it sets them to fire at will, and to hold fire otherwise. The idea is to make bombers easier and more intuitive to control.

The only issue it has is that bombers that have been given a fight command keep it persistently after the bomber drops its bombs, which can lead to unexpected behavior/suicide. (on the other hand that's not something this widget does, just something that it could potentially fix)

EDIT: I actually had to modify the bomber command gadget because otherwise it would override anything I tried to do with the widget. Now bombers will no longer treat fight orders as being on repeat unless the bomber is actually set to repeat orders.

### DIFF
--- a/LuaUI/Widgets/unit_smart_bombers.lua
+++ b/LuaUI/Widgets/unit_smart_bombers.lua
@@ -1,0 +1,131 @@
+-- $Id: cmd_unit_mover.lua 3171 2008-11-06 09:06:29Z det $
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--
+--  file:    cmd_unit_mover.lua
+--  brief:   Allows combat engineers to use repeat when building mobile units (use 2 or more build spots)
+--  author:  Owen Martindell
+--
+--  Copyright (C) 2007.
+--  Licensed under the terms of the GNU GPL, v2 or later.
+--
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+function widget:GetInfo()
+  return {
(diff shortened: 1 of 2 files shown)


LuaUI/Widgets/unit_smart_bombers.lua
```diff
@@ -0,0 +1,131 @@
...
+--
+--  file:    cmd_unit_mover.lua
+--  brief:   Allows combat engineers to use repeat when building mobile units (use 2 or more build spots)
+--  author:  Owen Martindell
+--
+--  Copyright (C) 2007.
+--  Licensed under the terms of the GNU GPL, v2 or later.
+--
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
```
GoogleFrog: Remove this line and above.

LuaUI/Widgets/unit_smart_bombers.lua
```diff
@@ -0,0 +1,131 @@
...
+end
+
+function widget:Initialize()
+	myID = Spring.GetMyPlayerID()
+	checkSpec()
+end
+
+function widget:GameFrame(frame)
+	if frame % 15 == 0 then
+		checkSpec()
```
GoogleFrog: checkSpec should be called in PlayerChanged (or the callin with a name similar to that). It calls every time the status (team, spec etc..) of a player changes.

PR #1579, opened 2016-11-12
Claw sound improvements

Anarchid:
Current claw: sounds almost exactly the same when fired, when hits ground, and then each time a bomblet is spawned (five per detonating mine).

PR claw: plays a *different* bomblet sound *once* per mine detonation.


GoogleFrog:
Sounds need to be mono. I can see its sort of got a spidermine vibe but it is too sharp for its spammability.

Anarchid:
Sharp as in short, or sharp as in high-freq? I think shortness is a virtue here because of spammability, so i'll try to further downregulate pitch.

GoogleFrog:
There are two relatively high frequency and high volume peaks of the sound, on in each ear. It was quite distracting. Sounds need to be mono for directionality and such a short spammy sound should be unimodal.

Anarchid:
https://github.com/ZeroK-RTS/Zero-K/commit/5ab559f01d4eb357a62cefa981b9e9411ce9f234

PR #1580, opened 2016-11-13
API-ified some widgets which should trivially not be disabled.

mtroyka:
-Chili docking, because most of chili depends on/uses it so it behaves like an api.
-The widget selector widget, because if you turn it off you can't turn it back on.
-The gadget icon drawer, because it serves gadgets which are not optional.
-Mex spot fetcher, because it only bridges synced and unsynced for other widgets and depends on the synced api. (like api_startboxes)
-Selection send, because turning it off will prevent other players/replays from receiving information that they normally would always receive, while conversely it has no effect on the person actually running it.


Licho1:
Also api-fi mex handler.

mtroyka:
are you talking about cmd_mex_placement?

Licho1:
no about handler without it mex cannot be placed at all

PR #1584, opened 2016-11-13
Modoption primarily for testing different raid defense building ranges

Sgathec:
All relevant buildings except for the Big Bertha, the superweapons, the missile  silo and the nuclear silo are affected. Additionally, the modoption only affects structures that can directly damage ground-based raids. That is, it doesn't do anything to things that can merely harm raiding. Ranges are multiplied by the modoption value but they cannot go below 1. The exact units the modoption touches are the Defender, the Lotus, the Stardust, the Gauss, the Faraday, the Stinger, the Newton, the Urchin, the Doomsday Machine, the Annihilator and the Behemoth.

### DIFF
--- a/ModOptions.lua
+++ b/ModOptions.lua
@@ -389,6 +389,18 @@ local options = {
     max    = 10000,
     step   = 0.05,
   },
+    {
+    key    = 'raidDefenseBuildingRangeMult',
+    name   = 'Range Multiplier for Raid-Damaging Buildings',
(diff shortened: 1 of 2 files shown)


sprunk:
bloat++

GoogleFrog:
Yes, it's easier to remove it in a few years than to have a conversation though.

Sgathec:
What exactly is your point? Difficulty maintaining modoptions?

PR #1585, opened 2016-11-13
Replace Central Build AI with Global Build Command

mtroyka:
GBC is an updated/maintained version of the old Central Build AI widget. It adds support for repair, reclaim, resurrect and terraform (including raise-and-build commands). It also has better options, fewer bugs and is more maintainable in general.

The current implementation includes:
-The GBC widget.
-Compatibility functions in gui_lasso_terraform and cmd_mex_placement.
-An interface via epicmenu config so that it can be enabled through Game/Worker AI (where its settings now reside) rather than through the widget menu.


GoogleFrog:
I prefer WG to Script.LuaUI. Two improvements stemming from WG.
 * WG enables you to group the functions in a table. You could have WG.GlobalBuildCommand as your table and this would make other widgets more readable.
 * The external handling check is too large, adds too many lines to widgets. With WG you could simply do "not (WG.GlobalBuildCommand and WG.GlobalBuildCommand.Function(stuff)) then <handle myself> end".

GoogleFrog:
This seems to be WIP as the UI is quite rough. However, its off by default and should get more work so it can be pulled once the external widget interface is improved.

Suggestions/bugs:
 * Don't override manually given commands, at least not partially like they are now. Instead, display the appropriate commands in the command panel and make those ones the global commands. See the initial queue widget for how to display and use commands independent of selection.
 * The widget is drawing something which breaks selection circles. The drawing seems to appear intermittently.
 * Get a better automation display than text under the unit. Use the unit_icons widget to easily do this.
 * You don't seem to be handling the persistent build height widget. It is also a source of terraform commands.
 * The global build commands could have a non-green outline to signify that they are global.

mtroyka:
>I prefer WG to Script.LuaUI. Two improvements stemming from WG.

I'm not sure if WG is a good solution, because if the widget crashes or is unloaded the entries will remain and cause errors. I think Script.LuaUI keeps track of the widget state.

>Don't override manually given commands, at least not partially like they are now. Instead, display the appropriate commands in the command panel and make those ones the global commands. See the initial queue widget for how to display and use commands independent of selection.

That would prevent the command panel from being hidden when no units are selected, which is annoying, and also would force a buggy looking panel like initial queue does. It would also require that no units be selected in order to use it, which is unintuitive.

>The widget is drawing something which breaks selection circles. The drawing seems to appear intermittently.

I need to check that.

>Get a better automation display than text under the unit. Use the unit_icons widget to easily do this.

Eh.. I don't have any icons on hand to use. I think the status would also be less clear with icons than with text. There are four states: idle, queue, move and direct order, none of which would make unambiguous icons.

>You don't seem to be handling the persistent build height widget. It is also a source of terraform commands.

I guess I need to check that as well. I couldn't get persistent build height to work at all so I couldn't test it.

>The global build commands could have a non-green outline to signify that they are global.

Maybe, but what color could I use that would not be confused with someone else's team color?

mtroyka:
Ok I confirmed the issue with unit shapes and also fixed it. (unit shapes, that is, since its drawing code was like wtf and still mostly is)

I also added compatibility for persistent build height.

I suppose I'll try some things with color and see if I can make something work.

mtroyka:
I didn't use WG but I did significantly simplify the compatibility code for the three widgets that support it. I wasn't able to get rid of the externallyHandled variable for lasso terraform, but only because the way the code is written requires it. I was able to make it more context local though. All the other widgets no longer need the extra variable and it's been removed.

Blue seems to work ok as an outline color.

>Don't override manually given commands, at least not partially like they are now. 

Actually the main reason for doing that is to allow other workers to assist on direct orders. If you don't want any intervention then you can always add the worker in question to the exclusion group so that its commands won't be handled.

I guess the main disadvantage is that you can't make short 'direct order' queues when you want a specific build order, but the cost model is supposed to minimize the need for that sort of micromanagement. The initial queue is an exception since ordering can be very important early game, which is why it's purposefully not captured.

LuaUI/Widgets/unit_global_build_command.lua
```diff
@@ -2194,6 +2194,14 @@ function GetNormalizedDirection(x1, z1, x2, z2)
 	return x, z
 end
 
+-- because lua is a shit language
```
GoogleFrog: You're doing it wrong.

GoogleFrog:
I don't trust your changes to not break unit shapes. There was a bit of work but in by Shadowfury333 to fix various edge cases involving water and reflection. The various draw worlds are probably related to it. Other widgets, such as team platters, don't break unit shapes and GBC is much lower in priority than unit shapes so I either expect a fix from the GBC side or a more convincing proof that unit shapes was broken.

> I'm not sure if WG is a good solution, because if the widget crashes or is unloaded the entries will remain and cause errors. I think Script.LuaUI keeps track of the widget state.

I'm pretty sure WG and Script.LuaUI are on the same footing here. With WG you can just set WG.GlobalBuildCommand to nil in shutdown while with Script.LuaUI I'm fairly sure you need to deregister each global.The main difference is that Script.LuaUI can be called by unsynced gadgets which isn't a thing you want to do. Using it implies in the code that you want the function to be called by gadgets. Not to mention that WG is the standard way to communicate between widgets and if you use a table the code becomes more readable. So just make the change.

mtroyka:
closing because git borked and need to refork.

PR #1588, opened 2016-11-14
Create OnyxCauldron1.7.lua

The-Yak:
startboxes for onyx cauldron 1.7


GoogleFrog:
I think you can include these directly in maps when they are released. I didn't write it though, I just expect that to be the case.

Anarchid:
That is indeed the case.

PR #1591, opened 2016-11-15
Replace Central Build AI with Global Build Command

mtroyka:
..a more maintainable and up to date version of the widget with additional features and fewer bugs.

-Add compatibility functions to cmd_mex_placement, gui_lasso_terraform, gui_persistent_build_height, and unit_initial_queue via WG.
-Add epic menu integration through epicmenu_conf.


mtroyka:
I found out that cmd_terraform_internal does in fact give commands to the unit that receives it. That basically means that cmd_terraform_internal will interrupt direct orders. Adding options such as meta seems to make this worse rather than better, and it doesn't work with cmd.insert at all.

It might be possible to circumvent the limitations by holding onto the tf command rather than plopping the terraunit immediately, but I suspect that due to the complexity of the tf command (and the built-in params that specify what units will be participating) that may not work properly.

However while it's less than ideal in general, the worst case occurs early game when a tf command might wipe out your pregame queue. Since not many people spam tf that early on it's not likely to be a major problem, and in other cases it doesn't behave any worse than lasso terraform does already.

GoogleFrog:
> I found out that cmd_terraform_internal does in fact give commands to the unit that receives it.

Then you've found a bug. The best way to report it is to make a small widget to reproduce the bug.

PR #1609, opened 2016-11-16
Added RTT option to grid view which merges the pylon circles

ashdnazg:
The code should discover whether it's supposed to run or not.
If you get reports that it doesn't work on certain GPUs, a settings toggle may be needed.


ashdnazg:
before:
oldodgraphics
after:
newodgraphics

If you want you can bump alpha a bit since it won't get too dark with the current system.

ashdnazg:
there's also this commit:
https://github.com/ashdnazg/Zero-K/commit/3c83e5c36ee4a5f777fe89ae5cad885f81bc9508
Which adds drawing the ranges of queued pylons in the current selected builder's queue if you're placing a grid building.

The two commits are mostly independent so I split it into a branch, I can add it here if you want.

PR #1612, opened 2016-11-17
Probably fix the range multiplier modoption for raid defense structur…

Sgathec:
…es by changing the camel case modoption key style to an underscore-separated one

### DIFF
--- a/ModOptions.lua
+++ b/ModOptions.lua
@@ -390,7 +390,7 @@ local options = {
     step   = 0.05,
   },
     {
-    key    = 'raidDefenseBuildingRangeMult',
+    key    = 'raid_defense_building_range_mult',
     name   = 'Range Multiplier for Raid-Damaging Buildings',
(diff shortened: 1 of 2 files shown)


Sgathec:
How does one test modoptions fully beforehand? Using spring.exe doesn't reveal all possible bugs.

GoogleFrog:
You can host dev in springie by doing !game zk:dev.

Sgathec:
Sadly that doesn't help either in this case because I think the lobby client doesn't recognize the new modoption key.

GoogleFrog:
That requires a dev version of the lobby. You can test is multiplayer by setting the key with the !command without using the lobby UI.

Sgathec:
I just did it again because I wasn't certain if I did it properly the last time. After setting the room to zk:dev, I tested using !setoptions and neither the old key nor the new one functioned. Are you sure this method works?

sprunk:
Setoptions probably fails since Springie does not actually have the mod so does not know the modoption. IMO just call it defenserangemult and be done with it.

PR #1633, opened 2016-11-22
Untested: General Commshare Fixes

### LINKED ISSUE #1592: Sharemode gadget doesn't work at all
[f=0000046] Error: LuaRules::RunCallIn: error = 2, GameFrame, [Internal Lua error: Call failure] [string "LuaRules/Gadgets/game_share_mode.lua"]:377: attempt to index global 'WG' (a nil value)
stack traceback:
	[string "LuaRules/Gadgets/game_share_mode.lua"]:377: in function 'GameFrame'
	[string "LuaRules/gadgets.lua"]:1008: in function <[string "LuaRules/gadgets.lua"]:1006>
	(tail call): ?

ArchShaman:
* Team leader is the only one who can accept or decline invites. Doesn't affect 1 player teams.
* Includes a little bit of efficiency (using i=1 instead of pairs)
* Fixes #1592 
* LUAUI exposure (Use GetTeamRulesParam("invites") and turn the string into a table)

### DIFF
--- a/LuaRules/Gadgets/game_share_mode.lua
+++ b/LuaRules/Gadgets/game_share_mode.lua
@@ -185,15 +185,15 @@ if (gadgetHandler:IsSyncedCode()) then
 			local targetspec = select(3,Spring.GetPlayerInfo(target))
 			local _,_,dead,ai,_ = Spring.GetTeamInfo(GetTeamID(targetid))
 			if player == target then
-				SendToUnsync("errors",player,"You can't merge with yourself!")
+				SendToUnsynced("errors",player,"You can't merge with yourself!")
 				return
 			end
 			if not IsTeamLeader(player) and targetid ~= player then
-				SendToUnsync("errors",player,"You can't invite players when you aren't the team leader!")
+				SendToUnsynced("errors",player,"You can't invite players when you aren't the team leader!")
 				return
 			end
 			if targetspec then
-				SendToUnsync("errors",player,"You can't merge with specs!")
+				SendToUnsynced("errors",player,"You can't merge with specs!")
 				return
 			end
 			if targetid == player then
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I'm not too worried about the widget side of things, I would merge and release it without a widget just so merge all is usable. Furthermore, any widget you write now will probably get rewritten later.

I'm more concerned with lagmonitor. That needs extending to support teams with variable player count.

ArchShaman:
I'm all for getting a rewrite. Mostly I want to include a widget so it's a complete feature instead of a partial feature.

How can I help with lagmonitor? If all it needs is extention to individual players instead of teams, wouldn't it be a good idea to keep a table like 
lagmonitorteams = {[1] = {afkcount = 0, afkplayers = {}} 
Then when someone goes afk, check afkcount against teamsize. Team is afk! :D?

only time an afk players could switch teams is when an afk player's invite is accepted. Perhaps if we add a function to lagmonitor to tell it when a player switches teams too? >.< I think that's the simplest solution, but not the best one.

GoogleFrog:
Lagmonitor detects and handles team activity. Inactive teams have their units shared and they are shared back to them when they become active. Lagmonitor tells overdrive which teams are active and overdrive then distributes them resources. Overdrive could actually know nothing about sharemode, just get quota-per-team from lagmontior. A team is active if it contains a non-afk player. 

Lagmonitor probably doesn't do what I said when you introduce sharemode. I'm saying what it should do.

ArchShaman:
WIP -> Untested.

Will change to ready after live test is done (need a big game for this).

ArchShaman:
Cleaning up PR after the https://github.com/ZeroK-RTS/Zero-K/pull/1633/commits/9b9c6f3ce06c235dfcf057d8ff9c82f8b61fd73d disaster.

PR #1640, opened 2016-11-25
[untested] Probably an even better fix for the lups error.

sprunk:
-2 means "all access" and -1 means "no access" ( https://github.com/spring/spring/blob/master/rts/System/EventClient.h#L53 )


GoogleFrog:
Might want to check around 828 too.

PR #1644, opened 2016-11-27
Add HDR capabilities to the deferred rendering shader.

mtroyka:
-Dynalights now get white shifted when any color channel exceeds 1.5, giving a stronger impression of brightness.
-Merged the old bloom shader into the deferred rendering widget, since it has to be run in HDR mode and gets blended with lighting anyway.

Settings are in Settings/Graphics/HDR. I expected it to require per-map settings for exposure but that turned out not to be the case. (and exposure was also strictly unnecessary so I omitted it as a setting)


mtroyka:
Performance in general is no different than without, but I've identified that the dynalights themselves are expensive to draw due to the potentially large number of gbuffer reads (from 2 gbuffers no less) as well as the large number of individual writes to the HDR texture.

Dynalights can end up being more expensive due to the higher bit width of the HDR texture but there's not much I can do about that. Tile-based deferred could theoretically solve that issue but it requires expensive per-tile frustum tests that I don't know if you could get spring to do and which would be expensive in lua. Apparently it might also be doable with a compute shader but I have no such capability.

Overall I'm pretty happy with the tradeoffs. It's definitely shiny++ and could be moreso with proper tuning of related effects.

PR #1649, opened 2016-11-29
Trees no longer autoreclaimed

sprunk:
Trees are only reclaimed if:
 * the center of the command is a tree, OR
 * CTRL is held and the center is not a unit.

There's a technical hax that makes trees worth 0.001 metal because else Spring disprioritises them.

Trees tend to live in wide forests or on the map edge a decoration. In both cases there's a fair amount of walking for a con with trees in its area reclaim command. Meanwhile Energy is abundant so it is wasted. This patch makes trees not reclaimed by default but still leaves ways to reclaim them if need be.

Compared to #1506 (not the recent PR but the one before that was merged and reverted) the feedback is implemented:
 * right-click and drag a tree reclaims trees,
 * trees are not deprioritised in the tree-reclaim mode
 * feedback related to force-reclaiming own units implemented separately (trees are a different issue), directly to the repo (because the existing behaviour was outright dangerous) - https://github.com/ZeroK-RTS/Zero-K/commit/379a19a14ab12a5662c20d788b06a8c8d281286d

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -1471,6 +1471,7 @@ local function GetResources(tooltip_type, unitID, ud)
 		if unitID then
 			local m, _, e, _, _ = Spring.GetFeatureResources(unitID)
 			metal = m or metal
+			if (metal < 1) then metal = 0 end
 			energy =  e or energy
 		end
 	else --tooltip_type == 'unit' or 'selunit'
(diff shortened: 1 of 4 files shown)


Licho1:
Looks good but why is the hack with 0.001 needed? Does it meant it woudlnt reclaim them at all if engine command is used (with 0m)?

sprunk:
It eats M==0 only after all M>0 are gone.

In theory that could also work as a good default but then you cannot have a way to make trees equal (which is needed for example when E-stalling).

PR #1658, opened 2016-12-03
Fixed HDR.

mtroyka:
Reinhard tonemapping was implemented incorrectly, and used color correction that it did not need. I added some notes so that nobody should get confused in the future.

Now there should be zero difference in color between HDR and LDR.


mtroyka:
I hate git. Piece of trash.

PR #1663, opened 2016-12-03
Unit AI: Fix tactical AI for arty

mtroyka:
Using skirmish AI for arty was something of an experiment, and one which did not work out particularly well. I decided to switch arty to use slasher-style flee-skirm instead of regular skirm and it seems to work much better, particularly for pillagers and penetrators which can't turn their turrets fast enough to aim while moving.

### DIFF
--- a/LuaRules/Configs/tactical_ai_defs.lua
+++ b/LuaRules/Configs/tactical_ai_defs.lua
@@ -910,19 +910,21 @@ local behaviourConfig = {
 	},
 	
 	["armmerl"] = {
-		skirms = allGround,
+		skirms = artyRangeSkirmieeArray,
 		skirmRadar = true, 
 		swarms = {}, 
 		flees = {},
-		skirmLeeway = 400, 
+		skirmLeeway = 400,
+		skirmOnlyNearEnemyRange = 350,
 	},
 	
 	["cormart"] = {
-		skirms = allGround,
+		skirms = artyRangeSkirmieeArray,
 		skirmRadar = true,
 		swarms = {}, 
 		flees = {},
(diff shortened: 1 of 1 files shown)


GoogleFrog:
skirmOnlyNearEnemyRange was only added for Slasher because nothing else would produce reasonable behaviour and the AI is off by default. It has poor side effects which I don't want to apply to other units, these artillery units should just stay at max range with a fight command. Fight means "stay at max range" for them and I don't want any complications overriding this behaviour because then there wouldn't be any way to automatically make your artillery stay at max range.

skirmOnlyNearEnemyRange means that you could put a Jack at the front of your army and easily kill artillery with longer ranged flankers when the arty fails to retreat.

mtroyka:
Flee doesn't work. I don't know why it doesn't work either.

>these artillery units should just stay at max range with a fight command. Fight means "stay at max range" for them and I don't want any complications overriding this behaviour because then there wouldn't be any way to automatically make your artillery stay at max range.

That's called "turning off unit AI". The adjusted behavior is also much more 'stay at max range' than the current behavior is.

>skirmOnlyNearEnemyRange means that you could put a Jack at the front of your army and easily kill artillery with longer ranged flankers when the arty fails to retreat.

what? Are you saying that because the jack is closer it will ignore the range of other things behind it? I'm not sure if I've observed that or not. In any case it's impossible to do worse than the current behavior, which for lots of arty either prevents them from firing properly or causes them to miss when they otherwise wouldn't have.

GoogleFrog:
Turning off unit AI does not cause your units to stay at max range.

PR #1680, opened 2016-12-05
Death explosion avoidance

sprunk:
* very low range units (Glaive and Dart) skirm windgen (chainsplosions tend to be lethal if adjacent)
 * medium range units skirm [Adv.] Geo and Pylon (to avoid death explosion: they already skirm Fusion and Singu for that reason)

### DIFF
--- a/LuaRules/Configs/tactical_ai_defs.lua
+++ b/LuaRules/Configs/tactical_ai_defs.lua
@@ -132,7 +132,6 @@ local veryShortRangeSkirmieeArray = NameToDefID({
 	"chicken_tiamat",
 	"chicken_dragon",
 	"hoverdepthcharge",
-	"armwin",
 	
 	"corgator",
 	"armflea",
@@ -219,13 +218,6 @@ local medRangeSkirmieeArray = NameToDefID({
 	"corgol",
 	"tawf114", -- banisher
 	"scorpion",
-	
-	"armfus", -- don't suicide vs fusions if possible.
-	"geo",
-	"amgeo",
-	"armestor",
-	"cafus", -- same with singu, at least to make an effort for survival.
-	"armbanth", -- banthas also have a fairly heavy but dodgeable explosion.
 	
 	"shipscout",
 	"shipassault",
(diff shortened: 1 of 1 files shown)


mtroyka:
You should probably add mexes to that, since they also explode. I did something like that for fleas but I can't remember if I added windgens or not.

sprunk:
Everything explodes, not just mexes or windgens. I chose windgens because they tend to appear in chainsploding clumps which tends to burst down the raiders.

Flea avoids every unarmed building (so in particular mex and windgen are already included).

mtroyka:
having that one hero glaive get killed randomly by a mex explosion sucks though.

GoogleFrog:
I disagree with tactical AI being used to make units keep their distance from exploding structures. You have to answer this question: how did the Flea end up in wWndgen range in the first place? I just put a Flea on fight vs a scattering of five Windgens and it defeated all of them without taking any damage. The Fight and leashing behaviour is already perfect for Flea vs structures (not considering sightlines, but tactical AI doesn't do that either). Dart does less well but it is perhaps the old unit for which this AI is reasonable. I would simply leave it out.

The main way to cause your units to die to Windgens is to tell them to walk next to the Windgen. This is easy behaviour to learn to avoid and the solution is the same whether there is tactical AI or not, attack Windgens with Fight. Therefore the main situation solved by this PR is when you move your Fleas next to a Windgen and they have time to automatically move away before it explodes.

The main downside of this PR is that your units will disobey your orders and potentially break. If someone give and order to move right next to a structure then maybe they wanted to be right next to that structure? If the structure explodes then they learn for next time. Sightline avoidance is a common requirement for raiding bases and when people start trying to do it they will run into the AI. This PR also adds some broken behaviour: imagine you have Hermits and want to kill a Geo in a hole. You tell the Hermits to go into the hole. They see the Geo and decide to walk out of the hole. Very frustrating. If someone has to touch a unit state then the UI has failed.

I think the small number of cases when this AI is useful is outweighed by the basic frustration of units disobeying orders.

GoogleFrog:
If you want this AI then I think the solution is to make it only active with Fight. Leashing should not cause units to run from explosive structures as that has too much potential to disobey move orders. Fight is acceptable because if you issue a fight command you are giving up fine control over unit positioning in favor of intelligent local behavior.

sprunk:
So, should the reverse happen? Remove fusion and singu from skirmiees?

GoogleFrog:
As I said, I think avoiding only on Fight is fine. See https://github.com/ZeroK-RTS/Zero-K/commit/9c6306f8430c30477be9710218dfae33e93f0906. I only tested it for Rogue so it all needs adding.

It's probably worth making a new PR.

PR #1686, opened 2016-12-06
EMP, Disarm and Slow now affect morphing speed and resource drain. Also partially fixes morph/transport interaction.

### LINKED ISSUE #1667: Morph is not hindered by status effects
EMP and disarm do not halt morph progress. Slow does not reduce morph BP. Ideally these things should affect morph.

Aquanim:
Fixes #1667. I have tested this but somebody else should probably check before this makes it into a stable.

### DIFF
--- a/LuaRules/Gadgets/unit_morph.lua
+++ b/LuaRules/Gadgets/unit_morph.lua
@@ -788,7 +788,18 @@ function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOpt
 	local morphData = morphUnits[unitID]
 	if (morphData) then
 		if (cmdID == morphData.def.stopCmd) or (cmdID == CMD.STOP and not morphData.def.combatMorph) or (cmdID == CMD_MORPH_STOP) then
-			if not Spring.GetUnitTransporter(unitID) then
+		
+			local morphCancelAllowedByTransport = 1
+			local transportID = Spring.GetUnitTransporter(unitID)
+			local transportUnitDefID = 0
+			if transportID then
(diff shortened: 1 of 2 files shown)


sprunk:
Looks legit. How does this work with transports (they apply a form of stun but not EMP)?

Aquanim:
That would be why I needed somebody else to check... I didn't think of all the edge cases. One moment.

sprunk:
Transports already break current morph so having them work would be more of a bonus than a requirement.

Aquanim:
Yes, from the outcomes of what I've tried so far "I'll fix that" was a bit over-optimistic, at least as far as tonight is concerned.

Aquanim:
Okay, this now _really_ needs testing to check stupid transporter edge cases.

From the commit:
Units do not complete the morph until they disembark to avoid problems if the target morph is no longer transportable.
(Commandeer morph cannot be cancelled on a Surfboard but I think that's a minor bug.)

Aquanim:
I believe that these commits now satisfy the requirements of #1667, as well as fixing some interactions with transports.

- Units morph slower and drain less resources when slowed.
- Units stop making progress on their morph when stunned or disarmed.
- Units can still cancel morph and refund resources when stunned or disarmed (this is consistent with behaviour of stunned factories).
- Combat morphs may occur normally on a fireplatform transport. Combat morphs on a non-fireplatform transport will be paused. Non-combat morphs on any transport will be cancelled. Morphs on transports can be cancelled by the user.
- A morphing unit which is fully stunned then brought aboard a fireplatform transport will continue to morph; this is a subset of #1687 and I do not think it is appropriate to address here.

PR #1701, opened 2016-12-09
Powerplant descriptions.

sprunk:
`Produces Energy (X)` -> `{Light, Medium, Heavy} Powerplant (+X)`


Aquanim:
I think I'd prefer {Small, Medium, Large} but it is a matter of taste.

sprunk:
Sounds better in hindsight, thanks.

PR #1702, opened 2016-12-09
[untested] Fix #1391 - Vote Display widget remembers position

### LINKED ISSUE #1391: Vote Display widget forgets position
Vote Display widget forgets position between games. Maybe needs to be dockable.

sprunk:


GoogleFrog:
It shouldn't be dockable.

sprunk:
I tested and it remembers position but won't dock.

PR #1705, opened 2016-12-10
Reduced Mistral flightTime (5->3) and remove weaponTimer from non-Starburst defs.

Aquanim:
Accounts for increased missile speed compared to Enforcer, and probably makes it a little worse against Swift.
Also redesignated Missile Frigate -> Missile Boat.


sprunk:
As far as I can tell weaponTimer does nothing on non-Starburst weapons. Is there any observable change?

Aquanim:
I *think* that if the weaponTimer expires the missile no longer does damage, but I can test that further by making it something extreme.

The flightTime is the important change though.

Anarchid:
The correct thing to do to both Slasher and Mistral's weaponTimer is to abolish them.

sprunk:
I believe 0-damage shots are due to impactOnly == false.

PR #1709, opened 2016-12-10
Terraform works smoothly.

sprunk:
Terraform update rate: 30 -> 3 frames

Makes terraform change the terrain very smoothly. Terraform is one of the few things where the update rate is directly seen by people. Large chunks of terrain (dis)appearing each second look very crude.

The performance cost should be fairly low. Terraform is used rarely and when it does it splits itself into smaller parts.

### DIFF
--- a/LuaRules/Gadgets/unit_terraform.lua
+++ b/LuaRules/Gadgets/unit_terraform.lua
@@ -106,7 +106,7 @@ for i = 0, 64, 8 do
 end
 
 local maxAreaSize = 2000 -- max X or Z bound of area terraform
-local updateFrequency = 30 -- how many frames to update
+local updateFrequency = 3 -- how many frames to update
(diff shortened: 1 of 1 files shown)


Licho1:
How big of a cost is this in practice? Wont it trigger repathing too often?

sprunk:
Repath only happens each slowUpdate ie. every 15 frames ( https://github.com/spring/spring/blob/723eb8f309b5ed0d1141fe1e68210df9613a3161/doc/changelog.txt#L625 ).

I see terra mostly used for low-scale stuff (eg. elevate a turret, make a small ramp for a cliff) so it should have low impact overall. If large terraform becomes popular, engine makes repathing occur more often or generally something happens that increases the cost then it of course could be tweaked further (eg use 10 frames).

PR #1710, opened 2016-12-10
Soft-remove Surfboard.

### LINKED ISSUE #1695: Surfboard
Surfboard has technical problems:
 * misbehaves with morph (see #1686)
 * misbehaves with EMP (see #1687)
 * units are teleported in front of it for loading and then the (un)loading process is just a slide (which allows units to slide through walls). This is bad but the model does not allow anything better.
 * the model is ugly and somewhat silly (pretty much an actual surfboard).
 * nubs might think it's a regular transport.

Ordinarily this would mean "those need fixing" but I'm unsure if Surfboard is well-designed from a gameplay POV either. It does have a role (anything a specialist land unit does but is unavailable in sea, Surfboard can provide) but I don't think the implementation is good.
 * control and management. The surfboard and its passenger are separate units of existing types. That means the interface for such a pair is subpar compared to regular units. There's an overhead in keeping the units paired, including getting them together physically, but also eg. keeping track of how many you have so there's a ~ 1:1 ratio. You can't for example ctrl+z (because if you ctrl+z a Surfboard you also select those with different passenger types, and if you don't you can't move), retreat won't work (a Surfboard might lose its passenger but won't go back because it is still healthy itself), unit counts are skewed (the UI says you have X units but half of them are sunken passengers whose surfboards died and are actually marooned underwater) etc.
 * balance. Passengers are tuned for the land game so the only controllable variable for a pair is the Surfboard, but that affects all other pairings as well. In practice this means there's a small handful of units that Surfboard works decently with and the other pairings are owled. Having regular units fulfil Surfboard's roles would mean they can be tweaked independently without the risk of owling something else.

sprunk:
Fixes #1695

### DIFF
--- a/LuaUI/Configs/integral_menu_commands.lua
+++ b/LuaUI/Configs/integral_menu_commands.lua
@@ -160,7 +160,6 @@ local units = {
 		shipassault = ASSAULT,
 		-- No Ship HEAVY_SOMETHING (yet)
 		subraider = SPECIAL,
-		shiptransport = UTILITY,
 	},
 
 
--- a/units/factoryship.lua
+++ b/units/factoryship.lua
@@ -18,7 +18,6 @@ unitDef = {
 	[[shipassault]],
     [[shiparty]],
     [[shipaa]],
-	[[shiptransport]],
   },
 
   buildPic               = [[FACTORYSHIP.png]],


GoogleFrog:
This is too soft. Who knows whether facpanel or some other random widget will be able to make surfboard. Someone might even use Spring command menu. It will be highly confusing for most players to be unable to build surfboard while some are able to. This could even result in heresay support which has people 'fix' the issue by enabling unsupported widgets.

Licho1:
Well generally removing it all in single commit is preferable "soft" remove because its easy to undo.

PR #1714, opened 2016-12-12
Fix #1713

### LINKED ISSUE #1713: Ghost Site: LoS hax
Regular buildings' ghost site persists if they die outside LoS, you need to see the location to learn they died.

Protector, Silencer and Gauss lose the ghost site immediately if they die, regardless of LoS.

sprunk:
Not pushing directly because it seems that GF explicitly did this ( https://github.com/ZeroK-RTS/Zero-K/commit/013492f213e93a30b745fe13f19074b98470269a ) so perhaps there was some other purpose for this.


GoogleFrog:
My guess is that I thought yardmap is a useless tag in that case and that Anarchid changed the model without fixing the yardmap.

GoogleFrog:
The standard for strings in the unitdefs is [[]].

PR #1715, opened 2016-12-12
Siren Nerf

ParzivalX:
Nerf to make siren a little less op. Note the changes besides damage to accomadate changes smoothly.


units/shipassault.lua
```diff
@@ -139,28 +139,28 @@ unitDef = {
...
+      areaOfEffect            = 30,
       cegTag                  = [[missiletrailyellow]],
 	  collideFriendly         = false,
       craterBoost             = 1,
       craterMult              = 2,
 
       damage                  = {
-        default = 400,
-        subs    = 400,
+        default = 360,
```
sprunk: Reflect in OKP.

units/shipassault.lua
```diff
@@ -139,28 +139,28 @@ unitDef = {
...
+        subs    = 360,
       },
 
       edgeEffectiveness       = 0.5,
       fireStarter             = 100,
 	  fixedLauncher			  = true,	  
-      flightTime              = 4,
+      flightTime              = 5,
```
sprunk: This effectively does nothing. Impart rationale.

GoogleFrog:
This is not how I receive balance changes. Balance changes should generally be discussed to the point of obviousness and a clear intention should be formed. The PR is then a specific implementation of an already fairly specific intention and can also include discussion. Balance PRs which just make changes would result in random walks around design space.

PR #1722, opened 2016-12-14
Pylon quadfield tweak

sprunk:
It would be good if @ashdnazg could look since he made the original code.
The comment implies the value should be 2x what it currently is.

However the max value is only used by the actual Pylon (at 500 elmo radius). All other structures have no more than 150 (so the "always <= 4 quads" already holds for them). They are much more common than Pylons (Solars and Winds especially) so perhaps doubling the current value would overall only do harm since it only benefits the rare Pylons. Maybe it would make more sense to set it to 2x the common max and let Pylon to be in >4 quads.

For comparison, the relevant size values:
 * 2x max except Pylon: 300
 * current value: 500
 * 2x max: 1000


ashdnazg:
I think this is why I didn't really make it 1000 originally
I agree that there are usually very few pylons, so a smaller value may have its merits

PR #1735, opened 2016-12-16
Global Build Command Update

### LINKED ISSUE #1630: GBC problem: floating structure blueprints are blue in Spring
The default structure drawing is blue for floating structures. This conflicts with GBCs blue outlines.

mtroyka:
-Added an option for setting the worker update rate, so you can choose between cpu/network usage and assignment efficiency.
-Added an option for auto-retreating units that are auto-repairing/reclaiming. Does not apply to units that are repairing porc.

-Fixed an issue where repair tasks would cause pathologically bad performance.
-Improved stuckness prevention for units being cleared off of active tasks.
-Changed the GBC outline color to orange. (Fix #1630)

-Renamed some globals to be less generic, and changed command id numbers to use the human-readable CMD.x format instead.

### DIFF
--- a/LuaUI/Widgets/unit_global_build_command.lua
+++ b/LuaUI/Widgets/unit_global_build_command.lua
@@ -35,7 +35,7 @@ function widget:GetInfo()
   }
 end
 
---  Global Build Command creates and manages a global, persistent build queue for all 
+--  Global Build Command creates and manages a global, persistent build queue for all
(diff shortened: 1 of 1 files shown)


GoogleFrog:
GBC has entered sort of the same space in the repo structure as COFC. I'll merge PRs for them fairly freely and then test every few months to see if they work.

Issue #1596, opened 2016-11-16
Rename the unitdefs


GoogleFrog:
The current unitdefs could be renamed to be more consistent. Two main advantages:
 * More convenient developing.
 * Whitewashing of mod-status history.

The already agreed upon method of naming units is ChassisRole and many of the more recently added units conform to it. Renaming was suggested in the past but could not be done because systems such as modstats and missions would break. At the moment modstats doesn't work and we may have relatively few working missions so now would be a time to rename the units. Places to rename:
 * ZK lua (easy find and replace).
 * Third party AI configs (just tell them rename is occurring)
 * Missions.
 * Anything else?

User effect:
 * Reset of everyones persistent build spacing and persistent build height settings. Minor.
 * If anyone has directly bound the construction of a unit to a key then they will need to update. No units are bound to keys by default.
 * Some obscure Ctrl+Key selection keys would break, for example the one that selects Vultures.
 * People who have custom widgets may need to change them if they hardcode the units they handle.

It looks like a bit of work but, if it should be done, it should be done sooner rather than later.

Licho1:
Also stuff in infrastructure, unit icons, avatars, unitguide and some
currently unused things - modstats, unlocks.

2016-11-16 4:39 GMT+01:00 GoogleFrog <notifications@github.com>:

mtroyka:
On one hand that's tempting, since zk units are often labeled 'core' and 'arm' when we no longer have those factions and a lot of units have little or no relationship to the historical names they carry. It certainly would make things easier on new devs. On the other hand it might break a lot of random things and the damage is difficult to predict.

Also it might be a good idea to reorganize unitdef params, since the default alphabetical order with no spacing/groups is about as good as random if you're trying to edit them.

Aquanim:
Fixing stuff that is broken by this will mean more work in the short term, but in the long run I think this will save a worthwhile amount of time and effort.

sprunk:
There seems to be agreement to do this and changes in this direction have been made (in the recent sea refactor) so any externals are already bork.

Anarchid:
What should be the scheme for economy, defense, intel, and superweapon buildings? 

While turrets and eco are reasonably straightforward (`turretriot`,`turretcancer`, `energysolar`,`energywind`,`energyfusion`), superweapons and striders are more complicated and tend to have proper names in their defnames (`reef`, `dante`, `funnelweb`, `zenith`). 

So for those i see three options:
[...]

PR #1736, opened 2016-12-17
Hard remove Razorback, old Reef, old Starlight

KingRaptor:
Also fixes Starlight referencing in lang files

See discussion on b0bbcbcfe4dac927a0369440947714955331d177


Anarchid:
New!Starlight  is actually](https://github.com/ZeroK-RTS/Zero-K/blob/master/units/starlight.lua#L350 ) `mahlazer` for mission compatibility purposes. IIRC those are now invalid, so may as well rename to whatever is proper new naming scheme. `superstarlight`, `superchisel`, or `utilitystarlight`?

GoogleFrog:
Fix the Starlight rule-breaking-naming.

units/starlight.lua
```diff
@@ -347,4 +347,4 @@ unitDef = {
 }
 
 
-return lowerkeys({ mahlazer = unitDef})
+return lowerkeys({ starlight = unitDef})
```
GoogleFrog: Doesn't this break everything?

KingRaptor: I parsed "Fix the Starlight rule-breaking-naming" as requesting exactly this change, was it supposed to mean [this](https://github.com/ZeroK-RTS/Zero-K/issues/1596#issuecomment-267147718)?

I forgot that buildlists, etc. also need changing. Will revert, may as well only touch it once we actually do the consistent unitdef naming for everything at once.

sprunk: I'd parse the request as to do the reverse (rename the file to mahlazer.lua).

Anarchid: I think it makes sense to interpret it this way. This is because `starlight` fits one of the proposed rename schemes (proper name) while `mahlazer` has no chance of fitting any of them.

GoogleFrog: The request does not have a specific way of fulfilling it. There is just a requirement to not break things.

units/mahlazer.lua
```diff
@@ -1,5 +1,5 @@
 unitDef = {
-  unitname                      = [[mahlazer]],
+  unitname                      = [[starlight]],
```
GoogleFrog: Did you run this to see if it works?

units/mahlazer.lua
```diff
@@ -11,13 +11,13 @@ unitDef = {
   buildingGroundDecalDecaySpeed = 30,
   buildingGroundDecalSizeX      = 11,
   buildingGroundDecalSizeY      = 11,
-  buildingGroundDecalType       = [[mahlazer_aoplane.dds]],
-  buildPic                      = [[mahlazer.png]],
+  buildingGroundDecalType       = [[starlight_aoplate.dds]],
+  buildPic                      = [[starlight.png]],
```
GoogleFrog: Breaks rules.

units/mahlazer.lua
KingRaptor: Yes.

(The unitname being `[[starlight]]` was already in the starlight.lua that's used in current stable and dev, it just looks like a change due to the git rename)

GoogleFrog: Regardless, it looks wrong to me.

sprunk: The value matters not to engine - there is no "unitname" unit def tag.

The value matters not to humans - it is a duplicate of the file name.
The tag should universally be removed in a separate commit.

GoogleFrog:
Clippy conflict. The PR tries to do too much at once now.

PR #1737, opened 2016-12-17
Sea balance changes

Aquanim:
Mistral missile replaced by a volley of four slightly wobbly Rogue-like rockets. 4x280 damage, 8s reload, 610 range (functionally outranges Urchin but not Defender/Stinger). Modified OKP is against statics and only 280 damage because the wobble makes its damage unreliable. Changed unit description to suit.
Ronin projectile speed 330 -> 400. Also cleaned up comments in its unit script. OKP timeout adjusted to suit.
Siren cost 590 -> 600, HP 4200 -> 4000, missile reload time 10s -> 16s (effective DPS 40 -> 25).


Aquanim:
I believe that the interaction between Mistral and Corsair in particular is likely to depend heavily on the larger-scale strategy of the game in question, and it would not surprise me to find that one or the other needs tweaking based on experience in real games.

PR #1774, opened 2016-12-26
UnitAI: Minor scorcher update

mtroyka:
-Fixed scorchers not diving vs facs. This isn't exactly optimal because they do get blown up, but a dead fac is usually worth that price.
-Improved (hopefully) the reliability of scorcher hug. It sometimes got stuck due to giving long skirm orders to unpathable spots.

### DIFF
--- a/LuaRules/Configs/tactical_ai_defs.lua
+++ b/LuaRules/Configs/tactical_ai_defs.lua
@@ -552,7 +552,7 @@ local behaviourConfig = {
 		swarms = lowRangeSwarmieeArray,
 		flees = {},
 		hugs = shortRangeDiveArray,
-		fightOnlyUnits = shortRangeExplodables, 
+		fightOnlyUnits = diverExplodables, 
 		localJinkOrder = false,
 		jinkTangentLength = 50,
 		circleStrafe = true,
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Who is the first change for? I think it would get in the way for anyone who needs to dive factories.

mtroyka:
No, scorchers -weren't- diving vs facs, this change makes it so that they do.

GoogleFrog:
I know. If you are a player who needs to dive factories then you're advanced enough to know when you don't want to dive factories. Often when raiding bases you'll give an attack command on the factory in order to apply set target. Diving factories is easy to implement with a move command so I think there is little benefit to the change.

mtroyka:
Without the change it takes set target and a move order, and with the change it just requires an attack command. Or just a fight order. AIs are generally only capable of the latter, since move orders directly on top of buildings cause pathing failure.

GoogleFrog:
You need to compare it ordering Scorchers to not hug:
 * With the change: Attack, Move
 * Without the change: Attack

It looks symmetrical but I think issuing a move command to make your units stay still is more annoying than issuing a move command change their position. In both cases the second move command is a refinement on how you want your units to attack.

GoogleFrog:
I tested reduced skirm order distance and it is a poor change. Scorchers are speedy and have noticeable deceleration so the shortened stopping distance (any less than 120) causes them to be slower when skirming things than they otherwise would be.

PR #1775, opened 2016-12-26
Improvements to drone controls

Aquanim:
- Set Target on a carrier results in the same drone behaviour as an Attack order. (Set-target is overridden by Attack, and is not cancelled if a Move order is given to the carrier, as you would expect.)
- Drones inherit their firestate from their carrier. (Probably most useful for a drone bomber.)
- Added a button to recall drones to the mothership.
-- EDIT: Recall no longer cancels orders given to the carrier; it just tells the drones to ignore them for a while.
-- I am not a graphics designer; improvements to the recall icon are welcomed.
-- Commanders get the recall drones button if and only if they have drones.
-- When the carrier is given a recall order, the drones will hover around the carrier for ~10 seconds before returning to normal operation.
-- An attack, fight, patrol or set-target order given to the carrier cancels the remainder of the drone recall time.

It would be a good idea for somebody else to test this before merge/stable, not only for bugs but also to see if this control scheme is intuitive to other people.


sprunk:
As far as I can tell you're setting the _recall_frame_start_ rules param to _-RECALL_TIMEOUT_ as a form of removal, including on newly created units. At the same time you're still checking it for nil.

If you're guaranteeing the param to be non-nil, consider dropping the nil checks.
Since GoogleFrog would probably raeg at dropping nil checks, alternatively consider explicitly setting the param to nil when removing.

Aquanim:
The second option seemed more reasonable.

LuaRules/Gadgets/unit_carrier_drones.lua
```diff
@@ -90,6 +107,8 @@ local function Drones_InitializeDynamicCarrier(unitID)
 		if drones then
 			carrierData[#carrierData + 1] = data
 			maxDronesOverride[#maxDronesOverride + 1] = drones
+			Spring.InsertUnitCmdDesc(unitID, recallDronesCmdDesc)
+			spSetUnitRulesParam(unitID,"recall_frame_start",nil)
```
GoogleFrog: If this is just initializing the unit (as the previous line suggests) then this line does nothing and should be removed. Setting the unit rules param to nil would only do something if it previously had a value.

LuaRules/Gadgets/unit_carrier_drones.lua
```diff
@@ -414,12 +436,27 @@ local function GetDistance(x1, x2, y1, y2)
...
 end
 
-local function UpdateCarrierTarget(carrierID)
+local function UpdateCarrierTarget(carrierID, frame)
 	local cQueueC = GetCommandQueue(carrierID, 1)
 	local droneSendDistance = nil
 	local px, py, pz
 	local target
-	if cQueueC and cQueueC[1] and cQueueC[1].id == CMD_ATTACK then
+	local recallDrones = false
```
GoogleFrog: Whitespace?

LuaRules/Gadgets/unit_carrier_drones.lua
```diff
@@ -540,6 +629,13 @@ function gadget:UnitDestroyed(unitID, unitDefID, unitTeam)
 	end
 end
 
+function gadget:UnitCreated(unitID, unitDefID, unitTeam)
+	if (carrierDefs[unitDefID]) then
+		Spring.InsertUnitCmdDesc(unitID, recallDronesCmdDesc)
+		spSetUnitRulesParam(unitID,"recall_frame_start",nil)
```
GoogleFrog: Surely this is not required.

PR #1791, opened 2016-12-30
Fix #1781

### LINKED ISSUE #1781: Translate Kloot Mantis 5262
https://springrts.com/mantis/view.php?id=5262

A fix for terrain texture might come out of it.

sprunk:
This does not work for me (still grainy) but perhaps it's just the crappy GPU that can't generate a mipmap so it could be nice if someone double checked.


GoogleFrog:
I'm not going down the path of testing different combinations of symbols for you.

sprunk:
I wanted to ask Kloot for feedback in the Mantis ticket but it's closed.
I asked in the Spring forums instead ( https://springrts.com/phpbb/viewtopic.php?f=23&t=35878 ).

sprunk:
This now works. I don't know the full performance implications but there wasn't any tangible difference even on very crappy hardware.

PR #1798, opened 2017-01-05
Blurry Halo Selections Improvements

mtroyka:
-Simplified options
-Fix: it should now work on ati cards.
-Now uses a real blur shader so it should produce smoother halos.
-Halos should now stay relatively constant sized vs zoom.
-Code cleanup.

A downside it that it runs slower. I haven't made any particular effort to optimize it, which would be fairly tricky since downsizing the textures tends to lead to ugly halos.

### DIFF
--- a/LuaUI/Widgets/unit_selectionblurryhalo.lua
+++ b/LuaUI/Widgets/unit_selectionblurryhalo.lua
@@ -28,7 +28,7 @@ options_path = 'Settings/Interface/Selection/Blurry Halo Selections'
 options_order = {
 	'showAlly',
 	'thickness',
-	'opacity',
+	'blur',
 }
 
 options = {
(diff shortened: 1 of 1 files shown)


mtroyka:
I found out after trying to optimize it that halos for small units like darts and daggers may completely disappear when zooming out if textures smaller than screen size are used. After further testing the performance seems ok regardless.

mtroyka:
I've decided to try using separate passes for feathering the unit shape and blurring, since blurring alone tends to thin the halos to the point of invisibility, however the old 'blur' shaders have serious issues so I'm marking this as WIP until I can make proper shaders for it.

mtroyka:
Comparison:
Old:
[https://s28.postimg.org/u1rpkkf4b/screen00025.png?dl=1](url)
New:
[https://s29.postimg.org/3zq59rklh/screen00206.png?dl=1](url)

Issue #1811, opened 2017-01-12
Unit start states nonsense


mtroyka:
Skirmishers (and certain arty units) are set to hold position by default in unit start states. This is bad because those units will not skirmish when set to hold position, leading them to suicide stupidly. They should just use 'inherit from factory' like everything else does.

Factories should probably be set to 'maneuver' by default, except for planes because their leash range is ridiculously short for their speed.

Also related: brawlers should not use air strafe by default, because it causes them to wander into enemy porc/units and die for no reason. Other gunships are probably fine as is.

I would just PR this but I don't even know where the start states defaults are stored. I checked the widget and found no mention of skirmishers or arty.

sprunk:
>I don't even know where the start states defaults are stored

[Here](https://github.com/ZeroK-RTS/Zero-K/blob/master/units/armrock.lua#L45)

Aquanim:
The start state in which a Rocko can be lured into range of a Stinger because it was chasing a constructor is an even more stupid start state. There is a reason why this is as it is. 

It is easier to make your units skirmish when desired by giving them a fight order than it is to stop them from chasing every time they are baited.

mtroyka:
Is that not exactly what every other unit in the game does? I don't see why skirmishers are special, and if anything it's even more stupid when a  rocko is standing there and then dies to a random warrior because it failed to skirmish due to holdpos, which is way more common than baiting is. With skirmishers that have more range than rocko (esp wolverine, which outranges stinger, and vs llts for most other skirms) it's significantly worse. I don't agree with a default setting that defeats tactical ai and if someone wants their skirms to hold pos then unit start state can still be customized to achieve exactly that.

This is also why I intend to set facs to maneuver by default, because it's significantly less bad vs baiting than roam is.

Aquanim:
>Is that not exactly what every other unit in the game does? I don't see why skirmishers are special
They have longer range, and by virtue of their role are frequently the first unit to engage the enemy. 

I don't park my Glaives just outside the range of a Stinger because there would be no point (they would be outranged by a unit protected by the Stinger. However, I would do that with Rockos more frequently.

>which is way more common than baiting is
I think you do not realise how much more frequently skirmishers would be baited than other units.

Licho1:
Player should use fight (or attack move in nuspeak).

Anarchid:
Skirmishers should skirmish.

DeinFreund:
[...]

PR #1812, opened 2017-01-13
Unit start state defaults

mtroyka:
-Removed hold position defaults for all skirmishers, arty, and AA units.
	Exceptions: slasher, crabe, tremor, which should not chase units.
-Set air-strafe to off by default by brawler because it suicides badly when enabled.
-Set outlaw to be activated by default.
-Set all factories to maneuver by default, except for plane fac which is set to roam.

-Added buttons to the unit start state widget for setting various unit types to hold pos. (skirms, arty and AA)
-Added a button to reset all ground units to their default move state.

(this ended up in GFX branch by mistake)

Has been tested and works.


units/cormak.lua
```diff
@@ -3,6 +3,7 @@ unitDef = {
   name                   = [[Outlaw]],
   description            = [[Riot Bot]],
   acceleration           = 0.25,
+  activateWhenBuilt      = true,
```
GoogleFrog: What?

mtroyka: This controls the 'on' state.

Aquanim:
>Removed hold position defaults for all skirmishers, arty, and AA units.
I believe this is not a good idea (see #1811).

mtroyka:
It was inconsistent before. Not all skirms, arty and AA were on hold pos by default. With this you can set them all consistently to hold pos with one button click even though they aren't by default.

Aquanim:
None of those statements is relevant to my statement that "skirmishers should holdpos by default". You raised buoy as an exception but I could make an argument that buoy is a flex-assault.

**IF** there is consensus that changing holdpos default for skirmishers is a good idea, then go ahead I suppose. If that consensus does not exist I object to this change in the strongest possible terms.

mtroyka:
Well the issue basically comes down to this, your skirmishers are standing there idle and some warriors/glaives/whatever walk up.

On holdpos your skirmishers will stand there like idiots and die. On maneuver they will skirmish properly and have a better chance of not dying.

On the other hand the situation is reversed wrt being baited into porc.

I don't think either one is necessarily better than the other, and neither handles every situation perfectly, however having skirms/arty/etc inherit from factory by default is more consistent with other units.

I've also made it extremely easy to set them to hold pos in bulk (and to reset defaults for move state), whereas currently if you -don't- want them on hold pos it's an extreme pain to dig through every stupid fac and change the settings for every unit.

If it's decided that default holdpos is preferable, it probably shouldn't be too difficult to filter arty/aa/skirms and set their movestate using the widget during its initial setup, and would probably be a much better solution than using the unitdefs as is currently the case.

Anarchid:
Skirmishers need to skirmish more often than they need to stand ground in my experience.

Aquanim:
However, if they do need to stand ground and they are presently not on hold position, the only way to achieve the desired behaviour is to put them on hold position. This requires you to change it back when you no longer desire that behaviour.

On the other hand, if your skirmishers need to kite/pursue something, you can give them a fight command and they will do the clever thing... then automatically revert to not chasing stupidly when the fight command is finished/cancelled, without any further required input from the player.

Licho1:
Hold pos is probably only good for arty stuff, for ordinary skimrs it should not be hold pos.

GoogleFrog:
Pillager cannot fire while moving.

mtroyka:
>Pillager cannot fire while moving.

That's true, but move state doesn't particularly fix that and would otherwise prevent it from skirmishing when enemies get close to it. If anything that's a tactical ai issue.

LuaUI/Widgets/unit_start_state.lua
```diff
@@ -310,17 +308,31 @@ local function addUnit(defName, path)
...
-            desc = "Values: inherit from factory, hold position, maneuver, roam",
-            type = 'number',
-            value = ud.moveState,
-            min = -1,
-            max = 2,
-            step = 1,
-            path = path,
-        }
-        options_order[#options_order+1] = defName .. "_movestate1"
+		if string.match(ud.tooltip, 'Skirm') or string.match(ud.tooltip, 'Arti') or (string.match(ud.tooltip, 'Anti') and string.match(ud.tooltip, 'Air')) then
```
GoogleFrog: Why are you doing it this way? The widget should not contain data, especially not hidden in code like this. It should read ud.moveState for everything.

mtroyka: Defining default in unitdefs is unmaintainable, thus why buoy and recluse and random AA units aren't set to holdpos already.

It would be better to have a defaults file like is done for practically every gadget that uses settings.. which I guess I could do instead.

GoogleFrog: Defining in the defs is one of the most maintainable places to define things. It must be there.

mtroyka: What? So that I have to go through 50 different files to change something that's being applied as a blanket case to whole classes of units? Please define 'maintainable'.

GoogleFrog: A resilience to breaking due to changes. Your change breaks as soon as someone changes a tooltip, moveState won't even break if a replacement widget comes along.

mtroyka: No, if someone changes a tooltip then it means the unit role changed, in which case it's only natural that the default move state also changes. Again, there are 3 or more units for which move states are not correctly set in the unit def, and zero units for which this code produces incorrect results.

GoogleFrog: The holdpos exception only makes sense in terms of this widget as it is used on the widget meta-settings. The don't fire at radar is laziness. These settings don't mean we should go backwards. I can't believe that you are advocating basing something on string parsing as a good solution.

mtroyka: I can't think of a more reliable way of determining a unit's class than parsing the tooltip. It doesn't necessarily work for all state defaults, but for this in particular I'm not aware of a better solution. I used the same solution in my AI for sorting out riots, skirms and arty as well, as it's a lot more efficient than naming every unitdef and just as reliable.

mtroyka:
Closing this to resolve internal conflicts and unnecessary file changes.

PR #1823, opened 2017-01-15
Unit State Defaults

mtroyka:
-Added options to unit_start_state to control move states over unit classes more easily.
-Moved constants from unit_start_state into LuaUI/Configs/unit_state_defaults.lua
-Cleaned up some old dead code.

-Added default holdpos to all ground skirms/arty/AA and similar units which did not previously have it.
-Removed default holdpos for Banisher, which is no longer a skirmisher.
-Added default maneuver movestate to all factories except planes, which is set to roam instead.
-Changed Brawler's default air-strafe state to off.
-Changed outlaw's on/off state to default to 'on'.


LuaUI/Widgets/unit_start_state.lua
```diff
@@ -16,31 +16,36 @@ end
...
-	[UnitDefNames["armanni"].id] = true,
-	[UnitDefNames["armmerl"].id] = true,
-}
-
---local rememberToSetHoldPositionPreset = false
+local alwaysHoldPos, holdPosException, dontFireAtRadarUnits = VFS.Include("LuaUI/Configs/unit_state_defaults.lua")
 
 local function IsGround(ud)
-    return not ud.canFly
+    return not ud.canFly and not string.match(ud.name, "factory")
```
GoogleFrog: There is a tag like 'ud.isFactory' which should be used instead of string matching.

PR #1829, opened 2017-01-16
[untested] fix #1821 and fix #1822 - CUS materials

### LINKED ISSUE #1821: Broken CUS with dev engine
Get 103.0.1-503-g09b43c2 and look at a commander with a skin. It does not break immediately. Perhaps drawing nanospray or some other model causes it to break.

screen00241
screen00242

### LINKED ISSUE #1822: Lua materials warnings
Look at this infolog http://pastebin.com/qRbSxppf

sprunk:


GoogleFrog:
I doubt this works.

sprunk:
Why do you think so? The way I understand the issue:

I looked at 49f2f41e8a84333790590fcbefba26575f5fca40 and it says "103 compatibility". According to my memory at the time 103 was the new version, so this is forward-compatibility and the check should've included 103.0 and onwards, but instead includes 103.0 and backwards.

103.0 is not newer than itself so passes the "not IsNewerThan(103, 0)" clause which is why it worked until now, but dev versions won't.

GoogleFrog:
We have not used 103.0 in a while. The current version for a few weeks is 103-298.

sprunk:
Makes sense.

PR #1830, opened 2017-01-16
properly fix lups allyteams

sprunk:
WIP because some raw LocalAllyTeamID is also used in particle subclasses


sprunk:
I wasn't able to test the cloak shader due to hardware but I managed to test on nanolasers and it seems to work.

Issue #1838, opened 2017-01-19
No Storage state shows low BP warning


dvr333:
When out of storage, the little warning under metal regarding not using enough constructors occasionally shows up, even if spending is greater than income.

http://i.imgur.com/JwSggRm.png

PR #1839, opened 2017-01-19
Fix for economy warnings with no storage (#1838)

dvr333:
* Metal warning when there's no storage triggered when net metal is positive
* Energy warning when there's no storage triggered when there's more metal income than energy income (i.e. energy is the bottleneck)

### DIFF
--- a/LuaUI/Widgets/gui_chili_economy_panel2.lua
+++ b/LuaUI/Widgets/gui_chili_economy_panel2.lua
@@ -612,11 +612,6 @@ function widget:GameFrame(n)
 		bar_energy:SetColor( col_energy )
 		bar_overlay_energy:SetColor({0,0,0,0})
 	end
-	
-	local metalWarning = mCurr > mStor * options.metalWarning.value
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_chili_economy_panel2.lua
```diff
@@ -781,6 +776,11 @@ function widget:GameFrame(n)
 	else
 		bar_overlay_energy:SetCaption(positiveColourStr..">>>>>>")
 	end
+	
+	local metalWarning = (mStor > 1 and mCurr > mStor * options.metalWarning.value) or (mStor <= 1 and netMetal > 0)
```
GoogleFrog: This could match the excess test on line 630 for consistency. I don't know how reliable netMetal is.

dvr333: The excess test also seems to happen when it shouldn't and unexpectedly, actually.

GoogleFrog: Could you also make the red text match the warning?

PR #1851, opened 2017-01-22
Emulate PlayerRulesParams.

sprunk:


GoogleFrog:
Currently I'd rather this be in a GG than in Spring. Function override is always a bit of a hacky thing and I don't want a function to look like its in Spring. I could be persuaded either way though.

sprunk:
My reasons for using `Spring`:
 * consistency with other rules params functions.
 * trivial transition in case this becomes available through engine.

GoogleFrog:
Ok, looks decent.

PR #1865, opened 2017-01-24
Ready: Autogroups fix

ArchShaman:
Should fix autogroups not working for commsharers.

### DIFF
--- a/LuaUI/Widgets/unit_auto_group.lua
+++ b/LuaUI/Widgets/unit_auto_group.lua
@@ -1,4 +1,4 @@
-local versionNum = '3.031'
+local versionNum = '3.032'
 
 function widget:GetInfo()
   return {
@@ -118,7 +118,6 @@ options = {
 }
 
 local finiGroup = {}
-local myTeam
 local selUnitDefs = {}
 local loadGroups = true
 local createdFrame = {}
@@ -155,7 +154,6 @@ function widget:Initialize()
 		widgetHandler:RemoveWidget()
 		return false
 	end
-	myTeam = team
 end
 
 function widget:DrawWorld()
@@ -181,7 +179,7 @@ end
 end
 
 function widget:UnitFinished(unitID, unitDefID, unitTeam)
-	if (unitTeam == myTeam and unitID ~= nil) then
+	if (unitTeam == Spring.GetMyTeamID() and unitID ~= nil) then
 		if (createdFrame[unitID] == GetGameFrame()) then
 			local gr = unit2group[unitDefID]
 --printDebug("<AUTOGROUP>: Unit finished " ..  unitID) --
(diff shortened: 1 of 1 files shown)


Licho1:
Why dont you apply global solution as suggested, its trivial to just call deinit/init on commshare enable.
You will need to patch similar things in every place now and in the future ..

ArchShaman:
Why not just grab the proper teamID?

Licho1:
Because its not just one widget, its probably most of them.

To init a widget call its initialize..

In the worst case or if you are lazy you could always just do "/reload" to re-init whole lua..

sprunk:
> its not just one widget, its probably most of them

In most cases it's just trivial global replace cachedMyTeamID -> Spring.GetMyTeamID().

>To init a widget call its initialize..

No. Most widgets initialize "constants" (such as own teamID) right on declaration outside Initialize(), and some expect Initialize() will never be called more than once.

>you could always just do "/reload" to re-init whole lua..

Don't, this causes a freeze and more importantly, discards any data other widgets might have collected.

ArchShaman:
> In most cases it's just trivial global replace cachedMyTeamID -> Spring.GetMyTeamID().

Which is pretty much what I did. I did try the initialize for auto-groups and that also worked apparently.

Licho1:
It's not just about that, it's also about internal lists of units many widgets maintain (like transport AI list of transports etc).

We should prefer general solution, rather than hand patching + maintenance.

If cachedteamid is often setup outside initialize, then let's just do /reload - this will force clean and correct state for all existing and future widgets at the 0 cost of maintenance.

ArchShaman:
And at the cost of user experience too. :/

Licho1:
Thats nothing, its going to be rare (commshare) + one time per game (share accept).

Compare that with user experience casued by existing and future bugs and the massive cost of dev time.

ArchShaman:
I can see reiniting for the individual buggy widgets, however I still think it's excessive to reload every widget.

> https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Widgets/unit_building_starter.lua#L18

Would this be easily fixed by Spring.GetMyTeamID() -> Spring.GetMyTeamID ?

Licho1:
Probably not..

 1) if you change it from call to call it can break expectation of the effect (if teamid associated with same item cahnges)

 2) its more expensive to call out to engine for it, that's why it's cached everywhere

 3) really it does not solve all classes of issues

As I see it by far the most valuable thing we have is dev time and historically it was proven that fixing bugs is hardest thing, especially in "other's" widgets (because people prefer to add new things instead).
There is absolutely nothing lost in reload causing slight delay (on my pc it's about 1s).

ArchShaman:
How expensive are GetMy____ callouts? As expensive as creating tables?

Licho1:
I still disagree, this does not solve MOST issues!!! 
Either fix all of the widgets, or it's cheaper to just remove commshare. We cannot add "bugged by design" features to game. 
Adding extra callin means updating and reviewing ALL widgets we have, current and future.

I cannot imagine a worse waste of work.

Seriously I would rather remove commshare than add buggy commshare which adds work to others. Already it burned some serious dev time.

GoogleFrog:
@Licho1 you know about this. What to do?

Licho1:
Don't know current state of things. I would recommend shutting down and initialize widgets on commshare change. This seems cleanest with least chance of issues (and yes it can cause a small lag, but it's universal solution that will work even with legacy widgetts).

It seems much simpler than doing and maintaining all widgets as "commshare aware".

hedgehogsthateat:
delete commshare. OP and makes disorganized teams

GoogleFrog:
It may work still.

sprunk:
See 01c62f6d34e9d5be81460a347917cb885f90b28b (same as this PR, done cleaner)

PR #1885, opened 2017-01-31
Slasher skirms turrets.

sprunk:


mtroyka:
No. Slasher only skirms near enemy range which only makes it practical for things which are short ranged (rather than everything with less range than it has). Any time a slasher moves it's expensive and potentially frustrating, because they cannot shoot while moving and thus are not skirmishers.

sprunk:
Fine. What about adding all shorter ranged turrets to the list? They won't cause a Slasher to reposition because they can't move, but I think it will prevent Slasher from driving underneath them.

mtroyka:
That's probably not a bad idea.

sprunk:
I tested and it looks like slashers escaping from under a turret will go pretty far away (instead of just behind its range) due to [skirmKeepOrder](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Configs/tactical_ai_defs.lua#L1025).

GoogleFrog:
I would be worried about this change if Slasher was not set to hold position with tactical AI disabled by default. My worry would be that Slashers would move out of turret range when told to attack the turret, dying pointlessly instead of shooting at the turret. However, Slasher tactical AI behaviour is already not recommended for general use so this looks fine.

Issue #1892, opened 2017-02-04
Buttons are hard to see


GoogleFrog:
Shadowfury333 made the buttons dark and now they are hard to see against a white background.

http://i.imgur.com/y3wiB2w.png

GoogleFrog:
Fixed.

PR #1898, opened 2017-02-06
Change Evolved button colours to fit less transparent panels

dvr333:
Possible fix for #1892


GoogleFrog:
Border color darker?

dvr333:
It seems to provide a bit of a contrast within the button itself, but I'm not entirely sure how that parameter works.

GoogleFrog:
Border color is for the foreground image. There are two images for the button.

dvr333:
Is that a major problem going forward, or is having it look visible on something bright and against a less transparent panel good regardless of how it gets there?

PR #1902, opened 2017-02-07
Use the nicer smoke.

sprunk:


GoogleFrog:
This smoke uses a more complicated callin than emitSfx and also makes the unit look like it is on fire.

sprunk:
alright

GoogleFrog:
Perhaps it can be used for units on fire if it's nicer than Lups? I am not sure if it is though.

sprunk:
I prefer the Lups fire too.

PR #1907, opened 2017-02-08
Moderator: sight = weaponRange*1.1

Anarchid:


Anarchid:
I couldn't remember the sight rule and i failed to infer it from other units because they were all bitrotten. 

Moderator is an egregious example of sight rule bitrot because it used to have a very long ranged slowbeam and now it has a very short range disruptor, but its sight hasn't changed.

Perhaps the true solution is to unitdefs_post it.

sprunk:
The rule is about minimum sight so morty does not break it.

Is Puppy decent at providing LoS? It's probably good to have some reliable sight provider in a fac.

GoogleFrog:
The rule is not strict. Maybe moderator is the sight provider of the factory. It has trouble getting sight in other ways.

Anarchid:
Pyro has 420 and Puppy has 425. 
It's good to have LoS providers but i think it is also good if it's possible to infer which things will have good sight before building them. The rule for finding those things ATM seems to be "things which shoot far, also scouts and radar units".  So i guess i'd buff Puppy range. Disruptorator fits none of the above parameters.

GoogleFrog:
True, it does not feel scouty.

PR #1996, opened 2017-02-27
[WIP] Fix #1988 - Spring burst disarm obedience

### LINKED ISSUE #1988: Spring bursts are not interrupted by disarm
Spring-side burst weapons are interrupted by EMP. The burst halts when the EMP is applied and resumes after it wears off. This should be implemented for disarm, possibly only for the units which have noticeably long burst (Dante, Bantha, Catapult etc..).

sprunk:
To do:
 * handle weapon > 1
 * test


sprunk:
40ab023c7d34ad928d7f0b8d3395df9ea5374ff3

PR #2011, opened 2017-03-01
Some adjustments to terraform look & internals:

lhog:
1) terraunit is shown using an icon, instead of current debris model
2) Removed terraTag from the gadget. No one used it, including myself (original author). commandTag should be good enough alone
3) Added terraForm type just in case someone finds a good way to display a unit icon, specific to terraform type (e.g. ramp, level, etc). I'd tried WG.icons.SetUnitIcon and couldn't get a good visual result.


quantumv:
Hurrah for icons! The debris were really ugly (I'm the one responsible for them!).

lhog:
Unfortunately, I couldn't make an icon switch depending on terraform type. One easy way would be to make 6 terraunit types with 6 different icons respectively. Sadly it'd break checks like ```unitDefID==terraDefID``` in various widgets & gadgets

GoogleFrog:
I dislike the unsynced section being in the gadget. Ideally you would set their icon distance so low that they will always be drawn as an icon. I don't remember the tag for icon distance but Ticks have it to make them draw as icons before every other unit.

lhog:
It looks so that it's no longer required to suppress engine's unit drawing.
Previously unit model used to clip through icon in some cases, now it doesn't seem to be the case anymore. Thus ecde74e

PR #2012, opened 2017-03-01
[PlayerList] Support for UserInfo server message.

### LINKED ISSUE #2010: Playerlist player multiplication
Screenshots are failing, but upon connecting to MM game, KingStad who was spec also appeared as player on team 1

DeinFreund:


LuaUI/cawidgets.lua
```diff
@@ -224,6 +225,7 @@ local callInLists = {
   'TextCommand',
   'CommandNotify',
   'AddConsoleLine',
+  'ReceiveUserInfo',
```
sprunk: I'd be wary of putting it like this because it's not an engine call. At least document this.

DeinFreund: Document where? If we have a documentation about this it'd be nice to know..

sprunk: inline

DeinFreund: done

PR #2014, opened 2017-03-01
Radar gets an option to morph to Adv. Radar. The upgrade is service-affecting.

lhog:


mtroyka:
That's not a bad idea but it should be disabled during morph like everything else that isn't a commander.

lhog:
I felt disabling radar would prevent people from performing a morph, because radar coverage is crucial in most of the cases. That being stated, it's of course possible to change.

mtroyka:
I kind of doubt that, since you wouldn't usually morph a front-line radar to adv.

GoogleFrog:
The upgrade should be service affecting for consistency.

This PR is one of those sudden balance changes with no discussion behind them. There should at least be discussion here. I don't see the benefit of this change:
 * You rarely want Adv Radar in the same spot as your Radar.
 * You rarely want Adv Radar.
 * Morph is not simply a convenience thing, it is a way of generating free buildpower and of circumventing the normal logistics challenges of moving constructors around. These logistics challenges are a testable skill in ZK.
 * The morph will occur at high priority by default because radar has high misc. priority so it keeps running during stalls.

I think this change removes a bit of logistical gameplay at best and is a noob trap at worst.

lhog:
It's a bit sudden I agree. The motivation was that I often found myself in the situation where I wanted to build  (and built) Adv.Radar right next to the existing, basic one.

As for morph priority I think it's possible to decouple running and morph priorities. But you seem to dislike the idea, so if you feel it's totally wrong, feel free to close, not a biggest deal anyway.

sprunk:
>As for morph priority I think it's possible to decouple running and morph priorities.

It is but then we have 3 separate priority states.

hedgehogsofZerok:
Are you sure we really need Adv morph? This was never stated in Forum, might want to discuss there

Licho1:
Personally I like morphs. Especially for experienced units or for units that are direct upgrades or depend on spot (geo).
Radar is direct upgrade so morph is ok imo.

GoogleFrog:
My concerns still stand. My primary worry is that it would be a noob trap since morph is an easy way to destroy your economy and adv radar has niche use.

lhog:
I understand your concerns.

It might become yet another noob trap.However the game have so many already:
* Geo morph (you can e-stall pretty hard during morph)
* Starting expensive mobile unit as a first unit in the game
* Start porcing up your 3 home turf mexes with DDM/Anni/Behemoth/Screamer
* Starting a superweapon, one cannot afford
* Unjustified singularity reactor construction
and the list goes on. If you think of it, basically whole game with it unit classes mechanics is a big noob trap.

I don't think another one will make a difference for "noobs", but I still think it's somewhat convenient for the rest of the players. 

I don't feel like I can defuse noob traps with code in this one nor I'd like to argue too much about this change request.

If you feel idea is wrong, plz close the PR.

mtroyka:
Adv radar is already a noob trap. Most of the other things you listed aren't noob traps, just dumb decisions.

A real noob trap is the current wyvern, which is now easy to kill using tacnukes and doesn't do jack damage for the ridiculous price tag.

Licho1:
I would be less concerned about nub traps of this particular type.
Morph is convenient and people loved it, still dont understand why we no longer have morphing high xp units..

lhog:
So: merge or close?

sprunk:
There doesn't seem to be consensus so probably the status quo is to be kept.

PR #2020, opened 2017-03-02
Quadrupled (configurable) Heavy Tank factory open/close animation speed.

lhog:
Currently it opens/closes very slow and thus delays unit production.

### DIFF
--- a/scripts/factorytank.lua
+++ b/scripts/factorytank.lua
@@ -41,39 +41,41 @@ local nanoPieces = {emit_1,emit_2,emit_3}
 local nanoIdx = 1
 local smokePiece = { piece "wing_1", piece "wing_2", piece "wing_3" }
 
+local animSpeed = 4
+
 --opening animation of the factory
 local function Open()
 	Signal(2) --kill the closing animation if it is in process
 	--SetSignalMask(1) --set the signal to kill the opening animation
 
-	Turn(wing_1, 3, -1.57, 1)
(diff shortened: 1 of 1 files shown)


Anarchid:
HT fac model is getting replaced.

lhog:
Last time I checked HT animation was slowest.
On the topic: feel free to close the PR if not needed.

sprunk:
The animation has to look good and variability between facs is not inherently bad so I'd support a change if the value was an extreme outlier. Having some other facs' values for comparison would be nice.

DeinFreund:
I thought this was long animation intentional to punish my lack of eco awareness.

lhog:
* Spiders, seafac, amphs - almost immediately (<1 sec wait)
* Shields, hovers, cloaky - ~2-3 sec
* Jumps - 3 sec
* GS - 4 sec
* Air, LV - 5 sec
* HT - 8 sec

mtroyka:
why does GS have any wait at all? :|

GoogleFrog:
Is this technically a balance change or not?

lhog:
Not so much a balance change. Rather getting rid of an annoyance.

DeinFreund:
This is very much a balance change, but I think we should aim for simplicity here. Unless there's a good reason for those values, standardization sounds like the way to go.

Anarchid:
I think standardizing fac startup time is not as much a balance change as it is slightly decoupling balance from the whims of crazy artists.

Issue #2025, opened 2017-03-03
Icon terraform does not have a healthbar


GoogleFrog:
The new terraform icons look much more professional than the wreck model and are better at conveying that the terraunit is not a real unit. However, they lack a progress bar. I don't want to use Spring.GetVisibleUnits to get all units on screen and then filter out most icons. Perhaps terraform could have a different style of progress bar or some other solution.

PR #2035, opened 2017-03-04
Another terraunit icons implementation. Look and feel is slightly diffferent:

lhog:
1) ETA & build progress bar are supported: #2025  
2) Icon shows what kind of terraform is blueprinted by a terraunit
3) Icon is placed under the ETA and "building" bar (previously it was placed slightly above the ground and moved with the ground if height was changing)

Implementation:
* Unsynced part of gadget suppresses terraunit model drawing (otherwise small sphere will be visible sometimes)
* New widget is responsible for the actual drawing of the terraform icons


GoogleFrog:
The previous implementation was more elegant as it did not contain any drawing in lua. I'd like it compared with allowing the ETA and healthbars widget to draw for icon terraunits. Also, Spring.SetUnitNoDraw is a callout that exists and may be relevant here.

lhog:
Well, yeah, it was more elegant. On the other hand, it also had it's own issues:
1) Icon was following ground level. And if this ground level was moving (due to terraform for instance) then a terraform unit and its icon would move away from each other in projection. If this was to happen, then clicking on the icon in order to select a unit wouldn't work.
2) ETA/healthbar would require handling terraunit as a special case
3) Same icon was used for all terraform types

Spring.SetUnitNoDraw also disables ETA/healthbar so was not directly applicable.

PR #2045, opened 2017-03-07
Add proper by-type sorting to selection window

dvr333:
* Also added option to group units if multiple types are selected.

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip_old.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip_old.lua
@@ -271,7 +271,7 @@ options_order = {
 	'showDrawTools',
 	
 	--selected units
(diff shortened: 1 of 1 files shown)


dvr333:
This uses a new option internally, and won't preserve behaviour from the old option. I'm not sure if doing otherwise is a good idea.

LuaUI/Widgets/gui_chili_selections_and_cursortip_old.lua
```diff
@@ -493,7 +493,14 @@ options = {
 		end,
 		path = selPath,
 	},
-	groupalways = {name='Always Group Units', type='bool', value=false, OnChange = option_Deselect,
+	groupbehaviour = {name='Unit Grouping Behaviour', type='radioButton', 
+		value=overflow,
```
GoogleFrog: Must be a string.

PR #2048, opened 2017-03-08
Dont repeat yourself

### LINKED ISSUE #2042: Puppies don't leave wreckage
http://zero-k.info/Battles/Detail/448850

I thought this got fixed a long time ago.

lhog:
D.R.Y. clean of unitDefs All removed tags get restored during _post stage.
Also should fix #2042


gamedata/unitdefs_post.lua
```diff
@@ -232,6 +232,56 @@ end
...
+
+			if not ud.terraformspeed then
+				ud.terraformspeed = bp * BP2TERRASPEED
+			end
+		end
+
+        --setting standard seismicSignature
+		if ud.floater or ud.canhover or ud.canfly then
+			if not ud.seismicsignature then ud.seismicsignature = 0 end
+        else
```
GoogleFrog: Whitespace.

gamedata/unitdefs_post.lua
```diff
@@ -232,6 +232,56 @@ end
...
+-- UnitDefs Dont Repeat Yourself
+--
+local BP2RES = 0.03
+local BP2TERRASPEED = 60 --used to be 60 in most of the cases
+local SEISMICSIG = 4 --used to be 4 in most of the cases
+for name, ud in pairs (UnitDefs) do
+        local cost = math.max (ud.buildcostenergy or 0, ud.buildcostmetal or 0, ud.buildtime or 0) --one of these should be set in actual unitdef file
+
+		--setting uniform buildTime, M/E cost
+		if not ud.buildcostenergy then ud.buildcostenergy = cost end
```
GoogleFrog: I don't want single line if then end.

GoogleFrog:
This scares me, there is a lot here so something is likely to be broken. Featuredef blocking is intentionally not standardized for death clones.

DEAD --> DEAD2 --> HEAP path may still exist for a reason.

lhog:
> This scares me, there is a lot here so something is likely to be broken.

Tags removal was performed with a number of ```sed``` scripts with manual control of for every changed unitDef. I had to undo only a few changes, specifically ```buildcostenergy & buildtime``` thingie for chicken units. I've never looked how chickens spawn works and if they have anything to do with traditional ZK eco, so I left them unchanged to not break things up.
I have some spare time now, so I bear responsibility to fix my things if they are broken.

> DEAD --> DEAD2 --> HEAP path may still exist for a reason.

Isn't it? I think I had gone through UnitDefs quite thoroughly before making the change and it was always the case that DEAD feature was blocking and HEAP was not. 

> DEAD --> DEAD2 --> HEAP path may still exist for a reason.

These were sparse and occasional cases and it didn't look like it was made with intention. If I'm wrong then what's the reason?

lhog:
I'm going to commit a fixup for items @GoogleFrog noticed and few typos I noticed myself.

Anarchid:
IIRC most light bot corpses are non-blocking, others can walk over them without crushing.

lhog:
Ok, some units indeed had non-blocking corpse and I've collected such unitDefs here:
https://hastebin.com/nocopunuze.lua

I'll restore non-blocking status for light units, as @Anarchid stated. Can other units get a second look?

sprunk:
The DEAD2 changes need to be updated in scripts too.

sprunk:
You removed some "levelGround = true" tags.

lhog:
@sprunk
1) Can you explain logic behind DEAD2? Isn't it a typo/leftover?
2) ```levelGround = true``` was present only for two PlanetWars structures. Not sure what the point for this setting was. Do you know?

sprunk:
2. I was just making sure you're aware of them. Removing both true and false looked suspicious.

1. It's a leftover from the past and should be removed. However scripts should reflect the change - units with DEAD2 currently return three wreck levels (instead of two) so if you just remove DEAD2 they will not leave any wreck when units usually leave debris heap.

lhog:
I walked thru the list of units that had ```blocking = false``` for DEAD corpse and here is the summary of my findings:
* Some land units with cost under <= 240 M (Leveler is the most expensive one).  For example it
  * Has Puppy, Flea, Bandit, Glaive, Pyro, Dagger, but doesn't have Duck, Scorcher and Kodachi
  * Has Leveler, but doesn't have any other riot, except Hover's Mace (see below)
  * Has Placeholder, but no other skirmishers.
* Crawling bombs: Tick, Roach & Skuttle
* Several hover units: Quill, Dagger, Mace, Flail, Claymore. Scalpel and Penetrator don't have it
* All seafac units: Mariner, Hunter, Mistral, Corsair, Siren, Ronin, Cutter, Zephyr, Seawolf. Also Reef, Shogun and Scylla. But not a single Amph unit.
* Some sea/amphibian buildings: Seafac, Hoverfac, Strider Hub, Caretaker, Urchin, Sonar. But again no Amph factory.
* Gnat and Blastwing
* Scorpion (WTF?)
* A bunch of currently unused units
* Some others I have missed

As seen above, there is a glimpse of some idea behind the findings, but the implementation is very inconsistent.

My proposal so far would be to only:
* Make non-blocking DEAD corpses of units with mass <=185. This will include most of the units you will likely call "light"

Also can someone kindly explain what was/were the reason(s) most sea units got non-blocking corpses?

sprunk:
I can agree that in some cases posts is bad, for example wreck blocking. These are the cases where there is no "hard" rule and everything is mostly arbitrary - light things generally don't block and heavy ones generally do but there are tons of exceptions.

But in some cases it doesn't take away anything while preventing mistakes. For example, metalcost = energycost = buildtime, so as long as one of those values is in the file you can still easily see and edit it, but there won't be any failures [like this](https://github.com/ZeroK-RTS/Zero-K/commit/21ee3515baf345a9a642920aff44e251b627888a) where there are 3 identical values and people forgot to make sure they are all the same.

Exceptions can still exist and be explicit. For example see income. There is a rule that cons have income = 0.03\*BP but for example Solar has no BP. So _posts makes sure the rule is kept for cons (so that [failures](https://github.com/ZeroK-RTS/Zero-K/commit/e359a5e2c444dbb215584b7b8d33f5b905537909) don't happen) but Solars still have explicit "energyMake = 2" and if there was an exceptional con it would just have an explicit value as well.

Something that currently works well as _posts is unit mass. It's a "hard" rule in that exceptions are real exceptions (to the point that there currently aren't any).

So, changes from this PR I think are good:
 * M = E = BT and income = 0.03\*BP: nice (see above).
 * terraformSpeed: nobody cares about tweaking this value per-unit but there is a general rule. Excellent for posts.

The ones that should stay out of posts:
 * wreck blocking: there is no rule, the values are arbitrary so should stay explicit for now. If a rule is created then posts would be good but I don't think a defs-based rule would even be desirable here (for example because model shape matters).
 * level ground: no rule on which buildings use it. Currently none do but that is not a rule and could arbitrarily change.

Others:
 * seismic tag is not used so I wouldn't even recreate it in posts.
 * DEAD2 removal does not involve posts. It's not really a "dont repeat yourself" thing because it affects gameplay but should still be done sometime though.

sprunk:
Looks good now

lhog:
Although I feel like wreck blocking and level ground could have been integrated pretty seamlessly into _post, it's now ready for merge without them.

GoogleFrog:
Wreck blocking is somewhat on a per-unit basis based on the size of the unit and the balance of having a blocking wreck. Vehicles and tanks tend to block since they are large and solid looking. Light raiders don't block to allow them to swarm unimpeded. Hovercraft and amph mostly don't block due to some engine requirement which would need retesting. Scorpion probably doesn't block due to copypasta.

PR #2052, opened 2017-03-09
Fix #2047 - allyteam names represent PW factions

### LINKED ISSUE #2047: Use faction name for allyTeam name in planetwars
There is a widget which generates allyTeam names based on startbox positions and clans. This widget should detect when a battle is a planetwars battle and display faction name.

sprunk:
should be able to test until weekend


Licho1:
Its silly hack why not sending it from server? I wanted to make new factions today..

Licho1:
use 266d8b713eae66960f0ef8ecce0de2fcdfa778a6

PR #2086, opened 2017-03-14
This widget and respective shader tries to remedy color blindness by …

lhog:
…alternating screen colors.

Should work fully-fledged for spring versions newer than 103.0.1-699 and in limited compatibility mode with earlier engines.
Currently implemented:
1) Three types of color blindness
2) Two remediation methods
3) Simulation vs Simulation & correction modes

### DIFF
--- a/LuaUI/Widgets/gfx_color_blindness_correction.lua
+++ b/LuaUI/Widgets/gfx_color_blindness_correction.lua
@@ -1,0 +1,158 @@
+function widget:GetInfo()
+	return {
+		name      = "Color Blindness Correction",
+		version   = 1.0,
+		desc      = "Corrects a screen colors for color-blinded people",
+		author    = "ivand",
+		date      = "2017",
+		license   = "GPL",
+		layer     = math.huge, --wanna draw last
+		enabled   = true
+	}
+end
+
+options_path = "Settings/Graphics/Accessibility/Color Blindness Correction"
+options_order = { "cbcType", "cbcMethod", "cbcOnlySim" }
+
+options = {
+	cbcType = {
+		name = "Color Blindness Type",
+		type = "radioButton",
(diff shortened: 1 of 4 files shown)


LuaUI/Widgets/gfx_color_blindness_correction.lua
```diff
@@ -0,0 +1,158 @@
...
+		((options.cbcOnlySim.value and "#define CORRECT") or "").."\n" ..
+		fragment_body
+		
+	cbcShader = gl.CreateShader({
+		fragment = fragment_complete,
+		uniformInt = {screenTex = 0}
+	})
+		
+	if cbcShader == nil then
+		Spring.Log(widget:GetInfo().name, LOG.ERROR, "cbcShader: shader error: "..gl.GetShaderLog())
```
gajop: Even if the shader compiled successfully, you should print out `gl.GetShaderLog()` with a warning if it's not an empty string. Warnings on some cards are sometimes errors on others.

lhog: Ops.... A bit too late. @GoogleFrog was unusually fast with this one.
Though the current shared code yields zero warnings on my system. Do you happen to see any on yours?

gajop: @lhog 
Sorry didn't see this message, no notification maybe? 
No I didn't see any errors or warnings (I haven't tried it yet), this is just good practice when compiling sharers, and it's a sad thing Spring doesn't enforce it. You never know if someone will modify these shaders in the future.

To simplify all this I've created a utility function for SB, maybe ZK should use it? https://github.com/Spring-SpringBoard/SpringBoard-Core/blob/master/scen_edit/utils/shaders.lua#L3-L19

PR #2107, opened 2017-03-22
Create Vittra v2.1.lua

The-Yak:
Standardized metal patches for vittra 2.1


GoogleFrog:
This should be inside the map when you made it.

The-Yak:
Yes, it should be, although metalValueOverride in particular doesn't work from within map. II have addressed this for Vittra 2.2 (not yet released)

sprunk:
When did you test metalValueOverride? It was recently changed: https://github.com/ZeroK-RTS/Zero-K/commit/e156c9e9d3c336c6c4d304dc6c54e9e295f655c3

PR #2111, opened 2017-03-22
Fix #2089 + better display with uneven amount of allyteams

### LINKED ISSUE #2089: Playerlist fails at team names when you spectate and change team
screen00405

DeinFreund:


DeinFreund:
Not 100% sure if 2088 is actually fixed, no teamgames today :/

LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -1420,12 +1422,13 @@ end
...
 local dtSum = 0
 local lastWindow = false
+local myAllyTeamID = -1
 
 function widget:Update(dt)
 	local f = Spring.GetGameFrame()
 	local alt,ctrl,_,shift = Spring.GetModKeyState()
 	if window and window.visible then
-		local showkey = string.lower(WG.crude.GetHotkey("epic_chili_share_menu_v1.24_sharemenu"))
+		local showkey = string.lower(WG.crude.GetHotkey("epic_chili_share_menu_v1.25_sharemenu"))
```
sprunk: Or actually maybe do change the name by removing the version number altogether because it makes this line dangerous (eventually someone will forget to update both places) and generally tempts people.

PR #2122, opened 2017-03-30
DEAD2 must die. It got resurrected by accident in 8f32756

lhog:


sprunk:
That is not DEAD2 which can in a HEAP lie
And with strange reverts even DEAD2 may die

PR #2125, opened 2017-03-31
snipers

sprunk:


Licho1:
Sorry I cannot accept bulelt being invisible to enemies. Game cannot hurt units randomly without giving proper indiciation what is it doing and where it is.

sprunk:
>proper indiciation what is it doing

Loud shot sound, hit explosion effect.

>and where it is

So snipers should decloak on fire.

Licho1:
Sniper decloak on fire is an ok solution.

PR #2130, opened 2017-04-02
Hacksaw: missile flight time 3 → 10 s

sprunk:
Makes the missile much more reliable; baiting now hard-requires to sacrifice the Swift (still makes cost; previously it could be saved with a timely boost).

### DIFF
--- a/units/missiletower.lua
+++ b/units/missiletower.lua
@@ -88,7 +88,7 @@ unitDef = {
 
       explosionGenerator      = [[custom:FLASH2]],
       fireStarter             = 70,
-      flightTime              = 3,
+      flightTime              = 10,
       impactOnly              = true,
       impulseBoost            = 0,
       impulseFactor           = 0.4,


Licho1:
Not good idea imo, this goes against "physics" aspect of the game. If the craft is fast enough to evade, it should evade.

I would like much less "exact" weapons that are working like in starcraft and more less reliable weapons.

GoogleFrog:
This is a lazy fix for the rare cases in which Ravens can avoid the missile due to the missiles large turn radius. The fix that preserves Swift boost (and other fast aircraft?) is to reduce the missile turn radius. If you want Hacksaw to randomly fail against Raven then no change is required.

This change is not just bad due to the Swift interaction. It would cause missiles to circle around in the air for 10 seconds if overkill occurs.

Licho1:
I think random fail to hit raven if it has some rules (like raven on the edge of range, changing course). If it's a raven coming straight at hacksaw it should hit it.
I never saw it miss under such conditions.

sprunk:
I thought the Swift interaction change was good. This is because Swift baiting currently makes Hacksaw practically useless (Swift is useful and not a niche unit, so forcing the planes player to build it is not a cost at all). If the Swift has to die then baiting still makes cost (and Hacksaw is still bad) but not extremely so (220 vs 150, compared to 220 vs 0).

Issue #2133, opened 2017-04-03
Fix attack ground Defender, HLT and LLT overshoot.


GoogleFrog:
Turrets can overshoot to hit slightly beyond their range. It is especially problematic for Defender. The vast majority of cases are blocked by blocking attack ground commands that occur more than (Turret Range - Fudge Factor) away from the turret. Fudge Factor can be fairly small for LLT and HLT and possibly around 10 or 20 elmos for Defender. This should be a simple gadget.

PR #2134, opened 2017-04-03
WIP: prevent offender hax (#2133)

sprunk:


GoogleFrog:
Looks complete too me. I wasn't even thinking of including a spherical check because those cases are rare and it is possibly confused by terraform and UHM.

sprunk:
eb9a5a21473b1f0ab68678d54ee771ef358228ad

PR #2182, opened 2017-04-22
COFC: Improve smoothmesh following

CrazyEddieTK:
Fixed math in smoothmesh following to reduce jumping when starting to scroll. It still jumps if the difference between the smoothmesh and heighmap countours are large and sharp, but this is a big improvement.


CrazyEddieTK:
... just to clarify - the math is the same (just slightly rearranged). The only real difference is removing the condition, which was only applying the correction when facing east.

GoogleFrog:
How is this @dvr333?

sprunk:
At a glance looks legit

Anarchid:
That can be done at merge time 
image

CrazyEddieTK:
Will resubmit in a new PR using a clean branch.

Issue #494, opened 2015-01-09
in-game options menu: GUI should give more feedback (v1.3.1.3)


knorker:
Some options give no feedback which way they are currently toggled.
Similar there are buttons that when clicked result in no obvious visible action.
-> This is not userfriendly.

Buttons labeld "Toggle SomeSetting" are bad because one can not see which value is active.
Instead should be checkboxes like:
SomeSetting - [ ]yes [ ] no [x] maybe

There are many such cases in options-menu, so can not detail all. The text only gives some examples:

![more options](https://cloud.githubusercontent.com/assets/3044855/5676539/c38b3008-97d7-11e4-9f3f-70627410c4e5.png)

Here it is not possible to tell which water setting (basic, reflective, dynamic,...) is currently active.

The buttons
"See More Trees"
"See Less Trees"
at least cause a message by engine in the chat but not so optimal either.

![toggle buttons that should be check boxes](https://cloud.githubusercontent.com/assets/3044855/5676540/c38df860-97d7-11e4-9982-a06faca369ec.png)
The buttons that toggle stuff for camera.
For example when "Toggle Sm**ooo**thScroll" is clicked then there is no feedback that something changed. (Only way to know is to actually try out scrolling.)
Compare with the "Camera Type" options (done as checkbox) as example of good GUI.

xponen:
Enhancement required was: Need the buttons to display feedback (example: like radiobutton do).

GoogleFrog:
I really dislike how many of these settings seem to be hardcoded into epic menu. It should just handle the menu framework and a sperate widget should then use the framework to implement a bunch of engine settings. I am not sure how to make the change in the ticket but it should be done.

At some point a significant number of these settings are going to be either removed or hidden behind the advanced toggle. Very few of the settings found in the menu are needed by new players.

knorker:
>Enhancement required was: Need the buttons to display feedback 

To avoid misunderstanding:
The button could be changed to give feedback: like changing the caption from "Turn SmooothScroll ON" to "Turn SmooothScroll OFF" when clicked.

But better is to use GUI elements in the standard ways: 
-Option with two states (like: ON & OFF) -> use checkbox.
-Options with many states (like: water setting) -> use radio buttons.
-Numerical value (like: ground detail) -> use slider.

sprunk:
Made some of the options more sensible:
[...]

PR #2188, opened 2017-04-25
More lenient noHotkey

sprunk:
Incentives: hotkeyability generally makes the menu look very messy (the button just fits the hotkey text but is still fairly large, this stretches the layout; it also displaces the checkbox/radiobutton) and not thought out ("why can you even set a hotkey for colorblind mode?"). Trying to remove a hotkey for a menu option and getting negative response from people can also signal UI failure - for example if you need to frequently toggle a menu option then probably it should just have a "active only on SHIFT held" mode or such. If someone really, really wants to bind a hotkey to some obscure function then we can consider them to be an advanced user and ask them to do it the hard way via zk_keys, rather than cluttering up the menu system. See also: #494.

--------

No hotkey for graphical settings. These are pretty much constant. Somebody could in theory try to manually lower the settings as the game gets laggy but any single option wouldn't do much; the proper way to handle that would either be a "general quality" setting that applies these in bulk or for the game itself to adjust it dynamically.
 * Shiny units (advunitshading)
 * Smoothly fade out projectile lights after they explode.
 
No hotkey for preferences that I think generally stay constant:
 * number/anvils as mex income representation
 * Radar color presets
 * square/rectangular pictures in selection window
 * whether to combine units of same type even before selection window is filled
 * whether to filter out OpenGL logs from chat.
 * whether to show group info on the side of selection window.
 * whether overlapping parts of grid circles are drawn twice over
 * color-blindness stuff.

No hotkey for meta-toggles for auto-toggling. If you want to fine-tune toggling then might as well hotkey the toggle itself, not the meta-toggle that controls auto-toggling.
 * meta-toggle on whether backlog / backlog-arrow auto-toggle themselves when chatting
 * meta-toggle on whether heightmap/pathmap toggle themselves when terraform/move command is active
 * meta-toggle on whether projectile lights toggle themselves off when FPS is low

No hotkey for misc things, with comments:
 * whether selection filtering is ignored unless ALT. The whole setting is pretty much "hold a key to temporarily disable" so I have no idea why someone would rather toggle the option rather than just hold ALT.
 * whether to use selection filtering in the first place. Since the ALT option exists it's effectively the same as above.
 * toggle healthbars. Actually still hotkeyable because of duplicate functionality, provided by engine /showhealthbars (bound by default to F9).
 * team colour config. Available through the global commands bar easily enough if someone needs it. I'm unsure on whether it should actually live in the bar because it's a preference that stays constant but that warrants noHotkey as well.
 * whether to draw eco circle for powerplants in queue if a powerplant blueprint is the active command. Already requires SHIFT to show the queue so nobody is going to also manually toggle this.

--------

In comparison to previous commit, the following are kept hotkeyable because it sounds legit to only want them sometimes:
 * show ally selected units.
 * show ally retreat zones.
 * show all commands.
However I think this signals an UI failure; they could instead get an "only show when SHIFT held" option (or a new mode for show all commands).

Misc things kept with comments:
 * whether mex circles use color of (owner team) or (first team of owner's allyteam). It's pretty much simplecolors for mexes so the whole setting should just be removed and obey regular simplecolors.
 * hide ally chat (points, labels). Regardless of hotkeyability I think being able to halt all allychat a priori is harmful for cooperation. Possibly remove the option and provide mute feature.
 * hide spec chat. Under current spec system it's legit and whether specs should be able to talk in the first place is a different issue.


GoogleFrog:
> How are you using show all commands? I think if you have the need to switch then it probably just warrants a new mode (that you could sit permanently on).

I have Show All Commands/Show On Shift/Show None bound to Ctrl+Shift+Alt (P/[/]). Sometimes I'll use it for switching in spectating and for screenshots. I often use the hotkey to reenable show all commands when playing.

I don't see much of a negative to such hotkeys. If you're testing out settings it is useful to be able to quickly toggle between options.

sprunk:
>I have Show All Commands/Show On Shift/Show None bound to Ctrl+Shift+Alt (P/[/]). Sometimes I'll use it for switching in spectating and for screenshots. I often use the hotkey to reenable show all commands when playing.

Fair enough. I think the optimal solution here is separating the setting for playing and for spectating.

>If you're testing out settings it is useful to be able to quickly toggle between options.

For "testing" as in when creating a new option then sure, have it hotkeyable for a period of time (but not forever).

For "testing" as in people discovering what the options do I think most wouldn't micro-optimize their time with a hotkey and would just manually toggle them (especially since currently radio-buttons can only have keys bound to setting specific values and not to cycling between them).

>I don't see much of a negative to such hotkeys.

I think checkbox and radiobutton menu options shouldn't in general have hotkeys. If needed, they should instead have an option to make a modifier do things.

For example Show All Commands has five states but in theory just three would be sufficient: Show All - Show Selected - Show None; with people binding a hotkey to toggle between them the way they currently use shift. I think this is worse than having the extra options that use a modifier key.

GoogleFrog:
There is a big difference between toggling and holding to enable. You can't add extra options to encompass all the hotkeys people might want. So far the only negative to hotkeyability that I have heard is menu clutter. The "None" hotkey button could be made smaller if that is really your concern.

sprunk:
Alright. Since there is disagreement and the issue is minor this is becoming a bike shed so I think skim through the changes, pick the ones you agree with and let's ignore the others (later fixing their clutter through reducing the hotkey button size or somesuch).

GoogleFrog:
I like to toggle things during the game to test UI, performance etc... This is a rare usage but I don't think the costs are high.

PR #2198, opened 2017-05-02
Range

hedgehogsthateat:
180-210

Flea beats this for some reason, even on flat ground

210 should help that


DeinFreund:
Could you describe in more detail the problem you are trying to solve?

sprunk:
This is not how @GoogleFrog receives balance changes. Balance changes should generally be discussed to the point of obviousness and a clear intention should be formed. The PR is then a specific implementation of an already fairly specific intention and can also include discussion. Balance PRs which just make changes would result in random walks around design space.

PR #2209, opened 2017-05-07
COFC: Better Tilt and Rotate Speeds

CrazyEddieTK:
. Separate Tilt Speed from Rotate Speed
. Set better default values for each


CrazyEddieTK:
Note that I rescaled the rotfactor / Rotation Speed parameter. So when this gets deployed, anyone who's using COFC with rotation turned on will find their rotation speed has been effectively reduced to almost zero, meaning they won't be able to rotate the camera. They'll have to go into Settings and bump the speed back up.

sprunk:
You can reset peoplers' defaults by renaming the option.

CrazyEddieTK:
Done.

Issue #2147, opened 2017-04-10
Endgame screen issues


GoogleFrog:
Here are some polish issues with the endgame window:
 * The horizontal bars between players in the awards screen have ellipsis (...) at their right side because the lines have a fixed size which is longer than the window. They should have a relative width (probably using 'right').
 * The numbers on the horizontal lines in the graph screen jitter up and down if you click a graph type button multiple times in a row.

CrazyEddieTK:
Regarding the jitter:

 - If you click back and forth between two different buttons, the labels never jump
 - If you click the same button multiple times, whether quickly or slowly, the labels sometimes jump
 - If you wait for the fifteen-second auto-update (in my new version that has the live updates), the labels always jump.

The jitter happens regardless of whether the button is an engine-builtin stat from GetTeamStatsHistory (like metalProduced) or the ZK-specific ones from GetTeamRulesParam (like metal_income).

Note that when you click the same button multiple times, the _same_ numbers are redrawn (and sometimes jump). But when you wait for the auto-update, _different_ numbers are drawn, and it's the _new_ numbers that jump - the old numbers are removed, then the new numbers are drawn slightly too high, then the new numbers immediately drop down into the correct position. Which means the jitter is happening when the new labels are drawn, not when the old labels are cleared.

---------

The code for drawing and redrawing these labels is pretty clean and simple: when you click a button (or wait for the autoupdate), the existing labels and lines are cleared with ClearChildren() and the new labels and lines are drawn with New(). The clearing is done in the button OnClick() and the new ones are created in the drawIntervals() function by way of the getEngineArrays() function. All three of the cases above exercise the same code path even though they have different results.

Therefore, I suspect this is a bug in Chili, or possibly the engine.

I can't troubleshoot Chili or the engine, but I can implement a workaround in the endgraph.lua. I can prevent the buttons from being clicked more than once in a row, with an override parameter for the fifteen-second auto-update call. This would solve the button spamming issue, but the labels would still jump when the graph auto-updates.

If desired, I can add this work-around on top of the new version I've PR'd. Otherwise, someone with more knowledge of Chili / engine will have to look into this more deeply.

CrazyEddieTK:
Regarding the horizontal bars:

I think this will be pretty easy to fix by changing the control from a horizontal stack of objects to a vertical stack of horizontal stacks of objects. I'll work on this and submit a PR against the master branch, since this should be orthogonal to my other PR.

[...]

PR #2228, opened 2017-05-15
Make endgame stats toggleable, including mid-game

CrazyEddieTK:
Added an action to toggle the endgame stats and awards display, bound by default to Alt+F1. The hotkey can also call up the stats screen during the game, which will display the stats and omit the awards panel. The screen now includes a toggle button, which also displays the currently-bound hotkey; this should allow players to discover the new functionality.

Other notes and features:

- The stats graphs auto-update every 15 seconds, allowing you to see a live feed during the game.
- The first time you view the stats, the first graph is displayed immediately rather than showing a blank screen until you click one of the buttons.
- The widget remembers which graph you were last viewing and displays it immediately the next time you toggle the stats on again.
- The button for the currently-viewed graph is highlighted.
- Stats for teams you cannot view are not included in the graph; i.e. if you are a player and view the stats mid-game the enemy stats will not be shown (rather than showing a fake line at zero). Stats for all teams are shown to specs, and to players when the game is over.
- When viewed mid-game, the Awards and Exit buttons are not shown.
- The screen can be toggled even after the game is over, so that you can view the battlefield without having to reposition the endgame screen.
- The screen was already moveable and resizable. Now, though, if you call up the screen mid-game and reposition it, when the game is over and the post-game delay counter has elapsed the screen is repositioned to its standard location so that it can still serve as an end-of-game splash screen.
- The caption during mid-game is "Game in progress - Statistics". At the end of the game, the caption reverts to the normal ones ("Victory!", "Defeat!", etc).


CrazyEddieTK:
I added a couple of WG globals to pass functions and flags between the two widgets gui_chili_endgraph.lua and gui_chili_endgamewindow.lua . I had to do this to work around problems with local variables being instantiated in a closure.

The two widgets could probably be refactored, cleaned up, and combined. I have not attempted this here; I've made only the minimum necessary changes to implement the mid-game toggling functionality.

I may have created memory leaks due to Chili objects being created and then removed but not garbage collected. I don't understand enough about how Chili works to know how to do this correctly.

CrazyEddieTK:
Paging @dvr333

GoogleFrog:
Since they are separate widgets perhaps the graph widget returns a control which draws the graph. If this is the case then a new widget could be created for midgame stats. This widget could call the graph widget in the same way. The benefit is that it is currently bad that midgame stats looks like endgame stats and that "Game in progress - Statistics" is fairly redundant. The midgame stats widget would create its own window that could be independently positioned, even shrunk manually.

GoogleFrog:
I've now had a look at the stats widget and the structure is pretty bad and it doesn't get better in this PR. Here are some things which would at least quarantine bad code and prevent it from spreading further:
 * Remove all loadpanel calls from endgaph.
 * Remove WG.statsPanel.
 * Make loadpanel return window0 and add loadpanel to WG in initialize.
 * Make loadpanel reuse window0 on subsequent calls. Some cleanup may be required, perhaps orphan window0.

The only change to endgamewindow should be replacing WG.statsPanel with the WG call that returns window0 of the graph panel. Then create a new widget to deal with the midgame stats, since I hate the code quality of endgamewindow and don't want to look at it further.

If you're delving deep into endgraph consider these issues: https://github.com/ZeroK-RTS/Zero-K/issues/2147

dvr333:
This sounds like a great idea. I like having midgame stats to pull up for analysis in downtimes (if there are any). The only thing I'd want to see, and this goes for the stats screen in general, is some kind of time interval markers on the x-axis. I don't know if they should be fixed (ex. every 5 minutes) or variable (ex. every 20% of the match, listed as time. For instance, a 10 minute match would have markers for 2:00, 4:00, 6:00, etc.), but something to show at what in-game time a particular part of the graph was relevant would be really handy.

Another nice-to-have would be if mousing over the graph showed the exact value for each player and the exact time in a tooltip window. There are a couple other minor things, but basically that would amount to copying Rise of Nations, which has the most detailed end-game stats screen.

CrazyEddieTK:
"The benefit is that it is currently bad that midgame stats looks like endgame stats and that "Game in progress - Statistics" is fairly redundant."

Could you expand on this? What would you want to see different between the midgame stats and the endgame stats?

Any thoughts on accepting this PR as is for now to get the new functionality live, with refactoring and clean-up to be done later? This PR doesn't make the code better, but it doesn't make it much worse.

I'll take the #2147 issues and @dvr333 's features and work on them with or without the refactoring.

LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -60,6 +66,20 @@ local awardDescs = VFS.Include("LuaRules/Configs/award_names.lua")
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
+
+options_path = 'Settings/HUD Panels/Stats Graph'
+options_order = {'togglestatsgraph'}
+options = { 
+	togglestatsgraph = { type = 'button',
+		name = 'Toggle stats graph',
+		desc = 'Shows and hides the statistics graph, which is normally only displayed at the end of the game.',
```
sprunk: >, which is normally only displayed at the end of the game

Implies this action is somehow not normal. Just say "toggles the stats graph" and that's it.

LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -188,10 +206,44 @@ local function ShowEndGameWindow()
...
+			addedStatsSubPanel = false
+		end
+		screen0:RemoveChild(window_endgame)
+	else
+		-- toggle on
+		
+		-- set the togglekey reminder here
+		-- this is dumb but I don't know of a hook for when it's changed in epicmenu
+		-- so this is as good a place as any to make sure it's set correctly
+		local toggleKey = WG.crude.GetHotkey("togglestatsgraph") or "no hotkey"
```
sprunk: I'd only append the hotkey part if there's an actual hotkey.

"Toggle (no hotkey)" -> just "Toggle"

LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -241,8 +293,22 @@ local function SetupControls()
 	}
 	
 	local B_HEIGHT = 40
-	awardButton = Button:New{
+
+	local toggleKey = WG.crude.GetHotkey("togglestatsgraph") or "no hotkey"
```
sprunk: ditto

LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -188,10 +206,44 @@ local function ShowEndGameWindow()
...
+		screen0:AddChild(window_endgame)
+	end
+	showingEndgameWindow = not showingEndgameWindow
+end
+
 local function SetupControls()
 	window_endgame = Window:New{  
 		name = "GameOver",
-		caption = "Game aborted",
+		caption = "Game in progress - Statistics",
```
sprunk: Just "game in progress"? Obviously it's statistics.

LuaUI/Widgets/gui_chili_endgamewindow.lua
CrazyEddieTK: Might be worth showing that the toggle IS hotkeyable even if the user doesn't have it hotkeyed? Without the hotkey the toggle button isn't very helpful, since it can only be toggled OFF and not back on again.

Then again, since it will be hotkeyed by default, if someone removes the hotkey we could presume they know what they're doing and leave it on them to put it back if they want to be able to use the feature.

sprunk: >Might be worth showing that the toggle IS hotkeyable even if the user doesn't have it hotkeyed?

I think if something is useful enough to have a not-hotkey reminder it's useful enough to have a default hotkey.

GoogleFrog:
> "The benefit is that it is currently bad that midgame stats looks like endgame stats and that "Game in progress - Statistics" is fairly redundant."

> Could you expand on this? What would you want to see different between the midgame stats and the endgame stats?

I want the midgame and endgame stats UIs to look so different that there is no question as to whether you are looking at the endgame stats. This would make "Game is progress" redundant. I want this because people are going to interact differently with each type of stats so there shouldn't be any confusion as to what type of stats they are looking at.

For example the midgame stats should be toggleable, the endgame stats should not. If the midgame stats are displayed and the game then ends it should be really obvious that the game just ended. Changing one bit of text on the UI does not make this transition highly obvious.

Midgame stats could be in a smaller, offset window with thinner borders than endgame.

> Any thoughts on accepting this PR as is for now to get the new functionality live, with refactoring and clean-up to be done later? This PR doesn't make the code better, but it doesn't make it much worse.

I don't expect that pulling this PR now will make it live particularly soon since I don't expect stable for a while.

CrazyEddieTK:
Okay, thanks. Makes sense. FWIW with this PR the transition from mid-game to end-game - should the user happen to have the midgame stats screen up when the game ends - has the following effects:

 - The caption text changes
 - The screen switches to display the awards instead of the stats
 - The screen repositions to the normal large window in the middle, if the user happened to have moved or resized it while viewing stats midgame

LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -222,10 +229,11 @@ local function ToggleStatsGraph()
 		-- set the togglekey reminder here
 		-- this is dumb but I don't know of a hook for when it's changed in epicmenu
 		-- so this is as good a place as any to make sure it's set correctly
-		local toggleKey = WG.crude.GetHotkey("togglestatsgraph") or "no hotkey"
-		if toggleKey == "" then toggleKey = "no hotkey" end
-		toggleButton.caption="Toggle ("..toggleKey..")"
-		toggleButton:Invalidate()
+		if toggleButton then
+			local toggleKey = WG.crude.GetHotkey("togglestatsgraph") or ""
+			toggleButton.caption="Toggle ("..toggleKey..")"
```
sprunk: Nitpick: I'd do this as
`local toggleKey = WG.crude.GetHotkey("togglestatsgraph") or ""
if toggleKey ~= "" then
  toggleKey = " (" .. toggleKey .. ")"
end
toggleButton.caption="Toggle" .. toggleKey
`

Having `Toggle ()` looks odd, was there supposed to be something in the bracket?

CrazyEddieTK: Thanks for the review. All is still WIP, much clean-up and fixing to be done. I was partway through implementing a decent feature and had to go to bed.

The ultimate intention is that if the screen can't be toggled (because no hotkey is defined), then the toggle button won't show at all. Still working on getting it right.

CrazyEddieTK:
Still needs work, but it's working well enough that @dvr333 could load it as a local widget and use it in today's tournament casts. You'll need both gui_chili_endgamewindow.lua and gui_chili_endgraph.lua . Bind to a hotkey via Settings/HUD Panels/Stats Graph .

GoogleFrog:
I still dislike that the gui_chili_endgamewindow.lua widget is being used to display the midgame stats.

CrazyEddieTK:
I'm submitting this PR as-is-now for review and approval.

@sprunk -

> - Just say "toggles the stats graph"
> - Just "game in progress"

Done.

> - I'd only append the hotkey part if there's an actual hotkey.
> - Possibly say "Close" on lack of hotkey?

What I ended up doing is preventing the toggle button from appearing entirely if there's no hotkey bound, or if it's the endgame and endgame toggling is not enabled. In other words, if the user can't summon the window back, I've prevented them from having a button that would dismiss it, lest they accidentally click on it and wonder how to bring it back.

@GoogleFrog -

> - I hate the code quality of endgamewindow and don't want to look at it further.

I've cleaned up endgamewindow.lua substantially. The code is better organized, pointless flag-passing and spaghetti code was removed, the logic is simpler and flows better.

> - The stats widget structure is pretty bad and it doesn't get better in this PR. [..] quarantine bad code

The interaction between endgamewindow.lua and endgraph.lua is now clean and simple as per your suggested changes:

> - Remove all loadpanel calls from endgaph.
> - Remove WG.statsPanel.
> - Make loadpanel return window0 and add loadpanel to WG in initialize.

Done.

> - Make loadpanel reuse window0 on subsequent calls. Some cleanup may be required, perhaps orphan window0.

Not done. With my changes, endgamewindow.lua only calls that global function once, so there's no need for it to have to reuse existing objects on subsequent calls since there will never be a subsequent call. And on the other hand, if multiple widgets want to do things with a stats window, they should each have their own instances of that window so that the things they do don't interfere with each other.

> - the midgame stats should be toggleable, the endgame stats should not.

This is the case by default. However, there is an option at Settings/HUD Panels/Stats Graph to allow toggling the endgame stats as well, so that (for example) casters can quickly hide the stats in order to display and review the battlefield (as @dvr333 usually does) without having to drag the window off to the side (as @dvr333 usually does) and can then quickly bring it back in order to review the stats (as @dvr333 usually does).

> - If the midgame stats are displayed and the game then ends it should be really obvious that the game just ended.

With my changes, if the screen is already displayed when the endgame arrives:
 - The screen changes from the stats display to the awards display
 - The caption changes
 - If the screen had been repositioned or resized, it is returned to the middle of the screen and returned to its original size

With my reworking of endgamewindow.lua it would be very easy to create even larger visual differences, but I'd need some further suggestions as to what those differences might be. I posit that switching from stats to awards is a pretty big change all in itself.

> - I still dislike that the gui_chili_endgamewindow.lua widget is being used to display the midgame stats.

If there were a separate widget just for midgame stats, there would be a substantial amount of logic to be duplicated between them. The common functionality is large enough that I don't think it makes much sense to split them up. Instead, the difference between midgame and endgame screens is largely one of configuration, and with my changes it's easy to create two very different configurations within the same widget and keep them cleanly separated from the shared logic.

> - If you're delving deep into endgraph consider these issues: #2147

I've looked into those issues and I have solutions for them. However, I'll address them separately in that ticket rather than folding them into this PR.

Likewise, I'll work on the features @dvr333 requested in separate branches rather than adding them to this PR. I'd like to get this one considered and accepted before building more features on top of it. My basic goal was to make the stats screen visible during the game, and I've done that here, along with a few cosmetic improvements (such as live auto-update and persistent button activation) and a good bit of code clean-up.

Thanks for your consideration.

GoogleFrog:
Some issues:
 * The caption should be removed for midgame stats. There should be nothing saying "Game in progression - statics" at the top of the window as it just takes up space and the contents of the window is obvious.
 * Manually making the window small can create a scrollbar on the panel which can be used to hide the graph. Seems to only occur before 15 seconds.
 * The position and size of the midgame stats window is not remembered between games. The midgame stats window needs to be a different window and to have docking enabled to implement this cleanly.
 * The playerlist has gaps in it for enemy players. It should reduce its own size.
 * The graph selection buttons take up too much space. The playerlist takes up too much space. Fonts are generally too large for something which is to be used during a game instead of for analysis at the end.
 * Toggle (Alt + F1) should be replaced with a button on the global commands bar. See the attrition counter widget for how to add such a button.

Scrollbar issue:
screen00398

I think this should be the end goal:
midgamestats
This window could be dragged even smaller which would allow you to check stats without obscuring the whole game.

I'll have time to have another look at it in a few days.

GoogleFrog:
The end goal doesn't need to be realized in this PR. There is useful cleaning up etc... Fixing the issues may require a restructure though.

GoogleFrog:
@sprunk your merge of the small change made this unmergable.

CrazyEddieTK:
I will fix any conflicts if you're ready to accept this PR. I can make updates later tonight (US time).

GoogleFrog:
Ok, but we've also had a regression with button highlighting.

CrazyEddieTK:
Surprised, but I'll check and fix.

CrazyEddieTK:
Conflicts resolved, ready to merge.

GoogleFrog:
There are conflicts.

CrazyEddieTK:
Conflicts resolved, ready to merge.

GoogleFrog:
I'll revert chicken dragon myself.

CrazyEddieTK:
Thanks for the merge. I assume the chicken reversion was unrelated?

> - Manually making the window small can create a scrollbar on the panel which can be used to hide the graph. Seems to only occur before 15 seconds.
> - The playerlist has gaps in it for enemy players. It should reduce its own size.

I'll fix these and submit a new PR.

> - The caption should be removed for midgame stats. There should be nothing saying "Game in progression - statics" at the top of the window as it just takes up space and the contents of the window is obvious.
> - The position and size of the midgame stats window is not remembered between games. The midgame stats window needs to be a different window and to have docking enabled to implement this cleanly.
> - The graph selection buttons take up too much space. The playerlist takes up too much space. Fonts are generally too large for something which is to be used during a game instead of for analysis at the end.
> - Toggle (Alt + F1) should be replaced with a button on the global commands bar. See the attrition counter widget for how to add such a button.

I'll open a new issue for a redesign of the midgame stats screen that can incorporate these changes.

I'll work on @dvr333's requested features and make a new PR for them when they're ready.

Issue #2236, opened 2017-05-18
Chili coloring broken for colors containing newline character 


DeinFreund:
This includes one of the default team colors, which is rgb(10, 100, 40). 10 is translated to \n which breaks Chili.

Affected UI:
![img](https://i.imgur.com/uqUcBiS.png)
![i](https://i.imgur.com/6u0lTjB.png)

sprunk:
Colour codes are engine, mantised. https://springrts.com/mantis/view.php?id=5836

GoogleFrog:
This change broke a lot.
![screen00586](https://user-images.githubusercontent.com/1318249/33237140-ee32fc8a-d2be-11e7-86b6-8af511d38bdc.jpg)

sprunk:
2ae299c874345fec3c750730d128116eca81ef04 reverted the fix so this needs to be checked again.

Sortale:
this is broken in some cases

example here:

https://cdn.discordapp.com/attachments/278908415756206080/418254876452061194/screen00000.png

replay

http://zero-k.info/Battles/Detail/488882

sprunk:
Probably fixed be7495852eafb9c02b6921197ed40a6cd402447c

PR #2237, opened 2017-05-18
Workaround #2236

DeinFreund:
As chili problems are usually solved with workarounds.. Of course a proper fix would be appreciated.


sprunk:
Brutal.

PR #2272, opened 2017-06-01
shieldfac remodel

sprunk:
image


GoogleFrog:
You should apply it to a different factory since I already hacked the current shield model into compliance.

sprunk:
I'm not sure if it fits any other fac. Not only visually but also the build area is tiny and anything larger than Felon won't fit.

GoogleFrog:
What about the vehicle factory?

sprunk:
Domi and Impaler won't fit the build area.

Anarchid:
Apparently this is now in need of rebase. I added the original decal and texture 2. I did not succeed in applying the tex2 though because that requires Upspring. Attempting to run Upspring on my machine hangs the desktop environment.

The decal is slightly misaligned. This is fine because it's getting replaced with a standardized decal anyway.

image

Anarchid:
The buildicon generator seems to choke on this s3o: 
image

GoogleFrog:
I might want to make some shieldbots 3x3 footprints.

sprunk:
With a model size increase? Footprint itself is probably okay.

GoogleFrog:
Not with a model size increase. Look at a stack of Thugs, they clip and have trouble firing.

sprunk:
I think footprint won't be a problem.

GoogleFrog:
screen00475

sprunk:
Is this picture demonstrating Thug's footprint being too small? I'm not saying "unit footprints are not a problem and their values are fine" but "sure, increase them, the new factory model should handle increased footprint well".

sprunk:
I checked and the remodeled fac can build 3x3 units without needing any change to accomodate them.

Anarchid:
image

Anarchid:
image

Anarchid:
I think this is pretty done aside from the merge conflict.

Anarchid:
image

Anarchid:
Verified upscaled thugs being mass-produced with no issue after rebasing to master. 
image

GoogleFrog:
Can you make the texture better? It looks comparatively grimdark with the dark streaked grey. Also the edges of inner structure are cut off a little weirdly. Perhaps I need to see it at normal zoom.

Anarchid:
I'll try to make it more rainbows-and-unicorns.

sprunk:
I've fixed the merge conflict.

Anarchid:
This version of the model has too small portal, and the model with the portal scaled to adequately house a felon fails to not be uglier than i am willing to commit.

sprunk:
sad!

GoogleFrog:
Portal? You mean the pad that the unit is built on?

Anarchid:
The whole pad / cover / door assembly.

I feel it's better to attempt a new model instead.

PR #2289, opened 2017-06-04
Fix #1675 - factory decals have OTA logos.

### LINKED ISSUE #1675: All factory ground decals have TA logos

sprunk:
screen00072
screen00073


GoogleFrog:
I think there are too problems:
 * They are too bright, they look like a halo around the factories. Ground decals are supposed to be background.
 * They are the wrong scale. The metal plates are large and have a very visible and large grain. This texture does not fix with the implied scale of anything else (maps, factory models etc..).

sprunk:
Do you think darkening the texture and applying a blur (to get rid of visible grain) would do the trick?

GoogleFrog:
I don't know. The metal panels themselves are quite large and still give the impression of smallness. I also just noticed that the textures have a reflection baked into them which makes them look a bit fake.

GoogleFrog:
It is worth a try though.

sprunk:
facs_take_2

facs_take_2a

mtroyka:
yeah that really doesn't look very good. :|

GoogleFrog:
It is hard to tell without advmodelshading.

Licho1:
It also seems to be missing AO plate shadow

sprunk:
Advmodelshading causes models to be invisible for me.

I don't know how to generate the shadow.

sprunk:
Will add proper decals to #2290

PR #2290, opened 2017-06-04
Tank/Hover/LV/Cloaky fac compliance

sprunk:
hover2
hover1
tankfac


GoogleFrog:
They require the solid_factory custom param to set the open area for prevent lab hax. This is required because enemy units are going to try to path through fully open factories.

sprunk:
`solid_factory` added,

GoogleFrog:
I don't want to apply tank factory before the ground decal is redone. Shield factory is already bad enough.

sprunk:
Added LV:

manglv

sprunk:
Only missing cloaky (in addition to the decal).

sprunk:
Cloaky added, with decal.

screen00106

Now missing just the decals.

GoogleFrog:
Are there any parts of the tank factory geometry that could be used to cover the construction yard? It is better when factory models have a covered construction yard while idle. It makes them look more reasonable as construction ghosts and nanoframes.

GoogleFrog:
I dislike the cloaky decal. I think the decals in your other PR were better. Can you put the source files for those decals in the artwork repo?

sprunk:
The middle part of tankfac (the rotating one with teamcolor) - the full model has three of those, with two on the sides. Should I readd those? They kinda obscure side exits though.

sprunk:
Decals: https://github.com/ZeroK-RTS/Zero-K-Artwork/pull/8

Anarchid:
Alternate cloakyfac decals for discussion. Style cheaply extensible to all decals. 

I think decals should have the build zone visually represesented on all factories for consistency.

image
image

GoogleFrog:
The second decal is better. The three big problems with the cloaky decal were that it wasn't square, was too bright and was too distinct (did not fade into background, bold lines too bold).

Anarchid:
image

Anarchid:
image

CrazyEddieTK:
Cool.

GoogleFrog:
Looks pretty good. I thought you had remodels for those factories lined up? Vehicle factory at least needs improvement.

Anarchid:
Yes, i have meshes in texturing for lv and tank. But these decals are quite easy to make and it may take a while to finish lv and tank. I can make them for the tacs unaffected by this PR as well. I kinda want one for when hover is built on ground, too.

sprunk:
Anything blocking this except the merge conflict?

GoogleFrog:
If tank and veh are near completion then it is worth waiting for them instead of polishing the current kitbash. If cloaky has a decal updated since the last screenshot then it may be complete. Also hover should be able to get a decal. Perhaps make a PR for cloaky and hover separately?

Anarchid:
> If tank and veh are near completion

Both are unwrapped but more or less untextured. Tank texturing is unpredictable because it's a new person. Rover texturing is on hold due to requirement to adjust Shield. So tank and rover are at least two weeks away.

>  If cloaky has a decal updated since the last screenshot then it may be complete.

Anything to do there besides adjust contrast? (The last screenshot is the one that has rv, tank and cloaky).

GoogleFrog:
I think cloaky just needs a contrast adjustment. The build are should not stand out too much so should be quite similar in colour to the rest of the decal. The rest of the decal could stand to be a bit darker. If you want to be really fancy the build area could have a less stark boundry.

Anarchid:
The build area in case of cloaky is not decal, but model geometry. I can try to find lighter bits of the UV map to assign there, though.

GoogleFrog:
Alternately I could pull this now and improvements could be supplied later.

Issue #2291, opened 2017-06-06
Add unit shield to overkill prevetion


GoogleFrog:
If a unit is holding a shield then add the shield to its health for the purpose of overkill prevention. I thought this was implemented a while ago but it turns out that the overkill prevention gadget does not mention shields. This should solve the edge case where some Scalpels can incorrectly hold their fire against an army of Thugs.

PR #2293, opened 2017-06-07
Better shield handling in OKP

lhog:
https://github.com/ZeroK-RTS/Zero-K/issues/2291
Seems to work....

Based on my old OKP code with shieldballs handling, which I tested thoroughly:
https://github.com/ZeroK-RTS/Zero-K/blob/84e18b21fd28f26418f35ace77968dd6007a5d81/LuaRules/Gadgets/unit_overkill_prevention.lua#L214-L276

### DIFF
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -246,9 +246,9 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
 
 				local expectedShieldPower
 
-				expectedShieldPower = math.min(shieldPower + (frame - gameFrame) * shieldRegenDef[unitDefID], shieldPowerDef[unitDefID])
+				expectedShieldPower = math.min(shieldPower + (frame - gameFrame) * shieldRegenDef[unitDefID], shieldPowerDef[unitDefID] or 0)
 				if disarmShieldDamage > expectedShieldPower then --disarming damage has passed through the shield
-					local disarmExtra = math.floor(disarmDamage/adjHealth*DECAY_FRAMES)
+					disarmExtra = math.floor(disarmDamage/adjHealth*DECAY_FRAMES)
 				else --otherwise the damage is absorbed
 					shieldPower = shieldPower - disarmShieldDamage --this can go below 0, but it's virtual, since calculus is done @frame (future)
 				end
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -37,6 +37,7 @@ local spGetUnitShieldState  = Spring.GetUnitShieldState
 local pmap = VFS.Include("LuaRules/Utilities/pmap.lua")
 
 local DECAY_FRAMES = 1200 -- time in frames it takes to decay 100% para to 0 (taken from unit_boolean_disable.lua)
+local DISARM_DAMAGE_MOD = 1/3 -- from armordefs.lua
 
 local FAST_SPEED = 5.5*30 -- Speed which is considered fast.
 local fastUnitDefs = {}
@@ -80,7 +81,7 @@ local HandledUnitDefIDs = {
 	[UnitDefNames["shiptorpraider"].id] = 1,
 	[UnitDefNames["shipskirm"].id] = 1,
 	[UnitDefNames["subraider"].id] = 1,
-	
+
 	-- Static only OKP below
 	[UnitDefNames["amphfloater"].id] = 1,
 	[UnitDefNames["vehheavyarty"].id] = 1,
@@ -92,7 +93,7 @@ local HandledUnitDefIDs = {
 	[UnitDefNames["striderdetriment"].id] = 1,
 	[UnitDefNames["shipassault"].id] = 1,
 	[UnitDefNames["shiparty"].id] = 1,
-	
+
 	-- Needs LUS
 	--[UnitDefNames["tankassault"].id] = 1,
 	--[UnitDefNames["vehassault"].id] = 1,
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -233,39 +239,60 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
...
+
+				--in rare case disarm and regular damages expected same frame, then disarm is applied first (arbitrary convention)
+				local disarmExtra = 0
+				local disarmShieldDamage = disarmDamage * DISARM_DAMAGE_MOD
+
+				local expectedShieldPower
+
+				expectedShieldPower = math.min(shieldPower + (frame - gameFrame) * shieldRegenDef[unitDefID], shieldPowerDef[unitDefID])
+				if disarmShieldDamage > expectedShieldPower then --disarming damage has passed through the shield
+					local disarmExtra = math.floor(disarmDamage/adjHealth*DECAY_FRAMES)
```
GoogleFrog: Nullop line.

LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -233,39 +239,60 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
...
-				adjHealth = adjHealth - fullDamage
-				
+
+				--in rare case disarm and regular damages expected same frame, then disarm is applied first (arbitrary convention)
+				local disarmExtra = 0
+				local disarmShieldDamage = disarmDamage * DISARM_DAMAGE_MOD
+
+				local expectedShieldPower
+
+				expectedShieldPower = math.min(shieldPower + (frame - gameFrame) * shieldRegenDef[unitDefID], shieldPowerDef[unitDefID])
```
GoogleFrog: Why is shieldRegenDef[unitDefID] necessarily not nil?

LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -233,39 +239,60 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
...
+				if disarmFrame > frame + DECAY_FRAMES + disarmTimeout then
+					disarmFrame = frame + DECAY_FRAMES + disarmTimeout
+				end
+
+				if disarmFrame - gameFrame >= DECAY_FRAMES then --shot above will disable shields, don't bother tracking when the shielded unit will become armed again
+					shieldPower = 0
+				end
+
+				expectedShieldPower = math.min(shieldPower + (frame - gameFrame) * shieldRegenDef[unitDefID], shieldPowerDef[unitDefID])
+				if fullDamage > expectedShieldPower then --regular damage has passed through the shield
```
GoogleFrog: What about salvo units?

lhog: Thanks for review.

1 & 2 are fixed in next commit.
Salvo is ignored at the moment, only cumulative salvo damage is sent to GG.OverkillPrevention_CheckBlock. If you want to get down to salvo level, it's doable but will require a slight change in GG.OverkillPrevention_CheckBlock parameters. Should I do it?

GoogleFrog: This complication is why I ignored the fact that shields can be penetrated by high damage projectiles. I don't think much is lost in this approximation and you get much simpler code.

GoogleFrog:
No update on dealing with salvo units?

lhog:
I read ```I don't think much is lost in this approximation and you get much simpler code.``` as indication that you were satisfied with the status quo code and would not like to continue with my implementation.

Did I misunderstand?

GoogleFrog:
I wasn't sure but now I think I would prefer there to be no change. My priorities here are:

1. Units should not be blocked from firing when firing might be useful.
2. Maintainability.
3. Performance.

The blocking that occurs is subject to these constraints. This PR increases the occurrence of shot blocking since the current approximation never says that less shots are required than the more exact shield handling in this PR.

I've looked back over the code and assumptions and I think your shield handling assumes incorrect things about the incoming data. Here are the false assumptions (ignoring disarm for the moment):
 * That the third argument to GG.OverkillPrevention_CheckBlock is the damage of the shot.
 * That the fourth argument to GG.OverkillPrevention_CheckBlock is the frame at which the shot will hit.

These two arguments to GG.OverkillPrevention_CheckBlock were never meant to correspond exactly to shot behaviour. I used weaker assumptions to leave room for behaviour tweaking. For example Slasher tells OKP that it deals 30 damage when it in fact deals 40 damage because repair caused issues with such low damage. Some units report reduced damage when they fire at a fast unit in order because a bit of overkill can be beneficial. The fourth argument is often set to about 80% of maximum flight time because it rarely matters that an additional shot is fired at the end of a battle.

The weak assumptions in GG.OverkillPrevention_CheckBlock simplify the code allow me to reduce the reported damage and flight time of a unit if it were found to be violating Priority 1, without much fear of consequences. However, the weak assumptions would really hamper your shield collision code, possibly to the point where you may as well have the approximation. When you add projectile count to the code do you just assume that each deals damage/projectiles? Do you rewrite large parts of OKP with stronger assumptions and then redo the configuration in each LUS? I think we've reached the point where changes would be large and introduce new bugs to fix.

Perhaps I should just say that OKP changes need grounding. I want to see a real example where OKP failed against shields in a way that was both important avoidable by code such as this PR. I added the shield health approximation because I saw Scalpels not firing enough shots against a small group of Thugs when they should have been able to kill a Thug.

lhog:
```I want to see a real example where OKP failed against shields in a way that was both important avoidable by code such as this PR.```

Actually I agree. Let's see if there will be failures with current approach and then get back to this.

PR #2304, opened 2017-06-11
Improved docs and added api status to win counter

dvr333:


dvr333:
I'm not entirely sure how to test this, but it doesn't seem to have disappeared from the widget list as it should have.

sprunk:
`hidden = true`. As far as I know `api = true` makes it load before non-api widgets.

PR #2308, opened 2017-06-11
Djinn: SPECIAL → UTILITY

sprunk:
Same as Aspis and Eraser. The main worry here would be that it makes the icon disjoint from the others.

A similar thing to consider would be Hawk: WEIRD RAIDER → ANTI AIR. All other planes fit decently into the slot roles but again this would disrupt the little Fighter vs Bomber grouping.

### DIFF
--- a/LuaUI/Configs/integral_menu_commands.lua
+++ b/LuaUI/Configs/integral_menu_commands.lua
@@ -148,7 +148,7 @@ local units = {
 		-- No Amph Artillery
 		amphaa = ANTI_AIR,
 		amphassault = HEAVY_SOMETHING,
-		amphtele = SPECIAL,
+		amphtele = UTILITY,
 	},
 	factoryship = {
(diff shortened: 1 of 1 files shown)


CrazyEddieTK:
I think role-based grid locations is most important for builder/raider/skirm/riot/assault/arty/aa. All the things that don't fit into those roles don't really fit neatly into any role, they're just sort-of shoehorned into these other oddball categories.

I think the visual aesthetic of not having an isolated icon outweighs the advantage of having Djinn be in the same category as Aspis and Eraser.

GoogleFrog:
I agree with @CrazyEddieTK. In the case of Hawk you could say that S means a specialized/weird version of W. There is an amphibious bomb model that may be worth adding (perhaps high HP, large AoE with high cost and no cloaking?) if you want to move Djinn.

PR #2335, opened 2017-06-24
OD grid colours improved

sprunk:
* raw mexes without any energy: pink → blue. A case of "efficient E usage, add more energy" so gets the same color as low OD.
 * raw energy without any mexes: pink → red. An extreme case of "inefficient E usage, add more mexes" so gets the high OD colour.
 * low OD: purple → blue. Differentiate more against red.

 * low OD colour no longer constant below ~3 energy per standard mex
 * backend: discard HSL in favour of simpler formulas (colours stay broadly the same)

### DIFF
--- a/LuaRules/Gadgets/unit_mex_overdrive.lua
+++ b/LuaRules/Gadgets/unit_mex_overdrive.lua
@@ -1115,10 +1115,16 @@ function gadget:GameFrame(n)
 
 			--// Calculate Per-Grid Energy
 			local maxGridCapacity = {}
+			local gridHasMex = {}
 			for i = 1, allyTeamData.grids do
 				maxGridCapacity[i] = 0
+				gridHasMex[i] = false
 				if not allyTeamData.nilGrid[i] then
 					for unitID,_ in pairs(allyTeamData.grid[i].pylon) do
+						local pylonData = pylon[allyTeamID][unitID]
+						if pylonData.mex then
+							gridHasMex[i] = true
+						end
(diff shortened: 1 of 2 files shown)


sprunk:
screen00107

GoogleFrog:
I like the distinct colours for grids which lack either energy or mexes. Can mexes with no energy be more purple? Energy without any mexes should not be so red because the two reasons for red here mean very different things. I don't want people to get used to seeing red.

sprunk:
I can see the qualitative difference between energy without mexes and heavy-OD energy (one wastes E, the other makes quite a lot of metal despite the bad ratio) so making energy have a different color sounds good after all (something beyond red? magenta?).

However I'm not sure if there's a qualitative difference between mexes with 0 and 0.1 energy, sure one is linked and the other isn't but both of them would benefit about the same from extra E. What's useful in having the 0e ones marked distinctly? Note that under the PR, one solar boosting one +2 mex results in cyan so in a realistic case raw mexes are already different (blue).

GoogleFrog:
I should have said that I am considering the information people gain from glancing at the colours on their screen. They should not need to look closely at the structure of their grids to gleam some basic information. I see the current system as this:
 * Red: you have built too much energy.
 * Pink: grids may need connecting. 
 * Any other colour: The energy efficiency that this colour implies.
Disconnected mexes are worse than disconnected energy so there is perhaps an argument for giving gridless mexes and energy their own colour. I think it would be overcomplicated though.

Grids with 0/0 energy (mexes) and 0/2 energy (a solar while E stalling) need to be distinct because the solar grid will produce overdrive when you have enough energy. If you are E stalling and have many choices of solar placement you would ideally place the solar next to a gridless mex.

Licho1:
There is never "too much energy"

sprunk:
Makes sense.

PR #2369, opened 2017-07-06
improved support commander walk animation

springraaar:
- improved support commander walk animation
- fix dynamicCommander.lua GetPace and GetScale methods

### DIFF
--- a/scripts/dynamicCommander.lua
+++ b/scripts/dynamicCommander.lua
@@ -55,21 +55,21 @@ for key, value in pairs(levelScale) do
 end
 
 local function CalculatePaceMult()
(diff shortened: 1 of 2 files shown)


GoogleFrog:
There are bugs with this animation that make it worse than the old one and I am disappointing with the attention to detail. A typo causes the shoulders of the commander to spasm. A foot jittered on one of the backswings. It took me about an hour to track down and fix these issues because the code is really annoyingly structured.

The new animation is better than the old one. There is a minor issue in that it looks jittery.

GoogleFrog:
The dynamic commander script fix was good.

springraaar:
A thing to note is that I added the pelvis tilting back and forth, changed the animation speed and made it dynamic, I didn't change the numbers on the poses on purpose.

This is one of those cases where the existing poses were already decent. The speed was what was making it look crappy.

PR #2380, opened 2017-07-09
Quick and lazy fix to the fact that dynamic commanders do not raise from dead in zombie=1 mode

lhog:
…

### DIFF
--- a/LuaRules/Gadgets/unit_zombies.lua
+++ b/LuaRules/Gadgets/unit_zombies.lua
@@ -31,7 +31,7 @@ end
 local modOptions = Spring.GetModOptions()
 
 local getMovetype = Spring.Utilities.getMovetype
-  
+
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_zombies.lua
```diff
@@ -285,13 +298,14 @@ function gadget:UnitCreated(unitID, unitDefID, teamID, builderID)
...
-  
+
 function gadget:UnitFinished(unitID, unitDefID, teamID)
 	UnitFinished(unitID, unitDefID, teamID)
 end
 
 function gadget:FeatureCreated(featureID, allyTeam)
-	local resName, face = spGetFeatureResurrect(featureID)
+	local resName, face = myGetFeatureRessurect(featureID)
+	Spring.Echo(resName)
```
sprunk: Extra echo.

Issue #2381, opened 2017-07-09
Give Hacksaw LUS burst


DeinFreund:
Angler is designed to miss about half of his shots against planes, so OKP causes it to leave all enemy bombers surviving but with half health.

Watch around 10 minutes:
http://zero-k.info/Battles/Detail/462317

DeinFreund:
I think angler missing shots is intentional. It's always been like that and it represents "angling" well, only sometimes do the fish take the bait.

lhog:
Interesting case.

Might be wrong, but it looks to me, that BlockShot is only called once per whole salvo (in the beginning), while it's possible that target gets changed between individual shots (and it's getting changed because units perceived as doomed are de-prioritized and cleared as attacker's current target).

Others may have better ideas, but mine is to make Angler shooting LUS scripted, so that each projectile is individually checked.

I might look into this thing when I have some spare time.

GoogleFrog:
I think do the LUS burst for Hacksaw as well.

lhog:
Ok.

PR #2388, opened 2017-07-11
Hopefully fixed Angler's OKP issue

lhog:
Rewrote Angler's shooting from engine's burst to LUS (similar to Trident). This allows for more precise OKP. See https://github.com/ZeroK-RTS/Zero-K/issues/2381

### DIFF
--- a/scripts/amphaa.lua
+++ b/scripts/amphaa.lua
@@ -7,11 +7,12 @@ local rthigh, rcalf, rfoot, lthigh, lcalf, lfoot = piece('rthigh', 'rcalf', 'rfo
 local lgun, lbarrel1, lbarrel2, rgun, rbarrel1, rbarrel2 = piece('lgun', 'lbarrel1', 'lbarrel2', 'rgun', 'rbarrel1', 'rbarrel2')
 
 
-local firepoints = {
-	rbarrel1,
-	lbarrel1, 
-	rbarrel2,
-	lbarrel2,
+local shot = 0
+local gun = {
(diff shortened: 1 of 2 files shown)


sprunk:
Looks good but is also a balance buff.

@GoogleFrog

GoogleFrog:
This is significant "balance stuff". Angler was intentionally implemented not like Defender/Trident in order to be burst AA like Hacksaw. Now you're looking at an individual Angler destroying 4 Blastwings or small drones. Angler not being entirely baited by Swift. If you make this change for Angler it shows that we're abandoning burst homing missiles so, for consistency, it should also be changed for Hacksaw. This may be ok since neither of those units are particularly useful at the moment. However, it is anti-quants rule because it is a buff to a weakness. Perhaps the weakness is not particularly interesting though.

Also, LUS burst removes the reload bar from the unit, a bar that is important due to the long reload time on each missile. One task would be to make a reload bar for units such as these. Defender and Trident are somewhat ok because their reload is shown on the model so if you zoom in and look you can learn what is going on.

lhog:
```Also, LUS burst removes the reload bar from the unit, a bar that is important due to the long reload time on each missile. One task would be to make a reload bar for units such as these. Defender and Trident are somewhat ok because their reload is shown on the model so if you zoom in and look you can learn what is going on.```

I thought to do same yesterday and took a look into healthbar widget. Damn, it was written so bad, despite made by jK. That said perhaps I will be able to hack LUS reload times in.

PR #2391, opened 2017-07-12
LUS-ed Hacksaw burst, tuned OKP timeout, increased vertical aiming an…

lhog:
…imation speed as it was too slow compared to heading one

https://github.com/ZeroK-RTS/Zero-K/issues/2381

### DIFF
--- a/scripts/turretaaclose.lua
+++ b/scripts/turretaaclose.lua
@@ -1,16 +1,19 @@
 include "constants.lua"
 
 -- Pieces
-local base = piece 'base' 
-local body = piece 'body' 
-local turret = piece 'turret' 
-local launcher1 = piece 'launcher1' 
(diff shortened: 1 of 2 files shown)


sprunk:
Technically the turn rate was a balance buff and unrelated to lusburst, but more importantly it's a reliability thing and as far as I can tell Hacksaw is still fairly bad.

lhog:
I noticed that after rotating in horizontal plane it still took extra unneeded time for hacksaw to start shooting. Think it was caused by slower vertical aiming as when I changed 100 --> 200 the annoying delay disappeared. This change barely changes balance, I think.

Hacksaw missiles still miss Raven's quite often, so in order to *really* change it, missile's tracking should be buffed.

Issue #2375, opened 2017-07-08
Deprecated things in CUS


GoogleFrog:
Occurs on all luarules loads.

[f=-000001] [Modular Comm Info] Warning: Modular Comms API warning: Comm data entry for player Player is empty or in invalid format
[f=-000001] Custom Unit Shaders:, 0(32) : warning C7555: 'varying' is deprecated, use 'in/out' instead
0(40) : warning C7555: 'varying' is deprecated, use 'in/out' instead
0(47) : warning C7533: global function shadow2DProj is deprecated after version 120
0(60) : warning C7533: global variable gl_TexCoord is deprecated after version 120
0(78) : warning C7533: global function textureCube is deprecated after version 120
0(111) : warning C7533: global variable gl_FragColor is deprecated after version 120

[f=-000001] Unsynced LuaRules: finished loading

sprunk:
https://github.com/ZeroK-RTS/CrashReports/issues/1484

sprunk:
ZeroK-RTS/CrashReports#916

PR #2405, opened 2017-07-15
Satisfied GLSL compiler. Should fix https://github.com/ZeroK-RTS/Zero-K/issues/2375

lhog:
Makes me wonder though, why ModelMaterials files are in use and ModelMaterials_103 are not.


ashdnazg:
103 are the old shaders afaik, some engine changes post 103 were done to fix loading.
After this PR, you should notice how loading times are horrid, as all unit models are loaded before game.

lhog:
It is minimally different to what old Material shader was. I don't have it in front of my eyes, but the differences were mostly look like ```modelpath --> model.path```

ashdnazg:
Indeed.
.model.path was redeprecated because accessing .model forced loading the model (to get info such as mins and maxs).

(So probably, it won't affect loading times, but will just crash)

lhog:
I ran a few Circuit vs Circuit games on that branch.
It worked just fine. No crashes or visual degradation.

Perhaps you can suggest/make changes, that you feel necessary, other than just criticizing.

ashdnazg:
yes, I didn't notice that it checks if modelpath isn't nil.
anyway, re. suggestions:
1) Add the line ```Spring.Echo(udef.name, udef.model.type)``` here: https://github.com/ZeroK-RTS/Zero-K/blob/master/ModelMaterials_103/1_normalmapping.lua#L73 and watch the infolog.
2) Change back to using ModelMaterials and not ModelMaterials_103

PR #2407, opened 2017-07-16
Fix #424 and #637 - show vertical lines for UW units

### LINKED ISSUE #424: Better Defense circles needed for amph warfare
Currently when playing with anything amphibian defense ranges drawn can not be relied upon for precise army movement because circles are only drawn above water and not on the ground surface below water.

To clarify:
Build two torpedo launchers in a deep ocean, build them far enough apart to leave a small gap between their range circles. Share them to an enemy player.
Now try to use the defense range drawn for them to move one or multiple Ducks through the gap without taking damage.

Another example is using Bouys and Scallops to kill torpedo launchers from afar without being inside the defense range of other nearby torpedo launcher (or when using Bouys even Defenders).

### LINKED ISSUE #637: Team color nearly invisible in sea
Team game battles are really hard to understand with default settings

http://i.imgur.com/gTCIEs5.jpg

sprunk:
Fixes #424, the lines point to the surface.

Also fixes #637, the lines are team colored. Not the best visuals but we get it for free when fixing 424.

### DIFF
--- a/LuaUI/Widgets/gui_vertLineAid.lua
+++ b/LuaUI/Widgets/gui_vertLineAid.lua
@@ -41,7 +41,7 @@ options = {
 		name = 'Show for enemy underwater',
 		desc = 'Draw a line perpendicular to the surface for enemy submerged units',
 		type = 'radioButton',
-		value = 'never',
+		value = 'always',
 		items = {
 			{key ='always', name='Always'},
 			{key ='radar',  name='In radar, not in sight'},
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Are these visible while in LOS?

sprunk:
There are two settings for enemy units, (only in radar) or (anywhere, ie either radar or los). The PR uses the second.

Licho1:
Meh would be better if radar didnt report any height (for air too, it gives too much info imo)

Skasi:
Hmm looks like this didn't properly close #424 and #637 like it should have.

sprunk:
This is a hanging PR. It won't close the tickets until (unless) it is merged.

Skasi:
Oh, I didn't see the "Open" thingie. My bad!

Issue #2392, opened 2017-07-12
Enable reload UI for LUS burst units


lhog:
@GoogleFrog, as I'm working on the topic, I'd like to discuss how LUS reload bar/UI should look like.

For reload bar I see a few simple options (Angler is used as example):
```
local gun = {
	[0] = {firepoint = rbarrel1, loaded = true, readyToAim = true},
	[1] = {firepoint = lbarrel1, loaded = true, readyToAim = true},
	[2] = {firepoint = rbarrel2, loaded = true, readyToAim = true},
	[3] = {firepoint = lbarrel2, loaded = true, readyToAim = true},
}
```
1) Display reload time for gun[0] projectile of LUS salvo. Cons: shows reload bar while other 3 projectiles might be available for shooting, a bit confusing. This is how it's implemented in my WIP code currently.
2) Display reload time for the projectile shot last. Cons: reload bar completes, when 3 other projectiles are already reloaded. Basically same issue as (1)
3) Display reload time for the projectile shot first. Again almost equal to (1), except not hardwired to be gun[0] projectile.
4) Display sum of reload times of all non-loaded projectiles normalized to 100% when all 4 are loaded. Cons: unit might start shooting even before 100% is reached. Once a shot is made, progress is reset by 25%.

Other ideas include:

1) @sprunk : ```https://i.imgur.com/vxbygHQ.png i thought that it (bar) could be 1 but overlaid. basically if one is at 20% and one is at 70% then 0-20 is one color, 20-70 is another color and 70-100 % is empty``` Requires quite heavy rework of unit_healthbar code and shader.

2) @Anarchid : ```could be a stockpile icon. either vertical lines for each shot, or even a bullet/missile icon + number. doesnt' Screamer do this already?``` Requires some adjustments to when stockpile icon is shown and some heuristic on how large the icon&text should be.

Current WIP code: https://github.com/lhog/Zero-K/commit/da69a22e79c49b4192784f3f5ad34bdc5184094b

![How it looks](https://i.snag.gy/aI9Xov.jpg)

Please discuss.

DeinFreund:
An even more popular unit which this should apply to is defender.

In the case of defender what I'd expect to see is a reload bar as soon as the third shot is fired, until the first shot is ready again. This is because a reload bar implies to me that the unit is currently unable to fire (I don't know of any shooting game where you can fire while reloading..). The disadvantage would be that you never know how many projectiles are actually ready.

[...]

PR #2410, opened 2017-07-16
LUS burst reload bar implementation

lhog:
https://github.com/ZeroK-RTS/Zero-K/issues/2392

Squashed and ready for merge

### DIFF
--- a/LuaUI/Widgets/unit_healthbars.lua
+++ b/LuaUI/Widgets/unit_healthbars.lua
@@ -34,7 +34,7 @@ local featureBarWidth  = 10
 local featureBarAlpha  = 0.6
 
 local drawBarTitles = true
-local drawBarPercentages = true 
+local drawBarPercentages = true
 local titlesAlpha   = 0.3*barAlpha
 
 local drawFullHealthBars = false
@@ -48,7 +48,9 @@ local infoDistance = 700000
 local drawStunnedOverlay = true
 local drawUnitsOnFire    = Spring.GetGameRulesParam("unitsOnFire")
 
-local TELEPORT_CHARGE_NEEDED = Spring.GetGameRulesParam("pw_teleport_time") or 30*60
+local gameSpeed = Game.gameSpeed
+
+local TELEPORT_CHARGE_NEEDED = Spring.GetGameRulesParam("pw_teleport_time") or gameSpeed*60
 
 --// this table is used to shows the hp of perimeter defence, and filter it for default wreckages
 local walls = {dragonsteeth=true,dragonsteeth_core=true,fortification=true,fortification_core=true,spike=true,floatingteeth=true,floatingteeth_core=true,spike=true}
@@ -95,11 +97,11 @@ end
 --------------------------------------------------------------------------------
 -- OPTIONS
 --------------------------------------------------------------------------------
-local function OptionsChanged() 
+local function OptionsChanged()
 	drawFeatureHealth = options.drawFeatureHealth.value
 	drawBarPercentages = options.drawBarPercentages.value
 	debugMode = options.debugMode.value
-end 
+end
 
 options_path = 'Settings/Interface/Healthbars'
 options_order = { 'showhealthbars', 'drawFeatureHealth', 'drawBarPercentages', 'debugMode', 'minReloadTime'}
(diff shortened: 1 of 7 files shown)


scripts/scriptReload.lua
```diff
@@ -5,31 +5,30 @@ local extenalFunctions = {}
...
+
+	if allReloaded then
 		return false
 	end
-	
+
 	local gameFrame = Spring.GetGameFrame()
-	local reloadFrame = gameFrame + reloadTime - highestProgress
+	local reloadFrame = gameFrame + reloadTime - maxReloadFrame
 	Spring.SetUnitRulesParam(unitID, "scriptReloadFrame", reloadFrame, ALLY_ACCESS)
```
GoogleFrog: You should be able to avoid calling Spring.SetUnitRulesParam most of the time by checking reloadFrame against the last reload frame set. In an un-slowed unit the reloadFrame should only update when a missile files.

LuaUI/Widgets/unit_healthbars.lua
```diff
@@ -902,6 +903,16 @@ do
 				end
 			end
 		end
+		
+		if ci.scriptReload and (ci.scriptReload >= options.minReloadTime.value) then
+			local reloadFrame = GetUnitRulesParam(unitID, "scriptReloadFrame")
+			if reloadFrame and reloadFrame > gameFrame then
+				reload = 1 - ((reloadFrame - gameFrame)/30) / ci.scriptReload
+				if (reload >= 0) then
+					AddBar(messages.reload, reload,"reload",(fullText and floor(reload*100) .. '%') or '')
```
GoogleFrog: Instead of having a percentage message how about a message which shows how many missiles are loaded? So it would just say, for example, '1/3'.

scripts/scriptReload.lua
```diff
@@ -39,19 +38,34 @@ local function UpdateReloadTime()
...
-	reloads[index] = false
+function extenalFunctions.GunLoaded(gunNum)
+	reloads[gunNum] = false
 	UpdateReloadTime()
 end
 
+--sleepDuration in (ms), must be multiple of 100
+function extenalFunctions.SleepAndUpdateReload(gunNum, adjustedDuration, sleepDuration)
+	local adjustedTimer = 0
+	local adjustedDuration0 = adjustedDuration
```
GoogleFrog: This various name is not at all descriptive.

scripts/scriptReload.lua
lhog: You mean if reloadMult has been equal to 1.0 (unslowed, unstunned) then reloadFrame doesn't need update. Right? If so, good, makes sense.

LuaUI/Widgets/unit_healthbars.lua
lhog: I had 'n/N' text along with percentage in my former implementation. The visual result was not great as text changes from 1/3 to 3/3 in very short period of time (2x engine's reloadTime). May only be useful if missiles were not fired 'in salvo'.
I'll make it and lets see how it works out.

scripts/scriptReload.lua
lhog: What name is bad exactly? Function? Variable? Which one?
Anyway instead of me guessing, feel free to propose a better one.

LuaUI/Widgets/unit_healthbars.lua
sprunk: I think that's okay, Defender shoots an incomplete salvo comparatively often.

lhog:
I made too many commits here. Let me know if you agree with the code and I will squash them into one before merge.

GoogleFrog:
There should be a smarter way to implement slow and pause reload than to update everything every frame. Perhaps add a UnitRulesParam for paused reload percent which is only set if the reload is paused and tells healthbars to display a particular reload amount.

I think extenalFunctions.SetupScriptReload should take a number of frames because it is misleading to have it take any number of seconds. This is similar to how I think that reload times in unitdefs should have been designed in terms of frames instead of seconds. Putting seconds there means that people can enter values that are not multiplies of 1/30 and get misleading results.

I saw with Angler and Hacksaw that you did not make the units weapon automatically reset which gun is the primary gun. I recall that a fully reloaded Defender will shoot gun 1 regardless of how many guns it shot in its last burst. I like this behaviour and want to at least keep it for Defender because volleys of shots look more coherent when the Defenders shoot from the same edge in unison.

GoogleFrog:
Also why not make SleepAndUpdateReload take frames as well? I like functions which work correctly if given integers and are phrased such that it looks like they should only accept integers.

lhog:
```There should be a smarter way to implement slow and pause reload than to update everything every frame. Perhaps add a UnitRulesParam for paused reload percent which is only set if the reload is paused and tells healthbars to display a particular reload amount.```
Let's see if I can do such shortcut, although it will require another UnitRulesParam call, so may not worth it  for example for slow. Unfortunately I had to make a tradeoff between making only one call to UnitRulesParam in most common case when the reloadMult stays 1.0 throughout the reload cycle and not make UI look jittery otherwise.

```I think extenalFunctions.SetupScriptReload should take a number of frames because it is misleading to have it take any number of seconds. This is similar to how I think that reload times in unitdefs should have been designed in terms of frames instead of seconds. Putting seconds there means that people can enter values that are not multiplies of 1/30 and get misleading results.```
Reload times are in seconds, reload_bursts are also in seconds, so I decided to make it more human readable and less error-prone.
Also all script_burst LUSes are already done. If someone to make yet another unit script bursted, I think we have established an easy to use framework, which is extremely easy to replicate and use

```I saw with Angler and Hacksaw that you did not make the units weapon automatically reset which gun is the primary gun. I recall that a fully reloaded Defender will shoot gun 1 regardless of how many guns it shot in its last burst. I like this behaviour and want to at least keep it for Defender because volleys of shots look more coherent when the Defenders shoot from the same edge in unison.```
Ok, this should be trivial to do.

```Also why not make SleepAndUpdateReload take frames as well? I like functions which work correctly if given integers and are phrased such that it looks like they should only accept integers.```
It's done for convenience mostly to preserve Sleep() semantic, although in SleepAndUpdateReload() actual sleep time will be greater or equal to supplied time. Anyway if it prevents from merging I can switch to frames, in my view it's mostly a matter of taste again.

**Comment by GoogleFrog** [comment]
Does slow cause your reload bar to jitter? Ideally only a zero reload speed would cause that and it would be the only case that requires reload frame setting.

I've come to the view that measuring anything in Spring in (fractional) seconds is wrong because doing so generates errors. Spring has no good handling of fractional reload times or burst rates so numbers which are not multiples of 1/30 will create behaviour (DPS etc..) different to what you would calculate by the numbers. Similarly, why does Sleep take milliseconds when you can only sleep for an integer number of frames? This mismatch between the accepted values and the non-identical values causes people to make changes much smaller or larger than they intended and hinders the calculation of anything. Yes, people think in seconds, but in my experience measuring things in seconds causes errors.

I think jittering will occur if you have a unit with LUS burst rate under your sleep time. This is because the adjusted reload time of the leading gun could be less than the trailing guns at the time that the trailing guns update.

**Comment by lhog** [comment]
With 100ms sleeps, slow could occasionally cause that reload % jumped for example from 38% to 37% and then back to 38%.

With EMP/Disarm the issue was that widget relied on reloadFrame - gameFrame, with gameFrame being increased every frame, while in LUS reloadFrame increased by 3 frames every 100ms (while unit is disarmed, disabled). This difference was enough to cause reload progress to tremble back and forth.

You seem to be adamant about frame based handling. Fine then I'll change to frames.

File: scripts/amphaa.lua [hunk 4]
```diff
@@ -303,7 +303,7 @@ function script.StopMoving()
 end
 
 function script.Create()
-	scriptReload.SetupScriptReload(4, 12)
+	scriptReload.SetupScriptReload(4, 12 * 30)
```
sprunk: `30` magic constant -> `Game.gameSpeed`

lhog:
I've done what was promised. However I'm less than happy scriptReloadPercentage thing. The use of it reduces the number of calls to SetUnitRulesParam, but the code reliability and readability suffer.

scripts/amphaa.lua
lhog: That's also why I also really disliked seconds to frames change. Now I need to insert boilerplate everywhere to replace `magic constant`

scripts/amphaa.lua
```diff
@@ -340,26 +343,22 @@ function script.QueryWeapon(num)
...
 	gun[num].readyToAim = false
 
-	local adjustedDuration = 0
-	while adjustedDuration < 9 do --unlock aiming
-		local stunnedOrInbuild = Spring.GetUnitIsStunned(unitID)
-		local reloadMult = (stunnedOrInbuild and 0) or (Spring.GetUnitRulesParam(unitID, "totalReloadSpeedChange") or 1)
-		adjustedDuration = adjustedDuration + reloadMult
-		Sleep(1000)
-	end
+	SleepAndUpdateReload(num, 9 * 30)
```
sprunk: 30 magic constant -> `Game.gameSpeed`

GoogleFrog:
Looks good. Sorry about the complication, I recalled that slow had no jittering issues but this is apparently because reload time is updated by the attributes gadget and read by healthbars. I thought the same system could be applied to script reload.

lhog:
Ready for merge?
If so, will you squash commits or I can do it by the end of my business day.

GoogleFrog:
Looks ready for merge.

lhog:
Squashed and ready for merge

Issue #2420, opened 2017-07-20
Shader errors


sprunk:
https://github.com/ZeroK-RTS/CrashReports/issues/1543
https://github.com/ZeroK-RTS/CrashReports/issues/1545

Different to all previous:
```
[f=-000001] Vertex shader failed to compile with the following errors:
ERROR: 1:1: error(#102) #error: Macro begin with GL_ is reserved
ERROR: error(#273) 1 compilation errors.  No code generated


[f=-000001] gfx_deferred_rendering.lua: Bad depth point shader, removing self.
[f=-000001] Vertex shader failed to compile with the following errors:
ERROR: 1:1: error(#102) #error: Macro begin with GL_ is reserved
ERROR: error(#273) 1 compilation errors.  No code generated


[f=-000001] gfx_deferred_rendering.lua: Bad depth beam shader, removing self.
```

PR #2421, opened 2017-07-20
Maybe fix to: https://github.com/ZeroK-RTS/Zero-K/issues/2420

lhog:
Let's see if enforcing version will fix the bug:
https://github.com/ZeroK-RTS/Zero-K/issues/2420


LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -289,6 +289,7 @@ function widget:Initialize()
 
 			depthPointShader = depthPointShader or glCreateShader({
 				defines = {
+				    "#version 120\n",
 					"#define BEAM_LIGHT 0\n",
 					"#define GL_CLIP_CONTROL " .. (Platform.glSupportClipSpaceControl and 1 or 0) .. "\n"
```
sprunk: The error says `Macro begin with GL_ is reserved` and this is still defined.

lhog: This could be an issue as well. Thx @sprunk

PR #2423, opened 2017-07-20
Claymore autorepair

tnodine:
Grenade jumps are fun; faster autorepair lets claymores grenade jump over ridges without permanently crippling themselves.

With these numbers, full recovery from a grenade jump should take between 1:00 and 1:40.


sprunk:
Grenade jumps have no good UI and rely on wonky engine physics. I'm don't think it's a good idea to encourage them.

sprunk:
Permanent(ish) damage also sounds like a fair price for skipping a cliff considering the unit is designed with flat area in mind. Puppy has a similar mechanic (though I think better implemented, in that its jump is a reliable, targetable command).

GoogleFrog:
This PR is similar to this one: https://github.com/ZeroK-RTS/Zero-K/pull/1715

This PR has more novelty in it though so it is a bit better. Even assuming we already had all the balance discussion I dislike your numbers. 200 frames is not a round number of seconds and 11 is a weird number.

Aquanim:
I don't think I've ever seen somebody deliberately grenade-jump with a Claymore (probably a combination of suitable cliffs being rare and unreliability). Claymore could stand to be better though and this is a direction at least worth contemplating.

tnodine:
@Googlefrog:
It looks like I used the wrong framerate for this, 10fps; my bad. The numbers were (should have been) chosen to hit the time targets: 1:00 to fully recover from being near the edge of its own shot and 1:40 to recover from a dead center hit. Factoring out the startup time, that means it recovers ~450 hp in the last 40 seconds, for 11-12 hp per second. 

@Aquanim 
I think it's rare only because it's too expensive in terms of HP - until recently a Claymore could only perform one ever, and if you bring a worker to repair it you might as well terraform a ramp instead - and because Claymores are rare. Claymores' new mechanics make land grenade jumps easier to set up.

@all 
 Even without deliberate grenade jumping, you should still consider giving Claymores faster than normal autorepair to compensate for their accidental self damage.

sprunk:
Is accidental self-damage a problem now that there is some time to escape (https://github.com/ZeroK-RTS/Zero-K/commit/5cdc4c1b2fde5ad4630d4bfb6d6cad1343713a01)?

Licho1:
Making friendly claymores a bit resistant to claymore explosions would be cleaner imo.

Anarchid:
There is exactly one precedent for a weapon that only damages enemies.

Licho1:
Not saying only, just less damage..

sprunk:
Less damage sounds inelegant, it adds a hidden special damages rule. I think the current delay is a very nice way to reduce suicide because it's 100% "physics" - it doesn't touch numerical stats at all and just avoids damage by allowing the unit to move away. I think if Claymore still regularly damages itself then the delay could be increased.

sprunk:
This doesn't seem to have gathered much interest, self-damage has since become less likely (5cdc4c1b2fde5ad4630d4bfb6d6cad1343713a01), and a year has passed.

Issue #2425, opened 2017-07-22
Excessive load in Mex Placement Handler :: DrawWorldPreUnit()


lhog:
![](https://i.snag.gy/vlraD8.jpg)

Will post fix.

GoogleFrog:
Fixed?

lhog:
I'm on AMD. It may matter.
If you are on Nvidia, please check if you see any improvement on any map with big amount of mexes.

PR #2426, opened 2017-07-22
Fix to https://github.com/ZeroK-RTS/Zero-K/issues/2425

lhog:
The issue was load and unload of "Iron" Texture. Made Texture rendering batched.
Reduces load 4-5 times.


GoogleFrog:
Ideally for changes like this you would have one commit which fixes the whitespace of the file followed by the commit that changes the code.

PR #2442, opened 2017-07-27
Racketeer: remove bad cats

sprunk:


GoogleFrog:
This change is fine, for clarity reasons, but does it actually do have any effect? I recall writing handling for FIXEDWING and GUNSHIP bad target cats into the target priorities gadget because the bad target cats were fully overridden by the priorities gadget. Did something change since then?

lhog:
It certainly has effect, as you can see in https://user-images.githubusercontent.com/1476261/28665783-7cbaec4e-72cd-11e7-8f16-3d2ff913a37d.jpg Crabe has like 1.75 priority and nanoframes like around 40. Still Crabe was fully ignored.

Reading the engine code you *might* get false feeling that badTargetCats are fully overridden:
https://github.com/spring/spring/blob/a076e51ca9cfa8122ea79d54eb742d2a3b469b2b/rts/Game/GameHelper.cpp#L711-L721

But in fact, targets, that are badTargetCats are never evaluated if there are "good" targets in place, regardless of priority you return from gadget:
https://github.com/spring/spring/blob/3ff8f21f556c7c038292e22157616ce976eb9503/rts/Sim/Weapons/Weapon.cpp#L641-L677

Issue #2439, opened 2017-07-27
Racketeers have wrong targetting priorities


lhog:
I watched a few team games yesterday and noticed racketeers had wrong targeting priorities.
I believe some experienced players know that and exploit such behavior. See Sigero spawn defender nanoframes here: http://zero-k.info/Battles/Detail/463696

I've taken a look myself and priorities are indeed not ok:

![screen00001](https://user-images.githubusercontent.com/1476261/28665783-7cbaec4e-72cd-11e7-8f16-3d2ff913a37d.jpg)

lhog:
Also racketeers prefer shooting radar dots instead of targets in LoS. I also saw notorious cases where this happened in real games.

![screen00002](https://user-images.githubusercontent.com/1476261/28666004-4c08559a-72ce-11e7-8cae-4bd5701bd52f.jpg)

mtroyka:
I tried to fix the second issue previously but my patch was shot down. They also ought to have some preference for shooting heavies, but the interaction with disarm/okp target priorities is very complicated. Racket behavior has always been pretty screwy though.

lhog:
The reason for 1st is this line https://github.com/ZeroK-RTS/Zero-K/blob/master/units/shieldarty.lua#L59
Since @GoogleFrog is the author of this line originally in https://github.com/ZeroK-RTS/Zero-K/commit/67ef847ddb66ab3761d68371fa280f2ce47de954 I'll ask him to explain rationale behind.

As for second picture, the behavior is in fact correct. I attached the wrong screenshot. The problem is different. It's with identified "radar dot" units. Racketeers overshoot them, because for overkill calculation purposes disarmFrame is set to -1 for units outside of LoS and it's a wrong assumption. I'll post a partial fix in a minute.

Also I noticed that anticipation of disarm expiration is not precise or broken. Basically target that should get a next disarm shot is not getting it timely. But it's a topic for another ticket.

GoogleFrog:
That was 6 years ago. I think I remember the reason though. Racketeer in 2011 shot non-homing EMP missiles because it was more intended as a way to disable turrets. It would rarely hit mobile land units, hence the bad target cats.

lhog:
![screen00003](https://user-images.githubusercontent.com/1476261/28743087-d0eb14b0-7449-11e7-9ca4-95b0a13c4eeb.jpg)

This is on latest code from https://github.com/ZeroK-RTS/Zero-K

lhog:
[...]

PR #2444, opened 2017-07-27
Fix 2439

lhog:
See discussion here:
https://github.com/ZeroK-RTS/Zero-K/issues/2439

### DIFF
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -204,17 +204,18 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
 			end
 		end
 
-		disarmFrame = spGetUnitRulesParam(targetID, "disarmframe") or -1
-		if disarmFrame == -1 then
-			--no disarm damage on targetID yet(already)
-			disarmFrame = gameFrame
-		end
+		--"or gameframe" means unit has not been hit with disarming shot yet
+		disarmFrame = spGetUnitRulesParam(targetID, "disarmframe") or gameframe
 	else
 		timeout = timeout * (radarMult or 1)
 
 		local ud = UnitDefs[unitDefID]
-		adjHealth = ud.health/ud.armoredMultiple + (shieldPowerDef[unitDefID] or 0)
-		disarmFrame = -1
+
+		-- math.huge means we don't know anything about the (unidentified) radar dot's health. Might be flea, might be terrible space monster, so OKP is effectively off
+		adjHealth = (targetIdentified and (ud.health/ud.armoredMultiple + (shieldPowerDef[unitDefID] or 0))) or math.huge
+
+		-- -math.huge means we don't want to speculate on what (unidentified) radar dot disarm status is, so OKP is effectively off
+		disarmFrame = (targetIdentified and gameFrame) or -math.huge
 	end
 
 	local incData = incomingDamage[targetID]
@@ -242,7 +243,7 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
 				end
 			end
 		end
-	else --new targe
+	else --new target
 		if not addToIncomingDamage then
 			lastShot[unitID] = targetID
 			return false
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -204,17 +204,18 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
...
-			--no disarm damage on targetID yet(already)
-			disarmFrame = gameFrame
-		end
+		--"or gameframe" means unit has not been hit with disarming shot yet
+		disarmFrame = spGetUnitRulesParam(targetID, "disarmframe") or gameframe
 	else
 		timeout = timeout * (radarMult or 1)
 
 		local ud = UnitDefs[unitDefID]
-		adjHealth = ud.health/ud.armoredMultiple + (shieldPowerDef[unitDefID] or 0)
```
GoogleFrog: What are these changes meant to fix? If you want to disable something I would much prefer you use checks instead of setting values to math.huge.

LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -204,17 +204,18 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
 			end
 		end
 
-		disarmFrame = spGetUnitRulesParam(targetID, "disarmframe") or -1
-		if disarmFrame == -1 then
-			--no disarm damage on targetID yet(already)
-			disarmFrame = gameFrame
-		end
+		--"or gameframe" means unit has not been hit with disarming shot yet
+		disarmFrame = spGetUnitRulesParam(targetID, "disarmframe") or gameframe
```
GoogleFrog: What is disarmFrame?

LuaRules/Gadgets/unit_overkill_prevention.lua
lhog: Previously disarmFrame erroneously set to -1 for both unidentified and identified radar dots:
https://github.com/ZeroK-RTS/Zero-K/pull/2444/commits/d36b7e095651b45fa592b3ae1dc22c01ad23e773#diff-5dee28e9aacee9a41318fbd4821647abL217

One change has been made to so that disarmFrame is set to the current gameFrame if target is radar dot of known type. This triggers more conservative shooting, since https://github.com/ZeroK-RTS/Zero-K/pull/2444/commits/d36b7e095651b45fa592b3ae1dc22c01ad23e773#diff-5dee28e9aacee9a41318fbd4821647abR256 now operates with gameFrame  in place of disarmFrame  and not with -1 as it was previously.

Another change is related to the fact that previously for radar dots you always queried unitDef.health and unitDef.armoredMultiple even if unit type was unknown (unidentified radar dots).  Therefore now health is set only if type is known. https://github.com/ZeroK-RTS/Zero-K/pull/2444/commits/d36b7e095651b45fa592b3ae1dc22c01ad23e773#diff-5dee28e9aacee9a41318fbd4821647abR215

Thirdly -1 was just arbitrary number negative number. As it's part of calculations that follow, I decided to make it more clean and put -infinity so everyone, who view the code later, understands that in cases when health or disarm frame are +-infinity the computations with them are singularities.

LuaRules/Gadgets/unit_overkill_prevention.lua
lhog: I don't get this question. You know for sure what "disarmframe" unit param is, so what is your question?

LuaRules/Gadgets/unit_overkill_prevention.lua
GoogleFrog: Negative infinity is a useless number for calculations. I have changed things so that now the calculations are simply not performed if the unit is unidentified.

LuaRules/Gadgets/unit_overkill_prevention.lua
GoogleFrog: I'm convinced that this change is an error. spGetUnitRulesParam(targetID, "disarmframe") is -1 when the unit has no disarm damage. This is because unitRulesParams would only take numbers of strings, this may have changed more recently though. If you want disarmFrame to fallback to gameFrame (which I think you do) then your change to these lines should be reverted.

sprunk: You can pass `nil` nowadays.

PR #2450, opened 2017-07-31
Remove old haxx.

sprunk:


GoogleFrog:
I still want to be able to benchmark back to 103.0 so I don't think this hax is sufficiently old. You could add a engine version check.

PR #2463, opened 2017-08-06
WIP factory remodel

GoogleFrog:


GoogleFrog:
- [ ] Hovercraft collision volume is not above ground.
- [ ] Hovercraft could do with a ground decal when on ground.
- [ ] Cloaky front is not centered.
- [ ] Cloaky clips out the side of cliffs.
- [ ] Tank is somewhat ugly so needs a new model.
- [ ] Veh is really ugly so needs a new model.
- [ ] The ground decal is a bit too dirty and uncentered.

GoogleFrog:
Alternately implemented.

PR #2504, opened 2017-09-02
Icon zoom transition

CrazyEddieTK:
Implements SupCom/PA-style transition between icons and models while zooming.


CrazyEddieTK:
How this works:

Above a certain zoom height, all units are rendered as icons as in the standard strategic view. Below that height, both icons and models are rendered. As you zoom in closer, the icons shrink in size. Below a lower threshold, the icons disappear completely and only models are rendered as in the standard tactical view.

The hotkey can switch between icons only, models only, and the dynamic mode described above. Pressing the hotkey toggles into and between icons only and models only. Double-tapping the hotkey switches back to dynamic mode.

The upper and lower thresholds and the size of the icons at those thresholds is configurable in the options menu (Settings -> Graphics -> Unit Visibility). The defaults I've chosen work well on my rig, which has a very large 4K monitor. I'd like feedback from others on what settings work well at various screen sizes and/or UI preferences.

You can also configure opacity at the upper and lower thresholds, but after experimenting with it I concluded it doesn't really work well, so I left it at 100%. But the controls are there, so users can try tweaking it to see if they like it.

CrazyEddieTK:
Performance considerations:

On my rig (GTX 670, i7-3770K @ 3.50GHz, 8GB RAM, 4K display) using a fresh install of ZK with all the defaults, there was no noticeable performance impact, even watching a 8v9 replay with 1000 units on DSD. Framerate and UI responsiveness remained high and steady.

However, on the same rig using my normal install of ZK (Custom graphics level, lots of extra widgets enabled, lots of settings tweaked) there was a definite performance impact. Without the widget, my performance was marginal, with medium-to-low framerates and occasional stuttering. Enabling the widget made the stuttering worse and more frequent, sometimes leading to outright crashes.

From my limited testing I've provisionally concluded that using this widget on a system that's running well should pose no issues, but using it on a system that's already overtaxed may push it over the edge.

GoogleFrog:
I'll see whether it can be enabled. There was an issue with setting high icon distance and I don't know how thoroughly it was solved. https://springrts.com/mantis/view.php?id=5672

GoogleFrog:
The mode toggle button only toggles between icons and non-icons. There is no way to set it to dynamic. The unit visibility menu is a mess of many widgets so you should not add your options to its root.

CrazyEddieTK:
Two options:

1) Change it to a three-way toggle: dynamic -> icons -> models -> dynamic.

2) Leave the single-tap / double-tap mechanism, and add another button to the menu that resets to dynamic (with no hotkey assigned by default, possibly not even assignable).

I like the double-tap mechanism; I've been using it in the iconheight widget for quite some time. I like it because if you're not in Dynamic, you usually want to toggle between Icons and Models, and then go straight to Dynamic from either state when you're ready to go back to it. The double-tap lets you do that without needing a second hotkey.

But perhaps it's a bad UI decision and should be abandoned.

sprunk:
I dislike double-tap because it is not discoverable.

GoogleFrog:
I didn't discover it.

CrazyEddieTK:
>I'll see whether it can be enabled. There was an issue with setting high icon distance and I don't know how thoroughly it was solved.

I've done some further performance testing. On my rig, with about 500 units in play, using a replay of an actual battle, using a fresh install with all config settings at default:

- Zoomed out, all units in view, all units seen as icons (/disticon 0) -> 70fps
- Zoomed out, all units in view, all units seen as models (/disticon 10000) -> 30fps
- Widget enabled, zoomed in slightly, all units in view, all units seen as models + icons: -> 30fps

So there's a pretty big drop switching from icons to models, although not as big a drop for me as you reported in the mantis ticket. And note that my test used an in-progress battle with LuaUI, LuaRules, LUPS, projectiles, moving units, and so forth.

Another critical point is where with the widget disabled and the icon distance at default (/disticon 151), the units in view have just switched from icons to models. This is the highest elevation where you can mostly see models, and where you can see the largest number of models. At that elevation, comparing with and without the widget enabled:

 - Widget disabled: 60 fps (showing only models)
 - Widget enabled: 50 fps (showing models and scaled icons)

So there's a modest drop, presumably due to the additional overhead of drawing the icons.

Finally, I saw no difference in performance with or without the widget when zoomed all the way out (showing only icons in both cases) or zoomed in fairly close (showing only models in both cases). Both zoomed-out and zoomed-in gave me my maximum FPS of 70.

Summing up the widget's performance penalty:
 - Zoomed out: no penalty
 - Zoomed in slightly: large penalty
 - Zoomed into medium range: small penalty
 - Zoomed close: no penalty

It would be good to collect feedback from other people trying out the widget.

Issue #2528, opened 2017-09-27
Retreat icon: chicken might be bad


sprunk:
https://github.com/ZeroK-RTS/Zero-K/commit/2a4917ef796324aaa10fabfd70f1ae52510fecd3 made retreat icon consistently the yellow chicken.

I think the chicken is a bad icon:
 * in ZK "chicken" generally represents the tower defense faction. This is extra bad because in those games there are two buttons with chicken pictures on the global bar.
 * potentially confusing for non-native speakers, requires the knowledge of the "chicken out" idiom.
 * weird shape. 90 only differs from 60 by the head and a small part of tail, it's somewhat hard to tell these two at a glance.

Anarchid:
Another idea: running man with or without fire or other associated hazards, as seen on the typical "escape" symbols:
![image](https://user-images.githubusercontent.com/3822768/30911009-9b7dc866-a38f-11e7-91f6-949b3eab441a.png)

PR #2529, opened 2017-09-27
Icon proposal for #2528 (retreat icon)

sprunk:
Basically the old icon for the global command, and monocolor versions for the state. #2528


GoogleFrog:
This solves the issue of the global command button looking nothing like the unit state. However, what does red yellow and green mean?

sprunk:
30% and 90% health, respectively (yellow being 60%). I think that's intuitive because these correspond to the colour of the healthbar at given level.

I think colours are preferable to "fill %" that the chicken state icon uses because it's easier to tell them apart at a glance but that is a fairly weak preference so if you'd rather have that I can do it too.

sprunk:
Do you mean the specific shades of the colours? Red, yellow and green match firestates (full saturation, hues 0/60/120 respectively).

The gray (off) approximately matches Transport AI Off but none of the various Off icons match one another and we have fifty shades of gray. For example Repeat Off is 0.27 brightness, OKP Off is 0.45, this new Retreat Off is 0.57 and Transport AI Off off is 0.60.

Issue #1018, opened 2015-08-23
Better naming for Sumo weapon toggle instead of "on / off"


knorker:
Sumo can toggle between attracting units with its beam or repulsing them.
The UI switch for this is labeled "on/off" and uses on/off graphic, which does not represent the function.

KlonZK:
problem seems to be that icons/tooltips are the same for all units and the on/off state for newton type units is the same on/off state as for any other unit.

ashdnazg:
this can be closed

sprunk:
What about GF's remarks from https://github.com/ZeroK-RTS/Zero-K/pull/2560#issuecomment-336775252 ?

ashdnazg:
https://github.com/ZeroK-RTS/Zero-K/commit/c9db22be2f229701abc99ddde77fc5abd1e5ccdf

PR #2560, opened 2017-10-14
Add CMD_PUSH_PULL + icons

ashdnazg:
I've kept the underlying on/off since I didn't want to touch unit scripts, especially when newton still uses bos/cob.

Changes without whitespace: https://github.com/ZeroK-RTS/Zero-K/pull/2560/files?w=1

#1018


GoogleFrog:
Why the colour change though?

ashdnazg:
to make it the same colour of the icon

sprunk:
I think the old colour pair was nice because it was unique (this looks a bit like heatray and mace?), possibly better to recolour the icons?

ashdnazg:
removed colour change

GoogleFrog:
There are a few lingering questions/issues but they are independent enough that I am happy with leaving this PR as is. If the issues are not addressed by the time I have a bit of free time I will pull the PR and fix them myself. Here is a list:
 * The name "Push / Pull (state)" is insufficiently descriptive. When I redid state tooltips my design was to give the command itself a name independent of the current state because the current state now occurs in the parenthesis. I would call the command "Impulse Mode" or "Gravity Mode".
 * Initial unit states seems to be unchanged in this commit, unless the widget gained some powerful parsing abilities. Sumo and Newton should have their On/Off state replaced with this command in initial unit states.
 * I recall that both players and widgets are unable to toggle On/Off for nanoframes, this makes the initial unit states widget a bit annoying. Testing on nanoframes is required. Potentially this gadget needs a workaround which applies the correct on/off to

PR #2585, opened 2017-10-25
Wind/Tidal Generator → Turbine

sprunk:
The reason I propose it is that [English names were explicitly cut to be max 20 characters at some point to fit a constraint](https://github.com/ZeroK-RTS/Zero-K/commit/842ac0a79267898c13350b3dd51e78a1d9994c8f) and I originally had problems localising the name under that constraint (would be 28) so I went with [Turbine](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Configs/lang/units.pl.json#L463) which I think is also a better name overall.

Of course the cost would be huge (likely larger than any of the renames that happened so far).


GoogleFrog:
I did not cut the names to fit some character limit constraint. I made the names short enough to render un-truncated on the construction tooltip, it says so in the commit. Anything that fits on the tooltip is fine by me. The name "WWWWWWWWWWWWWWWWWWWW" is probably not fine even though it has 20 characters.

sprunk:
Okay, though that doesn't change the point - the current name fits just barely and translated names don't.

Anarchid:
I like Turbine because i dislike the slash.

sprunk:
Doesn't seem to have happened.

PR #2590, opened 2017-10-30
Factoryveh remodel WIP

Anarchid:
Supercedes previous.

Todo:

- [x] decal
- [x] tune closing animation
- [x] more teamcolor
- [x] rotate buildee to face front
- [x] footprint
- [x] colvol
- [x] radius
- [x] dae is crap at making functional nanoframes.
- [x] there is a gap between the bottom of the factory and the ground.
- [x] death animation.
- [x] disable animation while stunned or inbuild.


GoogleFrog:
Very little teamcolour is visible from the top. Most of the colour is yellow lights so it looks more like a yellow unit than a green unit.

Impaler is square so the shipping container should be a bit longer to make the unit-pad square. Then units can leave the factory to the South without sticking over the edge of the pad.

Your division of tasks is unusual. I will make the ground decal. You need to rotate the model by 90 degrees. 

screen00530

Anarchid:
I did the turning withotu lenghtening the capsule. Impaler seems to fit reasonably well. 
image

sprunk:
How hard would it be to rescale? Looks odd because it sticks out.

Anarchid:
team: colored 
image

Anarchid:
Also i checked the closing anim after rotating the model back. It seems okay.

Anarchid:
> How hard would it be to rescale? Looks odd because it sticks out.

Scaling will require animation to be redone. Aside from that, easy enough.

GoogleFrog:
I think its worth trying with a square shipping capsule. It looks weird at the moment because most rovers are longer than they are wide. This makes them look misaligned when they are at right angles to a rectangle of similar dimensions. The back end of the capsule just needs extending towards the main body of the factory.

I think I am able to do the rest of the required things (decal, footprint etc..).

Anarchid:
Capsule lengtified to accomodate impalator superiorly
image

GoogleFrog:
screen00531

GoogleFrog:
@Anarchid it is your job to fix the dae nanoframe terribleness.

Anarchid:
Extents work. The coordinate order is spring ingame (y is altitude).

 i guess the issue without hardwired extents is with spring somehow erroneously detecting model's lowest point to be somewhere deep below. 

Possibly because because of conflict between Z-up in blender collada and Y-up in Spring, and extent calculation being done before the space is converted.

PR #2595, opened 2017-10-31
fkeys loadscreen update

The-Yak:
Removed obsolete LOS toggle info


sprunk:
That toggle still exists though, under `Ctrl+L`.

The-Yak:
Now that it is on by default user does not have to toggle it anymore. It's not necessary info.

GoogleFrog:
How did you update it? Are the source files somewhere? Can we make more?

The-Yak:
Source files are in ZK artwork repo under yak loadscreens. .xcf files still have their layers seperated.

The-Yak:
"Can we make more?"

Do you want more? There's already quite a few. With that said I would accept submissions of quality screenshots and at least a general idea of the text desired.

GoogleFrog:
I think the load screens are a great place to teach some of the useful but slightly more obscure features. I'll have to have a look over what we have already and possibly make a thread.

sprunk:
LoS toggle already lives on the global command bar so should be fairly discoverable.

I think the screenshot also has room for improvement as the different areas blend smoothly into each other without a visible border (so not immediately obvious there are just overlay modes) and the heightmap portion looks pretty ugly on the very high slope parts (where contour lines are wrongly vertical). Compare with the old one:
 
allside_windtut

Issue #1562, opened 2016-10-18
com immobilized in rezzed fac model


mojjj:
http://zero-k.info/Battles/Detail/433265

my com was standing in shady water inside a wreck of a hover fac. when rezzing named fac, the com found itself inside the fac hitbox trapped, not able to move. (it then died on site to gunships)

KingRaptor:
As I understand it, fixing this would require making hover wrecks not block, or a gadget to forcibly expel anything stuck in a factory. (The latter would be useful for factory-exit-clogging scenarios as well)

Sgathec:
What's the desired solution? A blocking wreck or a gadget?

Sgathec:
A blocking wreck would solve this particular problem, right?

Sgathec:
A unit also got stuck inside a ship factory in my test.

GoogleFrog:
When a structure is resurrected push all the units within it to the edges. That should fix all these cases.

Sgathec:
Ok, I'll do that.

Sgathec:
What if there's no room around the structure?

Sgathec:
Do units get stuck into structures due to the unpathable ground/water that is beneath them? The purple shape that Alt+B brings up is not the reason.

PR #2598, opened 2017-11-02
A gadget for removing stuck units from resurrected structures

Sgathec:
The units are moved to locations that are very close to the structure. Nearby wrecks and other structures are not taken into account.

The gadget checks north, east, south and west for a free path to a location near the building. That is to say, only four different directions are scanned. The gadget prioritizes the shortest paths.

I don't know how high or low the gadget layer should be. I didn't notice anything layer-related during testing.

This gadget is a response to the issue #1562.

### DIFF
--- a/LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
+++ b/LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
@@ -30,6 +30,7 @@ local spGetGroundNormal = Spring.GetGroundNormal
 local spSetUnitPosition = Spring.SetUnitPosition
 local spGetUnitCollisionVolumeData = Spring.GetUnitCollisionVolumeData
 local spGetUnitCommands = Spring.GetUnitCommands
+local spGetUnitHeight = Spring.GetUnitHeight
 local gameSquareSize = Game.squareSize
 
 local CMD_RESURRECT = CMD.RESURRECT
@@ -37,14 +38,13 @@ local CMD_RESURRECT = CMD.RESURRECT
 local Abs = math.abs
 local Min = math.min
 local Max = math.max
+local SortTable = table.sort
 
 local gapMin = 20
 
 local maxUnitHeightAboveGround = 1
 local maxUnitHeightAboveSea = 5
 local extraMaxWaterDepth = 5
-
-local elmosPerSizeSide = 8
 
 local extraGatherDistance = 125
 
@@ -90,7 +90,7 @@ local function IsPathTraversable(unitDef, startX, startZ, goalX, goalZ)
     if (goalZ - startZ) < 0 then
         stepZ = -stepZ
     end
-    
+
     -- Spring.MarkerAddPoint(goalX, 0, goalZ, "Possible goal")
 
     for x = startX, goalX, stepX do
@@ -104,10 +104,10 @@ local function IsPathTraversable(unitDef, startX, startZ, goalX, goalZ)
             else
                 --Spring.MarkerAddPoint(x, 0, z, "Slope traversable")
                 --if x >= goalX and z >= goalZ then
-                    --Debug("Slope " .. x .. ", " .. z .. " is traversable")
-                    --Debug("submerged = " .. tostring(submerged))
-                    --Debug("depth = " .. tostring(depth))
-                    --Debug("canMoveInWater = " .. tostring(canMoveInWater))
+                --Debug("Slope " .. x .. ", " .. z .. " is traversable")
+                --Debug("submerged = " .. tostring(submerged))
+                --Debug("depth = " .. tostring(depth))
+                --Debug("canMoveInWater = " .. tostring(canMoveInWater))
                 --end
             end
         end
@@ -115,20 +115,14 @@ local function IsPathTraversable(unitDef, startX, startZ, goalX, goalZ)
     return true
 end
 
-local function FindAccessibleSpot(diffs, unitDef, startX, startZ, requiredGap)
+local function FindAccessibleSpot(diffs, unitDef, startX, startZ)
 
     --Debug("StartX = " .. startX)
     --Debug("StartZ = " .. startZ)
+    --Spring.MarkerAddPoint(startX, 0, startZ, "startX = " .. startX .. ", startZ = " .. startZ)
     local failure = false
-    --Debug("requiredGap is " .. requiredGap)
     for i = 1, #diffs do
-        local offset
-
-        if diffs[i].value >= 0 then
-            offset = diffs[i].value + requiredGap
-        else
-            offset = diffs[i].value - requiredGap
-        end
+        local offset = diffs[i].value
 
         local goalX = startX
         local goalZ = startZ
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+local function IsGroundSlopeTraversable(unitDef, x, z)
+    local _, _, _, slope = spGetGroundNormal(x, z)
+    local traversable = unitDef.moveDef.maxSlope >= slope
+    --Spring.MarkerAddPoint(x, 0, z)
+    --Debug("traversable = " .. tostring(traversable) .. " (for \"" .. unitDef.humanName .. "\")")
+    return traversable
+end
+
+local function CanUnitSwim(unitDef)
+    return unitDef.springCategories.sub or unitDef.springCategories.ship
```
sprunk: Relying on categories is fragile. Ideally (cached) movedef would be used.

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+
+local function IsGroundSlopeTraversable(unitDef, x, z)
+    local _, _, _, slope = spGetGroundNormal(x, z)
+    local traversable = unitDef.moveDef.maxSlope >= slope
+    --Spring.MarkerAddPoint(x, 0, z)
+    --Debug("traversable = " .. tostring(traversable) .. " (for \"" .. unitDef.humanName .. "\")")
+    return traversable
+end
+
+local function CanUnitSwim(unitDef)
```
sprunk: Is this needed? Which units return `true` for `CanUnitSwim` but false for `IsUnitShip`? As far as I can tell submarines' movedef type is also `ship`.

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+                            WithinBounds(rectUY, -rectUDef.waterline - uHeight / 2 - extraMaxWaterDepth, 
+                                maxUnitHeightAboveSea, true))) then
+                        -- per the Spring wiki unit height determines if a ship can go over something that is underwater
+
+                        local requiredGap = EstimateRequiredStructureGap(units[i])
+                        --Debug("requiredGap is " .. requiredGap)
+
+                        if rectUX < (createdUnitMinX - requiredGap) or rectUX > (createdUnitMaxX + requiredGap) or
+                        rectUZ < (createdUnitMinZ - requiredGap) or rectUZ > (createdUnitMaxZ + requiredGap) then
+                            return
```
sprunk: This returns from the entire function, so any other units that haven't yet been processed will never be (ie. they will potentially stay stuck). This would make sense if each unit was processed individually in a function (see above).

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+
+local function WithinBounds(value, min, max, inclusive)
+    if inclusive then
+        return value >= min and value <= max
+    else
+        return value > min and value < max
+    end
+end
+
+local function EstimateRequiredStructureGap(unitID)
```
sprunk: As far as I know collision volumes are constant per unitdef in ZK so you can do this once at gadget load and use cached values afterwards.

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+                    def.zsize * elmosPerSizeSide / 2})), gapMin)
+
+    return estimatedGap
+
+end
+
+function gadget:UnitCreated(unitID, unitDefID, unitTeam, builderID)
+
+    local uDef = UnitDefs[unitDefID]
+    if builderID then
```
sprunk: This function has 7 levels of indentation yet is just a pyramid of `if`s. Consider splitting the inside of the loop into a function, and changing each `if X then ... end` into `if not X then return end`:

```
local function HandleUnit(unitID, ...)
  -- lines 232-246
  if not (bigass conditional from line 247-250) then
    return
  end
  -- lines 251-290
end

function gadget:UnitCreated(unitID, unitDefID, unitTeam, builderID)

  local uDef = UnitDefs[unitDefID]
  if not builderID then
    return
  end

  local command = spGetUnitCommands(builderID, 1)[1]
  if not (command and command.id == CMD_RESURRECT and uDef.isBuilding and uDef.name ~= "terraunit") then
    return
  end

  -- lines 204-221

  local units = spGetUnitsInRectangle(...)
  for i = 1, #units do
    if units[i] ~= unitID then
      HandleUnit(units[i], ...)
    end
  end
end
```

KingRaptor:
The license is fine, especially if the widget uses code from something also with the GPL v2 or later license.

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
GoogleFrog: Collision volumes are not constant. Also, I think they have little to do with unit-unit collisions with units which are stuck on the ground..

Sgathec:
@sprunk, thanks for the review. I don't have much time, but I'll fix the return problem.

LuaRules/Gadgets/unit_removal_from_rezzed_structure.lua
```diff
@@ -0,0 +1,362 @@
...
+local Min = math.min
+local Max = math.max
+
+local gapMin = 20
+
+local maxUnitHeightAboveGround = 1
+local maxUnitHeightAboveSea = 5
+local extraMaxWaterDepth = 5
+
+local elmosPerSizeSide = 8
```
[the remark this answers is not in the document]
Sgathec: It's half a footprint square side. "Size" in the variable name refers to the xsize and zsize of a unitDef. I empirically determined xsize and zsize are actually double the respective footprint values. That's what I figured based on your help and tests.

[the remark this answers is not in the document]
sprunk: >It's half a footprint square side.

Yes, that's `Game.squareSize`.

Sgathec:
Is there anything left that I'm required to do before this pull request is accepted?

sprunk:
I think it's good enough (could use a bit of refactoring but that can be done later). @GoogleFrog ?

GoogleFrog:
I'll look later.

Sgathec:
If you have the time, please look into it soon.

sprunk:
Don't worry, if something shows up but you can't do a fix, the PR can hang around and someone else will look at it later. The bug is pretty unimportant.

Sgathec:
In fact, I didn't mean that I don't have the time (but that is also true to an extent). Accepting the pull request would be fairly valuable to me due to an assignment.

Issue #2567, opened 2017-10-18
Startup Selector / COFC error


sprunk:
```
[f=-000001] [cawidgets.lua] Error: Error in Update(): [string "LuaUI/Widgets/camera_cofc.lua"]:1309: attempt to index field 'worldBegin' (a nil value)
[f=-000001] [cawidgets.lua] Error: Removed widget: Startup Info and Selector
```
https://github.com/ZeroK-RTS/CrashReports/issues/2035

sprunk:
@dvr333 could you look at the code? https://github.com/ZeroK-RTS/Zero-K/commit/e241e54caea11cb0791129fecf942f7981b5e4b7

It looks like it wanted the new `worldEnd` to be the previous `worldBegin` but it overrode the entire table first.

PR #2602, opened 2017-11-04
Fixed COFC not zooming to start box when choosing comm correctly

dvr333:


sprunk:
Crashes: https://github.com/ZeroK-RTS/Zero-K/issues/2567

PR #2515, opened 2017-09-14
WIP Tank Factory remodel

sprunk:
- [ ] no decal (@Anarchid)
 - [ ] there are normals but I don't know if they work.
 - [x] ~build area still clips underground. Factories built normally (or rezzed, /given etc.) don't flatten ground - only plop does.~ #2533
 - [ ] the pipe on the side has no collisions because `solid_factory` is 1-dimensional. Minor polish thing.

PR #2610, opened 2017-11-09
WIP Tank Factory remodel.

sprunk:
#2515


GoogleFrog:
This PR was much better than veh.

PR #2618, opened 2017-11-14
LUPS shield implementation.

GoogleFrog:
As far as I'm aware this PR is complete. Someone should look at it. I think it is relatively safe to merge this into stable for release because it impinges on few systems and LUPS is put together quite sanely. A luashader-free fallback exists.

These shields have better performance than engine shields and their visual quality is significantly higher, the old hex pattern was pixelated and the sphere was low on triangles. In terms of visual design I would like a well implemented hex pattern but it is not worth waiting for one to appear, someone can pick up that task later.

screen00573

screen00575

screen00576

oldshields


GoogleFrog:
Merged into temp develop, which will be merged to master.

PR #2645, opened 2017-11-26
Fix #2638 - default team colour scheme

### LINKED ISSUE #2638: Default color scheme not suitable for spectating
1) "default" makes this colors https://i.imgur.com/tK4AO5P.jpg
2) simplified does not alter colos based on current team selection, making all players appear same color when spectating

sprunk:
The early colours are now more vivid and distinct.
The first three colours are now all shades of blue so in an MM teamgame your allies are never green.

screen00014


sprunk:
There has been no feedback for a month.

Issue #2654, opened 2017-12-08
Revisit and reconsider loadscreen music


The-Yak:
Some experiments with the disabled loadscreen music nugget have yielded positive results with latest engine. https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaIntro/Addons/music.lua

"loading makes it choppy towards the end"

This appears to have been fixed. I remember the choppy transitions with fading in the past but can't reproduce anymore.

"also, volume cannot be adjusted"

~~Volume setting from Spring.SetSoundStreamVolume(1) is unreliable, appears to only work on first map load with chobby sometimes. This is also how fadeouts operate, so they only work the first time sometimes as well. When fadeouts don't work the transition is a hard cut, but still no static like before.~~

I have fixed fadeouts to be mostly reliable. Due to the way loadprogess is done there's no perfect solution here, it's reliant on cached map times which requires each map to be loaded once. Otherwise best guess at progress is made which often ends at 80%. Fadeouts now happen (a little) after 75% to compensate. 

Volume setting from Spring.PlaySoundStream's 2nd argument works reliably, but WG.music_volume can not be used, returns nil. Springsettings like snd_volmaster work, so music volume in luaintro can respect master volume. This is what I changed:

> function addon.Initialize()
>	Spring.SetSoundStreamVolume(1)
>	local volume = snd_volmaster
>	local musicfiles = VFS.DirList("sounds/music/loading", "*.ogg")
>	Spring.Echo("musicfiles", #musicfiles)
>	if (#musicfiles > 0) then
>		Spring.PlaySoundStream(musicfiles[ math.random(#musicfiles) ], volume or 0.8)
>		Spring.SetSoundStreamVolume(1)
>	end
>end

The-Yak:
On further testing I think unreliable fadeouts is caused by function addon.DrawLoadScreen() not drawing all the way to 100% always. Fadeout is timed only to happen after 90% progress, sometimes progress bar only gets to 80% and then the game finishes loading.

The-Yak:
Okay I think I've got this, PR time.

The-Yak:
https://github.com/ZeroK-RTS/Zero-K/pull/2655

PR #2655, opened 2017-12-09
Awaken loadscreen music

The-Yak:
Tweak fadeouts, fixed sound level to master volume

### DIFF
--- a/LuaIntro/Addons/music.lua
+++ b/LuaIntro/Addons/music.lua
@@ -8,7 +8,7 @@ if addon.InGetInfo then
 		license = "GPL2",
 		layer   = 0,
 		depend  = {"LoadProgress"},
-		enabled = false, -- loading makes it choppy towards the end; also, volume cannot be adjusted
+		enabled = true, -- loading makes it choppy towards the end; also, volume cannot be adjusted
 	}
 end
 
@@ -18,7 +18,7 @@ function addon.DrawLoadScreen()
 	local loadProgress = SG.GetLoadProgress()
 
 	-- fade in & out music with progress
-	if (loadProgress > 0.9) then
+	if (loadProgress > 0.75) then
 		Spring.SetSoundStreamVolume(1 - ((loadProgress - 0.9) * 10))
 	end
 end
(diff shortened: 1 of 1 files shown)


The-Yak:
https://github.com/ZeroK-RTS/Zero-K/issues/2654

LuaIntro/Addons/music.lua
```diff
@@ -18,7 +18,7 @@ function addon.DrawLoadScreen()
 	local loadProgress = SG.GetLoadProgress()
 
 	-- fade in & out music with progress
-	if (loadProgress > 0.9) then
+	if (loadProgress > 0.75) then
 		Spring.SetSoundStreamVolume(1 - ((loadProgress - 0.9) * 10))
```
sprunk: `1 - ((loadProgress - 0.9) * 10)` is >1 in the [0.75; 0.9) set so this either makes the music louder or does nothing.

LuaIntro/Addons/music.lua
```diff
@@ -30,10 +30,11 @@ end
 
 function addon.Initialize()
 	Spring.SetSoundStreamVolume(1)
+	local volume = snd_volmaster
```
sprunk: Are you sure this line does anything (ie. `snd_volmaster` is ever non-nil)? I don't think springsettings are exposed as globals that way.

As far as I can tell, setting `snd_volmaster` in Chobby will successfully lower the volume here but not because of this code but because `snd_volmaster` is already controls all the volumes.

The-Yak: Indeed it's nil, looks like this isn't needed

LuaIntro/Addons/music.lua
The-Yak: 0.75 results in a smoother auditory transition between loading track and briefing. With 0.9 there's a sharper cutoff. It definitely doesn't make the music louder.

The-Yak: Audio from luaintro always stops at "command console activated". Not sure how to keep the music playing ingame.

sprunk:
I think the fadeout change might've been okay.

PR #2720, opened 2018-01-13
Lups shields texture v1

lhog:
Here is an attempt to put hexagon texture onto the shield sphere and make shields look a bit better. This PR has two base methods to chose from:

First one mixes RGB:
https://github.com/ZeroK-RTS/Zero-K/commit/30b332c78bc70117f9b8e1a50e0224e359dd74a1
Method#1

Second one mixes alpha:
https://github.com/ZeroK-RTS/Zero-K/commit/26be3652cd52b9eeb3b29fa53047fc3e008c899d
Method#2

Second method seems to look better on dark maps, while first gives more consistent look on both dark and light maps.

So I'll probably recommend to go with slightly modified 1st version:
https://github.com/lhog/Zero-K/commit/1d3a6492b8676c33b6e100e66147005ba5e5d239
Method#1*

Also the perceived shield translucency area  around the unit body could be manipulated by applying ```pow()``` function to ```opac``` value in fragment shader. ```opacExp``` is the value you would want to change.


GoogleFrog:
I think the hex effect has too much of a sharp cutoff. Could the hexes be a bit fainter towards the edges and very slightly visible in the middle? It looks a bit too much like a cutaway.

The hex texture also needs improvement, but this is slightly tangential.

GoogleFrog:
I prefer the second one. Apply the wispy shield to chickens

https://i.snag.gy/5aGP8m.jpg

lhog:
I've made opacity window slightly less for all shield sizes and added some shield animation (texture ripple and slight diameter swing):
https://drive.google.com/file/d/1vSXksd-7Cjn7OK4eT5gfKajoZUIj_vq7/view?usp=sharing

Does it look like a right direction of improvement?

As for hex sharpness, I think it's best to rectify that with just placing a different texture. Will take some time to find or make a right one, though.

Licho1:
I like the animation!

lhog:
Would like to hear @GoogleFrog feedback before proceeding any further.

Also it's possible to take shield hitpoint into account and can cause local disturbance in the force.
Note:
* Here hitpoint is scripted, always on ripple from the first video is temporary disabled.
* If supporting infrastructure is in place, then in addition to messing with ripple, it'll be possible to play with alpha channel or blend in some other texture.
* The supporting infrastructure is probably going to be quite heavy in terms of code and performance: collection and storage of every hit on every shield, spatial accumulation, damage depreciation, etc
https://drive.google.com/file/d/1WYmbxqDpJqThKfb8jOMU4IB-n8mN1IAl/view?usp=sharing

IS it worth doing so?

GoogleFrog:
On the base visuals, I think that the white hex outlines against dark blue looks a bit weird. I think "second one" is much more fitting.

Impact wobble: Getting the impact location in worldspace for a projectile will be easy. Getting the impact location of a beam will be hard or impossible. How much work is it to convert a worldspace coordinate to the shield? I would say that adding hit location is much lower priority than polishing the hex grid but it should at least be enablable for Ultra.

Anarchid:
I bugged mantis for a way to detect beam-shiels interceptions exactly for this reason; iirc it was implemented.

lhog:
@GoogleFrog, as I wrote earlier "second option" doesn't look good on bright map. I didn't know how to make it look better, so I went for improved "option number one".  Would you like to ignore that and insist on option 2?

@Anarchid, yeah I had this code for a while in my branch:
https://github.com/lhog/Zero-K/blob/lupsShieldsTexture_v2/LuaRules/Gadgets/lups_shield.lua#L71-L100

> worldspace coordinate to the shield

It's as trivial as normalize(dx, dy, dz) in the fragment shield. Here I've just hardcoded the hitpoint, obtained via Spring.Echo in gadget:
https://github.com/lhog/Zero-K/blob/lupsShieldsTexture_v2/lups/ParticleClasses/ShieldSphereColor.lua#L228

The complicated part is hitpoints transfer from lups_shield gadget to the fragment shader. The way I see it:
1. The gadget should collect hits for all shielded units. and store them in the unit's shield spatial grid, something like 5x5 degrees. Hit frame and damage (or damage/maxShieldCapacity) are stored for each hit occasion.
a. Hit frame is used to attenuate hit effect over time and then remove it once it has become too weak.
2. As hits may come in succession (any hitscan weapon) or even simultaneously in the same grid cell, there is a need for damage accumulation over the time and space.
a. The cumulative damage or % in a cell is used to calculate the ripple magnitude and area of exposure (shader already has this code just in case).
3. Next for the sake of explanation say 4 topmost cumulative damages are somehow sent to unsynced part and where they are picked up as uniform arrays by shader.

@GoogleFrog, I've done the proof of concept and I see no obstacles in principle. That said whole sequence above is going to be quite heavy in terms of performance and amount of code. I don't think I'm going to do that unless I get a rock-solid guarantee it will be part of the code when I think it's done.

Licho1:
Can't wait to see some effect in action!

What exactly is "heavy"?
Afaik hardest part is getting information about impact, engine does that and gadget can collate over some time period - that can be fine, many gadgets process all damages. 
Sending to unsynced is relatively cheap if you only send stuff like coords + magnitude few times per second, it is basically just a serialization/deserialization locally. The effect can lag slightly so frequency does not have to be high.

As I see it the most expensive part is the actual ripple rendering. Instead of ripple it could be some sort of coloration too. In the worst case you would do shield segments, and light entire segment. That should be very cheap.

lhog:
Ripple in the fragment shader is basically "free" as a beer :smile:. I've made a benchmark with 50 shields:
While addition of every next shield drops FPS, comparatively the performance of the current shields, shields with static texture and shields with texture and always-on ripple was absolutely same. Four more ripple sources might have some impact but I expect it be marginal.

The hard part is that I will need to do iterations for every shield, then for every sector (which is lat,long), then for every damage this shield have absorbed in the recent frames then depreciate every such damage by a factor of how long ago it hit the shield, then make a cumulative damage per shield's sector/cell, then send all information to the shader, somehow, probably via SendToUnsynced. That is for every frame or at some pace, if there is some work to do.

lhog:
Screenshots on number 1 vs number 2:

screen00048
screen00049
screen00050
screen00052
screen00053
screen00054
screen00055
screen00056
screen00057
screen00058
screen00059
screen00060

Licho1:
Sounds like some sort of data structure problem..
You can make a widget to manage this sort of data and prepare final stucture for shader. 

To avoid iterations, you can for example maintain list of shield-space coords (or sectors) where ripple (or damage) occurs. You will only iterate these and update with some simple equation like x = 0.8*x + 0.2* value from synced.

You can completely ignore empty sectors. Not sure how well this goes with the shader, in the worst case you would maintain static list of sectors in a form of direct access array and on iteration of currently active ripples directly edit those values.

Thus complexity should never exceed o(n) where n is number of currently visualised ripples/sectors.

lhog:
@Licho1, Sure, I've never said it's impossible to do or optimize. I'm just warning in advance that the computation might cost something and code won't look too easy to understand, so it won't come as surprise when I have implemented this and spent a lot of time and efforts. As said I'm not going to make this, unless my effort is appreciated and the code is included into the main line. I've had unpleasant experience from the past when the code I had done went down the drain.

@GoogleFrog, you mentioned you didn't like hex texture. Here is the tool to make hex texture: https://www.shadertoy.com/view/4ljBzG You can play with #define values and get a texture as you like. Then either save the texture or tell me constants.

Licho1:
If your code can be turned off in graphics settings, so that weaker computers wont need to run it, then i think we can guarantee its going to be included.

Anarchid:
Uploading charge or recent damage shouldn't be that problematic i think, so i guess your real issue here is what happens when multiple damage sources affect one shield, causing multiple ripples.

There are two ways to go around it, i think. Either have a widget collect shield impacts like the dynlights widget that collects all lights and then renders them, or implement shield responses as a *separate* LUPS effect while maintaining intensifying global ripple in response to recent total damage.

So, on ShieldPreDamaged: 1) add damage to recent damage of this shield 2) spawn a properly oriented, scaled, and colored shield flash particle
On frame: decay recent damage on all shields
On shield draw: apply ripple controlled by one (uniform) variable.

lhog:
Just a WIP update.

It took me more than expected to get to a working prototype. Partly because I have very little understanding of what I am doing in the shader programming area and partly due to GLSL compiler bugs.

Anyway, now I collect shield impact points and deal with them over the space and time domains. It's done in the unsynced part of the gadget. I'm also able to disable shield hit handler in unsynced if Lups quality is less than Ultra.

ShieldSphereColor class in turn can pick up up to 4 biggest impact and render them on the sphere. For the sake of demo I left only the hex texture and disabled always on ripple. Impact and area of effect do not scale well at the moment, but it's a matter of fine tuning.

**Note this is very early prototype and WIP:**

[Video](https://drive.google.com/file/d/1lT6xS3dUif1R7yalWwZ-FYKH_R_ECXbo/view?usp=sharing)

lhog:
Another  WIP:
[Video_24.01.2018](https://drive.google.com/file/d/1LJ84iRGfmELGZ_vviQRqUJg4OW02DUDq/view)

Licho1:
Abrupt disappearance is a bit disturbing. It should fade out instead.

Anarchid:
Suggestions re: "Another WIP" video:
- Scale size and possibly initial alpha of the hit effect with damage. I think the damages close to 100% of shield's maximum capacity should reach half-sphere size.
- Make the duration of each impact much smaller and uniform. Something about 500ms should be quite enough. If you want more visibility, upscale size to compensate.
- Make the effect fade out with time 
- Drop lower-damage effects when higher-damage effects show up in the stack. 

Overall, my concern there is that the Glaive can spam the shield with so many events that the much more meaningful Rogue impact never even gets registered. Two of the above suggestions in concert should ensure that both units get reacted to (because there are less simultaneous effects because they are time-wise much shorter), and if there's contention, the higher-damaging effect has priority.

lhog:
There was an issue with ripple area of effect and magnitude scaling. Basically damage from llt and damage from DDM could be 3 magnitudes different. Therefore if I was to optimize for llt I would get terrible visuals on DDM, and If for DDM, then llt/glaive damage effect would be invisible at all.

So the quick solution was to clamp minimal area of ripple effect as well as affected area from top and bottom in really tight ranges. This gave a side effect of effect of sudden disappearance, because I forced minimal magnitude and AoE, despite the damage fades with time.

To get it done  right I need to figure out the scaling function that would scale from 0.5 dmg to something like 2000-3000 dmg, while only giving 2-3 fold difference on effect's magnitude and AoE. Yeah and it should rapidly fade out with time. so shield recovery is visible. Something to think about...

GoogleFrog:
The ripple looks distorted. For example at 0:17 the Glaive shield has a ripple that seems to be stretched upwards instead of forming a circle on the shield surface. To solve LLT vs DDM I think you may need some sort of cumulative damage effect that is caused by many hits from approximately the same direction as well as a nonlinear scaling function.

On the shield itself, some sort of ground intersection effect would be very good for usability and I am seeing an illusion which makes the shields look concave. Perhaps the hexes in the center of the shield are not large enough. I don't know how you are mapping your hex plane to part of a sphere.

lhog:
> The ripple looks distorted. For example at 0:17 the Glaive shield has a ripple that seems to be stretched upwards instead of forming a circle on the shield surface.

It might be distorted. Especially around the poles. I draw sphere in such way that poles are seen less, but it doesn't mean they don't exist. I may need to investigate options, as generally it's nearly impossible to map a 2D texture onto sphere without causing stretched and compressed areas. This webGL demo shows the problem well: https://www.shadertoy.com/view/4sjXW1 That said there might be some light in the end of the tunnel: https://www.shadertoy.com/view/4dBXWD I will investigate whether this could be ported onto the scene with defined geometry like in our case.

> On the shield itself, some sort of ground intersection effect would be very good for usability and I am seeing an illusion which makes the shields look concave.

I was going to play with stencil and depth buffer to outline the ground intersection figure and also possibly draw shields in a way that they form a single surface, possibly with some minor intersection highlight as well.

All of the above will take some time as much is going on IRL, so I can't put much time into that.
Do you want me to clean up the hex (+small animation) effect for now, and remove everything else that concerns with impact feedback so this could be merged?

lhog:
[Here is what I have now](https://drive.google.com/file/d/1XskxfwKJES42xuPF0asYCUhkMFrKMDID/view?usp=sharing)

For ground intersections and single surface drawing I think it's necessary to have at least two drawing passes. Not sure how it's done in LuPS, so except for fine tuning of colors, sizes, etc. this is as far as I can get in the nearest future.

GoogleFrog:
Looks good. Some thoughts:
- The effect of the ripple is looking really good.
- I think this would be good as a new particle class with the current shield as a fallback. Graphics settings could exist to choose between the three shields in case there is an issue.
- Does the shield rotate when their parent unit turns? I think rotating while turning looks stupid as it makes the shield look more 'physical' when it is supposed to be some sort of energy bubble. This is part of what stopped me implementing hex shield.
- The stark hexes look a bit bad. Perhaps we need a mix of slightly more transparent hexes as well as a flood colour. This should be doable in texture.

lhog:
> The effect of the ripple is looking really good.

Yeah, @ashdnazg wanted to look into the code and make some changes as well
> I think this would be good as a new particle class with the current shield as a fallback. Graphics settings could exist to choose between the three shields in case there is an issue.

Makes sense. Next week I'm traveling, I think I'll do it later in that week.
> Does the shield rotate when their parent unit turns? I think rotating while turning looks stupid as it makes the shield look more 'physical' when it is supposed to be some sort of energy bubble. This is part of what stopped me implementing hex shield.

No bubble does not rotate with a carrier unit. I tried an option when bubble rotated with the unit and it didn't look nice. The side effect as mentioned by @ashdnazg is that once camera is tilt when texture doesn't stay in the same place. Instead the sphere center is always in front of the camera (so poles, there texture is distorted, are less visible).

> The stark hexes look a bit bad. Perhaps we need a mix of slightly more transparent hexes as well as a flood colour. This should be doable in texture.

The mix is configurable in lua across all RGBA. This way it's possible to achieve what was formerly option 1 and option 2. The hex now is drawn without the use of texture, however it's totally possible to use texture (if texture is set then texture willl be used  by shader, if not - then hex is drawn).

GoogleFrog:
I think it is bad to have the texture move when the camera tilts or scrolls. Could you throw in some pentagons to fix stretching?

Anarchid:
You can do intersect detection in the same pass in the fragment shader by binding screen z-buffer or heightmap to the shader and then raytracing the sphere to see whether the world position of sphere surface at this fragment is close to the other coordinate.

lhog:
I was reminded about this PR yesterday.

Given the real life situation, I can't "throw in some pentagons to fix stretching" nor "do intersect detection ...". At least not in the foreseeable future.

I understand the ZK code is frozen for the new features before the Steam release, therefore I'm closing this PR. The code stays in my repo and I might issue another PR with the existing content (no intersect, no fixing of stretch, but coded as a new Lups class as it was agreed) once ZK is unfrozen and I have time.

Anarchid:
Neither intersect detection nor pentagons are/were hard requirements, i think.

PR #2754, opened 2018-02-06
WIP WaitForTurn POC.

GoogleFrog:


sprunk:
Did it work? I wrote it blindly on github without even checking if it compiles.

PR #2853, opened 2018-03-16
Sea balance changes

Aquanim:
Shogun:
Cost 4K -> 5K
HP 12K -> 8K
Reload time 10s -> 15s (DPS 450 -> 300)

Corsair:
Cost 220 -> 240 (might be overkill)
Speed 3.2 -> 2.7
Shotgun damage 16x32 -> 9x26 and spray angle decreased (both the same as Scallop)
Reload time (for each shotgun) 2.4s -> 2.0s

Hunter:
Increased model and collision size by a bit less than 50%.

Mistral:
Damage 280x4 -> 200x4

Siren:
HP 4000 -> 3750

Envoy:
HP 2500 -> 2000


Aquanim:
Don't pull this yet, the dead Hunter model is still borked.

Aquanim:
That should have fixed it.

PR #2914, opened 2018-04-18
Reinstate zombie sounds

Anarchid:
A user complained about zombies being silent in #zk:

image

They used to be quite noisy before https://github.com/ZeroK-RTS/Zero-K/commit/0b30472feba0745a5731516c3f22ebb03867bcf0#diff-1c6ecb7d455a178bbe6f22a8cd44856e, when many SendToUnsynced actions were mass-removed because of desync concerns. IIRC the desync was found elsewhere, but zombies and rez stayed silent since then.

Unlike in 2014, today zombies are a campaign feature - it makes sense for them to be slightly prettier. This commit restores the original sound behaviour with several alterations:

1) It does not restore the athena/commander resurrection sound. This is because i don't want to touch that gadget at this time.
2) It uses the fogged sound api instead of implementing its own SendToUnsynced actions.

While technically not a new feature, since this affects campaign, it may be best to wait until after release before pulling.


Anarchid:
Sound api is awesome, this is much cleaner. I wish it played sounds in AirLos rather than in straight los, but that's another matter.

LuaRules/Gadgets/unit_zombies.lua
```diff
@@ -74,6 +74,13 @@ local zombiesToSpawnMap = {}
 local zombiesToSpawnCount = 0
 local zombies = {}
 
+local ZOMBIE_SOUNDS = {
```
sprunk: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/api_fog_hidden_effects.lua#L16

You need to add the sounds to the list in the API (sounds pretty lame when I think about it but perhaps it's not the time to rewrite the api)

Anarchid:
Oh wow. Retracting that awesomeness opinion! 

I guess between my preference for AirLos and the file list thing it may be worthwhile to update the api. Eventually.

LuaRules/Gadgets/unit_zombies.lua
```diff
@@ -308,7 +315,8 @@ function gadget:GameFrame(f)
 					spSpawnCEG(CEG_SPAWN, x, y, z, 0, 0, 0, 10 + r, 10 + r)
 
 					if steps_to_spawn == WARNING_TIME then
-						--SendToUnsynced("zombie_sound", x, y, z)
+						local z_sound = ZOMBIE_SOUNDS[math.random(#ZOMBIE_SOUNDS)]
+						GG.PlayFogHiddenSound(z_sound, 10 x, y, z)
```
sprunk: Typo? `10 x`

PR #2916, opened 2018-04-19
WIP Selection and visible memory optimization

GoogleFrog:
The new widget profiler reveals some allocation pressure issues. UnitShapes and Healthbars are particular offenders and it is likely due to the large array created and destroyed on every GetVisibleUnits call. It is likely better to keep an array of all units and to provide an interface which allows widgets to iterate over visible units, with the API calling and caching Spring.IsUnitVisible. The same API could store the list of selected units that is sent to widgets in the SelectionChanged event.

UnitShapes also creates and destroys many tables every update as it filters the visible units into various categories.


sprunk:
Reopen if this becomes ready (I won't delete the `memoryOptimize` branch on which this lives).

PR #2939, opened 2018-04-22
Fixed caretaker building not blocking ground construction upon load

lhog:
See http://zero-k.info/Forum/Thread/25100?page=1


LuaRules/Gadgets/api_saveload.lua
```diff
@@ -254,7 +254,7 @@ local function LoadUnits()
 			spSetUnitVelocity(newID, unpack(data.vel))
 			--spSetUnitDirection(newID, unpack(data.dir))	-- FIXME: callin does not exist
 			
-			if not UnitDefNames[data.unitDefName].isBuilding then
+			if UnitDefNames[data.unitDefName].canMove then
```
sprunk: I think this should be `not UnitDefNames[data.unitDefName].isImmobile` (IIRC factories have `canMove = true` because they can queue up the order)

PR #2947, opened 2018-04-23
Make engine infobox (ctrl+i) and debugcolvol (alt+b) require alt+ctrl

sprunk:
Newbs can enable those by accident and not know how to get back. They will never need those. Devs can bind manually.


GoogleFrog:
I am unsure. They should be bindable through some advanced setting because devs (and modders) may want it. There is effeminately a usecase for colvol view by mappers, modders, devs and people reporting bugs so I would keep the hotkey. The info hotkey is useful to asking people their game and engine version.

What if we changed both the hotkeys to Ctrl+Alt+Letter?

Issue #2941, opened 2018-04-22
Line command behaviour when clicking meanwhile


sprunk:
Inconsistent with area commands: canceling area commands requires to press+release the other button, line commands just press.

Also:
 * hold down left mouse to make a selection box
 * press right mouse button to start a line draw command
 * release right mouse button
 * then release left mouse button
 * now with no mouse buttons pressed, press left button to make selection box

-> you wont get a selection box at your first click. You have to click once again to get a selection box

(reported by Godde, contact him for details if more info needed)

PR #2948, opened 2018-04-23
WIP Basic selection and command improvements

GoogleFrog:
A selection box that has the mouse start and end on the same unitID should behave identically to click selecting the unit. This is difficult to do without single-frame UI flicker.

Line commands should work like this https://github.com/ZeroK-RTS/Zero-K/issues/2941


sprunk:
Reopen if this becomes ready (I won't delete the `LineCommand` branch on which this lives).

PR #3004, opened 2018-05-03
Ruin balance for the sake of pretty numbers.

sprunk:
Health values made round because such numbers are easier to read and learn:
LLT 785 → 800
HLT 2475 → 2500
Artemis 1570 → 1600
Likho 2360 → 2350
Airpad 1860 → 1850
Detriment 85800 → 86000
Tremor 2045 → 2050


GoogleFrog:
I don't want to ruin balance for numbers.

sprunk:
That's good because the description was in jest and balance wouldn't actually be ruined.

I backed out of touching LLT.

Issue #3014, opened 2018-05-04
bky noticed that OKP doesn't work correctly with armored targets


lhog:
We have incorrect armor multiple read in the code. Status quo is such that in-LoS armored units are read as wearing armor no matter if they are open on closed.

Will post a fix in a minute.

PR #3015, opened 2018-05-04
Fixed https://github.com/ZeroK-RTS/Zero-K/issues/3014

lhog:
We were reading armorMultiple but never checked if armored state is enabled or not.
I hope it's not intentional.


GoogleFrog:
It may well have been intentional, since the failure state of OKP should be to fire too often instead of too little. But lets see what happens.

Issue #3011, opened 2018-05-04
Fix out-of-range targeting


GoogleFrog:
Apparently the HLT not firing error was fixed in the dev engine by making out of range enemies always have lower target prioritization. We will likely need to put something in allowWeaponTarget to replicate this. Perhaps the default priority will indicate whether the enemy is in range or not.

sprunk:
The default priority gets multiplied x100'000x for those targets.

GoogleFrog:
So the default priority should be easy to detect?

PR #3016, opened 2018-05-04
This is a performance light way to fix https://github.com/ZeroK-RTS/Zero-K/issues/3011

lhog:
The reason behind this particular implementation is that even with new engine, it's hard to tell what defPriority will be if target is out of range.
This implementation will work with the current engine and future engines equally well.

Alternative implementation would be to guess if defPriority is high enough and then deprioritize target.
The issues are following: 1) Needs recent enough engine 2) Needs a good guestimate on defPriority value (so engine wants to tell us the target is out of reach).

P.S. One downside is that attacker will oscilate between similar out of reach targets, because, as we all are aware, defPriority has random() inside. To fix this it is required to run priority calculation the way ZK does that and only then apply out-of-reach multiplier.
Seeking guidance if this is something we want to do.

### DIFF
--- a/LuaRules/Gadgets/unit_target_priority.lua
+++ b/LuaRules/Gadgets/unit_target_priority.lua
@@ -278,10 +278,14 @@ end
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 -- Priority callin
+local DEF_TARGET_TOO_FAR_PRIORITY = 100000 --usually numbers are around several millions, if target is out of range
 
 function gadget:AllowWeaponTarget(unitID, targetID, attackerWeaponNum, attackerWeaponDefID, defPriority)
 
 	--Spring.Echo("TARGET CHECK")
+	if defPriority > DEF_TARGET_TOO_FAR_PRIORITY then
+		return defPriority --hope engine is not that wrong about the best target outside of the range
+	end
 	
 	if (not targetID) or (not unitID) or (not attackerWeaponDefID) then
 		return true, 25


LuaRules/Gadgets/unit_target_priority.lua
```diff
@@ -278,21 +281,42 @@ end
...
-	
+
 	if (not allyTeam) then
 		return true, 25
 	end
-	
+
+	if remReachable[unitID] and remReachable[unitID][attackerWeaponNum] and remReachable[unitID][attackerWeaponNum][targetID] ~= nil then
+		if remReachable[unitID][attackerWeaponNum][targetID] == false then
+			return defPriority * UNREACHABLE_MULT
```
sprunk: Why not just return `UNRECHABLE_MULT` (without multiplying by `defPriority`)? This way there won't be any oscillation and the number is large enough to always be the worst anyway.

sprunk:
I think something like
```lua
if defPriority > SOME_MAGIC_THRESHOLD then
    return math.huge
end
```
should be enough because the leeway we get is enormous (multiplier for outrange is 100k). The task would be to find out what the ordinary `defPriority` values are and set the magic appropriately.

LuaRules/Gadgets/unit_target_priority.lua
lhog: One reason is ```defPriority``` is only semi-random. It has some sane value inside and we want to target some "valuable" unit. Another reason, I'm not sure how stable sort is and what is the tie breaker in the engine, so it might oscillate as well.

lhog:
Yeah as stated, it's a valid option also. There are several obstacles though:
1. We will need a very fresh engine, that does 100k mult.
2. We don't know exactly SOME_MAGIC_THRESHOLD. Shouldn't be too hard to figure it out.

Let's see what @GoogleFrog thinks of it and I will act accordingly.

sprunk:
1. doesn't sound like a problem since the engines between the introduction of the feature and the 100k mult can just be skipped.

GoogleFrog:
Why should we do duplicate work when the engine already (allegedly) sends a high enough default priority for us to distinguish the cases. Also does this change work on units which are in range but blocked? Does the engine even work for such units?

Issue #3058, opened 2018-05-13
map option : min wind multiplier


springraaar:
on maps where the playable area is on high plateaus, the windmill gadget behavior leads to very high min wind values (1-2) which lead to worse game experience as they remove the solar vs wind decision making.

a min wind multiplier option would allow map devs to lower that to more reasonable values if they want to.

sprunk:
#3059

PR #3059, opened 2018-05-13
map option : min wind multiplier #3058

springraaar:
map option : min wind multiplier #3058


lhog:
Shouldn't you mess with `groundMin` instead?

sprunk:
Personally I'd rather see the whole height mechanic removed and some minimum being given to windgens baseline (say 0.5). As I see it the effects would be:
```diff
+ wind becomes viable on flat maps because there is no longer the chance you get screwed by 0 wind.
+ non-wind sources (in particular, solars and fusions) become viable on hilly maps because wind is no longer so absurdly efficient. Currently the most extreme case, 2.5 minimum, is more efficient than singu and mohogeo while not exploding and being gradual.
+ simplicity. One less mechanic that doesn't provide interesting decisions (and actually reduces them, see above).
- less "realism" (not that the current implementation is any good: in reality, the wind is actually more reliable at sea level while in the mountains the maximum increases as well).
- less variety on windmaps. Pretty legit argument but I think the layout (heightmap, mexes) already provide enough there.
```

springraaar:
note: there's the forum thread [here](https://zero-k.info/Forum/Thread/24711?page=1)

springraaar:
I think the multiplier is easier to use than groundmin.

forcing 0.5 min wind on all maps could work too, i haven't seen people complaining about the tidal values.

GoogleFrog:
I don't want to reply on mapinfo. I would much prefer a separate file.

springraaar:
why is mapinfo.lua not reliable? all maps have it.

and it has a nice custom subtable where misc attributes can be defined

Aquanim:
I'd prefer a solution which relies less on the map developer to do something sensible (particularly because the vast majority of maps have no active maintainer).

springraaar:
Note that the solution i provided doesn't change existing maps. The current behavior is kept unless the map dev explicitly adds the attribute.

springraaar:
so...

springraaar:
thanks, I'm about to make some map releases.

That way i can add the "mapInfo.custom.zkminwindmult" and keep the wind values sane.

lhog:
IMO this change is a mistake. There were more sane proposals, like checking height of mexes and setting the lowest mex's height as a wind baseline height.

springraaar:
> There were more sane proposals, like checking height of mexes and setting the lowest mex's height as a wind baseline height.

That was on my first post on the forum thread about it, and got no support...

GoogleFrog:
I don't know if I am going to support maps with such a mult, I guess I got worn down by the discussion and merged the PR.

KingRaptor:
It doesn't cause any first-order harm to have the option, does it?

lhog:
Perhaps I'm reading that wrong, but setting minWindMult to low or two high makes slope to flat or too steep. This should break the assumption that maximum wind is achieved in the areas of maximum height, shouldn't it? And I've already mentioned that multiplier is not intuitive compared to setting height of minimal wind or judging that based on mex placement. 

One can, of course, ban maps abusing wrong values, but why open pandora box?

Issue #3071, opened 2018-05-19
autorepair module is insane


mojjj:
http://zero-k.info/Battles/Detail/535225

Firepluk 's trollcom at level 11 had +391hp/s regeneration. please fix asap someone!

sprunk:
culprit https://github.com/ZeroK-RTS/Zero-K/commit/23326cdd08b315671cdb1fca6947540311e31542#diff-018196cfdc4b966a0b4c7a2886586809L43

springraaar:
waaaait...so commander regen scales with build power increase %?

sprunk:
:owl:

PR #3072, opened 2018-05-20
Quickfix #3071

sprunk:
Not very elegant using `slowState` directly but using `totalBuildPowerChange` is incorrect (due to commanders) and nothing else makes much sense.


GoogleFrog:
As long as stun and disarm are also checked I think we are fine. Possibly zombie slow state should also impact autoregen.

PR #3084, opened 2018-05-22
Textured luaintro progress bar

The-Yak:
Progress bar with darkhive theme, now also draws all the way to 100% even on first time map load.

progbarpreview


sprunk:
Looks good.

It probably shouldn't reach 100% if there is still anything to load though (especially since first time loads can take a while to calculate pathing)?

The-Yak:
That's how it works currently on first map load, in the first screenshot the pathcache is only being read, not generated, so it happens in an instant before Loading Skirmish AI (the true last step).

334920_screenshots_20180522132311_1

sprunk:
thanks!

Issue #3078, opened 2018-05-21
Fancy shields shader compilation minor stuff


sprunk:
https://github.com/ZeroK-RTS/CrashReports/issues/8251
```
[f=-000001] LUPS->Shield: shader warnings & errors: WARNING: 1:88: 'assing' : implict conversion between types allowed from GLSL 1.20 

No errors.
```

The code probably needs `#version 120` somewhere. However this was just a warning and the code compiled but it got rejected:
https://github.com/ZeroK-RTS/Zero-K/blob/a0fa21b36a5b2f812b9d25f8e0b22af872fb35ee/lups/ParticleClasses/ShieldSphereColorHQ.lua#L340-L344

It should still always print the log if it's not empty but only abort if `shieldShader == nil` (also because it leaks otherwise, by the way).

@lhog

PR #3085, opened 2018-05-22
A few improvements to the current fancy shields code:

lhog:
1) Warnings no longer cause Lups class failure. See https://github.com/ZeroK-RTS/Zero-K/issues/3078
2) Should a GLSL compiler produce errors or warnings, whole code of the shader is dumped along with the error/warning. This allows for easier matching between the error line reference and the actual code
3) I think I found what was causing "WARNING: 1:88: 'assing' : implict conversion between types allowed from GLSL 1.20" error. Probably it was MAX_POINTS variable, that got presented as float to the GLSL compiler
4) Put "#version 120" in case I found the cause above wrong
5) Fixed terribly looking distortion on the back side of the sphere. This was caused by wrong 2D coordinate of ripple center for the back side.
6) Pace of shield animation no longer depends on the drawframe, but on the current gameframe. As a result, shields no longer animate on pause and the animation pace has become independent from FPS.

### DIFF
--- a/lups/ParticleClasses/ShieldSphereColorHQ.lua
+++ b/lups/ParticleClasses/ShieldSphereColorHQ.lua
@@ -73,7 +73,7 @@ function ShieldSphereColorHQParticle:Visible()
 	return self.visibleToMyAllyTeam
 end
 
-local PACE = 1200
+local PACE = 400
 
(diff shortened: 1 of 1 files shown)


sprunk:
thanks!

PR #3101, opened 2018-05-27
Added overkill prevention to OKP gadget and to Faraday

lhog:
See: http://zero-k.info/Forum/Thread/25574

### DIFF
--- a/LuaRules/Gadgets/unit_overkill_prevention.lua
+++ b/LuaRules/Gadgets/unit_overkill_prevention.lua
@@ -329,7 +329,7 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
 	incomingDamage[targetID].disarmed = disarmed
 	incomingDamage[targetID].emped = emped
 
-	block = doomed or disarmed or emped --assume function is not called with both regular and disarming and EMP damage types
+	block = doomed or disarmed or emped --assume function is not called with more than one of: regular, disarming and EMP damage types
 
 	if not block then
 		--Spring.Echo("^^^^SHOT^^^^")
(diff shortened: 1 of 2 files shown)


LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -290,11 +322,14 @@ local function CheckBlockCommon(unitID, targetID, gameFrame, fullDamage, disarmD
...
 	local disarmed = targetIdentified and disarmFrame and (disarmFrame - gameFrame - timeout >= DECAY_FRAMES) and (disarmDamage > 0) --for disarming projectile
+	local emped = targetIdentified and empFrame and (empFrame - gameFrame - timeout >= DECAY_FRAMES) and (empDamage > 0) --for EMP projectile
+
 
 	incomingDamage[targetID].doomed = doomed
 	incomingDamage[targetID].disarmed = disarmed
+	incomingDamage[targetID].emped = emped
 
-	block = doomed or disarmed --assume function is not called with both regular and disarming damage types
+	block = doomed or disarmed or emped --assume function is not called with both regular and disarming and EMP damage types
```
GoogleFrog: Comment should say 'more than one of'

GoogleFrog:
I don't get it, why is emp computation done for non-emp damage?

GoogleFrog:
Don't included commented-out GG functions as unused code is liable to stop working in the future and just cause hassle in the meantime.

lhog:
Without seeing my code, but based on the comments, I will say that some of them might make sense (pending review), other are likely rubbish (pending review again).

I'll review once I am back from vacation in a few days.

lhog:
OKP code doesn't work correctly if target is hit with projectiles with different effect timeouts. It was written with the assumption in mind that timeouts of projectiles target is git with are same, otherwise extra unsolicited shot happens.

Good example of different effect timeout are for example widow's EMP and faraday's EMP. These two effects don't interact with each other correctly in the current framework of OKP. The effect is marginal (if I recall it right from July timeframe): an extra redundant shot from faraday.

In principle I fixed what @GoogleFrog pointed out and the code lies somewhere in some branch, but given how obsessed I am with perfectionism, I'd rather close this one and see if I get an inspiration to patch OKP for proper solution.

PR #3144, opened 2018-06-24
Fixes Resign/Exit causes entire squad to resign

ArchShaman:
Currently there's a bug introduced April 30th in which any person on a squad that resigns causes the entire squad to resign. This is due to forceresign being sent to luarules without any commshare stuff going on in the resign gadget.  This bug fix introduces a function that is called on resign button pressed or when user presses exit that calls for the squad to be unmerged if leader or just self to leave the squad before they are resigned.


ArchShaman:
Oh. Whoops. Yeah you're right.

Issue #3212, opened 2018-08-13
Replace resign on quit by improved inactivity win


DeinFreund:
- [ ] Pause the game when inactivity win is available
- [ ] Assure inactivity win window is always in foreground, so it can be clearly seen
- [ ] Allow inactivity win once ping exceeds 10 secs as well as on a dropped connection
- [ ] Remove resign when clicking "Exit to Lobby" in the main menu

GoogleFrog:
Pausing when inactivity is detected is a regression. Imagine 8v8.

If resign on quit is removed then inactivity should be flagged immediately upon quitting.

DeinFreund:
This is for the inactivity win. You can see in the PR that it only kicks in if all players of the enemy team are dropped or have a ping > 10 seconds. This is nothing I ever see happening in 8v8.

GoogleFrog:
I'm just reading what the ticket says without stopping to load too much of the context back into my memory.

PR #3213, opened 2018-08-13
Replace resign on quit by improved inactivity win

DeinFreund:
See #3212

### DIFF
--- a/LuaRules/Gadgets/game_over.lua
+++ b/LuaRules/Gadgets/game_over.lua
@@ -534,9 +534,9 @@ local function ProcessLastAlly()
 					local playerlist = spGetPlayerList(t) -- active players
 					if playerlist then
 						for j = 1, #playerlist do
-							local name,active,spec = spGetPlayerInfo(playerlist[j])
+							local name,active,spec,_,_,ping = spGetPlayerInfo(playerlist[j])
 							if not spec then
-								if active then
+								if active and ping < 10 then
 									hasActiveTeam = true
 								else
 									hasDroppedTeam = true
--- a/LuaUI/Widgets/gui_chili_inactivity_win.lua
+++ b/LuaUI/Widgets/gui_chili_inactivity_win.lua
@@ -149,6 +149,10 @@ function widget:Update(dt)
 	if desiredActive ~= windowVisible then
 		if desiredActive then
 			windowVisible = true
+			local _, _, paused = spGetGameSpeed()
+			if not paused then
+				Spring.SendCommands("pause")
+			end
 			if not mainWindow then
 				mainWindow = SetupWindow()
 			end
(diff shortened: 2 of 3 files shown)


LuaUI/Widgets/gui_chili_inactivity_win.lua
```diff
@@ -149,13 +149,21 @@ function widget:Update(dt)
 	if desiredActive ~= windowVisible then
 		if desiredActive then
```
sprunk: put `Spring.SendCommands("pause 1")` (without `if !paused` because `1`) before this line since it happens regardless of the taken branch

LuaUI/Widgets/gui_chili_inactivity_win.lua
```diff
@@ -149,13 +149,15 @@ function widget:Update(dt)
...
 			windowVisible = true
+			Spring.SendCommands("pause 1")
 			if not mainWindow then
 				mainWindow = SetupWindow()
 			end
 			screen0:AddChild(mainWindow)
 		else
 			windowVisible = false
 			screen0:RemoveChild(mainWindow)
+			Spring.SendCommands("pause 0")
```
sprunk: not sure if good idea, IMO let the rejoinee unpause manually to make sure he's really there

DeinFreund: then we need an unpause button. currently, pausing and unpausing is a poweruser feature. it should be possible to play without knowing how to operate it.

DeinFreund: we can still remove it if it turns out this still leads to people missing out on matches.

sprunk: alright, keep it for now then

LuaRules/Gadgets/game_over.lua
```diff
@@ -534,9 +534,9 @@ local function ProcessLastAlly()
 					local playerlist = spGetPlayerList(t) -- active players
 					if playerlist then
 						for j = 1, #playerlist do
-							local name,active,spec = spGetPlayerInfo(playerlist[j])
+							local name,active,spec,_,_,ping = spGetPlayerInfo(playerlist[j])
 							if not spec then
-								if active then
+								if active and ping < 10 then
```
sprunk: Something to think about: AFK giveaway happens at 30, possibly this magic constant should be global and shared with AFK monitor.

DeinFreund: AFK != ping though. I don't care if the opponent goes afk. AFK time also accumulates differently from ping.

sprunk: AFK takes ping into account though, and possibly somebody who is considered dropped enough to cause game over should be eligible to get his units given away as well.

sprunk:
Lacks the "always on top" for the wait-or-win window to fully solve #3213 (use `window:BringToFront()`).

Issue #3175, opened 2018-07-19
Consider Cyclops nerf


Aquanim:
The feeling that Cyclops is currently too strong seems to be fairly prevalent among 1v1 players currently. Making an issue so that it gets seen.

My hot take is to decrease its movespeed to 50 or so. 
A decrease to slowbeam range or main cannon damage could be considered although I don't know exactly what the current numbers are tuned to defeat.

https://zero-k.info/Battles/Detail/583454
https://zero-k.info/Battles/Detail/582792
https://zero-k.info/Battles/Detail/582724
https://zero-k.info/Battles/Detail/582579
https://zero-k.info/Battles/Detail/581393
https://zero-k.info/Battles/Detail/546363

Anarchid:
Adjusting the range of one of the weapons will make it annoying to use, so if you nerf them, nerf both.

Sgathec:
What exactly makes the Cyclops too effective?

sprunk:
Cyclops has high range because back in OTA larger units generally had higher range and Goliath was the biggest tank in the game (see the range progression between Ravager -> Minotaur -> Cyclops). As far as I can tell the range was never tuned against anything specific (the last time it was touched was in CA so any tuning would likely be obsolete regardless).

ArchShaman:
Maybe make the main cannon and slow gun more alpha-ish by increasing damage&reload time. Ideally this would make it fare much worse against raiders (which _should be_ its mortal enemy), while keeping the idea of a "tankbuster".

Sgathec:
I like ArchShaman's idea for the same reasons as he mentioned, although it's difficult to recommend anything because I don't know why the Cyclops is overpowered.

anmeldekram:
> ArchShaman's idea

Is the lobbyaccount "_Shaman"?
All eleven 1v1 matches of the last 30 days were against new noob accounts. Seeing this I did not bother to check if Cyclops were even used. Should balance ideas be based on matches where the opponent had a 5-10%win chance?

Aquanim:
Do you have anything useful to contribute @anmeldekram?

GoogleFrog:
I'd like to see the 1v1 matchup data before making changes. Lower speed sounds quant within the factory but anti-quant if you take all the factories into account, since tanks are relatively fast. Perhaps lower turn rate could come along with slightly lower speed, as vehicle pathing units have low turn rate as a weakness.

sprunk:
[...]

PR #3224, opened 2018-08-22
Convert Cyclops weapon to use Disruptor Cannon

Anarchid:
Implements #3175 by removing Cyclops' ability to slow fast units and then kill them. Now it kills them first, then slows.

Removes obsolete weapons (flamer, slowgun sidearm) - they're in git history anyway, screw this bloat.

This also includes CEGs and sounds for heavy disruptor cannon, and consequently also removes Cyclops/Bertha sound collision. 

The timeslow factor has been tuned to maintain Cyclops' previous 600 slow-dps. However, with much lower accuracy and rate of fire on this slow, it gets applied less often, and decays more between hits, so it could be buffed to a standard 3 if everyone finds Cyclops entirely useless from utterly dominating after this.


DeinFreund:
Merge please, I want to be able to play interesting 1v1 again

Anarchid:
One like and i'm merging this

Anarchid:
Okay

PR #3226, opened 2018-08-27
fix #3220 (rezzed uw gauss) by giving gauss legs

### LINKED ISSUE #3220: rezzing non-underwater units underwater.
http://zero-k.info/Battles/Detail/599598

underwater gauss is too strong. also x4 armour. also underwater has no artillery....

suggestion: underwater rezzed units are constantly disabled or max slowed or rezzing not possible or terraspike autocreated to lift building over water or sth else...

Anarchid:
Gives gauss model an underbelly and 6 pillar segments. Underbelly is 35 elmos tall and pillars are 45 elmo per segment, giving it a total maximum depth of 305 elmos, which should be enough for everyone.

The minimal required number of these pieces are `Show`n, so this plays nicely with halo selections (unlike mex).

Several other units remain which can be rezzed underwater and still be useful. But they are all Starlight, making the exploit massively impractical.

image


Licho1:
I like this

sprunk:
Buildpic needs to be updated.

PR #3247, opened 2018-09-21
Making the funnelweb a heavy shield support strider

danfireman:
Initial WIP MR to get feedback on this: http://zero-k.info/Forum/Post/190307#190307
Mostly I'm wanting feedback on what I've done wrong code-wise, thoughts on this from a gameplay perspective and any tips for how to fix the outstanding issues (at the bottom). Even if this never merges I'd be quite grateful for any feedback given. :)

**Unit role:**
This new Funnelweb is a genuine support strider. Weaponless, it is intended to protect fragile units on attacks on heavy porc and as a means of quickly taking advantage of the resulting victory in battle.

**Unit abilities:**
A large and strong shield of 28000hp that regenerates at a modest 200hp/s rate allows regeneration in 2 mins+20 secs, but requires 15 seconds after being hit before it will start to regenerate.
Additionally, it has 40bp with good range, intended for primarily for battlefield repair and reclaim.
Spider movement is slightly faster than the Funnelweb 45->54.
HP at 6000 (so it's vulnerable to widows, amongst other things).
Costs 3500, the usual price for a strider.

**Unit analysis and impact on battlefield:**
* Its very high energy consumption and high cost make it a late game unit.
* It requires a large army to prevent enemy units getting under its shields, but has enough hp that assassinations are less likely.
* There are two main ways of using it: Protection or negating attrition.
It can either protect an assault force for a quick and devastating strike OR it can hang around after an attack, repairing units and reclaiming, thus mitigating attrition. Doing the latter is likely to deplete its shield and render it useless for unit protection.

**Comparison with existing units with similar roles:**
Its shield has more hp per unit of metal than the Aspis (equivalent to 4666 metal of Aspis), but it's regeneration is weaker (equivalent to 2400 metal worth) and it is much more vulnerable to sustained bombardment due to its regeneration delay.
Its cost and lack of regen on being attacked make it a worse choice for maintaining a continual presence at the front and it's lack of shield linking mean it can't supply shield power for felons.
Its large shield size make it worse at defending smaller forces against attack.

Its battlefield reclaim, repair and construction is equivalent to 1150 worth of welders.
Its higher cost, lesser speed and inability to be in multiple places at once mean it is mildly worse at battlefield repair and reclaim than the welder.

**Initial playtesting thoughts:**
Fairly large micro requirements at present, but does it's job well. Likely to be far more useful in Campaign, FFA and 16v16 matches than 1v1 or 2v2s.

**Thoughts on balance:**
~~I am considering removing the resurrection entirely due to its snowballing effect. (it depends whether people consider snowballing good or bad, I've heard arguments either way for RTSs)~~ Now removed at the advice of people in the thread. (may return depending on people's thoughts)
~~Its high hp makes assassinations hard and allowing that could add interesting gameplay. Considering reducing hp to a fragile 6000.~~ Done.

**Known TODO/issues:**
* ~~Shield links (it shouldn't transfer shield strength or at least should stop regenerating when it does)~~ Now doesn't link.
* ~~Shield should take energy (aprox 72) to charge (currently doesn't charge at all if it's a builder and has shieldPowerRegenEnergy set :( )~~ Now works.
* ~~It should do automatic rez/repair/reclaim with a prioritization setting allowing a user to pick what it should prioritize (to reduce micro burden).~~ DeinFreund thinks this unwise
* ~~Should be able to move and rez/repair/reclaim simultaneously (so it can reclaim without slowing down the assault).~~ DeinFreund thinks this unwise
* ~~Fix rez and reclaim graphics (currently getting two styles of nano fog). The nano-related code is all cargo culted (copy pasted frankenstien) and probably partially nonsense.~~ Now only one fog style!
* ~~Regen delay needs to work with energy prioritization.~~ Done!
* ~~Shield becomes invisible when the Funnelweb is turned off. This is apparently also a problem in Master...~~ Should now be fixed
* ~~Shield and nano seem to be centered in the wrong place in the model, this is also a problem in Master.~~
* ~~Reduce weight~~ Done
* ~~Funnelweb should have a round number of BP~~ Made 40
* ~~The nanospray should emanate from somewhere on the surface, not just somewhere internal. Perhaps its face could be used.~~ Now coming from the eye.
* ~~The wobbly purple shield stuff is clipping. Perhaps this stuff could emit from the gap that held its guns. A different type of effect could be used.~~ A different type would be preferable (iteration 2), but for now commented out
* ~~There is no visual indicator that the shield is blocked from charging after being damaged. Perhaps the wobble could reflect this, as well as the colour or transparency of the shield.~~ Indicator now present
* ~~Smooth the shield sphere.~~ Done
* ~~Shield texture is ugly.~~ Made prettier
* ~~Shield color unpleasant.~~ Made prettier
* ~~Indicator of shield recharging not very clear.~~ Now noticeable

**Issues for iteration 2:**
I don't want to do a ton of polishing on a concept that we don't even know works yet!
* There should be some sort of emanation from the Funnelweb to help indicate that the Funnelweb is the source of the shield.
* The model needs changing to be more obviously a heavy shield strider.
* The Funnelweb needs some (heavy duty!) nanospray emitters.
* Needs a new map icon.


danfireman:
I've removed the rez, standardized buildpower and the shield now consumed energy to charge.

ShadowWolfTJC:
Although I've got a few critical bones to pick about the unit's gameplay concept, for the time, I'll focus on the unit's balance.

For starters, I'm worried about how OP a 4500-cost unit with more shield health than a Paladin would be, even if the shield couldn't self-repair at all. I mean, the Aspis's shield has 3600 HP, and it costs as much as a Jack, which has 6000 HP, but unlike the Jack, it can use that shield health to help protect other, more fragile units, such as a few Rogues, from harm. Imagine needing to bring enough firepower to bring down a freaking Paladin, just to bring down such an OP shield, from a unit that costs less than half of a Paladin!

Licho1:
I wanted to make proper weapon-less trider - the tunneler.
It was meant to be a strider that will terraform its way slowly through everything, with a shield in front (might just be high hp/armor plate) that could defend units behind it. Armed only with nano turrets. 

Abilities like ress or repair would fit this tunneler much better. As a bonus it would be useful against terraformed megaporc (mostly seen in FFA)

GoogleFrog:
This sounds like a bit of a nightmare for consistency. How do you show people that this shield has 10x the HP of other large shields? How do you deal with linking?

GoogleFrog:
Is the intention that it can only reclaim and repair? There is also an inconsistency there, as all other sources of BP can do everything. This sounds bad for the UI.

GoogleFrog:
@Licho1 I am talking about the PR, not your off-topic terraformer. I am aware of the idea of the ramp-tank from discussions over some years. I wonder if we ever made a ticket, or if such a ticket was made on this particular repository.

Licho1:
Well personally I don't want to change funnel

danfireman:
@GoogleFrog : Thanks for taking the time to look at this! :)
In order:
The shield has a different look (SmokeDragon has stated he may be able to improve on the current placeholder big green shield) and has now been made to not link with other shields.
As per recommendations from @DeinFreund and others it now has fairly standard buildpower, though, like the Reef, it cannot terraform (happy enough to change this if you think it's important).

I'll be updating the top post to actually describe current state...

DeinFreund:
In order to make such a mostly shield-based strider viable, the shield has to function differently from existing shields. Otherwise you'd either just be making an upscaled aspis or an effective anti-artillery defense. The modified recharge mechanics implemented here allow for the shield to temporarily push against artillery, while still losing to it during prolonged defense. In order to allow for such different recharging behavior, the shield has to be clearly (visually) distinguishable from other shields either way. At that point I think it's justifiable for linking to behave differently as well (not share).

danfireman:
One possible solution to the shield regen issue is to use Spring's shield regen delay system and have the unit_shield_charge code monitor how much regeneration actually happened last frame and set an energy cost accordingly, reducing the shield strength according to the energy deficit.
This will technically mean that it won't charge quite as much energy as it should in the frame following a hit when it's recharging, but due to the regen delay such situations can only occur at most once every 60 seconds (and likely a lot less) and should have no noticeable effects on gameplay or UI.

GoogleFrog:
You should not use any shield charging provided by Spring if this shield costs energy to change, since it needs to be handled by the shield gadget to implement priorities. Adding a shield charge delay is fairly easy with the shield hit callin.

danfireman:
@GoogleFrog : I didn't know there was a "shield hit call-in", my previous attempt to find such a thing failed.
I was planning on using the prioritization code in Zero K to determine energy expenditure (shieldpowerregenenergy itself would be 0) and reduce shield hp by the amount charged last frame if that frame didn't manage to use any energy.
Your idea sounds much less hacky though, so I've done some poking and implemented it. :)

gamedata/weapondefs_post.lua
```diff
@@ -319,9 +319,11 @@ for _, weaponDef in pairs(WeaponDefs) do
 		
 		weaponDef.customparams.shield_rate = weaponDef.shieldpowerregen
 		weaponDef.customparams.shield_drain = weaponDef.shieldpowerregenenergy
+		weaponDef.customparams.shield_recharge_delay = (weaponDef.shieldrechargedelay or 0)
```
GoogleFrog: shield_recharge_delay should never be set to 0. The gadget and widget code should first check whether shield_recharge_delay exists, and it if does, then do the additional checks based on it.

danfireman:
I've made a number of tweaks after trying out the campaign mission with some starting unit tweaks:
Build range 550 -> 450 (guards at proper distance, the 550 wasn't necessary anyway)
Can now terraform (rather than just assist terraforming started by a random con)
Movement speed 45 -> 54 (can now keep up with Dante)
Hitpoints 11000 -> 6000 (can now be Widowed)

danfireman:
I think this is basically ready barring feedback...
I also have modifications to the campaign mission to have it work with the new Funnelweb, I'll make an MR for that too.

GoogleFrog:
I like the idea of a big shield because it may make small units more useful. The inconsistencies required seem a bit steep though.

- The shield has more HP than a nuke, so it should intercept nukes. Currently it does not.
- The nanospray should emanate from somewhere on the surface, not just somewhere internal. Perhaps its face could be used.
image
- The wobbly purple shield stuff is clipping. Perhaps this stuff could emit from the gap that held its guns. A different type of effect could be used.
image
- There is no visual indicator that the shield is blocked from charging after being damaged. Perhaps the wobble could reflect this, as well as the colour or transparency of the shield.
- The shield texture has a very low resolution. Using a large version of the usual texture messes with the sense of scale.
image
- 35 BP is a weird number, why not 30 or 40?
- The health colour progression is not consistent with smaller shields, I think it is probably better to retain consistency. Perhaps it could be consistent at low levels and then go to bright green.
- I don't really like a lack of link. Perhaps another link rank could be added.

danfireman:
Thanks @GoogleFrog !
A few questions:
1. What do you mean by nuke interception? A nuke will hit the shield, but the blast wave will go through, doing 60-90% of its normal damage to things under the shield.
2. You thinking red/yellow -> blue/green -> green/X for color? I worry that looking too much like other shields will make its different linking behavior less obvious (though see below).
Should the blue/green be around where it currently is for Aspis (eg. ~4000hp)?
3. I'm not too sure what you have in mind. But I think it's important to note that any extra behaviors are added complexity. Admittedly shield linking is already too complex for anyone who hasn't read the wiki/code to understand and even then having a good model for in-game use is difficult, so it might not make too much difference.
Were you thinking something like shield transfer going at 1/10th the normal rate? That would still result in some rather fast (150hp/s) transfer to normal sized shields, but not the Felon-breaking amount it might otherwise get.
Or were you thinking something like any shield charge over 4000hp is treated like 4000 for linking purposes?

springraaar:
In my opinion, it could be just a bigger shield on a tougher chassis that also does other stuff, but it should keep the same mechanics as the other big shields, just bigger and stronger (but with less shield/cost).

The end result of this may be more people realizing that aspis are crazy good if they have very high energy income and making groups of them more often, which still have pretty good shield/cost while having no such restrictions and inherent vulnerability single units have.

danfireman:
@springraaar 
Same mechanics is bad for two reasons:
1. Linking is proportional to the hp difference between shields, leading to a high hp shield being quite OP for a short time and then spending the rest of the game hovering at 10%.
2. Why get Funnelweb when you can Aspis? They need to be distinct in important ways.

GoogleFrog:
Linking consistency is problematic. On balance I think we're better off without Funnelweb linking but I don't like the inconsistency involved.

The biggest issue for me is that a Funnelweb at the back deletes Missile Silo. Shockley only deals 10k damage and this Funnelweb has 24k charge so not even Shockley is particularly effective. The only answer I can see is sustained Bertha bombardment to keep the shield down.

danfireman:
So, to confirm, you'd rather no linking than a linking cap (at around 4000hp) or linking scaling?

As I mentioned in the thread, the Missile Silo interaction is a concern.
I am unsure whether the extra layer of defence the Funnelweb adds is a good thing or a bad thing. It's still possible to cost effectively penetrate with missile silo (e.g. 2 shockleys will bring the shield down enough the 3rd can go through) and Bertha is indeed a hard counter if you try and have your Funnelweb defend your base (and naturally Funnelweb does nothing against air attacks).
I considered giving Shockley shield penetration.
Though I will also note that the fact missile silo is pretty much uncounterable currently, and this has troubled me in the past. On mid-large maps things can get very porcy, with the silos preventing the setting up of defences (and other silos) within its range. With Funnelweb you could actually porc creep (or even better, launch a proper assault), rather than being stuck in an indefinite stalemate.

GoogleFrog:
I don't know about linking as all the options seem bad. If it links there should be a new link rank added to reduce the drain rate. Should linking disrupt charging? With multiple Funnelwebs you could hold some in the back to charge the damage-receiving shield in front. Bad links will be hard to avoid which such a large radius.

On balance I think I want to avoid linking.

danfireman:
@GoogleFrog : Can we defer the majority of the visual improvements to a later date?
I don't want to do a large amount of work I have little aptitude for (visual things) when we don't even know if the concept works. Once the players have decided whether the idea has merit it can have polish added.
What do you think? Are there visual things that are Must Have before trialing it?

Unrelated: It didn't make sense to have an expensive, high buildpower, mobile unit that needed another unit to start building near it, it just added extra tedious micro. So I made it able to build like a normal constructor.

Lasica:
If you want to make it be able only to repair and reclaim it could have old-style funnelweb drones that can't be controlled and do just that. Repair all around funnelweb automatically and reclaim when nothing to repair is near.

danfireman:
@Lasica : That means the shield gets taken down by AA.

Lasica:
They could be in form of ground units - little spiders or be radar jammed so you need vision to target them.

danfireman:
Little spiders is beyond my level of skill and energy to make. I can imagine them being quite difficult to do.
Radar jamming doesn't help thanks to flea, owl, scythe, etc.

Aquanim:
I have concerns that this unit will be very effective on maps like Icy Run (i.e. where there is only one narrow front to shield) and not very effective elsewhere. I also have some concerns about the interaction vectors it potentially nullifies. I'm not really sure how to verify anything other than "suck it and see" though.

Anarchid:
> 2 shockleys will bring the shield down enough the 3rd can go through)

However, to get an Eos through you'll need two Shockleys and another Eos (because the third Shockley would go through without damaging the shield). That's half of the unit's cost in missiles, and worthless without a follow-up of some sort.

> Bertha is indeed a hard counter

I don't think so. For one Bertha, you could have two of these super-shield funnelwebs and rotate them out, especially if they don't link. A single Funnelweb's recharge rate is 200, vs a single Bertha DPS of 285. While this still will wear down the Funnelwebs over a long run, adding even a single Aspis/Aegis to take 1-2 shots meanwhile will tip the balance.

danfireman:
I could really use a bit of help making the shield graphics a tad less sucky if anyone has any GLSL experience...?

@Aquanim : I am not completely sure myself, but think it worth a shot for a strider slot that has languished for over half a decade.  I found it pretty fun to use, for whatever that's worth

@Anarchid : The third shockley is to hit and paralyze the Funnelweb itself (and likely most of what it was guarding). Yes, a follow up would definitely be required.

Hmm...
I hadn't realized that the weight reduction drastically changes the calculus with respect to Bertha neutralization. It's still cost effectively neutralized (particularly as Funnelweb shields start uncharged, so your 5K metal bertha neutralizes and eventually kills 7K of 2xmicrod Funnelweb+singu/4) but it's a soft counter rather than a hard counter.
I still think people are unlikely to just keep Funnelweb's around for base protection and it'll prove most useful for assaults/raids. (If they do keep it as base protection a lot then it's time for another unit rethink/redesign.)

Note: Aspis is still much much more cost effective against Berthas so I wouldn't really count "can survive Bertha's with Aspis protection" as a flaw.

lhog:
> I could really use a bit of help making the shield graphics a tad less sucky if anyone has any GLSL experience...?

Care to explain?
I did the current shields implementation in GLSL. Find me on Discord (ivand) if you need help.

Anarchid:
> I still think people are unlikely to just keep Funnelweb's around for base protection and it'll prove most useful for assaults/raids. (If they do keep it as base protection a lot then it's time for another unit rethink/redesign.)

In teams, maybe. In FFA i expect it to be a feature of base defense at nuke tier and above. Not only does it keep your antis safe from Shockley and Scylla, it also softens the hits from nuclear missiles that can get through.

danfireman:
@lhog : Mostly I just don't have much of a sense of aesthetics. :/

@Anarchid : I do agree that supershield!Funnelweb is a rather effective missile silo counter. Though I think in FFA you'll frequently have it out on raids due to it aiding in attritionless overwhelm.
Do you think the Funnelweb's anti-silo abilities merits one of the following:
1. Shockley doing 2x damage to shields so a single Shockley almost neutralizes a Funnelweb's protection.
2. Shockley penetrating shields, returning Shockley to it's one-shots-anything role (which I personally am not much of a fan of).

Anarchid:
Both Shockley penetration and Shockley double damage would be annoying exceptions.

I guess my preferred solution a type of Shockley that ignores shields by driving under them before firing, and has a massive AoE. 

I guess supershield!Funnelweb gives a niche to the nuke tank.

:P

danfireman:
I think Silo+3xSchockleys vs Funnelweb is something of a false comparison as Silos are just generally awesome and the Shockleys will probably have frozen whatever the Funnelweb was trying to protect (like that anti-nuke).

That said, I've upped the cost to 3500 and the shield hp to 28000, so the Funnelweb is now a bit worse in every respect.

danfireman:
I'm currently trying to decide between two different shield effects, 3D perlin noise going upwards or 4D perlin noise going timewise.
Thoughts?

danfireman:
So, I think this is good enough now.
Unless anyone has any further feedback?

GoogleFrog:
I don't see any impact location distortion on the big shield. Is this a bug?

The charge/not charging state change is perhaps a bit subtle. Although it is easy to spot once I know that there is a difference. The charging shield state looks cooler, but will likely be rare. I think all of this would be fixed with a mounted LUPS wobble that changes colour/intensity when able to charge. This would prompt people to notice that there is a difference between the two shield shaders.
image

danfireman:
I see impact location on the shield:
gunshipattack

I've added indicators now.

For the curious, here's what the Funnelweb shield looks like overall:
funnelwebstatic

Anarchid:
> I don't see any impact location distortion on the big shield. Is this a bug?

> I see impact location on the shield:

The word distortion effect is missing though, and it looks really good on the textured version.

lhog:
Impact distortion doesn't work because the 3D noise version doesn't sample with or use distorted coordinates. Also distorted sampling coordinates are 2D so they are not applicable as inputs to Perlin noise.

While certain heavy face-lifting is possible, overall, I think it's worth rewriting the shield shader, possibly in non-LUPS way.

Anarchid:
You can shift the point where you sample the noise from based on its distance from the real-world impact point, thus distorting even 3d noise.

danfireman:
Well I've added in a distortion effect, though because of the fact the shield surface isn't regular I had to go with a very different feel (feel free to play around with the distortion constant to get better effects):
funnelwebdistortion

shadoxxhd:
It might be cool if the shield generator effect would be located and shaped like the wing cases of beetles (energy effect or hull armor case over back of body; when recharging it separates and elevates somewhat, and energy wings fold out)
this is how it would look like closed:
ladybug closed
beetle closed
and opened:
ladybug opened
the wings would be purple energy wobble, of course; the wing case either normal hull or a "hard energy" hull section

shadoxxhd:
Shouldn't it be able to heal everything under it's shield?

Anarchid:
No.

danfireman:
@shadoxxhd 
Thanks for the love. :)

I don't think Spring supports non-spherical shields unfortunately. :(

The current design has the Funnelweb with normal constructor abilities. This is to minimize the number of unique behaviors it has. A good rule if thumb is that a unit should have no more than one unique behaviour, maybe two if it's an expensive special unit. I very much doubt the dev team would allow something with three or more unique traits.
Additionally, units need to be as simple as possible and have intuitive traits. A shield with two unique traits (no linking and recharge delay) and a construction ability is close to the limit of complexity for a mid-weight strider. Could potentially justify the outlaw weapon in terms of complexity, but ideally a unit should also have some big weaknesses to make balance easier. With an outlaw aura balance will become a lot harder as planes and raiders would no longer be quite as big a weaknesses.
Of course, depending on how it plays we may end up having enough leeway to add in that extra ability.

You are correct that a construction radius equal to the shield size would make more intuitive sense, however I feel that mitigating porc creep and making the guard behavior good (and low micro!) outway that.

shadoxxhd:
I don't want to make the shield non-spherical, merely change the shield active/shield charging energy wobble FX. Having wings as a visual shield status indicator would look better than just turning the flower petals FX from the lotus upside down/scaling them up.

Both outlaw and slow-on-enter weapons wouldn't significantly help it survive against air attacks; they would just make an attack against a funnelweb with heavy AA support more costly. Raiders would be nerfed a bit by it, though I find some slow necessary to prevent enemy armies from getting under the shield too easily (I haven't played with your branch of the game yet though, so take it with a grain of salt).
If you don't like the slow idea, perhaps you could add a slow firing fast projectile gun with low or no aoe (maybe felon gun with much lower fire rate? otherwise maybe some railgun-themed weapon) and a range slightly bigger on than the shield radius, just to prevent a single glaive from killing a 4500m unit, while swarms and air (except maybe gunships?) would still be a major threat.

Having a smaller con radius than the shield radius would certainly make writing a good guard AI more challenging, since you wouldn't want it to extend it's shield over enemy defenses just to heal your assault units, but would normally expect it to heal all units under its shield. Removing the nano feature completely and making the unit a bit cheaper might be preferable to it having a low range.

shadoxxhd:
Also, the intention of the 'apply slow when entering' weapon was to allow cloaked units to stay cloaked, but still impair then slightly by reducing their speed for a bit while greatly impairing enemy assault groups trying to get under the shield (though the amphibious slow bombs could probably be used to the same effect).

danfireman:
@shadoxxhd : 
Wings may be viable, but sound like more work than I'm willing to put in for the relative gain.

I think simplicity is a key aspect here. This is supposed to be a support strider, it's supposed to move with an army. If AA is needed it should be in the army, if slow is needed there should be outlaws in the army, etc.
And a Glaive certainly won't be killing this 3500 metal unit as it can just make a lotus in two seconds to kill the Glaive. Its construction abilities are good enough to fend off small assaults of <500 metal given a few seconds of notice.

In practice nothing but melee fighters will put defenses under the shield when attacking. And neither Jacks, Scythes or Halberds benefit from the shield (for different reasons) and shouldn't be used with it.

Slow-on-entering is a new mechanic. It will both be a bit tricky to code and is yet another thing to introduce new players to. I'd definitely want to see how the Funnelweb plays before coding it.

@GoogleFrog : Any more feedback?

shadoxxhd:
> 
> @shadoxxhd :
> Wings may be viable, but sound like more work than I'm willing to put in for the relative gain.

If you point me to where the particle magic happens, I might be able to make a nice looking effect (as long as it doesn't just render a model - I'm absolutely awful at 3d-moddeling, but decent with algorithmic visuals).

danfireman:
@shadoxxhd :
The relevant files are for the charging wobble are:
lups/ParticleClasses/ShieldSphere.lua
LuaUI/Configs/lupsUnitFXs.lua
The main shield is:
 lups/ParticleClasses/ShieldSphereHQ.lua

shadoxxhd:
@danfireman Where is the movement/initial velocity controlled? I'd need to assign each particle a specifically randomized main vector, and a parameter based on that. Then the individual particles would have to move depending on time, main vector and parameter.
lupsUnitFX only sets emitter location, particle class and colors; the ShieldSphere doesn't seem to update its location either.
How is particle movements controlled with lups?

danfireman:
@[shadoxxhd](/shadoxxhd): I don't know (I'd assume the vertex code), but I'd recommend asking in #zkdev.

GoogleFrog:
Here are some things to do.
- The wobbly shield effect immediately struck me as "laggy" because it does not interpolate between frames.
- The red balls of "no charging" look low quality and out of place compared to the usual shield balls.
- Small projectiles, such as Sling, have so little effect on the shield wobble that it is hard to tell if a wobble occurred. They should show an effect, perhaps there should be a minimum. Look at the other shield shader and compare Sling to larger projectiles.

GoogleFrog:
The "laggy" problem has gone away, and I don't know why.

GoogleFrog:
I rebased.

GoogleFrog:
The Slig magnitude problem is likely best addressed inside the shader code, so I'll not touch it for now.

GoogleFrog:
https://springrts.com/mantis/view.php?id=6146 makes me a little wary about this Funnelweb.

GoogleFrog:
Two remaining issues that I'm aware of:
- High occurrence of https://springrts.com/mantis/view.php?id=6146
- Sling barely makes a shield wobble. See the medium shields for a good scale of aoe+damage compared to wobble size.

GoogleFrog:
Sling impact wobble appears to work now.

GoogleFrog:
I was mistaken, single Sling impacts do not work, as they are barely visible, but multiple Slings can add up to a visible impact.

GoogleFrog:
The sphere of the graphics does not match the sphere of the collision volume for the shield.

GoogleFrog:
image

danfireman:
The collision volume and shield graphics mismatch is puzzling. Investigation has just made me more confused as to how shield centre positions are calculated, particularly as the centre seems to orbit depending on the orientation of the unit if the origin of the shield is not the unit's centre.

sprunk:
The physical centre is defined here https://github.com/danfireman/Zero-K/blob/bigShieldFunnelweb/scripts/striderfunnelweb.lua#L114 while the visual center is defined here https://github.com/danfireman/Zero-K/blob/bigShieldFunnelweb/LuaUI/Configs/lupsUnitFXs.lua#L108

danfireman:
@sprunk 
The initial position of the physical centre is indeed defined there, but it doesn't stay there and moves about depending on the orientation of the Funnelweb (as if the angle from the centre of the unit model (which is NOT the same as the origin for the shield graphic) is being doublecounted somehow).
The visual centre is defined here as far as I can tell: https://github.com/danfireman/Zero-K/blob/bigShieldFunnelweb/LuaRules/Configs/lups_shield_fxs.lua#L107

GoogleFrog:
I think that the shield should emit and draw from (0,0,0). If (0,0,0) is on the ground, unfortunately the case for some models, then it should be vertically moved to be closer to the models center of mass.

danfireman:
As far as I can tell the main shield visual/hitbox mismatch issue goes away with 104.0.1-1060-g7f3541f.
I've now adjusted the aimpoint to give as close a match as I can.

Regarding slings and hits, it looks like the damage output of a sling bullet is so low compared to the shield's hp that it rounds to 0 most of the time. This is similar to glaive bullets not always triggering a wave in an Aspis.
I'm struggling to think of a solution to this that doesn't have annoying implications...
Ideas?

lhog:
The gadget works in a way that hits get accumulated over time, so the visual feedback is eventually drawn even for low burst, but high DPS units. Since new shield has crazy charge values, my previous handcrafted constants might work in such way that damage is depreciated faster than it's accumulated.

I think the most reliable way be to retune the common constants so they work for all range of units. For example you might want to introduce tiered or progressive damage decay, so it's slow at low damage values and is getting higher the higher the accrued damage values becomes. 
If this is not possible then it's likely that some special path/unique constants should be used for high HP shields.

lhog:
Also ideally ZK should update the gadget code as the previous one was a bit redundant and partly wrong. 
Also @GoogleFrog wanted to include the terrain/units outline I put together for other game.

Let me know if you feel like you can take that. "No" is the perfectly valid answer ofc :)

GoogleFrog:
The new engine seems to work, which fixes some of the shield issues that I feel would destroy a Funnelweb which is so focused on its shield. I'll look into changing the shield damage constant and merging.

GoogleFrog:
The shield effect occasionally looks laggy https://youtu.be/vGSn9G4jxqo

GoogleFrog:
It appears to be related to unitID seeding. 30276 reproduces the lag.

Licho1:
nooo not without repair drones

lhog:
@Licho1 my man :+1:

PR #3250, opened 2018-09-24
Widget to remove mearth locations

ArchShaman:
This widget can remove crap that's broken from maps.

### DIFF
--- a/luaui/widgets/bad_widget_disabler.lua
+++ b/luaui/widgets/bad_widget_disabler.lua
@@ -1,15 +1,30 @@
+--[[Shaman's Policy:
+Do whatever you want with this code.
+]]
+
 function widget:GetInfo()
 	return {
 		name      = "Disable Bad Widgets",
 		desc      = "Disables broken widgets based on config.",
 		author    = "_Shaman",
 		date      = "09/24/18",
-		license   = "Resignateer license",
-		layer     = -9999999999999999999999999999999999999999999999999,
+		license   = "Shaman's Policy",
+		layer     = 5,
 		enabled   = true,
 		alwaysStart = true,
 	}
 end
+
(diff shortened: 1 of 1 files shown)


sprunk:
Tested? Layer looks suspicious.

ArchShaman:
>Tested? Layer looks suspicious.

Tested. Code executes as intended. I tried to make it load last, just to be on the safe side. Tested with layer = 5, works as intended too. So I'll change to 5.

>License is pretty much Horse again. If you wish to include files under the "Resignateer License", you should also provide text of this license.

For all intents and purposes, all my code, future, past, and current, falls under the policy of "do whatever with it". I'll change it to reflect this, if necessary.

ArchShaman:
Can I commit future code with "Shaman's policy" now that it has some text reference?

Anarchid:
"Do whatever" sounds very much like "public domain", so why not just use that?

ArchShaman:
Gonna add reasons to this.

>"Do whatever" sounds very much like "public domain", so why not just use that?

Cuz I don't/didn't know what "Public domain" is. I'll add it in with my reasoning thing I want to add.

ArchShaman:
[f=-000001] Disabled '*Mearth Location Tags*1.0' (Reason: Causes black ground on some cards, possible copyright issues.)
[f=-000001] Loaded widget:  Disable Bad Widgets  <widgetdisabler.lua>

ArchShaman:
Only problem that this may have is that there's no `widgetHandler:FindWidget` function, so on other maps it will complain about the bad widgets not existing. I may have to include a cawidgets.lua change to fix that.

sprunk:
e6f3c516faec7c088a17c084b7df852a4ffb29ce

PR #3282, opened 2018-10-17
Cyclops hit effect aoe

danfireman:
Changing the Cyclops visual hit effect less dissonant with what it is by game mechanics.
This has the unfortunate side effect of making the Cyclops bullet explosion a lot less cool. :(


Anarchid:
A screenshot wouldn't hurt ;)

Anarchid:
Looks good.

PR #3291, opened 2018-10-21
WIP: add the phytophyte hunter model 

Anarchid:
-[x] aims 
-[x] shoots units
-[x] dies
-[ ] leaves rek
-[ ] shoots terrain (WTF?)


Anarchid:
Not shooting terrain seems to be a feature. Zero initial projectile velocity is either a Spring bug or something the current Hunter does also.

This leaves the wreck.

PR #3309, opened 2018-10-31
Fixes solar opening too early when attacked by flea

### LINKED ISSUE #3304: Solar tries to open too early
Get a Flea to attack a Solar. Every 7 or so seconds it will attempt to open. Ideally each new damage instance should reset its automatic opening timer.

n42k:
Closes #3304.

When the solar is in armored form, the flea attacks it for 0.88083333 damage, which isn't enough to trigger the defensive maneuver thread, which starts a cycle of the solar eventually going into unarmed form, taking a hit, triggering it, and going back into armored form.

### DIFF
--- a/scripts/energysolar.lua
+++ b/scripts/energysolar.lua
@@ -117,7 +117,7 @@ end
 
 -- this happens before PreDamaged but only in 97.0+
 function script.HitByWeapon(x, z, weaponDefID, damage)
-	if damage > 1 and not (weaponDefID and noFFWeaponDefs[weaponDefID]) then
+	if damage > 0 and not (weaponDefID and noFFWeaponDefs[weaponDefID]) then
 		StartThread(DefensiveManeuver)
 	end
 end


sprunk:
Some non-weapon stuff might have 0.0001 damage for technical reasons (kinda like this https://github.com/ZeroK-RTS/Zero-K/blob/master/units/bomberprec.lua#L185 but not negative), make sure this behaves correctly. Alternatively just make the magic constant `0.8`.

n42k:
I looked through all units with `/luarules give`, especially the ones that don't do standard damage, and this is what I found:

Imp: closes, ~2033 damage
Gnat: closes, ~1000 damage
Thunderbird: does not close, 0 damage
Dominatrix: does not close, does not call damage function
Racketeer: does not close, 0 damage
Widow: closes, ~8000 damage

I conclude that it is safe to set the value to 0. Do note that EMP units close the solar panel, which might be unwanted behavior, but this should be dealt with in another Issue/PR.

sprunk:
Great, thanks! (If any case is still found it is probably the offending weapon that should get changed.)

PR #3363, opened 2018-11-29
Commander slow damage buff

danfireman:
Currently the "Disruptor ammo" upgrade for the commander machine gun and shotgun are substantial downgrades and disappointing to all the newbies who try them (myself included).
Proposal:

Machine gun damage: 15 + 7.5s -> 15 + 30s  (compare to 30 damage for normal machine gun)
Shotgun damage: 16 + 8s -> 16 + 32s  (compare to 32 damage for normal shotgun)

Puts it more in line with the disruptor upgrade for the light particle beam:
165 dps -> 82 + 165s dps


GoogleFrog:
Sure, we can see what comes of it.

PR #3365, opened 2018-11-29
strikecom model shows its primary gun 

Anarchid:
This is intended to showcase the feasibility and request feedback for the general idea of commanders showing what they are.  Actual weapon models and textures are rapid-prototypes and to be iterated on before cleaning up and making them properly nice. 

Code is same, though possibly more reusable.

Play around and see at which zoom levels you can tell commander guns apart, and how this can be improved.

Equipping the other commanders with similar stuff should be possible, but i'm not sure this is better than going full omnicom.

image


GoogleFrog:
There is a merge conflict.

Anarchid:
Fixed merge conflict.

GoogleFrog:
Minor things.
- Beamlaser, Shotgun, Lighting seem fine.
- LPB has two nozzles for no apparent reason.
- Machine gun lacks an elbow joint.
- Missile launcher does not have size consistency. It looks like the model for the missile launcher is a large version of a much smaller model.
- All of the firepoints are broken.

image

The weapons are all too long. Apart from making the comm look like it will overbalance it creates technical problems. Once weapons start poking outside the hitvolume you run into trouble with occlusion and aiming. I have fought the bugs caused by the Engineer Chassis long barrel for quite a while and the current solution is neither perfect nor pretty. The Strike weapons can probably be a bit longer than in master, but not by much.

The other major issue is how the secondary weapon is going to be handled. The other hand is also used for nanospray.

Anarchid:
> The weapons are all too long

One of the concerns when the idea was aired was that guns will be too small to be visible. If they can be made smaller while considered visible enough, that's fine.

> The other major issue is how the secondary weapon is going to be handled. The other hand is also used for nanospray.

This is one of the reasons i'm not sure doing this rather than properly designed from-the-ground omnicom is a good idea.

I don't think the idea that second arm shares the purposes of holding the second gun and nanoturret a good decision. I'd move either the second gun to be shoulder-mounted, or move the nano to be shoulder-mounted.

Shoulder-mount nano makes more sense IMO because it's extremely common for a commander to be facing one way to shoot while building something in another direction, which makes the under-arm nanos on all commanders look ridiculous.

Anarchid:
Was intended mainly as demo; the purpose is served.

PR #3378, opened 2018-12-08
Altered 1v1 start boxes for Into Battle v4

Aquanim:
1v1 is locked to south-west vs north-east. Start boxes for games with more than 2 players are unchanged.

### DIFF
--- a/LuaRules/Configs/StartBoxes/Into Battle v4.lua
+++ b/LuaRules/Configs/StartBoxes/Into Battle v4.lua
@@ -1,0 +1,72 @@
+local duelbox = {
+	[0] = {
+		nameLong = "West",
+		nameShort = "W",
+		startpoints = {
+			{1250,4800},
+		},
+		boxes = {
+			{
+				{800,5030},
+				{1320,4360},
+				{1470,4420},
+				{1635,5055},
+				{900,5140},
+			},
+		},
+	},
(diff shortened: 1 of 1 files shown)


LuaRules/Configs/StartBoxes/Into Battle v4.lua
```diff
@@ -0,0 +1,96 @@
...
+				{6144-0,6144-0},
+				{6144-1536,6144-0},
+				{6144-1536,6144-6144},
+				{6144-0,6144-6144},
+			},
+		},
+	},
+}
+
+-- there's probably a better way to get player count... but this is known to work
```
sprunk: `Spring.Utilities.GetTeamCount()`

Aquanim:
Don't pull this just yet, I think it doesn't actually work.

sprunk:
In that case `if #Spring.GetTeamList() == 3` should do the trick (specs don't have a team, each real player has a team, the 3rd team is Gaia) or if that doesn't work for some reason then what Adansonia does:
```
if #Spring.GetTeamList(0) == 1 and #Spring.GetTeamList(1) == 1 then
```

Aquanim:
Verified that 2v2 and FFA results in the old boxes while 1v1 results in new boxes after this commit.

PR #3393, opened 2019-01-12
Prevent Halberd armoring up while disarmed

### LINKED ISSUE #3376: Disarmed blockade runners should armor up
The first few minutes of this battle has a Halberd attack on the north-east island that could have done serious damage if the Halberds had armored up once disarmed.
http://zero-k.info/Battles/Detail/653049

mvollrath:
Fixes #3376 by getting around to fixing the LUS.

Normalize Halberd armor up behavior with other units.


mvollrath:
Sleeps are probably not necessary, copied this from Razor.

PR #3395, opened 2019-01-12
Cloakbots animations

mvollrath:


mvollrath:
Here's a [video of the new animations](https://youtu.be/lKbzvkce65E).

mvollrath:
Here's [a video of some more tweaks to Glaive, Ronin, Gremlin, and Phantom animations](https://youtu.be/ZbYt9vVEjIs).  I've added idle animations to Glaive and Gremlin and some other subtle tweaks like being able to lead with either the left or right foot.

Anarchid:
I think thr new glaivr and scythe run cycles leg it a bit too much, skipping the phase where the character has both feet off the ground.

The walk cycles - which is all units except scythe and glaive here - look much inproved.

mvollrath:
Here's [the new Glaive](https://youtu.be/WC-BE9_M_tQ).

mvollrath:
[Had to do the Conjurer too](https://youtu.be/Fje5c4-Cdfg).

mvollrath:
[Video](https://youtu.be/Yv1-wyT4Z0U)

Revise animation scripts for most units in Cloakbot factory.

Overall, I tweaked the animation speeds to match the movement speeds of the units, so they look more natural. There's also new logic to prevent future balance changes (movement speed) from breaking the animation rate.

Many units now have additional motion in their running animations or completely redone animations. Each unit has its own character.

A recurring bug was running animation speed not being affected by movement speed modifiers. Should be fixed for all of them now.

GoogleFrog:
Looks good and I like the style in the code changes. The Glaive is especially improved since the WIP video I saw earlier. You've applied a bit more attention to detail than I have previously used in similar cases, and it is impressive as a weekend of work. I hope to see more things like this, perhaps commanders or striders could be improved.

sprunk:
I really like those as well, CJ!

PR #3410, opened 2019-02-12
Making lotus fire graphics better fit their damage

danfireman:
As discussed in this thread: http://zero-k.info/Forum/Post/196343#196343
Figured I'd put my money where my mouth is and make an effort on some of the issues identified, so here is an attempt to make the Lotus' graphics better reflect it's actual DPS by making it red (a color generally associated with lower energy than green/white).
Old lotus:
oldlotus
redlotus


sprunk:
I like the idea of changing LLT colour from the same as Commander Beamlaser to the same as Locust because it is a fair bit closer to the latter in performance (objective). Red LLT is also super oldschool and I like the energy connotation (subjective).

On the other hand: LLT used to be red in the past (and HLT green, and Lucifer deep blue - most lasers followed the energy level association you mention, and this was OTA legacy). I believe the reason for the change to cyan was to differentiate the factions which is no longer relevant so in that regard the change should be fine though.

sprunk:
Things recently said in chat (so that they don't become tears in the rain):
 * GoogleFrog wants LLT to look scary
 * the precedent for a red continuous beamlaser is Locust, but all other continuous beamlasers use a cool colour scheme; likely it is Locust who should become cyan instead

GoogleFrog:
This is one of those PRs where I see little reason for the change. The cyan colour seems fine to me. It feels stronger than Locust and weaker than Mace, which is a good feeling to convey.

danfireman:
> The cyan colour seems fine to me.

Would it be worthwhile for me to do a poll to get what sentiment people have towards colour coding?
This is one of those things that's likely to be slightly offputting to newbies and seem perfectly natural to veterans.

sprunk:
I think what makes LLT feel weaker than Mace is the (comparative lack of) sound and thickness rather than the colour.

About PR: what about Commander beamlaser? That one's also cyan so would need to be changed under your "energy levels going red-green-blue" design.

danfireman:
I think I should point out that the issue isn't just colour, it's beam thickness, hit effect width and how many sparks fly out (and how far they fly).
@GoogleFrog : Would you be happy to merge in something that just toned down the hit effect? Making the floor patch smaller than the com laser patch and with less sparks.

@sprunk : It's also 2x the damage of the Lotus.
Lotus is 26% more damaging than Locust, definitely in the same weightclass, while the com laser is 112% more damaging than the Lotus. Lotus is MUCH more similar to com than to Locust.

For reference laser beam DPS is:
Locust: 59
Lotus: 71
Commander: 150
Mace: 297
Stinger: 1062
Lucifer: 4000

There are a number of colour schemes in use in media, but most go red->...->blue, likely due to black-body radiation.

I'm honestly not hugely fussed about what the colour scheme is as long as it's consistent and will communicate the relevant information (how damaging is this laser?) to newbies.
Though "GoogleFrog wants LLT to look scary" makes me think... Do we want to slightly mislead newcomers about how powerful a Lotus is? LLTs aren't particularly scary mechanically and are avoided much more by newbies than veterans (the latter having learnt that a single scorcher can kill a Lotus).

sprunk:
This doesn't seem like it's going to get anywhere in its current form.

PR #3432, opened 2019-02-24
Improved DoF widget

dvr333:
Incorporated a separable Frostbite 2-style DoF implementation into the DoF widget.

### DIFF
--- a/LuaUI/Widgets/Shaders/dof.fs
+++ b/LuaUI/Widgets/Shaders/dof.fs
@@ -90,7 +90,7 @@ float LinearizeDepth(vec2 uv){
 
     float n22 = viewProjectionInv[2][2];
 
-    return abs(((1 + n22) * (-1 + 2 * n22 + depthNDC)) / (2 * (n22 + depthNDC)));
+    return abs(((1.0 + depthNDC) * (1.0 + n22))/(2.0 * (depthNDC + n22)));
 }
 
 vec2 GetFilterCoords(int i, vec2 uv, vec2 stepVal, float filterRadius, out int compI)
@@ -140,7 +140,7 @@ void main()
     gl_FragData[0] = fragColor;
   }
 
-  else if (pass == HORIZ_BLUR_PASS)
+  else if (pass == HORIZ_BLUR_PASS) //Really Vert
   {
     vec2 stepVal = 1.0/resolution.xy;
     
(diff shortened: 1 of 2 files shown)


LuaUI/Widgets/Shaders/dof.fs
```diff
@@ -90,7 +90,7 @@ float LinearizeDepth(vec2 uv){
 
     float n22 = viewProjectionInv[2][2];
```
lhog: The formula below expects projection matrix, not the inverse of (viewProj)

dvr333: Oh, right, the name is misleading. If you look at the Lua you'll see it's taking the "projection" matrix now, but I didn't want to deal with renames at 2AM.

lhog: Sorry missed that.

sprunk:
Got a comparison screenshot?

LuaUI/Widgets/Shaders/dof.fs
sprunk: At least comment in the code would be good.

lhog: I think the @dvr333 answer implies he's going to fix name when it's not 2 AM :)

dvr333: It's no longer 2AM, and the problem is fixed, along with a more pressing problem of the correctness of the blur, both at the edges of in-focus objects and in general. Turns out glsl's "inout" is the equivalent of C#'s "out", which threw me off.

lhog: Not quite equivalent. All values are still copied (not passed by reference), so `inout` means an input and output copy of value is made. `out` means only output copy is made. `in` AFAIK has no separate meaning from just default behavior (input copy is made)

dvr333:
> 
> 
> Got a comparison screenshot?

Here's an [album](https://imgur.com/a/0y10VbT). The first 8 shots are more artistic shots to really show off the effect (I did actually have the old effect on in all the shots that look normal, but it's really subtle), the last 4 are taken from angles that are representative of how the camera will be placed in-game while playing normally.

GoogleFrog:
I'm adding WIP until it is clear that this is ready. Make sure you can rename it, and rename it when it is ready.

dvr333:
I can't rename it, but I'll put a comment here when I consider it ready. It's pretty close now. I have an experiment to improve the look of objects nearer to the camera than the focus plane, and depending on that I'll decide whether autofocus should be realistic (success) or more of a tilt-shift effect (failure).

sprunk:
@dvr333 pinging to ensure you're aware of the comment above.

dvr333:
Yep, I got it. I'm going to do a bit of testing and tweaking over the weekend (though to be fair working on this did hurt my hands a bit more than expected), if it's done, the "WIP" will come off.

dvr333:
Sorry, misclicked on the "Close and Comment" button.

dvr333:
Todo before removing WIP:

- [ ] Ensure effect is resolution-independent
- [ ] Add proper blur for near-to-camera objects

dvr333:
Currently debating whether to actually have the proper blur for near-to-camera objects. Currently objects nearer to the camera than the focus plane don't look half bad. It's not perfect, but it's still pretty clearly a blur. I might give myself a couple days to see if I can work out why my current attempts aren't working, but if I can't I'll just take off the WIP anyway, as this near-to-camera object blur is intended as a "high quality" default-off option anyhow, and I don't see the point of holding up releasing this on an optional default-off feature.

PR #3471, opened 2019-03-18
Fixes for new Hunter model 

Anarchid:
- load shinies, bringing texture closer to others (gunmetal vs plastic)
- turret reloading animation 
- fix torpedo orientation 
- wreck was inside you all this time


Anarchid:
image

sprunk:
Now the Hunter is quite ovoid while the other boats are fairly angular.

GoogleFrog:
image
Torpedo comes out the side of the launcher and does not have a matching model. The torpedo can be seen in the launcher immediately after it fires.

GoogleFrog:
Why have there been such dramatic balance changes to the projectile?

Anarchid:
This is the reason for the physics changes:
image

 Fwiw current Hunter does the same, but noone bears to look at it anyway, so it's tolerated.

Anarchid:
- fixed torp reappearing immediately
- slightly toned down the shiny
- replaced projectile model 

Further texture fixes will require a full re-unwrapping of the model because of too many repeating elements - e.g. it's not possible to paint one of the hulls without also painting the superstructure.

cc @Phytophyte

GoogleFrog:
image
Still fires out of the side.

Anarchid:
Side torp is no longer, also at 120% original scale it's now about as long as Cutter while slightly wider, appropriately for its slightly larger cost.

Texture feedback so far is that it needs more greeble of similar quality and quantity as Cutter, plus Cylon robot eyes. This wil need a re-unwrapping. If @Phytophyte can do that, that's great, otherwise i'll probably deal with that over the weekend.

Phytophyte:
Re-unwrap the Hunter model with minimal overlap and redo the texture?  I can give it a go.

I'll ask again, is the 512px size fine, or should we go for 1024px if we're redoing the texture?

Anarchid:
512px is fine. I did a test where i cut away a symmetric half of the model and auto-unwrapped it in Blender (which is not very efficient, but very effective) - there were enough pixels to fit the Cylon scanlines into the "face", so i think that's enough pixels.

sprunk:
Any recent advances? Somebody else has just started yet another Hunter remodel.

danfireman:
Any chance this'll get merged soon? Even as it stands it seems to be a decent improvement on the current Hunter model.

Anarchid:
Unlikely given criticism, low mana, aesthetic superiority of the [other attempt](https://www.youtube.com/watch?v=rnweNG5N8uY&feature=youtu.be) and cursedness of Hunter.

GoogleFrog:
I was unaware of the other attempt. I like the shape a lot better.

Anarchid:
Obsolete.

PR #3476, opened 2019-03-21
WIP: fix nuke explosion not appearing on unit hit

Anarchid:
Untested blind edit.

Bug seems to occur here: https://www.youtube.com/watch?v=6mTYu0A04uk


Anarchid:
Testing confirmed issue, but this fix doesn't work. And nothing else i tried could make the CEG work.

PR #3483, opened 2019-03-23
fix #3477 (nuke explosion not spawning for shield and unit hits)

### LINKED ISSUE #3477: Nuke impacting a flying unit at altitude over 20 elmo spawns no effect
See https://www.youtube.com/watch?v=6mTYu0A04uk

To replicate, place an enemy Nimbus over the silo and launch. Krows and Athenas will still trigger the CEG because they're low enough that the `ground` bit triggers.

The attempted solution in https://github.com/ZeroK-RTS/Zero-K/pull/3476 does not work.

Anarchid:
Gadget solution. Automatically also handles Funnelweb shields (which were also bork, like any other non-anti intercept above altitude 100).

Increased altitude valid for a "ground" hit because imo nuking a Nimbus should still trigger a ground version of the explosion.

### DIFF
--- a/LuaRules/Gadgets/exp_nuke_effect_chooser.lua
+++ b/LuaRules/Gadgets/exp_nuke_effect_chooser.lua
@@ -1,10 +1,8 @@
--- $Id: exp_no_air_nuke.lua 3171 2008-11-06 09:06:29Z det $
-
 function gadget:GetInfo()
   return {
-    name      = "NoAirNuke",
-    desc      = "Disables the custom nuke effect, if the nuke is shoot in the air.",
-    author    = "jK",
+    name      = "Nuke Explosion Chooser",
+    desc      = "Chooses which nuke explosion to spawn based on altitude.",
+    author    = "jK, Anarchid",
     date      = "Dec, 2007",
     license   = "GNU GPL, v2 or later",
     layer     = 0,
@@ -15,6 +13,8 @@ end
 local GetGroundHeight = Spring.GetGroundHeight
 
 local nux = {}
+local defaultSuccessExplosion = [[LONDON_FLAT]]
+local defaultInterceptExplosion = [[ANTINUKE]]
 
 if (not gadgetHandler:IsSyncedCode()) then
   return false
@@ -27,7 +27,7 @@ for i=1,#WeaponDefs do
 	local wd = WeaponDefs[i]
 	--note that area of effect is radius, not diameter here!
 	if (wd.damageAreaOfEffect >= 800 and wd.targetable) then
-		nux[wd.id] = true
+		nux[wd.id] = wd.damageAreaOfEffect
 		wantedList[#wantedList + 1] = wd.id
 		if Script.SetWatchExplosion then
 			Script.SetWatchExplosion(wd.id, true)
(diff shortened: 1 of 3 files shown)


LuaRules/Gadgets/exp_nuke_effect_chooser.lua
```diff
@@ -42,8 +42,10 @@ function gadget:Explosion_GetWantedWeaponDef()
 end
 
 function gadget:Explosion(weaponID, px, py, pz, ownerID)
-	if (nux[weaponID] and py-GetGroundHeight(px,pz)>100) then
-		return true
+	if (nux[weaponID] and py-GetGroundHeight(px,pz)>200) then
+		Spring.SpawnCEG(defaultInterceptExplosion, px, py, pz, 0, 0, 0, nux[weaponID])
+	else
+		Spring.SpawnCEG(defaultSuccessExplosion, px, py, pz, 0, 0, 0, nux[weaponID])
```
sprunk: IMO leave the success case to be unhandled. This way somebody can tweak the value in the weaponDef and it will take effect even without being aware of this gadget.
```suggestion
		return false
```

units/staticantinuke.lua
```diff
@@ -87,7 +87,8 @@ unitDef = {
         subs    = 75,
       },
 
-      explosionGenerator      = [[custom:ANTINUKE]],
+      --spawning the intercept explosion is handled by exp_nuke_effect_chooser.lua
+      --explosionGenerator      = [[custom:ANTINUKE]],
```
sprunk: IMO it would still be good to have some sort of effect here because this projectile can also be bodyblocked by units (and instagibs most aircraft).

sprunk:
The `unit = true` bit should probably be propagated to `LONDON_FLAT`'s sub-explosions as well while we're at it.

units/staticantinuke.lua
Anarchid: There's a default explosion that antinuke has if not specified, but i guess `lrpc_expl` is prettier indeed

Anarchid:
This approach does not require or make use of the `unit` bit, and the unit bit does not trigger on funnelweb shields anyway.

LuaRules/Gadgets/exp_nuke_effect_chooser.lua
```diff
@@ -42,8 +42,10 @@ function gadget:Explosion_GetWantedWeaponDef()
 end
 
 function gadget:Explosion(weaponID, px, py, pz, ownerID)
-	if (nux[weaponID] and py-GetGroundHeight(px,pz)>100) then
-		return true
+	if (nux[weaponID] and py-GetGroundHeight(px,pz)>200) then
```
sprunk: Funnel shield is pretty tall (weaponDef says 550 radius at 45 emitHeight) so 200 can still result in the interception gfx:
image


```suggestion
	if (nux[weaponID] and py-GetGroundHeight(px,pz)>600) then
```

Anarchid: I think 600 is too much - at some point a Funnelweb interception actually *is* an interception - even if partial one, with a reduced rather than absent ground aoe radius.

sprunk: Alright. Another question is sea: `GetGroundHeight` returns negative values there so at deep enough sea you can get the interception fx  when nuking ships:
image
```suggestion
	if (nux[weaponID] and py-math.max(0, GetGroundHeight(px,pz))>200) then
```

Anarchid:
I think it makes sense to commit additional refinements directly tbh

LuaRules/Gadgets/exp_nuke_effect_chooser.lua
Anarchid: Good catch!

PR #3492, opened 2019-03-28
Fixed several DoF issues

dvr333:
* Fixed issue where the autofocus would incorrectly weaken when zooming in after a certain point
* Fixed colours blending to a mush (unfortunately this means some banding in dark regions, not sure how to fix without ruining the more pronounced circles)
* Normalized depth calculations to Spring unit distances, so blur power is no longer map-dependent
* Fixed projectile lights rendering on top of DoF (this involved changing priority, and this seems like a blunter fix than necessary)
* Expanded blur radius limits so near-to-camera objects blend more nicely

### DIFF
--- a/LuaUI/Widgets/gfx_dof.lua
+++ b/LuaUI/Widgets/gfx_dof.lua
@@ -6,7 +6,7 @@ function widget:GetInfo()
 		author    = "aeonios, Shadowfury333 (with some code from Kleber Garcia)",
 		date      = "Feb. 2019",
 		license   = "GPL, MIT",
-		layer     = -1,
+		layer     = -100000000000,
 		enabled   = true
 	}
 end
@@ -127,6 +127,8 @@ end
 -- Global Vars
 -----------------------------------------------------------------
 
+local maxBlurDistance = 10000 --Distance in Spring units above which autofocus blurring can't happen
+
 local vsx = nil	-- current viewport width
 local vsy = nil	-- current viewport height
 local dofShader = nil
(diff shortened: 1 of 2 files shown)


LuaUI/Widgets/gfx_dof.lua
```diff
@@ -6,7 +6,7 @@ function widget:GetInfo()
 		author    = "aeonios, Shadowfury333 (with some code from Kleber Garcia)",
 		date      = "Feb. 2019",
 		license   = "GPL, MIT",
-		layer     = -1,
+		layer     = -100000000000,
```
sprunk: You said this is to make it run before projectile lights but `gfx_projectile_lights.lua` has `layer = 0`?

https://github.com/ZeroK-RTS/Zero-K/blob/958ce4d4569d985e48f8a6af023e2a1f69351ebe/LuaUI/Widgets/gfx_projectile_lights.lua#L12

Regardless, add a comment about the value since it is meaningful (can be just `-- before gfx_projectile_lights.lua` or such)

LuaUI/Widgets/gfx_dof.lua
```diff
@@ -185,11 +188,11 @@ function InitTextures()
 
 	baseBlurTex = glCreateTexture(blurTexSizeX, blurTexSizeY, {
 		min_filter = GL.LINEAR, mag_filter = GL.LINEAR,
-		wrap_s = GL.CLAMP_TO_EDGE, wrap_t = GL.CLAMP_TO_EDGE,
+		 wrap_s = GL.CLAMP_TO_EDGE, wrap_t = GL.CLAMP_TO_EDGE,
```
sprunk: Random whitespace change here and below.

LuaUI/Widgets/gfx_dof.lua
dvr333: The problem case is actually gfx_deferred_rendering, but I've added a comment for that now.

sprunk:
thanks!

Issue #2847, opened 2018-03-13
Feature request: count puppies and missiles spent in attrition widget


ashdnazg:

GoogleFrog:
And other bombs?

ashdnazg:
of course

PR #3526, opened 2019-04-19
Attrition counter now handles capture and suicide units

Anarchid:
Capture:
- Captured units are considered as losses for the original owner on capture event
- Deaths of units that are currently slaves are not counted anywhere because they were already worse than dead
- Units returned to original owner due to Domi death are considered as reverse losses for the original owner.

Suicides:
- Health check on unit death - used to handle factory cancel, morph, and accidentally also suicide - now only prevents *incomplete* units from being counted. 
- Morph is detected via `wasMorphedTo` UnitRulesParam instead
- Self-destructing and suicide units are counted as casualties.

### DIFF
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -454,7 +454,7 @@ function widget:UnitDestroyed(unitID, unitDefID, teamID, attUnitID, attDefID, at
 	
 	-- prevents factory-cancel from counting as kill
 	local buildProgress = select(5, GetUnitHealth(unitID))
-	if GetUnitHealth(unitID) > 0 && buildProgress < 1 then return end
+	if GetUnitHealth(unitID) > 0 and buildProgress < 1 then return end
 	
 	-- don't count morphed units
 	local wasMorphed = Spring.GetUnitRulesParam(unitID, wasMorphedTo);
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -376,8 +376,6 @@ function widget:UnitTaken(unitID, unitDefID, unitTeamID, newTeamID) --//will be 
 			doUpdate = true
 		end
 	else
-		-- this is most likely dead code because of the weird relationship between UnitGiven and UnitTaken
-		-- however there's no documentation i could find that explains it, so i feel unsafe otherwise.
 		if capturedUnits[unitID] and capturedUnits[unitID] == unitDefID then
 			Echo("<AttritionCounter>: unitTaken: captured by a new team while already captured, how peculiar")
 			-- unit has been captured before and is now captured by another owner. 
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -384,7 +384,6 @@ function widget:UnitTaken(unitID, unitDefID, unitTeamID, newTeamID) --//will be 
 			-- do nothing; shouldn't happen unless FFA, and this widget shuts down in FFA.
 		else
 			-- unit has now become mind-controlled for the first time; unitTeamID lost it
-			Echo("<AttritionCounter>: unitTaken: captured for the first time, file as casualty")
 			capturedUnits[unitID] = unitDefID; -- verify unitdef in case of ID recycling 
 			local buildProgress = select(5, GetUnitHealth(unitID))
 			local worth = Spring.Utilities.GetUnitCost(unitID, unitDefID) * buildProgress
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -449,6 +449,9 @@ function widget:UnitDestroyed(unitID, unitDefID, teamID, attUnitID, attDefID, at
 	-- if its also the same kind of unit, its safe to assume that it is the very same unit
 	-- else it is most likely not the same unit but an old table entry and a re-used unitID. we just keep the entry
 	-- small margin of error remains
+	
+	-- prevents morph and factory-cancel from counting as kill
+	if GetUnitHealth(unitID) > 0 then return end
 
 	if teamID == gaiaTeam then return end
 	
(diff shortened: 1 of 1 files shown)


Anarchid:
I have no idea why the health check was there originally, if anyone knows comment pls. Otherwise i'll merge in a day or two

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -353,27 +353,126 @@ end
...
+			allyTeam.lostMetal = allyTeam.lostMetal - worth		
+			
+			doUpdate = true
+		end
+		-- unit is not presently or formerly mind controlled and is thus not a casualty.
+	else
+		if capturedUnits[unitID] and capturedUnits[unitID] == unitDefID then
+			Echo("<AttritionCounter>: unitGiven: captured by a new team while already captured, how peculiar")
+			-- unit has been captured before and is now captured by another owner. 
+			-- do nothing; shouldn't happen unless FFA, and this widget shuts down in FFA.
```
sprunk: >shouldn't happen unless FFA

Zombies. Also maybe rover mirror?

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -353,27 +353,126 @@ end
...
+	end
+end
 
 function widget:UnitDestroyed(unitID, unitDefID, teamID, attUnitID, attDefID, attTeamID)	
 	
 	-- if its also the same kind of unit, its safe to assume that it is the very same unit
 	-- else it is most likely not the same unit but an old table entry and a re-used unitID. we just keep the entry
 	-- small margin of error remains
-	
-	if teamID == gaiaTeam or GetUnitHealth(unitID) > 0 then return end
```
sprunk: Morph (unit not actually lost). Factory cancel (unit lost but 100% refunded). Maybe terrorunit.

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -353,27 +353,126 @@ end
...
+			team.lostMetal = team.lostMetal - worth
+			
+			local allyTeam = allyTeams[team.friendlyAllyTeam]
+			allyTeam.lostUnits = allyTeam.lostUnits - 1
+			allyTeam.lostMetal = allyTeam.lostMetal - worth		
+			
+			doUpdate = true
+		end
+	else
+		-- this is most likely dead code because of the weird relationship between UnitGiven and UnitTaken
```
sprunk: `UnitTaken` is called when the unit is still in the old team, before all transfer logic happens, and can be thought of as a `UnitDestroyed` equivalent.
`UnitGiven` is called after the unit is already transferred, at the end of the transfer logic chain, and can be thought of as a `UnitCreated` equivalent.
Otherwise they are the same. Use one or the other (probably `UnitTaken` because that's a loss).

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -353,27 +353,126 @@ end
...
+		-- however there's no documentation i could find that explains it, so i feel unsafe otherwise.
+		if capturedUnits[unitID] and capturedUnits[unitID] == unitDefID then
+			Echo("<AttritionCounter>: unitTaken: captured by a new team while already captured, how peculiar")
+			-- unit has been captured before and is now captured by another owner. 
+			-- do nothing; shouldn't happen unless FFA, and this widget shuts down in FFA.
+		else
+			-- unit has now become mind-controlled for the first time; unitTeamID lost it
+			Echo("<AttritionCounter>: unitTaken: captured for the first time, file as casualty")
```
sprunk: No echoes.

LuaUI/Widgets/gui_attrition_counter.lua
sprunk: Also add `fix #2847` to commit message.

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: That was my initial plan, but then i noticed that only UnitGiven is fired on Dominatrix capture events while only UnitTaken is fired when Dominatrix loses control due to existence failure.

The probably-dead code bits are "unit returned to owner" in UnitGiven, and "unit captured for the first time" in UnitTaken because of that.

I have no idea why it is this way, though, therefore the code and the comment.

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: Good catch!

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: Morph should be possible to deal with reading URP's. Unsure about cancel?

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: In a rover mirror, capturing the enemy's captured slaves frees them, without creating a control link.

Zombies could be a thing, but gaia is not counted as a team (ignored in all other calculations). I'll check though.

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: Factory cancel cannot be on a complete unit, so the factory cancel check is now "health over 0 and buildprogress under 1", while morph is handled directly.

LuaUI/Widgets/gui_attrition_counter.lua
Anarchid: Checked recursive capture, zombies, and recursive zombies. All of this does not trigger that code.

Anarchid:
Fixed errors, should be good now

Self-destructing nanoframes via ctrl+D will still not count them as losses (even when it's a non-factory abort, so no refunds) - but that's a way rarer case compared to puppy, snitch, eos, and generally self-destructing completed units.

Anarchid:
Turned out the "dead code" thing was not dead:
- `UnitTaken` with a `capture_controller` happens when an enemy Dominatrix steals our units;
- `UnitGiven` without a `capture_controller` happens when enemy Dominatrix dies and returns the units.

sprunk:
Right, I forgot this is widget code which does not receive events for enemy units. However it should still fire up both events for spectators which would result in double counting, check this.

Depending on what the check result is:
 * no double counting, everything is correct? At a glance `UnitTaken` and `UnitGiven` look the same so you can define just one body and then do `widget.UnitTaken = widget.UnitGiven`.
 * double counting? In that case IMO add `Unit{Taken,Given}ByTeam` to https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/api_widget_events.lua and use just that new callin in Attrition Counter.

Anarchid:
I checked it with spectator and there doesn't seem to be any double counting. 
UnitTaken and UnitGiven cannot be completely shared because they have different parameters - `oldTeamID` in one and `newTeamID` in another. But that's minor, most of the logic stays the same so yeah.

PR #3529, opened 2019-04-24
Pork chop's metal glow stuff (F4 reclaim view)

sprunk:


LuaUI/Widgets/gui_metal_features.lua
```diff
@@ -15,63 +15,133 @@
...
+		noHotkey = true,
+	},
+
+	intensity = {
+		name = 'Highlighted Reclaim Brightness',
+		desc = "Increase or decrease visibility of effect",
+		type = 'radioButton',
+		value = '1',
+		items = {
+			{key ='1', name='High'},
```
sprunk: This should probably be a numerical slider.

LuaUI/Widgets/gui_metal_features.lua
```diff
@@ -15,63 +15,133 @@
...
+options = {
+	showhighlight = {
+		name = 'Show Reclaim on Economy Overlay',
+		desc = "When to highlight reclaimable features",
+		type = 'radioButton',
+		value = 'always',
+		items = {
+			{key ='always', name='Always'},
+			{key ='constructors',  name='When Constructor Selected'},
+			{key ='reclaiming',  name='When Reclaiming'},
```
sprunk: The `reclaiming` option doesn't seem to be a thing so far.

Porkch0p: The current gui_metal_features widget has two cases where it can activate. It highlights on the economy overlay glow, but also provides the glow on reclaim commands with the overlay disabled. Turning off the widget would disable both.

`reclaiming` keeps the glow for issuing commands, but hides it from the overlay.

sprunk: Right!

LuaUI/Widgets/gui_metal_features.lua
Porkch0p: This has been changed to a slider with more meaningful values (percentage of original brightness)

GoogleFrog:
I dislike the gameyness of always highlighting wrecks. I think I would make it default to show when constructors are selected.

PR #3532, opened 2019-04-25
Reset button for Chili Profiler

Porkch0p:
QoL update - adds a button to reset the profiler, instead of needing to unload and reload the Chili Profiler widget to profile again

### DIFF
--- a/LuaUI/Widgets/dbg_chili_profiler.lua
+++ b/LuaUI/Widgets/dbg_chili_profiler.lua
@@ -110,6 +110,15 @@ local function AddSedateProfiler()
 	debug.sethook(trace, "l", 16000000)
 end
 
+local function ResetProfiler()
+	if profiling then return end
+	tree0.root:ClearChildren()
+	sample_tree = {}
+	samples = 0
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/dbg_chili_profiler.lua
```diff
@@ -110,6 +110,15 @@ local function AddSedateProfiler()
 	debug.sethook(trace, "l", 16000000)
 end
 
+local function ResetProfiler()
+	if profiling then return end
+	tree0.root:ClearChildren()
+	sample_tree = {}
+	samples = 0
+	label0:SetCaption("Samples: " .. samples)
+	profiling = false
```
sprunk: Why do `profiling = false` here if the earlier `if profiling then return end` ensures it is false anyway at this point?

Porkch0p: You're right it's not needed... An earlier version didn't have the `if profiling then return end` and it got left in.

sprunk:
thanks!

PR #3534, opened 2019-04-26
Fix Deferred Rendering Widget to listen to Chobby

Porkch0p:
This allows Deferred Rendering (HDR+Bloom Effects) to be turned off, saving FPS, especially important on low-end machines.

### DIFF
--- a/LuaUI/Widgets/gfx_deferred_rendering.lua
+++ b/LuaUI/Widgets/gfx_deferred_rendering.lua
@@ -79,6 +79,8 @@ local spGetSmoothMeshHeight  = Spring.GetSmoothMeshHeight
 -- Config
 
 local GLSLRenderer = true
+local springMapState = nil
+local springUnitState = nil
 
 options_path = 'Settings/Graphics/HDR (experimental)'
 options_order = {'enableHDR', 'enableBloom', 'illumThreshold', 'maxBrightness'}
@@ -97,6 +99,9 @@ options = {
 local initialized = false
 
 local function OnchangeFunc()
+	springMapState = Spring.GetConfigInt("AllowDeferredMapRendering")
+	springUnitState = Spring.GetConfigInt("AllowDeferredModelRendering")
+
 	widget:Initialize()
 end
 for key,option in pairs(options) do
@@ -187,6 +192,10 @@ local collectionFunctionCount = 0
 --------------------------------------------------------------------------------
 
 function widget:ViewResize()
+	if (springMapState == 0 or springModelState == 0) then
+		return
+	end
+
 	vsx, vsy = gl.GetViewSizes()
 	ivsx = 1.0 / vsx --we can do /n here!
 	ivsy = 1.0 / vsy
@@ -257,9 +266,16 @@ local function DeferredLighting_RegisterFunction(func)
 end
 
 function widget:Initialize()
+	local mapState = Spring.GetConfigInt("AllowDeferredMapRendering")
+	local modelState = Spring.GetConfigInt("AllowDeferredModelRendering")
+
 	if not initialized then
 		initialized = true
 		OnchangeFunc()
+		return
+	end
+
+	if (mapState == 0 or modelState == 0) then
 		return
 	end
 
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -597,6 +610,10 @@ local function Bloom()
 end
 
 function widget:DrawScreenEffects()
+	if (springMapState == 0 or springModelState == 0) then
```
sprunk: These conditions don't often change so it would be better to just disable this call-in when they do, see above.

LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -97,6 +99,9 @@ options = {
 local initialized = false
 
 local function OnchangeFunc()
+	springMapState = Spring.GetConfigInt("AllowDeferredMapRendering")
+	springUnitState = Spring.GetConfigInt("AllowDeferredModelRendering")
+
```
sprunk: You can disable the `DrawScreenEffects` callin here so that it doesn't get called at all.

```lua
if springMapState == 0 and springUnitState == 0 then
    widgetHandler:RemoveCallIn("DrawScreenEffects")
    widgetHandler:RemoveCallIn("ViewResize")
else
    widgetHandler:UpdateCallIn("DrawScreenEffects")
    widgetHandler:UpdateCallIn("ViewResize")
end
```

LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -257,21 +266,25 @@ local function DeferredLighting_RegisterFunction(func)
...
+		return
+	end
+
 	if (glCreateShader == nil) then
 		Spring.Echo('Deferred Rendering requires shader support!')
 		widgetHandler:RemoveWidget()
 		return
 	end
 
-	Spring.SetConfigInt("AllowDeferredMapRendering", 1)
```
sprunk: Won't removing these potentially cause settings changes to be forgotten? What about parallel Spring instances (I think this might be possible when launching a replay from a different engine)?

LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -257,21 +266,25 @@ local function DeferredLighting_RegisterFunction(func)
...
-	Spring.SetConfigInt("AllowDeferredModelRendering", 1)
-
 	if (Spring.GetConfigString("AllowDeferredMapRendering") == '0' or Spring.GetConfigString("AllowDeferredModelRendering") == '0') then
```
sprunk: This line seems to read the setting as a string, maybe that causes it not to be read correctly?

LuaUI/Widgets/gfx_deferred_rendering.lua
Porkch0p: These seem to mainly serve the purpose of preventing errors as far as I can tell. If removed and not accounted for, the widget would immediately crash and remove itself.

LuaUI/Widgets/gfx_deferred_rendering.lua
Porkch0p: Done

LuaUI/Widgets/gfx_deferred_rendering.lua
Porkch0p: Modified it to an integer, doesn't seem to cause problems.

Porkch0p:
23 vs 14 FPS on Lowest, paused during a large teams game replay when Deferred Rendering is disabled
Lobsterpot Replay

LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -257,22 +270,26 @@ local function DeferredLighting_RegisterFunction(func)
 end
 
 function widget:Initialize()
+	local mapState = Spring.GetConfigInt("AllowDeferredMapRendering")
```
sprunk: What are these locals for, why not use the global `springMapState` and `springUnitState`? `widget:Initialize` calls `OnChangeFunc` and vice versa (lines 278 and 113 respectively, after your commit) so in line 282 the globals should already have the values they are supposed to (and the same as locals).

GoogleFrog:
How dynamic are the settings in this widget intended to be? One part of the widget removes it if the settings are 0, while another removes or adds callins based on the state of the settings. I dislike adding anything to the already bad initialisation code of this widget, and this is a bad foundation on which to tackle dynamic vs. static settings.

I'll reformulate the initialisation code and solve the original bug. If it was your intention to make the settings respond dynamically then that is a different PR. It probably requires chobby work and the creation of a framework to track such changes.

PR #3535, opened 2019-04-26
Healthbar Widget - Maximum Height Limit

Porkch0p:
When fully zoomed out in a large game, health bar widget can have high CPU and upwards of 1+ MB/s memory by processing all units and features on the map. 

The existing widget already stops rendering bars at a distance of around 3000 units. This adds a slider to place an height limit over which health bars will not be processed or drawn. Entirely based on 'Icon Height' widget. When set to 3000 units, testing shows memory peaking at 300KB/s and zero when viewed from higher.

It would need to be checked for side effects with other camera modes.

### DIFF
--- a/LuaUI/Widgets/unit_healthbars.lua
+++ b/LuaUI/Widgets/unit_healthbars.lua
@@ -197,6 +197,25 @@ end
 
 local paralyzeOnMaxHealth = ((lowerkeys(VFS.Include"gamedata/modrules.lua") or {}).paralyze or {}).paralyzeonmaxhealth
 
+local function IsCameraBelowMaxHeight() 
+	
+	local cs = Spring.GetCameraState()
+	local gy = Spring.GetGroundHeight(cs.px, cs.pz)
+	local tolerance = 25 --// as defined in iconheight
+
+	if cs.name == "ov" then
+		testHeight = options.drawMaxHeight.value * 2
+	elseif cs.name == "ta" then
+		testHeight = cs.height - gy
+	end
+
+	if testHeight >= options.drawMaxHeight.value - tolerance then
+		return false
+	end
+	return true
+end
+
+
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 
@@ -249,7 +268,6 @@ local gameFrame = 0;
 local empDecline = 1/40;
 
 local cx, cy, cz = 0,0,0;  --// camera pos
-local tolerance = 25 --// via iconheight
 
 local paraUnits   = {};
 local disarmUnits = {};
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/unit_healthbars.lua
```diff
@@ -242,6 +249,7 @@ local gameFrame = 0;
 local empDecline = 1/40;
 
 local cx, cy, cz = 0,0,0;  --// camera pos
+local tolerance = 25 --// via iconheight
```
sprunk: What is this for? Isn't this equivalent to just adding +25 to the max height (which won't make a difference if the values are in the thousands order of magnitude)?

LuaUI/Widgets/unit_healthbars.lua
```diff
@@ -1180,6 +1204,22 @@ do
 	local sec2 = 0
 
 	function widget:Update(dt)
+
+		-- Test camera height before processing
```
sprunk: Same as the earlier check, could be extracted to a separate function and called in both places.

Porkch0p:
Zoomed Out 18800
Zoomed Out 18600
Zoomed In 3000
Zoomed In Above 3000

LuaUI/Widgets/unit_healthbars.lua
Porkch0p: It was an undocumented magic number in the other widget, so I left it in, in case it had some significant meaning.

LuaUI/Widgets/unit_healthbars.lua
Porkch0p: True, it's extracted now.

sprunk:
Looks decent, if you'd like to leave this as-is I can merge.

Here are some further potential improvements:
 * there's an `infoDistance` constant which seems to the squared distance at which healthbars get simplified (no text drawn, just bars). Currently its value is about 830 elmos, this could become controllable as well.
 * I've found the hardcoded 3000 distance constant (seems to be about 2500 for features). Possibly your new option could just directly affect that one instead.

https://github.com/ZeroK-RTS/Zero-K/blob/60bde14fdef008c3e6f84faaee843f9749bdd96c/LuaUI/Widgets/unit_healthbars.lua#L48
https://github.com/ZeroK-RTS/Zero-K/blob/60bde14fdef008c3e6f84faaee843f9749bdd96c/LuaUI/Widgets/unit_healthbars.lua#L562-L564
https://github.com/ZeroK-RTS/Zero-K/blob/60bde14fdef008c3e6f84faaee843f9749bdd96c/LuaUI/Widgets/unit_healthbars.lua#L623-L632
https://github.com/ZeroK-RTS/Zero-K/blob/60bde14fdef008c3e6f84faaee843f9749bdd96c/LuaUI/Widgets/unit_healthbars.lua#L1170-L1176

Porkch0p:
I've exposed the infoDistance value as a slider, and tied the max draw distance and feature draw distance to the Max Height slider (in the same ratio they were present in the code).

It looks like this widget has a huge chance of shooting yourself in the foot if you set both sliders to the max, at 20000 draw distance. At full map zoom, performance was hitting 20% CPU. 

I've placed a cap on the max distance of 9000 elmo. There are warnings on the sliders not to touch them if you care about performance. Rendering simplified health bars at full distance is much less taxing on CPU than full health bars.

Porkch0p:
Simplified at 9000
Simplified at 9000
Standard at 9000
Standard at 9000

sprunk:
Thanks!

PR #3540, opened 2019-04-28
Corrected DoF autofocus calculations for focus spread

dvr333:


dvr333:
AFAICT This is final and working with basically no artifacts.

PR #3542, opened 2019-05-01
Fixed LuaUI crash bug when DoF or deferred rendering shutdown on load

dvr333:


sprunk:
Why not make the layers `0` and `1` for simplicity?

dvr333:
The comment is right as far as the ordering of DrawScreenEffects. It's bit weird, but it seems like the ordering of DrawFoo calls is the opposite of widget load order. Since I want DoF handled last (to make sure any other post effects meant to form the final world image get incorporated into the blur), it gets loaded first.

PR #3543, opened 2019-05-02
Factory differentiation [WIP]

mvollrath:
Various special abilities for factories.


mvollrath:
So far:

### Amphbot Factory
* +40 underwater regen.

### Cloakbot Factory
* Radar jamming, slightly larger radius than Cornea.

### Jumpbot Factory
* Is now fireproof.

### Air Factory
* Has radar.  Same radius as Funnelweb but emitted from a lower height, making it a little more useful for detecting air targets.

### Ship Factory
* Has sonar.
* Slightly increased sight radius.

### Tank Factory
* +2000 health.

### Gunship Factory
* Can morph to Mobile Gunship Factory and back.

The morphable Mobile Gunship Plant is a lot of fun and was really easy to implement, but there are some major bugs which I'm not sure how to resolve:
* When morphing back to static Gunship Plant, the ground below is instantly terraformed flat for free.
* When morphing back to static Gunship Plant while hoving over another building, the Gunship Plant will be created inside the other building.

There are some more visual details in the Mobile Gunship Plant unit script I intend to resolve if the bigger issues can be worked out.

### Rover Factory
* Nothing yet, need good thematic ideas.

### Hover Factory
* Nothing yet, need good thematic ideas.

mvollrath:
[Here's a video](https://www.youtube.com/watch?v=rvSKeS-0aaQ) of the Mobile Gunship Plant.

Licho1:
Lol what is the game purpose of flying gunship factory? How do you imagine its uses?

Also fire proof - meh .. we don't do this much in ZK.

mvollrath:
Example flow of Mobile Gunship Factory:
* Morph to mobile.
* Scout with Blastwing.
* Defend with Locust/Gnat.
* Eco with Comm and Wasps.
* When you want a second factory, your Gunship Factory can help build it.
* If you need more GS build power, use Wasps like Caretakers.
* Mid/late game per usual Tridents and Nimbus spam.

mvollrath:
Fixed the Mobile GS morph back to static like so:
* Mobile GS Factory can build a static GS factory.
* As soon as a Mobile GS Factory places static GS Factory nanoframe, that nanoframe is removed and the mobile factory moves to its position.
* When it arrives, if the build site is still clear, it instantly morphs.

[Here's a video](https://youtu.be/xd1kKox_4HM) of this iteration.

TODO:
* Generalize the gadget used for morphing back to static factory.
* Hide the morph button from mobile factory integral menu.
* Visual FX for flying factory, jet exhaust etc.
* Something for hovers.
* Something for spiders.
* Something for hovers.

GoogleFrog:
Gunship factory:
- Use the callin that is called something like 'AllowUnitCreation' to block the creation of the nanoframe. This callin should contain all the information you need and avoids the flickering nanoframe. Be careful to call it after the Prevent Lab Hax usage of this callin, as otherwise you could end up landing inside factory yardmaps.
- Don't use the command for constructing a gunship factory for the command that lands a gunship factory. This could cause trouble when multiple units are selected.
- Put the land command in the orders tab instead of the factory tab. The approach I have in mind for this (and the previous point) is to make the mobile gunship factory able to construct itself, and to add parameters to integral menu to customise the appearance and placement of the "construct mobile GS factory" command.
- The factory should land instead of snapping onto the ground. Perhaps Spring.SetUnitLandGoal is appropriate (it is new, I've never used it), of if that fails, MoveCtrl. It should be possible to cancel the land sequence while it is in progress.
- The unitdef naming scheme suggests that 'factorymobilegunship' is a factory that produces mobile gunships. I would prefer a new class for mobile unit swaps of structures, like 'mobilefactorygunship'. The Aegis and Iris morphs don't need to fit this naming scheme because the mobile versions are distinct (worse) versions of the static versions.

Other notes:
- I'm against making arbitrary things fireproof. We've been trying to cut down on the number of special rules and are not looking to add more.
- How does a lower radar emit height make the airplane plant relatively better at spotting planes? All radar towers do is determine which bits of the terrain are in radar coverage. To check whether a unit is visible they check the underlying terrain.
- +2000 health for tanks is quite boring, and dilutes that as the Shipyards special ability.

Licho1:
Some random ideas for lab skills:

 * healing ability. Could be implemented as a nanoturret without increasing actual BP

 * slow crawl with ground leveling ability - automated area terrain smoother. Always wanted a thing that would flatten the ground for heavy tanks and auto tear down terraform without micro

 * shelter mode, close up to get armor like gauss or crabe, while protecting the unit being built. 

 * factory toss - after unit is built, it is tossed short range, this allows factory to build stuff up-cliff (jumpjet lab)

Anarchid:
Mobile gunship factory allows you to use the gunship factory to repair your gunships and other units, reclaim metal, possibly terraform, become cloaked, start Krow as a nanoframe.

mvollrath:
> * How does a lower radar emit height make the airplane plant relatively better at spotting planes? All radar towers do is determine which bits of the terrain are in radar coverage. To check whether a unit is visible they check the underlying terrain.

By emitting at a lower height the radar covers less terrain.  To phrase it better, this is relatively worse at spotting land units.

mvollrath:
> * +2000 health for tanks is quite boring

It's my goal to make all of the factories thematic, not all of the features exciting.  Do you have any suggestions for exciting thematic features for Tank Foundry?

GoogleFrog:
> By emitting at a lower height the radar covers less terrain. To phrase it better, this is relatively worse at spotting land units.

That isn't how radar works. Visibility on radar is informed purely by whether the terrain is covered.

mvollrath:
I found that SetUnitLandGoal didn't seem to have any effect.  Right now the landing depends upon the unit being toggled to land, which I intend to control.  It does SetUnitPhysics to kill any inertia, which so far is reliable.  If it lands in the wrong spot it won't morph, and the next command will cancel the premorph.

This will probably only work for Gunships, to do something like Swift morphing to Radar maybe better to make an invisible air pad.

darloth:
> > * +2000 health for tanks is quite boring
> 
> It's my goal to make all of the factories thematic, not all of the features exciting. Do you have any suggestions for exciting thematic features for Tank Foundry?

It could become armed, like the Welder.  If it came with its own defensive turret at least equivalent to the Lotus, that would certainly be a unique factor.

darloth:
Hover factory being always slowly mobile might fit the hover ideology?  Wouldn't give it the other stuff of gunshipfac, just ability to slowly move while building over flat terrain.  Maybe it can't even rotate, to keep simplicity?

sprunk:
Wacky design change with large balance ramifications that nobody asked for. Accumulated some merge conflicts over time.

PR #3559, opened 2019-05-07
Fixes to your wipShader branch

lhog:


ModelMaterials/1_normalmapping.lua
```diff
@@ -30,10 +34,11 @@ local materials = {
 			[0] = '%%UNITDEFID:0',
 			[1] = '%%UNITDEFID:1',
 			[2] = '$shadow',
-			[3] = '$specular',
+			--[3] = '$specular',
```
sprunk: Would be good to leave a code comment here explaining why this is disabled.

lhog: Yeah, sure, later today

PR #3569, opened 2019-05-14
Shader updates

GoogleFrog:


GoogleFrog:
Seems mergable?

lhog:
Hold on when I'm back from vacation. I think I forgot to include one important fix from my local rep.

lhog:
https://github.com/ZeroK-RTS/Zero-K/pull/3573

PR #3576, opened 2019-05-25
Fixed team name side for spectator widgets

dvr333:
* Player/Team names now show on the side of the attrition and spectator panels corresponding to their start location on the map.
* Spectator Panel no longer messes with Economy Panel settings.

### DIFF
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -137,10 +137,44 @@ local function GetTeamName(teamID)
 	end
 end
 
+local function GetLeftRightAllyTeamIDs()
+ 		if Spring.Utilities.Gametype.isFFA() or Spring.Utilities.Gametype.isSandbox() then
+        -- not 2 teams: unhandled by spec panels
+        return {}
+    end
+
+
+    local myAllyTeamID = 0--Spring.GetLocalAllyTeamID()
+    local enemyAllyTeamID = 1 -- FIXME assumes teams are 0 and 1
+    local myTeamID = Spring.GetTeamList(myAllyTeamID)[1]--Spring.GetLocalTeamID()
+
+    local myBoxID = Spring.GetTeamRulesParam(myTeamID, "start_box_id")
+    if not myBoxID then -- can start anywhere
+        return {0, 1} -- players see themselves on the left, maybe should `return 0, 1` (see fixme above) so everybody sees everything the same?
+    end
+
+    local myBoxRepresentativeSpotX = Spring.GetGameRulesParam("startpos_x_" .. myBoxID .. "_1")
+
+    local comparisonX
+    if Spring.GetGameRulesParam("shuffleMode") == "allshuffle" and Spring.GetGameRulesParam("startbox_max_n") > 1 then
+        -- there are multiple boxes the enemy can be in so just look where we are on the map
+        comparisonX = Game.mapSizeX / 2
+    else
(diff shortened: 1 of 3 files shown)


LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -137,10 +137,44 @@ local function GetTeamName(teamID)
 	end
 end
 
+local function GetLeftRightAllyTeamIDs()
+ 		if Spring.Utilities.Gametype.isFFA() or Spring.Utilities.Gametype.isSandbox() then
```
sprunk: screwed up indentation here

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -137,10 +137,44 @@ local function GetTeamName(teamID)
...
+        -- not 2 teams: unhandled by spec panels
+        return {}
+    end
+
+
+    local myAllyTeamID = 0--Spring.GetLocalAllyTeamID()
```
sprunk: If somebody is in allyTeam 1 they won't be able to read team 0's box ID in case of shuffle

LuaUI/Widgets/gui_chili_spectator_panels.lua
```diff
@@ -904,10 +905,43 @@ local function GetWinString(name)
 	return ""
 end
 
+local function GetLeftRightAllyTeamIDs()
```
sprunk: this seems to be the same function as in the other file, should be extracted to some header

LuaUI/Widgets/gui_chili_spectator_panels.lua
```diff
@@ -983,6 +1017,7 @@ function option_CheckEnable(self)
 	if not allyTeamData then
 		return false
 	end
+
```
sprunk: random space

LuaUI/Widgets/gui_attrition_counter.lua
```diff
@@ -165,9 +199,9 @@ local function GetOpposingAllyTeams()
 		return
 	end
 	
-	if returnData[1].allyTeamID > returnData[2].allyTeamID then
-		returnData[1], returnData[2] = returnData[2], returnData[1]
-	end
+	-- if returnData[1].allyTeamID > returnData[2].allyTeamID then
```
sprunk: leave a comment on why it's disabled and how it will be useful in the future (or if it won't, remove)

LuaUI/Widgets/gui_chili_spectator_panels.lua
dvr333: Right, I meant to do that earlier. Do we have headers for team selection helper functions already?

LuaUI/Widgets/gui_attrition_counter.lua
dvr333: I've resolved this for non-spectators (since spectators should have everything), but from what I could tell things worked fine in my testing this way.

LuaUI/Widgets/gui_chili_spectator_panels.lua
dvr333: Okay, I tried this, but I don't think I know how Lua include works. Everything just broke afterwards (for some reason attrition panel is so important it breaks LuaUI if it doesn't work).

dvr333: Okay, I figured it out. I needed to return the function name.

PR #3577, opened 2019-05-26
Exposed SunChanged to gadgets

lhog:


GoogleFrog:
Did you test this? I would have expected you to need to edit the actions table.

lhog:
My bad, will fix

PR #3583, opened 2019-06-01
The great equalizer

lhog:


GoogleFrog:
I see no visual change upon enabling the new outline widget.

[infolog.txt](https://github.com/ZeroK-RTS/Zero-K/files/3244626/infolog.txt)

lhog:
Did you check out this branch and then launch the game?

GoogleFrog:
Yes.

GoogleFrog:
image

image

lhog:
The reason I was asking was because I although I checked it from mobile (so can be mistaken), but I didnt see any mentions of Custom Unit Shader gadget being loaded or failed.

GoogleFrog:
image

CUS seems to be working. I think the issue is just with the outline gadget. (You can spawn the skinned commanders with dynfancy_)

lhog:
Can you also check the deferred rendering debug widget, I updated it in this branch a bit, so it should show material indices (red color of varying intensity). It should draw bleak red shapes for units and mid-intensity or high-intensity red for features.

GoogleFrog:
image

It seems to draw everything except cloaked units, which is expected. There is a tree on the right of the screenshot.

lhog:
I'm not around my home PC to check exact pixel colors, but overall the picture looks as expected.

Since I see no critical complaints from the outline  shader or widget, and I think you have same environment (Win/Nvidia/1244 engine), then the only other likely option is that some other widget, that enabled for you and not for me, tampers OpenGL state. I'll post some debug instructions when I get back home.

lhog:
Sadly, I don't have any time today or any time soon till weekend.
Bu t can you just try to replace https://github.com/ZeroK-RTS/Zero-K/blob/3fe6243cac2cf6cfe893e9183fc0921bc8a5fa06/LuaUI/Widgets/gfx_outline_shader.lua#L409-L418
with something like this:
```
function widget:DrawScreenEffects()
	gl.Blending(false)

	gl.Texture(0, dilationColorTexes[pingPongIdx])
	gl.Texture(0, shapeColorTex)
	gl.TexRect(0, 0, vsx, vsy, false, true)
	gl.Texture(0, false)

	gl.Blending(true)
end
```

Change OUTLINE_COLOR to something like `local OUTLINE_COLOR = {1.0, 0.0, 0.0, 1.0}`

And then comment out `gl.Texture(0, dilationColorTexes[pingPongIdx])` and `gl.Texture(0, shapeColorTex)` sequentially. These two should show you the similar red-tinted picture of the same shape, except one of them will be a little bit bigger/dilated. This should prove the widget/sequence of draw calls work and the issue is elsewhere (most probably some silently broken widget of ZK).

GoogleFrog:
The strings you want me to comment out don't exist, so I assumed you meant the first two gl.Texture calls in DrawScreenEffects. Here are the results of the various combinations, where a `0` means that the line was commented out.
- 11 - Black view
- 01 - Black view
- 00 - White view
- 10 - Black view
The view here is the entirety of worldspace. If I disable the UI then the entire window is one uniform colour.

image

Are these shader errors not worrying?

lhog:
Ok let's try the following:
1) Update to https://github.com/ZeroK-RTS/Zero-K/pull/3583/commits/1f9c790220ec8a8c6d5447942076a8fa3e82db23
2) Test if it still doesn't show outline for you. Make sure you use "Outline Shader" widget
2) Un-comment https://github.com/lhog/Zero-K/blob/theGreatEqualizer/LuaUI/Widgets/gfx_outline_shader.lua#L111 or https://github.com/lhog/Zero-K/blob/theGreatEqualizer/LuaUI/Widgets/gfx_outline_shader.lua#L187 or both of them. This should break a few things in the game, but at least you should see a white outline around units. If so it's the other ZK widget/widgets that don't clean up state after themselves
3) If actions listed in the bullet above still won't render outline please go ahead and comment/uncomment this block: https://github.com/lhog/Zero-K/blob/theGreatEqualizer/LuaUI/Widgets/gfx_outline_shader.lua#L415-L437 and post screenshots of both stages here. Zoom in to one big unit, like Grizzly, so I can see if dilation and shape shaders work.

GoogleFrog:
I had some leftover files in my local luaUI/Widgets. I deleted them and outline no longer loads at all as Include is missing.

`[f=-000001] Failed to load: gfx_outline_shader.lua  ([LuaVFS::Include(synced=0)][loadvfs] file=LuaUI/Widgets/Include/LuaShader.lua status=-1 cenv=0)

lhog:
Yes it is missing in the repo.  Dunno how and when this happened. I'll fix it my evening.
Thanks for the heads up.

GoogleFrog:
Seems to work. The outlines are quite wide and jagged.

image

lhog:
Yes they are going to look a bit jagged,  because default frame buffer uses multi sampling (MSAA) and deferred buffers don't. This is engine limitation due to performance concern s and concerns related to the fact that almost no one in the community knows how to deal with MSAA textures. There is not too much I can do about it (I tried), except perhaps I can run blur of the edges. The problem - it is another full screen render pass and the widget is supposed to be fast. 

Also glResetState is a terrible hack. It was meant to be there only to prove that some other widget craps into the state without cleaning after themselves. It is going to break the rendering of many things including outline.  Point is don't use it in production.  Rather we should find the offending widget and patch it up.

GoogleFrog:
As far as I can tell this is done, apart from removing the gl.ResetState at the end of widget:DrawWorld(). I found that outlines would break startboxes, custom formation dots, and facplop (with an unselected commander) without it.

lhog:
Interesting... It should be possible to fix these widgets with one or a few lines of code. I will have a look today evening or over the weekend.

GoogleFrog:
Shouldn't outlines clean up its state? I suspect that it is changing something that no widget has expected to be changed in the gl state, so Outlines could break any number of widgets based on load order.

lhog:
It should and it does as far as I'm aware.  The fact that other widgets break doesn't prove much to me. Each case needs to be reviewed individually.  It is totally possible that uncleaned state in one widget cause outline to output something that breaks third widget.

If I don't clean something up it is going to be a bug.

GoogleFrog:
To be clear, the bugs disappear when `gl.ResetState()` is added to the end of DrawWorld in the outline widget.

lhog:
I think I found the reason for unexpected behavior. Other widgets relied on gl.Blending to stay on.
Please check yourself as I couldn't reproduce startboxes and plop widgets issue locally (startboxes only appear in multiplayer for me and for plop I couldn't see what the issue was).

In any case I also intend to rework default CUS shader quite a bit in two ways:
1) Place more attachment points to "plugins", so materials have more freedom to define the way they are supposed to look
2) Enable widget to CUS feedback, so I the highlight of metal features can be integrated into wreck materials.

GoogleFrog:
How are you running ZK? You should be launching in a chobby environment so you have the correct settings and can change map/AI easily.

PR #3590, opened 2019-06-04
add transifex client config

Anarchid:
Having this hidden file in the repository allows the transifex client to be used to easily pull translations from transifex.

It is necessary because the newfangled zero-effort github integration transifex claims to have seems to be one way only - it doesn't open any PR's when translations get completed as it's supposed to.

Additionally, this approach is more flexible, since it allows more than one repository to be configured (Chobby i'm looking at you).

Usage:
* Download all languages and all files, ignoring timestamps:
```
tx pull -af
```
  The `-f` param is necessary because https://github.com/transifex/transifex-client/issues/22. 
* Download only a specific language:
```
tx pull -fl zh_CN
```

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,3 +1,12 @@
+# This is the ZK config for the Transifex online translation platform 
+# The job of this file is to allow the Transifex CLI tool to fetch translation 
+# files from Transifex for you to commit to the repository with minimal effort.
+# Here's how to use it:
+# * Fetch one language: `tx pull -fl zh_CN`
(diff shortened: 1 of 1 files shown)


sprunk:
Would be good to add a comment at the top explaining:
 * what this is/does
 * what tool it works with (eg. is "the transifex client" still `tx` on windows?)
 * example usage (`tx pull -af` or `tx pull -fl zh_CN`)

GoogleFrog:
I'm assuming that everything translation related is somebody elses project.

PR #3591, opened 2019-06-04
Update polish translation from transifex using the transifex client thingy.

Anarchid:
This is mainly intended as a test for this pipeline. 

See if it breaks things in any way.


LuaUI/Configs/lang/interface.pl.json
```diff
@@ -1,8 +1,6 @@
 {
 	"idle_cons" : {
 		"one" : "%{count} wolny konstruktor",
-		"few" : "%{count} wolnych konstruktorów",
```
sprunk: These should not have been removed. I don't think they are present on Transifex though. I believe they are missing because the English source does not define `few` and `many`.

LuaUI/Configs/lang/interface.pl.json
```diff
@@ -103,8 +101,6 @@
 	"cheatsheet_cat_rejects" : "Techniczne śmieci",
 	"x_units_in_queue" : {
 		"one" : "%{count} jednostka w kolejce",
-		"few" : "%{count} jednostki w kolejce",
```
sprunk: Ditto

LuaUI/Configs/lang/interface.pl.json
Anarchid: Yep, that was because of missing keys in english file.

After i added `few` and `many` to source, the integration automatically found them in the `pl` json file and put them into transifex to be gotten with `tx pull`.

Anarchid:
Restored things erased by the forced integration push, rebased to be mergeable to current master.

sprunk:
I'm pulling your strings.

Issue #3596, opened 2019-06-14
Highlighting reclaimables in eco overlay doesn't work


Cat-Lady:
Since last change (making highlighting reclaimables being possible when having constructors selected and making that the default), it is not possible to bring back old behavior (reclaimables highlighted when eco overlay active) - despite such option being present.

Selecting the "when eco overlay enabled" option makes reclaimables being highlighted all the time (even without eco overlay enabled).

Porkch0p:
Looks like there are some quirks in zero-k around the economy overlay and the 'metal map viewmode'. A number of code paths refer to the map mode being set to 'metal' but in practice it doesn't seem to ever get set that way, and the economy overlay for geothermals has additional conditions to trigger that this widget didn't have added. I would have sworn that it worked correctly before, but this could have slipped through as well.

PR #3597, opened 2019-06-14
Fix to 'always' highlighting reclaim outside of economy overlay

Porkch0p:


Porkch0p:
Re: #3596 
One question - should the 'when constructor selected' preset also be changed to only show on the economy overlay view? That one would change default behavior for users.

Cat-Lady:
Maybe a checkbox (under all options) "only when economy overlay enabled", that would affect all variants? So, without it checked, it would apply globally - with, would work only on eco overlay.

GoogleFrog:
I thought the current behaviour was the intended behaviour for 'always'.

PR #3616, opened 2019-06-20
Tweaks to commander modules

Cat-Lady:
1. Light particle beam:
DPS - 165 -> 180

It already have lower range than starting beam laser (310 vs 330), and divides damage between 3 shoots per second, instead of being "continuous" laser beam - which means more damage gets "wasted" while finishing off damaged targets (most notable in case of its intended target - raiders).
To give any incentive for paying to swap beam laser into this shorter range weapon, damage got minimally boosted.

2. Riot Cannon:
Range - 275 -> 300
Reload - 2s -> 1.63s

Reduced reload time and increased range to exactly match ripper's riot canon - now both weapons are exactly the same DPS and range (without module upgrades), their also match CEG exactly.

Riot Cannon was filling same role as Heavy Machine Gun - crowd control weapon with AoE. Hoever, Riot Cannon have lower range 275 vs 285 and very high damage granulation - shooting once per 2 seconds - "wasting" much more potential on finishing damaged targets. Then, it had *much* lower DPS than HMG (110 vs 180), resulting in being strictly worse than HMG in every aspect. On top of that, it is available only to Guardian Chassis (akin to heavy rockets), versus HMG available to all except Engineer. Last but not least, practical AoE was *worse* than those of HMG, despite being theoretically bigger - HMG's bullet spread and fast fire rate (with every bullet having 2/3 of Riot Cannon's AoE) increased practical AoE of HMG dramatically, elaving much to be desired for Riot Cannon.

This change aims to make it weapon at least worth considering. Before, it had worse range and worse DPS than even retired peashooter (+worse aspect of damage granulation), for which you had to pay for an "upgrade".

Now, it have minimally higher damage than Heavy Rocket, with less granulation (reload 2s vs 3s), and much lower range (300 vs 430), at same chassis restrictions (guardian only) and upgrade availabilities (napalm).

3. Heavy Rocket:
AoE - 75 -> 96

Made AoE at least as big as single bullet from Heavy Machine Gun... No problems with CEG size.

4. Disruptor Ammo:
Heavy Machine gun with Disruptor Ammo:
DPS - 90 -> 135

Shotgun with disruptor Ammo:
DPS - 96 -> 144

Unified direct damage reduction to 25%.
Previously, this particular upgrade was single worst deal in modules department, making weapons affected suddenly unable to kill weakest raiders (Glaive) in one burst. Now, it isn't a strict downgrade, and might find some utility.

5. Napalm Warhead:
Riot with Napalm:
DPS - 86 -> 100
AoE - 144 -> 170
Heavy Rocket with Napalm:
DPS - 180 -> 135
AoE - 128 -> 144
Cratering - 0 - > 1

Unified direct damage reduction to 25% - previously, it was highest on the weakest weapons (riot cannon), making it a strict downgrade. OTOH, on already very good heavy rockets, it was reducing direct damage by 0% (sic!).
Unified AoE gain. Added cratering to Heavy Rocket with Napalm, due to that version having sensibly big AoE (128 previously, now 144), and similar (and smaller) rockets on units doing cratering.


GoogleFrog:
I like consistency across weapon types and don't want to break it when not necessary, especially if the weapon type is rare.
- HMG had the same craterBoost and Mult as Stardust and Reaver. There doesn't seem to be a good reason to change it.
- Riot Cannon had the same AoE and damage as Ripper. Why not leave those the same and reduce its reload time instead? Otherwise people will see the same effect one-shot some units and not others. Also, a new CEG would be required.
- I don't see light particle beam among the changes.

Generally, when you change AoE you should check whether the CEG needs changing (search for the explosiongenerator names).

The concussion shell and cluster bomb buffs are fairly extreme, and mainly look like Recon buffs. A weapon being limited to only a few chassis is not something that makes a weapon less powerful. If anything, it just makes it more problematic.

We didn't get into the problem with special weapons and I think they are one of the more problematic aspects of commanders. In my experience, whenever a special weapon is good it is simply rushed and becomes the entire focus of the early game. If you want to buff special weapons I would like to make them more expensive and harder to rush. To do this:
- Make all morphs more expensive and reduce module cost by the appropriate amount. This level 4 rushers save less metal compared to people who morph up slowly.
- Make morphs take longer to complete (reduce their effect buildpower).

sprunk:
Riot cannon 220 -> 250 is a massive change because now it 1-shots Bandit.
230 and above would allow it to 1-shot Bandit at level 2 with one damage module.

Cat-Lady:
@Cat-Lady
1. Reverted changes to HMG cratering, as requested.

2. Reverted damage and AoE increase of Riot Cannon. Instead, reduced reload time and increased range to exactly match ripper's riot canon - now both weapons are exactly the same DPS and range (without module upgrades), their also match CEG exactly.

Thank you for the insight - it turned out to be much better idea, indeed.

3. This time, particle beam change included as described (Ooops.)

4. For now, completely reverted the Cluster Bomb and Concussion shell changes. I will put them in separate PR, alongside the proposed modifications to morph and modules cost - this require a little more work and double-checking that everything fits, so I don't want it to be clustered with the pure modules tweaks.
---

Is his now ready to be pulled?

sprunk:
Squashing the 3 commits would be good. The content looks okay from a technical PoV.

Cat-Lady:
I lack write permission to squash commits and an already published PR. Hoever, for future reference, commits from PR can be squashed into a single one automatically, during pulling:
https://stackoverflow.com/questions/11396069/squash-to-only-one-proper-commit-for-github-pull-request/36377439#36377439

GoogleFrog:
I'm leaving this PR until it becomes clear that there aren't more graphics bugfixes releases. This PR would make a release require a picture.

Cat-Lady:
Sure thing, that sounds sensible (and it helps to know whats the status). I'll try to rig some picture candidates.

Cat-Lady:
Few screenshots that could be used for showcasing balance changes focused on Commanders:

https://i.imgur.com/lb2AjDF.jpg

https://i.imgur.com/zoBpWBv.jpg

https://i.imgur.com/o146jx9.jpg

https://i.imgur.com/IhSyOvw.jpg

https://i.imgur.com/EnDKYyb.jpg

sprunk:
Squashed (I know it can be done on pull but GoogleFrog doesn't seem to ever do that and those commit messages were quite confusing with all 4-5 kept individually)

PR #3625, opened 2019-06-21
CUS rework v1

lhog:
Massive rework of CUS, default shader and material files.
Features so many advantages I can hardly enumerate, but here are the most important ones:
1) Graceful fallback - if material is not compiled then default engine shader continues to do its job
2) Ability to toggle on/off shader options instantly (shadows, flashlights, metal wreck highlight, vertex AO, etc) by means of `/luarules <option>` command and thus through `Spring.SendLuaRulesMsg()`
3) API_CUS is now more or less readable

TODO:
* ~~make PR mergeable~~
* ~~leverage material based metal wreck highlights from widget~~
* ~~restore removed material files (right now they are inactive and moved to ` ModelMaterials/x/` directory)~~
* test everything

It'd be nice if someone knowledgeable tested this PR on AMD/Intel GPU


LuaUI/Widgets/gui_metal_features.lua
```diff
@@ -36,34 +36,21 @@ local spGetGameFrame = Spring.GetGameFrame
 --------------------------------------------------------------------------------
 
 options_path = 'Settings/Interface/Map/Reclaimables'
-options_order = { 'showhighlight','pregamehighlight','intensity','minmetal'}
+options_order = { 'showhighlight', 'pregamehighlight', 'minmetal'}
 options = {
 	showhighlight = {
```
GoogleFrog: This reverts recent changes to the showhighlight option.

GoogleFrog:
I was sure this had a WIP tag. Removing the tag did not send me a notification. Comments on the ticket do.

lhog:
I was almost certain this would be ready for tests within an hour and it indeed was, however the tests revealed that DrawFeature() call-in is very slow, such slow that it makes whole thing work even slower than the current version.

I disabled metal features rendering and slow parts of the code, so for now code does more or less the same what the existing CUS & related code do, except the code is much-much better.

Now I have a dilemma if this code is worth merging and testing alone as it naturally brings only one improvement - ability to fallback to the engine default shading in case of shader compilation failure. The rest of the advantages are not immediately tangible.
In the same time the amount of changes is quite impressive and it's always risky to put changes like that into the production.

In order to progress I need new engine features, which I don't think I can author myself, so I guess this branch is going to stay in the dormant mode for some time.

lhog:
Some good news to share.
@rtri added necessary features to engine-develop and I managed to port them back to engine-maintenance. Hope my PR gets through https://github.com/spring/spring/pull/465

I've tested stuff locally and results are quite promising in terms of the performance. Looking forward for aforementioned changes to get into the maintenance so we can move forward with this PR too.

ModelMaterials/128_features_special.lua
```diff
@@ -44,6 +44,42 @@ local function DrawFeature(objectID, objectDefID, mat, drawMode, luaShaderObj)
...
+	end
+
+	if highlightActive then
+		local fs = Spring.GetAllFeatures()
+		--local fs = Spring.GetVisibleFeatures(-1, 30, false)
+		for _, fID in ipairs(fs) do
+			local metalHere = Spring.GetFeatureResources(fID)
+			metalHere = ((metalHere >= metalWreckTreshold) and metalHere) or 0.0
+			--Spring.Echo("GameFrameSlow", fID, metalHere)
+			frSetMaterialUniform[isDeferred](fID, "opaque", 3, "floatOptions[1]", GL_FLOAT, {metalHere})
```
ashdnazg: maybe reuse table here?

lhog: Thanks for the hint. I assume you meant {metalHere}?
Generally there are a lot of things to improve as far as metal highlight optimization.

Overall I'm questioning myself if the uniform approach I took here is optimal solution. Pros: options toggle is lightning fast, Cons: lots of shader code might mean heavy register pressure on the GPU side, especially for old drivers/GPU. I know windows/Nvidia makes recompilation for for uniforms conditionals, so it is likely not the issue on "premium" setups, might be not the case on other systems.

I think I will try how fast recompilation might be if options are toggled with #define. Maybe it is not that bad and supporting code will be much simpler.

ashdnazg: Indeed what I meant.
Dunno about other optimisations, but this is big chunk of unnecessary allocations every few frames, so as far as lua goes, this has an effect that is quite easy to cancel.

lhog:
Please test

GoogleFrog:
Ok. Also for future reference I take WIP to mean you are not looking for testing/merging at this time.

GoogleFrog:
image

The only potentially big issue I found is that my test game said "Warning: [LuaMatUniforms::Validate] missing "TeamColor" uniform" 71 times. 

The wreck highlights look unchanged. There is no occlusion so overlapping bits of geometry can cause bright patches.

[infolog.txt](https://github.com/ZeroK-RTS/Zero-K/files/3413376/infolog.txt)

lhog:
https://github.com/ZeroK-RTS/Zero-K/pull/3667

Issue #3574, opened 2019-05-25
Update outline widget


GoogleFrog:
ivand has written shader based unit outlines: https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets_BAR/gfx_outline.lua

The old widget should remain as an automatic fallback for players without the requisite shader abilities.
- Rename the old widget file and widget name.
- Add the old settings to the new widget.
- Make the new widget enable the old widget if the new widget is unable to run.

dvr333:
Two questions:

Does this work underwater with all Refraction and Reflection settings?

Does this also make rocks and trees get outlines?

lhog:
> Does this work underwater with all Refraction and Reflection settings?

That is actually a very good remark by @dvr333 and I wasn't sure how it behaved before I checked.
Currently outline seems to ignore other effects of top, such as shields, water, smoke particles
![screen00760](https://user-images.githubusercontent.com/1476261/58372017-0ae56e80-7f20-11e9-9d09-3008370c6dff.jpg)
![screen00761](https://user-images.githubusercontent.com/1476261/58372018-0b7e0500-7f20-11e9-9166-e65fdf55092f.jpg)

I'm not sure it's desired, also I'm not sure if I can change much to the way it behaves. 


> Does this also make rocks and trees get outlines?

The widget can be configured to skip certain "materials", where material is defined by material index:
https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/materials/Shaders/default.lua#L364
https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/materials/1_normalmapping.lua#L30
https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets_BAR/Shaders/outlineShape.frag.glsl#L26

The current ZK CUS and default CUS shader are very close to what BAR has, so implementing these changes should require ~10 LoC across several files.

dvr333:
Awesome, that looks perfect.

GoogleFrog:
Smoke and underwater outlines could look fine. Is the thickness of the line able to be modified on the fly? The current widget has some settings that slightly reduce the thickness of the line when you zoom out.

I was also unable to get the BAR widget working myself.

lhog:
> I suppose if you wanted it to respect shields/fog/smoke you could write a depth value and have that preserved across the blurring, but I'm not sure how well that would work in the first place.

[...]

PR #3631, opened 2019-06-25
Fixed outline shader from BAR.

lhog:
1. Took Outline Shader v2 from BAR's 05ad3c1d599163bd79e4e2d6125f8810f5c134a9
2. Made outline artificially thicker (key values to play with in Lua code: BLUR_PASSES, BLUR_HALF_KERNEL_SIZE, BLUR_SIGMA, OUTLINE_STRENGTH), so it looks more or less like the one on the screenshots here: https://github.com/ZeroK-RTS/Zero-K/issues/3574
3. Added small piece of GLSL such that outline is also drawn around terrain. See outlineApplication2.frag.glsl on how to turn it off

Basically it looks same jagged as the existing version. But I'm keeping this flavor of outline so it's available for experiments.

The real breakthrough in terms of visual quality lies in https://github.com/spring/spring/commit/cd53d96da2bc6956f3ff62ab10d0d6d6c63a9a01 though

MSAA enabled deferred buffers should produce far better visual quality, although they come with the price:
* All shaders dealing with deferred buffers will need to be redone. Not a big deal, but require some knowledge only a very few in the community might posses.
* Increased GPU compute load/memory footprint might upset potato owners

### DIFF
--- a/LuaUI/Widgets/Shaders/outlineApplication2.frag.glsl
+++ b/LuaUI/Widgets/Shaders/outlineApplication2.frag.glsl
@@ -2,16 +2,17 @@
 
 uniform sampler2D tex;
 uniform sampler2D modelDepthTex;
-
-uniform float strength = 1.0;
+uniform sampler2D mapDepthTex;
 
 void main() {
 	ivec2 imageCoord = ivec2(gl_FragCoord.xy);
 
 	vec4 color = texelFetch(tex, imageCoord, 0);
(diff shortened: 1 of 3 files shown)


lhog:
Screenshots for lazy:
screen00864
screen00865

lhog:
Interestingly enough it's entirely possible to change outline v2 into something like that.
https://drive.google.com/open?id=1zewgCM78Hq8vS4WBz5RhdCV4zGR-aFfR

PR #3636, opened 2019-06-26
Outline shader v3

lhog:
Astrologers proclaim the week of outline shaders. Population of outline shaders doubles!

Depth discontinuity version.


GoogleFrog:
I don't know if I can use this without breaking it. I don't see a way to implement width changing.

lhog:
It is not easy to do currently.
But I thought I tuned existing values such that outline looks good at any zoom. No?

PR #3211, opened 2018-08-12
Update ru translation

N0U:

PR #3642, opened 2019-07-03
Update russian translation

Anarchid:
Mainly adds missing translations for units. `pw_units` and `campaign_units` are incomplete, but who cares about them anyway.

Bonus: remove duplicate section from tx config.


sprunk:
It would be good to figure out how #3211 stands in relation to the current state of Transifex translations and close or integrate that PR.

Anarchid:
Almost none of the stuff in #3211, most of which seems to be stylistic fixes and weird name swaps (emissary/spy, other highlighted issues), is present on transifex. 

I see no way to integrate them non-manually, and i think the manual integration could proceed on tx, maybe partially using stuff in #3211 as a hint.

Anarchid:
Brought up to date by using the Force. Since #3211 is obsoleted, also merging.

PR #3654, opened 2019-07-15
Context menu shield damage display improvements

danfireman:
### Overview
This PR covers making the context menu state how much a weapon's on-paper damage is multiplied by when calculating shield damage.

Hopefully this will prevent disappointment when people fire a Shockley at a Funnelweb's shield and discover it only does 10K damage...

### Examples
* A Cyclop's slow beam will show " - Shield damage:   33%" because slow damage is 33% effective against shields.
* For multi-damage-type weapons (like a Scorpion's lightning gun) the damage is calculated for each damage type then summed. So a reader will see " - Shield damage:   50%" in the case of the Scorpion.
* Weapons with no special damage adjustment vs shields, such as a Dominatrix capture beam or a Ravager's cannon will continue to not state anything regarding shield damage.

Additionally, the shield damage percentages for gauss and flamethrowers will now come from the actual damage calculation, rather than being recalculated (and possibly getting out of sync).

### Regarding shield_damage_total 
I would like to rename `wd.customparams.shield_damage` to `wd.customparams.shield_damage_per_frame` in all the places it is used to better reflect its actual value and rename `shield_damage_total` to `shield_damage`.

However, I am unsure if this could potentially break some people's widgets.

Thoughts?

### DIFF
--- a/LuaUI/Widgets/gui_contextmenu.lua
+++ b/LuaUI/Widgets/gui_contextmenu.lua
@@ -493,6 +493,23 @@ local function weapons2Table(cells, ws, unitID)
 		else
 			dam = val
 		end
+		
+		-- shield damage multiplier
+		local shieldDam = dam
+		if wd.customParams and wd.customParams.shield_damage_total then
+			-- Use the "real" damage value if possible
+			shieldDam = wd.customParams.shield_damage_total
+		else
+			-- Fallback on trying to calculate damage here
+			shieldDam = dam + damc + damd / 3 + dams / 3 + damc + damw / 3
+			local lowerName = name:lower()
+			if lowerName:find("flamethrower") or lowerName:find("flame thrower") then
+				shieldDam = shieldDam * 3
+			elseif lowerName:find("gauss") then
+				shieldDam = shieldDam * 1.5
+			end
+		end
+		local shieldDamPc = (100 * shieldDam) / (dam + damd + dams + damc + damw)
 
 		-- get reloadtime and calculate dps
 		local reloadtime = tonumber(cp.script_reload) or wd.reload
@@ -591,14 +608,11 @@ local function weapons2Table(cells, ws, unitID)
 			cells[#cells+1] = ' - DPS:'
 			cells[#cells+1] = dps_str
 		end
-		
-		local lowerName = name:lower()
-		if lowerName:find("flamethrower") or lowerName:find("flame thrower") then
-			cells[#cells+1] = ' - Shield damage:'
-			cells[#cells+1] = "300%"
-		elseif lowerName:find("gauss") then
-			cells[#cells+1] = ' - Shield damage:'
-			cells[#cells+1] = "150%"
+		if shieldDamPc ~= 100 then
+			if not cp.damage_vs_shield then -- Badger shield damage is handled above
+				cells[#cells+1] = ' - Shield damage:'
+				cells[#cells+1] = numformat(shieldDamPc, 2) .. "%"
+			end
 		end
 
 		if (wd.interceptedByShieldType == 0) then
--- a/LuaUI/Widgets/gui_contextmenu.lua
+++ b/LuaUI/Widgets/gui_contextmenu.lua
@@ -493,23 +493,6 @@ local function weapons2Table(cells, ws, unitID)
 		else
 			dam = val
 		end
-		
-		-- shield damage multiplier
-		local shieldDam = dam
-		if wd.customParams and wd.customParams.shield_damage_total then
-			-- Use the "real" damage value if possible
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Why do you create a customParam when wd.damages.shield likely has a corresponding value in the lua weapondefs table?

danfireman:
I tried using `wd.damage.shield` initially, but it doesn't seem to exist once you get to `gui_contextmenu.lua`... :confused: 

Both `wd.damage and wd.damage.shield` and `wd.damages and wd.damages.shield` are falsy inside `gui_contextmenu.lua`, even though the former has the proper values inside `armordefs.lua`.

GoogleFrog:
The WeaponDefs table available to lua is not the same as the tables that are fed into the engine to define the weapons.

defs_pre -> def files -> defs_posts -> engine -> WeaponDefs/UnitDefs

See Defs: https://springrts.com/wiki/Lua:Main

danfireman:
That's useful to know.

And indeed when looking at the properties available on `wd.damages` it looks like property number 4 (however you access that...) is usually the shield damage:
```
Electro-Stunner
1=1200
2=1200
3=1200
4=400
5=1200
0=0.001
paralyzeDamageTime=2
impulseBoost=0
impulseFactor=0
craterBoost=0
craterMult=0
```
However, for beams like the Cyclops slow beam, shield damage is set to 0 in `armordefs.lua`:

`wd.damage.shield = 0`
`https://github.com/ZeroK-RTS/Zero-K/blob/2aa36c23bb82689113c8ca271514661e84e800ad/gamedata/armordefs.lua#L156`

And indeed:
```
```

sprunk:
>property number 4 (however you access that...) is usually the shield damage

The map from `armordefs.lua` keys like `shield` to LuaUI keys like `4` is the `Game.armorTypes` table (see the example below).

A new customparam wouldn't strictly be needed because you can do this:
```lua
local shieldArmorTypeID = Game.armorTypes.shield
local GAME_SPEED = Game.gameSpeed

local function GetTotalShieldDamage(wd)
  if wd.type == "BeamLaser" then
    -- beamlasers have engine damage set to 0 so we can't just use wd.damages everywhere
    return wd.beamTime * gameSpeed * wd.customParams.shield_damage
  else
    -- non-beamlasers don't have beamTime so we can't just use wd.customparams.shield_damage everywhere
    return wd.damages[shieldArmorTypeID]
  end
end
```

but I agree a new customParam achieves this more cleanly.

GoogleFrog:
Ok, I am satisfied that the customParam is worth adding.
- The customParam should be called stats_shield_damage to indicate that it is only used for the UI (there are other similar custom params that start with stats_).
- Remove the "-- Fallback on trying to calculate damage here" branch because this branch should never execute. It just looks like somewhere to hide bugs.
- Display mixed or unclear shield damage as a number instead of a percentage. Your code tells me that Knight does 52% damage against shields, but this number is not enlightening. The player is liable to ask the question "52% of what?".
- To be clearest about the mechanic you could write "Shield Damage: 440 (240 + Blue200Blue)". Only do it for mixed damages. This shows enough of the working to teach people a general rule about shield damage.
- Flame is continuous so it may be best to leave its modifier as a percentage. See how it looks.
- Badger is not adequately handled above as it is unclear where the hit and mine damage go when hitting a shield, it may as well be included.

danfireman:
* Done
* Removed, but see below.
* Mixed damage now displaying as number.
* Note that this requires reintroducing the duplication of calculating shield damage as the breakdown of where damage came from is lost in `armordefs.lua`. I've added it in because I think improved player experience trumps code quality.
* I can't find any mixed flame weapons and I doubt they'd be made given the thematic issues. When they are, their creator can add in the code to make them display nicely.
* Looks a bit redundant but I've put it in anyway:
BadgerMineStats

Also, badger specific code now checks for mine presence.

LuaUI/Widgets/gui_contextmenu.lua
```diff
@@ -570,7 +579,7 @@ local function weapons2Table(cells, ws, unitID)
 			show_dps = false
 		end
 		
-		if cp.damage_vs_shield then -- Badger
+		if name:lower():find("mine") then -- Badger
```
GoogleFrog: Since we're doing things properly you shouldn't base logic on weapon names. Checking for the existence of the customParams spawns_name and damage_vs_shield should be sufficient here.

GoogleFrog:
I don't like the style still but I explaining why is more work than I am willing to put in.

GoogleFrog:
See https://github.com/ZeroK-RTS/Zero-K/commit/5a3474a860c35207ce8700d2ecc8dd6f184214a1.

PR #3656, opened 2019-07-17
Shield damage context menu bugfixes

danfireman:
Fixing bug that made shield damage always be displayed.

Also improved shield damage display for multi-shot weapons.


GoogleFrog:
Thanks, I seem to have only tested with units that show DPS instead of damage.

PR #3671, opened 2019-07-27
WIP: Changing shield transparency at low hp so they are visibly weak

danfireman:
Initial transparency changes to low hp shields.

Still need to do more testing, but need computer fixed first.


GoogleFrog:
Won't `(0.2 + charge / (charge + 200))` exceed 1?

danfireman:
Yep.

I figured it will be 1 at 800hp and so not make shields look too different for their low hp.

Once I can do more testing I will probably change to `(0.2 + X * charge / (charge + 200))` for 0.8 <= X <= 1. Also got tweaking of the 0.2 and 200 to do.

Just blocked on computer issues.

PR #3674, opened 2019-07-30
Synchronize in-game ranks with lobby ranks

esainane:
These are significantly lower resolution, but I can't seem to find
equivalent higher resolution sources elsewhere. These are the versions
used in the lobby, and appear to also be visually identical to those
used in the website.

Hopefully this should result in less confusion of the form 'silver is
higher than gold?'


sprunk:
@DeinFreund might have the full size versions.

esainane:
I think that the versions including the silver icon are used in more places than I initially realized. I still think the lobby light blue -> dark blue is clearer, but this probably needs extended discussion since changes one way or another would affect more than I initially thought.

DeinFreund:
https://github.com/ZeroK-RTS/Zero-K-Artwork/tree/master/eloRanks/unshaded

esainane:
This was resolved in the other direction, which had higher resolution graphics anyway.

PR #3681, opened 2019-08-02
WIP: Dynamic player list

Aquanim:
This playerlist widget is "dynamic" in that it reuses as many GUI elements as possible when a player's status changes, rather than recreating everything. This will hopefully improve overall game performance.


Aquanim:
screen00030

The same widget handles the minimal list in the lower-right (displaying player names, who is excessing/stalling/storageless/lagging) and the more detailed window which can be opened in the upper-left.

In the detailed window, clicking on the resource bars gifts resources. Ctrl-click on player name gifts units. I intend to handle as much interaction in this kind of way as possible since buttons take up screen space. At the moment I can play a game with the large playerlist active but I suspect it would annoy at least some people to do so; making it yet larger is contraindicated.

Many things about the playerlist display are customisable.

Tasks for this to not be WIP:

-  [ ] General code cleanup. I didn't know much about Chili (or for that matter Lua) when I started and I am hardly an expert now.
- [ ] Ensuring that it plays well with the GUI as a whole (e.g. disable conflicting playerlists).
- [ ] Adding commshare controls. This may be optional if tab-playerlist coexists with it. (At the moment the widget can display the *consequences* of commshare accurately but does not control them.)

Additional things that would be nice to have:

- [ ] Adding whisper and ignore controls. I believe both of these require work outside the widget to accomodate.
- [ ] Alternative playerlist layout more similar to the current tab playerlist for those who might prefer it.

GoogleFrog:
How are units shared? How is metal and energy shared in a non-poweruser way?

Aquanim:
At the moment there is no sharing through the minimal list to reduce risk of misclicks. A suitably stringent set of required modifiers could work around this.

Aquanim:
Looking back at it now I am feeling like the code in this widget is too kludge-y for me to want to be responsible for maintaining it. A large scale rewrite is probably required.

PR #3683, opened 2019-08-04
Experimental autonormal feature of the unit/feature shader

lhog:
As I couldn't do anything productive this weekend I came by shadertoy to see if anything inspires and it truly did (like always). Here is the port of https://www.shadertoy.com/view/ttlSRS to ZK units/features that have no artist normal map. Call it crazybump for poor && lazy.

Autonormal on
Spring Screenshot 2019 08 04 - 13 53 36 64

Autonormal off
Spring Screenshot 2019 08 04 - 13 53 32 78

Individual units and wrecks might require some hand tuning on how pronounced the effect is.


lhog:
Toggle on and off with `/luarules autonormal` in the game

Issue #3339, opened 2018-11-16
Scallop shoots depthcharges at the land that a Claw used to exist on


GoogleFrog:
Reproducible on Red Comet. I suspect the Claw is being sent underwater and Scallop then decides to shoot at it. This would also break missile homing. This is likely solvable by applying the same in-place invulnerability and invisibility of Puppy, which now hides itself from its launch location.

sprunk:
Maybe related (haven't had time to check that replay) #3197

GoogleFrog:
The PR is untested.

Issue #2214, opened 2017-05-10
Underground unit hiding leads to homing issues


GoogleFrog:
Watch Scalpel fire at Claws. The shots will sometimes go underground as the Claw is sent underground when it fires (for stats purposes). Outlaw, Puppy and missile silo missiles use this form of unit hiding. Changing their aimpoints to counteract the underground offset may well work around it.

GoogleFrog:
This issue can occur when a Scalpel fires at a living Claw, then the Claw fires.

Anarchid:
if the PR works well enough, the Claw could be also made invincible (block all damage) in a similar way to `untargetable`, and then kept in place rather than moved.

I guess active glints can be also kept invincible (rather than selectively invincible only to their parent Starlight).

PR #3702, opened 2019-08-15
Claw is `untargetable` and `neutral` after firing

### LINKED ISSUE #3413: Dagger shoots the ground where a Claw recently fired.

Anarchid:
Takes a few tricks out of Glint's playbook and applies them to Claw after it fires.

The `untargetable` URP prevents everything that could affect the Claw while mediated through a command (with gadget help).

The `neutral` property prevents auto-targeting by units that don't have an explicit command to target that unit.

This should fix #3413, fix https://github.com/ZeroK-RTS/Zero-K/issues/2214 and fix https://github.com/ZeroK-RTS/Zero-K/issues/3339.

Untested. If this works, should also be applied to tac silo missiles and Puppy.


GoogleFrog:
I am not convinced that this clears a target that is already acquired, but I suppose the PR can't hurt.

PR #3703, opened 2019-08-15
Clean up text file encodings, and some formatting

esainane:
This expands the set of definite text files, allows linux users to check out files with their native line endings without interfering with the ability of windows users to check out files with their native line endings, ensures files in the index are consistent in having LF (git can automatically convert them to CRLF on checkout for platforms that want this), ensures text files have a trailing newline at the end of file, encodes files as UTF-8 where possible, and strips trailing whitespace for lines that have other content.

- 4 file extensions are added to the list of definite text files, .h, .glsl, .fs and .json, joining the existing list of .lua, .txt, and .tdf.
- 53 files were previously checked in as CRLF without conversion, and are now converted to LF in the index - windows users can still automatically checkout as CRLF, git will handle conversion automatically.
- 509 files were previously missing a trailing newline.
- 6 files were encoded in iso-8859-1 and could be trivially converted to UTF-8.
- `sounds/music/license.txt` appeared to have been misencoded at some point - I've replaced what appears from context to originally have been a hyphen with an ASCII hyphen.
- 757 files previously contained trailing whitespace on lines which had other content.

Not touched:
- Text files in `LuaUI/Fonts`. These contain literal bytes between \x0 and \x255, so their "encoding" is left in binary.
- Encoding of `LuaUI/Widgets/cmd_area_attack_tweak.lua` and `lups/headers/tablebin.lua`, detected as '`unknown-8bit`'. The former file has binary bytes as part of the widget description; I suspect the correct solution is adjusting it to use escaped codes, similar to what eg. Global Build Command uses, but that is probably best discussed in a follow-up PR. The latter file ends with CHILLCODE\x99 in a comment; I have no idea what this is supposed to indicate or if there is an alternative we can use that leaves files friendlier to external tooling.
- Lines consisting entirely of whitespace. Working out what the correct level of indentation is and adjusting is going to be slightly more involved. There are many lines which consist entirely of whitespace but are not meant to have any indentation at all, for example, still in the index.

Overall, this makes for a much cleaner working tree which plays nicely with external tools.


GoogleFrog:
bos files were changed, but I suppose it is ok. The chili changes should be applied to Chobby immediately.

sprunk:
>I don't know why so many files have changes and pushing elsewhere seems to do nothing

PR #3707, opened 2019-08-16
Renamed Puppy to Bunny

claw4oo:
Renamed Puppy to Bunny, using google translate for language localizations.

Basically used sublime text to find and replace every instance of Puppy I could find.

Based on an earlier chat in the ZK main chat, where a suicidal bunny fits better than suicidal puppies.
I just did the renaming to try it.
It'll probably need some testing incase I missed anything.
(sprang) @Cerin#7618
@Anarchid#7320


LuaUI/Configs/lang/units.pl.json
```diff
@@ -990,7 +990,7 @@
 		"helptext" : "Jugglenaut posiada imponującą wytrzymałość i podwójne promienie grawitacyjne, które mogą przyciągać lub odpychać inne jednostki. Może również skakać, co w połączeniu z jego dużą masą pozwala mu zgniatać wrogie jednostki, skacząc na nie."
 	},
 	"jumpscout" : {
-		"name" : "Puppy",
+		"name" : "królik",
```
Anarchid: Wrong case on first letter

LuaUI/Configs/lang/units.ru.json
```diff
@@ -1014,7 +1014,7 @@
         "name": "Drone"
     },
     "jumpscout": {
-        "helptext": "Щенок (Puppy) - быстрый рейдер-камикадзе, эффективно уничтожающий слабозащищённые цели. Находясь рядом с обломками постепенно производит из них новых щенков.",
+        "helptext": "кролик (Bunny) - быстрый рейдер-камикадзе, эффективно уничтожающий слабозащищённые цели. Находясь рядом с обломками постепенно производит из них новых щенков.",
         "description": "Шагающая ракета",
         "name": "Щенок"
```
Anarchid: Desc is updated but name is still old.

CrazyEddieTK:
Please do not rename Puppy to Bunny.

Aquanim:
I don't think any further renames in general are a good idea.

claw4oo:
I'll make a forum post to get some more opinions on it. Though, this pull request can remain open until it's needed or desired

LuaUI/Configs/lang/units.ru.json
claw4oo: I've updated it, but I am not currently interested in setting up transifax myself right now. Thanks for pointing it out though.

sprangg:
I agree with no renames. The OTA names are gone which is the most important part. I'm a bit incredulous Puppy -> Bunny hasn't gone through during the 8 years it's been occasionally suggested, but its window is now past. And Puppy is a decent name anyway.

sprunk:
I'm also disappointed that this did not happen during the rename rounds but I agree with others that this ship has sailed. Keep in mind there are multiple places that cannot be modified and will still refer to the unit as Puppy (for example youtube videos or veteran players) so a rename would come with a massive confusion cost. Our current stance is that this cost would be too high and I don't expect this stance to change anytime soon.

Issue #3655, opened 2019-07-16
Bug: Sudden accelerating (and exploding) Funnelweb


danfireman:
At 19:29 in this battle: http://zero-k.info/Battles/Detail/745563

A Funnelweb is given a command to move a little bit away from the spire it's on.

Shortly before it would reach the base of the spire it seems to lose its grip on the spire and fall (a short distance), crushing the units below it, then accelerate at high speed away from the spire, crashing into a Jugglenaught about 200 elmo away from the spire and exploding.

danfireman:
The Funnel is going sufficiently fast that, if I move the Juggle out of the way, the Funnelweb will hit the valley wall and then go flying about 2000 elmo up into the air...

danfireman:
Moving the little units at the base stops the Funnelweb losing its grip and shooting off.

sprunk:
Engine 104, many pathing changes since then.

PR #3709, opened 2019-08-18
Wreck hitbox improvements and some non-wreck ones too

danfireman:
Adding proper wreck hitboxes to:
* Athena
* Phoenix
* Wasp
* Moderator
* Raptor
* Owl
* Tarantula
* Venom (previous intangible)
* Recluse (massed Recluse may now be more vulnerable to attack, previously their HUGE wreck hitbox completely blocked off attacks from ground based units)
* Scorpion (this means Scorpions can fire over their own wrecks)
* Ravager (this means Ravagers can fire over their own wrecks!)
* Dominatrix

Tweaks to live collision boxes to:
* Athena (lenthened and squashed to better match model)
* ~~Moderator (now a box)~~
* Scorpion (Scorpions can now fire their lightning over each other, and even fire their lasers if against something sufficiently tall)


GoogleFrog:
- Non-spherical ellipsoid hitvolumes require testing if they are to be used widely. At one point they were de-supported in the engine due to bugs. I have not paid attention to how/if they were fixed.
- Check the selection volumes of the modified units. There is a way to do this via unitdefs_posts.
- I would like pictures of the changed units.

danfireman:
What kind of testing would you want for the non-spherical hitvolumes? Would an image showing lotuses hitting the ellipsoid from the four cardinal directions suffice?

GoogleFrog:
Testing non-spherical ellipsoids is an open-ended task that requires a bit of creativity. I don't know all the situations that may cause a collision volume to misbehave. Whoever does it will need to fiddle around and try things until they are satisfied. A fundamental part of this process is seeking out tests with unknown results or even that you may expect would fail. Setting up a few situations that you expect to succeed is insufficient.

Some things to test would be the ability to aim past allies and the ability to not friendly-fire allies. Maybe a large group of jiggling Moderators with ellipsoids would help.

Here is a video of me testing whether rapidly moving aimpoints can friendly fire: https://www.youtube.com/watch?v=jR1H8U6IjFo

Something similar could be done.

GoogleFrog:
- Nothing shows collision volumes by default. There is a constant in unitdefs_posts that can be set to make collision volumes show the selection volume.
- Units with collision volumes.

danfireman:
So I've tried a variety of things:
1. Crowds of moderators with variable sized hitboxes (e.g. very long, very tall, very flat and/or positioned in completely the wrong place such as a long way from the model) trying to hit things on various elevations while themselves at various elevations.
2. Crowds being lobbed places.
3. Crowds being pushed places.
4. Projectile units (ravager and glaive) trying to fire through the crowd.

I was actually expending this bug to make an appearance, but it didn't: https://github.com/ZeroK-RTS/Zero-K/issues/3655

### Results:
#### Lobbing and pushing
Lobbing seemed to do damage, which surprised me, but upon inspection is seems that lobbing can normally do damage when units collide, it was just the very long hitboxes making collisions more likely.

Pushing via Newton or Archer similarly showed effects of large hitboxes, but otherwise did not look unusual.

#### Friendly fire (hitscan)
No instances of friendly fire from the moderators or lotus.

#### Friendly fire (projectile)
Ravagers DID show serious friendly fire issues with extreme collision boxes, frequently failing to determine that their trajectory was blocked. The collision sphere the Moderator has also leads to friendly fire, but at a much reduced rate. The proposed box Moderator hitbox may have a higher friendly fire rate, but the margin of error is too high to be certain (could be anything from 1x to 2x the friendly fire rate).

I also examined the proposed Scorpion hitbox and compared to the current one. The new hitbox shows a roughly 50% reduced rate of friendly fire from ravagers compared to the old one. This is mostly due to its lower profile (the corners take plenty of fire, but nothing compared to what the invisible hemisphere was taking).

I also got derailed for a couple of hours because apparently having a `modelradius` a lot smaller that the collision volume causes the latter to malfunction. :(

### Actions
I am backing out the change to Moderator hitbox. The potential extra friendly fire is not worth the rather minimal gain (the original round hitbox is good enough really).

I am keeping the Scorpion hitbox change in the PR for now. The reduced friendly fire, ability to fire over itself and far better match between the visuals and what actually gets hit makes it the superior choice in my view.

The new Athena hitbox should be fine. Friendly fire isn't much of a concern with flying units and fire passing through the Athena will happen less with the newer hitbox, improving look-and-feel (the fast speed and small size of the Athena means any individual shot is not very noticeable, but what happens to the average shot is noticeable). Its main balance effect would be to make Glaives, Reavers and Razors more damaging to Athenas.

danfireman:
Regarding the two remaining things:
* I've checked the selection volumes of the units. They're their usual big spherical selves (notably much larger than collision volumes). Anything you want me to do here?
* Collision volumes:
SkorpHitboxes
Scorpion
AthenaHitboxes
Athena

PR #3721, opened 2019-09-03
Replace the AA laser sound with something better

Anarchid:
The current AA laser sound is basically a simple sine wave. It completely conflicts with most of the rest of the soundscape, is low quality, and is very loud and thus feels like it breaks immersion.

This replaces it with something a bit easier on the ear and a bit more interesting in construction.

Here's a video that showcases the replacement sound https://www.youtube.com/watch?v=NTBwS0PYXXg&feature=youtu.be


GoogleFrog:
I have not had a chance to look at this. I don't want to base my perception off the video because all the sounds on the video sound different and worse.

sprunk:
I can hear mouse clicks so the sound is probably from a microphone.

Anarchid:
Yep, it looks like OBS hooked onto mike rather than just capture the stream.

Anarchid:
Bump.

sprunk:
I think GF expects a video where OBS didn't hook onto Michael and captured the stream instead.

Anarchid:
Clarifying that would help.

GoogleFrog:
Any kind of comparison would be nice, otherwise this is left until I can get around to it and are reminded.

[aaCompare.zip](https://github.com/ZeroK-RTS/Zero-K/files/4258893/aaCompare.zip)

This is eight Gremlins firing. The new sound seems to be worse than the old one.

Anarchid:
I guess the higher frequencies in the candidate sound could be annoying. I'm still liking it a lot better on the grounds of actually having any texture.

sprunk:
I think the PR sounds better when there's like 1 or 2 units shooting (video in the first post) but the current sounds better when there are many units shooting (`aaCompare.zip`).

sprunk:
Sounds worse for multiple units shooting; reopen if a different sample appears

PR #3732, opened 2019-09-28
Fixed depth linearization for DoF

dvr333:


LuaUI/Widgets/Shaders/dof.fs
```diff
@@ -128,9 +128,9 @@ float LinearizeDepth(vec2 uv){
 
     float n22 = viewProjection[2][2];
 
-    return ((abs(((1.0 + depthNDC) * (1.0 + n22))/(2.0 * (depthNDC + n22)))
-      * (distanceLimits.y - distanceLimits.x)) + distanceLimits.x) / BLUR_START_DIST;
-    // return abs(viewProjection[3][2] / (viewProjection[2][2] + depthNDC)) / BLUR_START_DIST;
+    // return ((abs(((1.0 + depthNDC) * (1.0 + n22))/(2.0 * (depthNDC + n22)))
```
sprunk: It would be good if there was some comment on why this line is disabled and what can be done about it, something like "disabled in favour of the one below because of X, reinstate if Y".

dvr333:
Not sure why there is a conflict, but the dof-mouse-focus version is the correct one.

sprunk:
Conflict was due to d644d6caadee1ab5b6e74c32ba0a761fea640821 screwing up mergeability.

PR #3737, opened 2019-10-09
Don't offer beamlaser on lvl 2 (already baseline)

sprunk:


sprunk:
will test when i get back home in 10ish hours

Issue #3734, opened 2019-10-08
Unit shadows aren't being drawn properly.


aeonios:
I have both unit and terrain shadows enabled.

I see shadows for terrain, smoke, and nanoframes show up properly, but as soon as a nanoframe is completed its shadow disappears. There seem to be some weird effects as the nanoframe approaches completion, although I can't tell if that's just the nanoframe doing its final fill-in or if the shadow actually starts to disappear as the nanoframe is replaced by the unit model.

I don't think explosion debris is casting shadows either, so it's probably an issue with the solid models not being drawn to the shadow map.

[infolog.txt](https://github.com/ZeroK-RTS/Zero-K/files/3700288/infolog.txt)

Anarchid:
ZK has been drawing everything using with CUS shaders for a while (this is why you may notice some emissives pulsing etc). Possibly this fails for you?

Can you provide a screenshot of the issue for completeness?

aeonios:
Sure.
![screen00017](https://user-images.githubusercontent.com/7027766/66400209-0a211680-e9af-11e9-9d96-198b53c41b54.jpg)

I've seen the glowing unit thing and that seems to work fine. I've also noticed some new projectile lights on flaming debris. :P Aside from weird screen flickering from duck and to a lesser extent fencer missiles I haven't really noticed anything else being off though.

Anarchid:
Does the problem persist if you `/luarules disable` or `/luaui disable`?

aeonios:
[infolog_full.txt](https://github.com/ZeroK-RTS/Zero-K/files/3707152/infolog_full.txt)
Hopefully a better infolog.

/luarules disable does fix the shadows. /luaui disable does not.

sprunk:
@aeonios you can try the `bug3734wa` springsetting workaround, see #3739.

aeonios:
No, this workaround does not work. I've had it in my local zk $VERSION the whole time without realizing it and I still don't have shadows.

sprunk:
@lhog ^

lhog:
Also with disabled custom shadow map depth shader you should see terrain shadows casted by units. Do you see those?

aeonios:
[...]

PR #3739, opened 2019-10-10
Missing CUS shadows workaround

lhog:
https://github.com/ZeroK-RTS/Zero-K/issues/3734

In the future we should try to find someone else who doesn't see custom shadows in order to confirm the current theory of a buggy old NVidia driver.

### DIFF
--- a/LuaRules/Gadgets/api_custom_unit_shaders.lua
+++ b/LuaRules/Gadgets/api_custom_unit_shaders.lua
@@ -585,18 +585,6 @@ local function ObjectFinished(rendering, objectID, objectDefID)
 end
 
 
---[[
-
-	drawList            = {},
-	materialInfos       = {},
-	bufMaterials        = {},
-	bufShadowMaterials  = {},
-	materialDefs        = {},
-	loadedTextures      = {},
-
-
-]]--
-
 local function _CleanupEverything(rendering)
 	for objectID, mat in pairs(rendering.drawList) do
 		local DrawObject = mat[rendering.DrawObject]
--- a/LuaRules/Gadgets/api_custom_unit_shaders.lua
+++ b/LuaRules/Gadgets/api_custom_unit_shaders.lua
@@ -253,7 +253,6 @@ local function _CompileMaterialShaders(rendering)
 				matSrc.shadowDefinitions,
 				string.format("MatName: \"%s\"(%s)", matName, "Shadow")
 			)
-
 			if luaShader then
 				if matSrc.shadowShader then
 					if matSrc.shadowShaderObj then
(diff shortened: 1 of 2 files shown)


sprunk:
thanks!

Zlodeiko:
> #3734
> 
> In the future we should try to find someone else who doesn't see custom shadows in order to confirm the current theory of a buggy old NVidia driver.

Well, I have exactly that problem like in #3734, which I can’t cope with for several months now. I don’t really remember if the shadows worked before, but now they don’t work (the shadows are visible only from the relief, some effects and nanoframes). 
I don’t have any buggy old nvidia drivers because I have radeon rx570 with latest drivers.
I tried some common solutions - reinstalled the game, reinstalled the video card drivers - with no results.
I read threads on this issue here - understood almost nothing. What are these .lua files and what to do with them? Can anyone explain in detail to a simple lobster how to make these damned shadows work?

lhog:
I can send you some instructions how to work it around. In the same time I really want to understand why it might be happening and fix it properly. The fact that it doesn't work only for a few people suggests  me that something tiny is missing.  Are you tech savvy by a chance in a way that you can run custom widgets and upload the text or pictures they produce? If so contact ivand (another nichlkname of mine) on Discord.

Zlodeiko:
Figured out a little. Added that "bug3734wa = 1" entry in springsettings.cfg and shadows appeared. 
But I also noticed another problem that may be related to #3734. I can see bumpmapping water only by turning it off in "menu - settings - graphics - map detail" and turning it back on. But after a restart, it does not appear again.

lhog:
Yes, you could add a workaround that turns off custom shadow mapping, but it is what it says a "workaround". The proper solution is to find the cause and fix it. This issue is not reproducible on my GPU/drivers as well as for most of the players,  so I need someone with the issue and ability to put some extra files in the right places, run them and upload me results.

Are you qualified to do the above and willing to help?

Zlodeiko:
Sounds like something not very complicated) I will try to. I'll write to Discord when I'm ready.

lhog:
Thanks. Note, during the weekdays I'm only available for something more sophisticated than a quick chat in the evening after 6-7 pm (UTC+3)

lhog:
I managed to identify the issue. At least for AMD. The issue was caused by drivers to not accept fragment shader of shadow map program. Normally this shouldn't be happening, but compatibility profile of OpenGL in spring and/or some FFP legacy setting in the engine cause the described behavior. Perhaps this combination is described somewhere, but it's like looking for a needle in a haystack.

In the end I worked around that issue the same way it's done by the engine. BAR has patched version of api_cus and material files. I should be able to make a PR for ZK once I have a spare minute.

PR #3741, opened 2019-10-12
Elevating centre of Bertha so that terraformed Berthas are targettable

danfireman:
Currently Redbacks, Scorpions and Fleas can't target Berthas that have been terraformed up on high spikes. This is because the targetting centre of the Bertha is so low that the ground of the platform it's on blocks line of sight.

This pull request elevates the Berthas centre so that Fleas, Redbacks and Scorpions will be able to fire at it, while leaving the collision volumes and selection volumes untouched.

### DIFF
--- a/units/staticheavyarty.lua
+++ b/units/staticheavyarty.lua
@@ -10,15 +10,13 @@ unitDef = {
   buildingGroundDecalType       = [[staticheavyarty_aoplane.dds]],
   buildPic                      = [[staticheavyarty.png]],
   category                      = [[SINK]],
-  selectionVolumeOffsets        = [[0 -40 0]],
-  collisionVolumeOffsets        = [[0 -40 0]],
+  collisionVolumeOffsets        = [[0 0 0]],
   collisionVolumeScales         = [[70 194 70]],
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Change the aimpos not the midpos.

sprunk:
There's a way to keep the aimpos low on un-terraformed buildings (for scythes etc) and only raise it when terraformed. See https://github.com/ZeroK-RTS/Zero-K/blob/0372f4db69bc6a7c2d55f7bf370ab31bba42f645/scripts/turretheavylaser.lua#L50

danfireman:
@GoogleFrog : Changed to aimpoint. Also properly centred aimpoint and hitbox. Also added wreck hitbox while in the area.

@sprunk : As discussed on Discord, not too sure how to get that working.

GoogleFrog:
I don't think the GG is required since BB is rarely damaged as part of a knife-edge battle. Does this aimpoint work at all build orientations? I recall there being an issue with horizontal offsets for aim and midpoints.

danfireman:
Aimpoint dot, collision volumes and where a Redback fires seem consistent and sensible in all your orientations.

Issue #3500, opened 2019-03-31
Look at Zephyr vs planes


GoogleFrog:
https://zero-k.info/Forum/Thread/28092

sprunk:
Zephyr is trash against planes. This is because it has a non-homing projectile which planes often dodge through their normal movement patterns. Engine leading being unreliable (#1881) doesn't help. Zephyr also has 0.25° inaccuracy (compared to Razor's 0.1°). Having no alpha means that getting lucky won't result in meaningful damage. Plane hitvolumes seem fairly small so missing is easy.

Missiles (present on the other sea AA - Flail and Angler) have a random hax that makes them ignore radar wobble. Zephyr doesn't:
![image](https://user-images.githubusercontent.com/2573076/66119284-a9719280-e5d8-11e9-9850-aa6b799da0f0.png)

Here's some unit-specific situations. Likho has a forward-facing missile which reduces the time it spends flying forward and a relateively large turning circle which increases the time it spends turning. 
![image](https://user-images.githubusercontent.com/2573076/66118449-ad041a00-e5d6-11e9-9552-3a10396087e3.png)
![image](https://user-images.githubusercontent.com/2573076/66118461-b42b2800-e5d6-11e9-8dbd-abfb743297e5.png)
![image](https://user-images.githubusercontent.com/2573076/66118473-bab99f80-e5d6-11e9-9140-283173efb32e.png)
![image](https://user-images.githubusercontent.com/2573076/66118485-c0af8080-e5d6-11e9-93cb-5bcdbcd5cfc0.png)
![image](https://user-images.githubusercontent.com/2573076/66118495-c73df800-e5d6-11e9-92de-6da7e7db3088.png)

Normally, flying straight on approach gets Zephyr clean hits but with a bit of manual jinking you can dodge that too:
![image](https://user-images.githubusercontent.com/2573076/66119518-0ec58380-e5d9-11e9-94a1-e30a4f694db4.png)

Raven diving is a turning point that introduces misses.
![image](https://user-images.githubusercontent.com/2573076/66119180-76c79a00-e5d8-11e9-8783-e64aa1e099d4.png)

Swifts seem to usually move in arcs when chasing other planes (screenshot tbd).

sprunk:
I am unsure whether this can be solved without any major change. The ideal would probably be to have some sort of ship flex AA that can handle planes but this involves touching other units. Alternatively Zephyr's weapontype could be changed but that loses uniqueness.

[...]

PR #3742, opened 2019-10-20
Laser DPS moved to EMG; Laser range moved to sight.

Oflameo:
This Toad will do what it already does, but less janky.

### DIFF
--- a/units/jumpaa.lua
+++ b/units/jumpaa.lua
@@ -40,7 +40,7 @@ unitDef = {
   objectName          = [[hunchback.s3o]],
   script              = [[jumpaa.lua]],
   selfDestructAs      = [[BIG_UNITEX]],
-  sightDistance       = 660,
+  sightDistance       = 820,
   trackOffset         = 0,
   trackStrength       = 8,
   trackStretch        = 1,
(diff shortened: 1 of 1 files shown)


sprunk:
How is the machinegun less janky than the laser (which as a hitscan weapon won't randomly miss, see #3500)?
Why would Toad be the exception to the "sight caps at 660" rule that other AA conforms to?

I assume the underlying "what problem are you trying to solve" question is answered by "Toad generally sucks" as outlined here https://zero-k.info/Forum/Thread/29054 but that thread has exactly zero replays so far. It would be good to distill the most important points for GoogleFrog here since he is absent for now so probably isn't following that thread but ultimately is the one you'll need to convince.

Oflameo:
This Toad is a specialist unit for seeing air units before they hit, therefore is allowed of see over 660 elmos. It is also why it is more expensive than other AA. Having just one weapon removes aiming issues caused by having 2 weapons with different ranges which have the same effect (does normal damage to damage air).

GoogleFrog:
I took the LOS increase suggestion, but kept both weapons. https://github.com/ZeroK-RTS/Zero-K/commit/9aeb920208e960c2b02a67699f39abfcadb5862f

PR #3743, opened 2019-10-20
WIP: Fixes/Improvements to DoF, bloom, and HDR

aeonios:
-Fixes DOF shader not working properly (reset matrices).
-Fixes the deferred rendering widget so that it no longer trashes the gl state with its texture binds.
-Moves bloom into DoF where it shares downscaled/blurred textures.
-Fixes bloom so that it works the way you'd expect it to.
-Adjusts the HDR tone mapping formula to provide smoother white shifting.
-Adjusts widget layers for some gfx widgets so that things are gathered and drawn in the proper order.
-Removes boke blur until that shader code can be separated out and fixed to give proper variable boke (to be re-introduced as a 'high-quality' option).


sprunk:
Would be good if @dvr333 checked this because I don't want a tug-of-war style back and forth.

dvr333:
@aeonios So, is the issue with DoF that it was set up with a fixed set of samples that can't be easily computed on the fly? I mean, it's not a huge ask to set it up as a one-pass approach to make variable sized bokeh easier, but I'm curious what exactly you have in mind for a proper DoF effect. I've heard your problems with the DoF effect as is, but I'm not sure entirely what you are looking for.

aeonios:
Pretty much the shader and widget were a mess, full of loose ends and spaghetti that I couldn't work with. I figured out basically how the bokeh blur shader works from the shadertoy example well enough to reimplement it, and with only 3 steps since you can write the to the R G and B textures in a single pass using MRT. Now that I've got a decent working codebase to start from I can implement it separately as a high-quality option rather than always being on.

Your shader didn't seem to be producing variable blur widths either, which is something I already tested trivially on the shadertoy sample and which works fine. It interpolated between blurred and not but the blur width seemed to be constant and I never saw any bokeh (which may be since that only appears as you approach 3.0 filtersize while you had it capped at 2.0 for some unknown reason).

I started working on this because your widget was failing to draw, probably because some other widget messed up the texture matrix, but I have no idea which. Then I wanted to fix the excessively narrow field of view, but that was buried under a mountain of magic constants and global variables, and the function I wanted was too simple to be sorting through all of that junk for it. It was simpler just to start over from scratch.

Another thing that made no sense was why you split it into near and far textures. Given the way it works you wouldn't save anything that way, just add to the texture bandwidth usage.

I'll reimplement bokeh blur tomorrow. The implementation isn't really difficult but integrating all of its components properly is going to be a pain.

dvr333:
Okay, that makes sense.

So, the MRT thing was already in there, but yeah, it was buried in a few other things. The variable blur width happened through scaling the sample, but I realize that's hard to spot. Basically, during the GetFilterCoords() function, the filterRadius parameter scales the hardcoded sample parameters, so if the filterRadius parameter is 0.5, then the blur has a radius of KERNEL_RADIUS * 0.5, but that isn't straightforward, and I wasn't a huge fan of it anyway. 

I'm kinda surprised you didn't see a bokeh in my shader, but it also isn't set up for very wide blurs (about half of the shadertoy sample) so that's probably it.

As for near vs far, the reason for that is to keep straight the blur for objects nearer to the camera than the focal plane, and objects farther away than the focal plane. Objects farther away just need to check during the blur pass that they aren't sampling less out-of-focus pixels, and if they are to instead sample some other pixel from the equally out-of-focus region; but objects nearer to the camera need to make sure, while going over in-focus pixels, that they sample any out-of-focus pixels that would be out of focus enough that the blurry bits would overlap the in-focus pixels. This also requires blurring an alpha channel that starts out with the edges of the different focus kernel values, so you can blend the near-to-camera pixels properly.

LuaUI/Widgets/gfx_projectile_lights.lua
```diff
@@ -9,7 +9,7 @@ function widget:GetInfo()
 		author    = "GoogleFrog (beherith orgional)",
 		date      = "5 March 2016",
 		license   = "GPL V2",
-		layer     = 0,
+		layer     = 999999,
```
GoogleFrog: This is a NOOP because the widget barely uses the widget interface. Revert it.

aeonios: done.

aeonios:
> As for near vs far, the reason for that is to keep straight the blur for objects nearer to the camera than the focal plane, and objects farther away than the focal plane. Objects farther away just need to check during the blur pass that they aren't sampling less out-of-focus pixels, and if they are to instead sample some other pixel from the equally out-of-focus region; but objects nearer to the camera need to make sure, while going over in-focus pixels, that they sample any out-of-focus pixels that would be out of focus enough that the blurry bits would overlap the in-focus pixels. This also requires blurring an alpha channel that starts out with the edges of the different focus kernel values, so you can blend the near-to-camera pixels properly.
I see what you were trying to do there but you can't do it that way. It doesn't actually work. Getting bokeh from nearby objects to obscure far away pixels requires a totally different (and probably even more expensive) approach. All this shader can really do is provide a variable circular blur effect.

Here's a decent example of multilayered bokeh blur: http://cg.skku.edu/pub/papers/2008-lee-pg-psdof-cam.pdf
note that they have to render each model into multiple buffers in order to sort near/far/middle depths and handle overlapping objects.

dvr333:
Yes, I'm aware of the multi-layer approach, but that's not an option, so I went with the next best thing. For distant objects it's not a problem, as there is no missing information, and you need to make sure it isn't sampling in-focus points anyway; but for close objects I realized that, while you can't get the missing data to fill in where the blurry edges would get transparent, that only affects half of the pixels on the blurry edges (where the object would, if in focus, be opaque), and with just the other half of the pixels (where the object would, if in focus, not be drawn at all) it still makes the close blurry objects not have weird hard edges. 

Also, you're right that the two-pass approach I used only does a circular bokeh, which is a bit of a drawback of it.

LuaUI/Widgets/gfx_projectile_lights.lua
aeonios: Ok this is not correct. If projectile lights are not set to a higher layer than deferred rendering then no projectile lights are drawn. Projectile lights must run first, and deferred rendering should be pretty much the first thing to be applied in drawscreeneffects since it's basically a carryover from drawworld.

aeonios:
Your shader didn't actually fix the "weird hard edges" issue.
https://cdn.discordapp.com/attachments/561223041414332436/592985666724429839/unknown.png

Further downscaling isn't an option simply because quality is nil at that level. Fine for bloom, not fine for anything the user actually sees with any level of detail.

dvr333:
Hrm, that's why I'm turning on the two-pass approach. If you look closely the hard edges only show up on one edge, while everything else looks fine. This is an artifact of the fact that it has to be selective about blurring, but the two-pass approach means it can miss some edges due to their orientation.

As for downsampling, consider that the DoF effect is basically just a linear average across multiple pixels, and all downsampling does is do some of that work in dedicated hardware. Downsampling a 1920x1080 image to 960x540 and then that to 480x270, sampling a ring of 8 pixels in a 1.5 pixel radius, and re-compositing that is identical to taking a 6 pixel radius blur at 1920x1080, but with only about 1m samples (2m if you are being careful about edges, 4m if you are being careful about nearby edges) instead of 100m/(200m/400m) samples. It's basically what I did for Achron.

aeonios:
Well, feathering the blur amount stored in the alpha channel is sort of possible, but you'd want to ensure that it only expands by at most the blur radius or else you run the risk of blurring things that you shouldn't. Actually you may end up blurring things that you shouldn't anyway. The real problem though is that you end up with a nasty branch in the shader that gfx cards may not handle well. There's no other way to go "ok if that pixel has at least radius amount of blur assigned to it, then feather to this pixel, otherwise ignore". That's why I haven't bothered attempting a variable dilation. Not to mention it isn't separable, and if you use sparser samples then you end up with undersampling artifacts.

dvr333:
Yeah, I generally count the samples in every branch when counting samples for the effect, because of that aspect of gfx card branch management. How are you planning to handle different levels of blur, and especially handling the edges of those regions?

aeonios:
The way I handled it with gaussian blur was just to interpolate between non-blurred and unblurred downscaled for low amounts of blur, and between downscaled and downscaled + 1px gaussian blur for higher levels of blur. For bokeh blur I'm just setting the blur amount per-pixel in the preprocessing step so it'll use variable kernel widths in the blurring stage. One thing I don't like is that large parts of the image will end up using a zero-width blur since those areas either won't be blurred in the final image or will just use a downscaled sample to get a smooth transition between no blur and full blur. However I can't skip the samples because the seperable blur still requires them to be present in order to correctly reconstruct the final blur.

As for edges, I don't know if I'll handle them at all. If I had a way to do it without branching then maybe. If you can find any relevant papers or think of anything let me know. :|

dvr333:
I just wouldn't be so scared of branching. Yes, it increases the number of samples taken, but to my knowledge that's all you've got. The only other way I could see dealing with it would be that multi-layer approach you were talking about earlier.

aeonios:
The problem with branching is that it stops the entire thread block. If every thread in the block has to stop and wait for one thread to finish the entire shader every time there's a divergent branch, it's like increasing the number of samples exponentially.

Using the sampled point's blur radius is nonsensical since you'd only replace it's blur amount if you ran across a nearby point with a larger radius, meaning you have to take the offset's radius and search at the max radius. I'm going to check and see if I can find more information about branching and how bad it'd actually end up being. If the branches can join at the following loop iteration then it's no problem, I'll just use an 8-direction search at max radius. The main problem then is just that sampling the full-res image for that is perhaps prohibitively expensive due to the doubled radii.

dvr333:
Actually, using the sampled point's blur radius is to shrink the blur radius, for when pixels in the out-of-focus background try to sample less out-of-focus regions (including the in-focus region) to make sure that they only sample points that should be used in the blur.

For the stuff near to the camera it does start by doing an 8-sample check in a cross shape to get a likely max blur radius, then continues as normal. Since that is blurred on a different texture, the key part is the alpha channel, so if it blurs out stuff it shouldn't those parts won't be in the final image.

aeonios:
Oi that's a good point. Apparently a trivial branch like what this would require doesn't even count as a branch since it'll just use conditional execution to no-op the instructions if it branches the wrong way.

There should not be any condition where it would normally blur something in the foreground too much due to background pixels being blurred. Blurring occurs on a per-pixel basis, there's no bleed. Only taking a higher blur if our source pixel falls within the blur radius of another pixel with a higher blur AND that pixel is also nearer to the camera should be doable. However the 48 pixel blur radius in every direction is an issue. That's too expensive to sample and an 8 direction sample may not be sufficient to prevent artifacts. I'm not sure if it can be made separable, either.

dvr333:
48 pixel radius? I mean, downsampling would work, and separability has been managed (you'd need to recompute the kernel tables), but I'm not sure the motivation.

Anyway, the foreground isn't getting blurred when blurring naively, but it does get sampled into the blurry background pixels if you don't check for it, and that's the artifact being dealt with by scaling the blur radius.

aeonios:
Oh you mean the "halos" that show up around depth contrasts. That's sort of the inverse problem.

Seperability isn't possible here since you need to know the depth and distance to the original pixel in order to decide whether to modify the blur or not. Something like SATs might work, but I dunno if I could manage that. Guess it's back to the literature.

dvr333:
Good luck with the literature, but AFAIK this two-pass approach is only a few years old, with most DoF implementations using a one-pass approach and some kind of poisson disc sampling distribution. I get your criticisms about separability (which is exactly why that edge issue from earlier happens), but in practice I've found my approach works reasonably well most of the time.

aeonios:
Well, it turns out that an 8 radius blur is ridiculously excessive. I had to cut it down to 0.25 max filter scaling just to get it to look like anything, and that basically meant that what I needed was a 5-tap filter instead of a 17-tap filter. I don't know why you were using 11 taps, that's way too much. Like you could easily max blur a full-res screen image with that and still have samples to spare. I also found out that squaring the color in the first step before applying the filters is probably not optional. idk if you can remove that and just remove the sqrt at the final step, or whether both are required as part of the filter's mathematical space. Something to test anyway..

As usual the literature is just frustrating. I think what I'll try is applying a 1px gaussian blur over the alpha channel of the downscaled image so that the blur is smoothed out around edges. That'll take some experimentation but I think I can at least reduce the halos around focused objects with out-of-focus backgrounds. The question is whether I can get smoother edges for highly blurred foreground objects without also causing ugly inward blurring on focused objects.

dvr333:
Huh, I was using an 11-tap filter to cut down on the samples, but whatever minimum works is cool. The shadertoy example was definitely way too much either way.

As for the power/sqrt, that's done to make sure the blur circle has a defined edge. I've tested with it off, and with it reversed, and basically the higher the power, the more defined the blur edge is, but the tradeoff is that darker sections become squashed in the colour space, which can lead to dark regions looking like they lost a ton of texture resolution. A power of 1.8 seemed to work best in practice, but squaring/sqrt-ing is faster for the GPU.

Oh also, I was slightly mistaken about my earlier comments regarding taking the target point's blur radius for final sampling. The algorithm actually takes the minimum blur radius between the starting blur radius and the first target point's blur radius, and uses that for the final sample. The key there is that in the background blur/far blur pass the blur radii are treated as absolute values, while in the near pass they are signed, with negative values representing blurry pixels close to the camera.

aeonios:
Hm.. well I think boundary edges could be handled by feathering the blur amount by taking the blur amount of the adjacent pixel with the lowest depth, then doing a 1px gaussian blur over the blur amount channel. That should fix the sharp edges issue for near/blurry objects as well as the halo issue for focused objects with blurry backgrounds. The main issue is that openGL doesn't give you a lot of options for dealing with depth, so I need to figure out some way of taking a min filtered downscaled version of depth, or some other means of sampling the depth map, like maybe treating it as a shadow sampler.

dvr333:
The issue with that approach is one of directionality. The edges between blurry and in-focus pixels are not going to affect the blurry pixels when sampling equally blurry pixels, so I would expect you'd just end up with reversed halos (i.e. an area of erroneously in-focus pixels near the blurry/in-focus edge, where the halo would otherwise be)

aeonios:
The nearest-depth feathering prevents that, by taking the low blur and extending it out over the edges. Basically the blur amount from nearer pixels overrides the blur amount from higher-depth pixels. That means nearer-blurred pixels will feather out over focused ones, and focused pixels will feather out over background blurry pixels. Then I do a gaussian blur over the blur amount channel to smooth it out.

dvr333:
That still sounds like it would make the areas that should be blurry just less blurry than they should be, which would again create an inverse halo of in-focus regions.

aeonios:
That's entirely possible. The real question is how badly and whether it looks worse than the alternative. And also whether I can manage to implement it at all. :P

aeonios:
Well I got some results.

Focused foreground/Blurred Background:
screen00020

Blurry Foreground/Focused Background:
screen00021

Focused Foreground/Blurry Background:
screen00022

Middle-depth objects still get some haloing on unfocused backgrounds, which is probably a downscaling artifact. Not downscaling for HQ mode is a definite possibility, although I loath the thought of the texture size hacks I'll need to make that work seamlessly.

dvr333:
The middle-depth problem is something I struggled with for a while, though I agree with the downscaling hypothesis. Overall that looks quite good, well done. Do you happen to have a perf comparison between that and what I had?

aeonios:
Hm.. I don't know how to do ms timings in lua.

dvr333:
Fair, I just do A/B tests with "/fps" on and divide that into 1000.

aeonios:
My fps tends to be single digits so it's not much of a comparison. :P

aeonios:
Well I feel stupid. Turns out the feathering pass wasn't working, or rather downscaling the depth tex wasn't working. All it was doing was applying a 1px blur to the alpha channel which smoothed out the blur radii along depth contrasts.

After changing it to just direct sample the full res depth tex I got those nasty sharpening halos you were worried about, and even going half and half between the original blur value and the overriding blur value in the feathering step still produced them.

So I just got rid of the feathering step and applied the 1px blur over the blur map instead. :P

GoogleFrog:
Based on the comments I am assuming this is WIP. Also I'd like @lhog to comment on whether it makes sense.

aeonios:
Yeah, testing in a more realistic game situation showed it doesn't really look good. The transition to downscaling clashes with the look of the bokeh blur. I'm gonna have to try some things to see if I can get bokeh blur at full resolution at a reasonable cost. There are some other loose ends to tie up as well.

lhog:
DoF is a very domain specific task. As I haven't done it myself, I think @dvr333 is a better authority here.
The code itself has mistakes, I'll point them out in the review.

GoogleFrog:
What about HDR and bloom, should this be a few smaller PRs?

lhog:
I thought about the same. The PR is a bit too big and unfocused.

sprunk:
There are conflicts and the PR is large. I think it would benefit from being a handful of small, easily-reviewable PRs for each fix.

sprunk:
See the comment above.

PR #3747, opened 2019-10-28
make `untargetable` units also untargetable by weapons

Anarchid:
This should make ducks and scallops stop shooting at spent claws even after acquiring them as targets

Untested.


sprunk:
Checking URP all the time sounds expensive (string, iterating the urp map) while it's not really necessary for widgets to know about this state. Maybe this should be a GG table?

sprunk:
`AllowWeaponTarget` is only called on acquisition and not ran again afterwards.

PR #3748, opened 2019-10-29
Preserve build queue from an incomplete builder

esainane:
The unit must be an immobile builder capable of starting construction,
so this only applies to an incomplete Strider Hub.

When shift was pressed, cmd_guard_remove.lua triggers, removing the
old patrol command. Because the Strider Hub is still a nanoframe and
can't do anything, UnitIdle was immediately called for
unit_auto_patrol_nanos, which calls SetupUnit and overwrites the newly
queued build order with an unshifted patrol command.

Checking that the Idle unit was complete and therefore has a meaningful
idle state resolves this problem.

### DIFF
--- a/LuaUI/Widgets/unit_auto_patrol_nanos.lua
+++ b/LuaUI/Widgets/unit_auto_patrol_nanos.lua
@@ -36,6 +36,7 @@ local spGetMyTeamID     = Spring.GetMyTeamID
 local spGetTeamUnits    = Spring.GetTeamUnits
 local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
 local spGetUnitDefID    = Spring.GetUnitDefID
+local spGetUnitIsStunned = Spring.GetUnitIsStunned
 local spGetUnitPosition = Spring.GetUnitPosition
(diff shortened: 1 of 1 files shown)


sprunk:
Looks good.

Semi-nitpick: what happens if you queue a duplicate build order, cancelling it (not using Stop because that explicitly prevents restarting the self-patrol)? Is it consistent with how a live unit would behave?

Reading the code I would expect the live unit to restart self-patrol but not the nanoframe. There should probably be a `UnitFinished` callin that also checks for an empty queue and initiates self-patrol.

GoogleFrog:
A Caretaker nanoframe can be given an order that is completed before the Caretaker is finished. The desired behaviour is for the Caretaker to resume patrolling. What will happen with this PR is that it will idle forever. This could be circumvented with UnitFinished, but then you need to check for whether Stop has been used. Also, the lack of a patrol order until UnitFinished fails to indicate the 'Stop state' of the Caretaker to the player. I have a less hacky fix.

sprunk:
Could just check whether last command is nil probably.

Issue #3730, opened 2019-09-23
Widgets do expensive things when changing which team is spectated


GoogleFrog:
Select various enemy units to change which team is spectated. Note that widgets now act as if you just resigned. This behaviour is new.

sprunk:
Selecting any unit calls `PlayerChanged` due to indiscriminate `/specteam` commands:
https://github.com/ZeroK-RTS/Zero-K/blob/d644d6caadee1ab5b6e74c32ba0a761fea640821/LuaUI/Widgets/gui_spectate_selected_team.lua#L75-L81

PR #3750, opened 2019-10-29
Only specteam if selected team actually changed

esainane:
Since spring/spring@9d3cfda, widgets get a PlayerChanged event whenever
a spectator changes spectating teams. This causes a lot of recomputation,
so to avoid unnecessary load we only issue a specteam command if the new
team differs.

See #3730


sprunk:
What if I manually `/specteam x`? There should be some sort of getter.

sprunk:
Probably `Spring.GetMyTeamID`.

esainane:
@sprunk Updated with the proposed amendments

sprunk:
Thanks!

PR #3755, opened 2019-10-31
Fix #3511 (OOM during catching up)

### LINKED ISSUE #3511: crash - not enough memory (catching up replay)
linux 64, ubuntu 16.04, steam, 32G ram, running command

cd ~/.local/share/Steam/steamapps/common/Zero-K && ./Zero-K_linux64.sh

#####
infolog
https://pastebin.com/5hFq8SPk

sprunk:


GoogleFrog:
This may make catchup very slow. More people need to test. It could turn on or off depending on load.

GoogleFrog:
Stuff happens in the console while spectating.

image

GoogleFrog:
This widget makes catching up significantly slower. It should not be active for all players all the time.

sprunk:
I don't think I'll be able to test this well. I'll close the PR but keep the branch around for better times.

PR #3758, opened 2019-11-03
Fix commander upgrade window closing unexpectedly

esainane:
Sort a copy of the modules that we expect our commander to own before
comparing them to the sorted copy of modules that a selected commander
has.

ModuleSetsAreIdentical checks that the items in each index is the same.
The list of the modules the selected commander had was sorted, and
the list of modules that the commander we have the window open for
was not.

Therefore, unless the modules were already sorted by internal ID,
ModuleSetsAreIdentical would return false.

This caused the widget to incorrectly conclude that there was a new
selection that did not contain a commander with the same set of modules,
and close itself.


sprunk:
Just add `table.sort(alreadyOwned)` at line 942.

PR #3776, opened 2019-11-26
Only restrict non-allied projectile information

esainane:


LuaUI/cache.lua
```diff
@@ -36,12 +39,17 @@ local function FilterOutRestrictedProjectiles(projectiles)
...
 	while i <= n do
 		local p = projectiles[i]
-		local pID = spGetProjectileDefID(p)
-		if filteredWeaponDefID[pID] then
-			projectiles[i] = projectiles[n]
-			projectiles[n] = nil
-			n = n - 1
-			i = i - 1
+		local ownerID = spGetProjectileTeamID(p)
+		local ourID = spGetMyTeamID()
```
sprunk: Put this one outside the loop and make it have `teamID` in the name (ideally `myTeamID` because that's the callout but `ourTeamID` or such is also fine).

esainane: Amended!

GoogleFrog:
1. Is FilterOutRestrictedProjectiles called by any default widgets?
2. This seems like an expensive way to implement any UI that I would want to include in the game. If a nuke impact indicator is desired, then gadgets should implement some interface for it (probably a rulesparam). See the nuke launch warning.

sprunk:
1. Not used by any default widgets.

2. The "unit fired" event that Fire Once uses could just be passed to LuaUI? Alternatively ProjectileCreated could become exposed.

GoogleFrog:
2. Is synced iirc, I don't know what sort of impact that would have.

PR #3786, opened 2020-02-07
Fix #3243, Domi leash obscured by terrain

### LINKED ISSUE #3243: Consider enhancing the Dominatrix capture link looks
The current line looks a bit dull and it can penetrate the terrain geometry. It could be replaced with team-colored glowing spheres going into the captured unit that stay above the ground.
balls

sprunk:
Domi leash doesn't draw for me at all due to potato so I can't test.


Arch-Shaman:
I'll test this sometime later tonight.

sprunk:
Thanks!

PR #3796, opened 2020-02-17
Welder hitbox fixing: Making hitbox a box and increasing unit footprint.

danfireman:
Also adding more sensible wreck hitbox.

Should fix https://zero-k.info/Battles/Detail/836765

Unit packing comparison:

Old:
oldWelderPacking

New:
newWelderPacking

Hitbox comparisons:

Top:
weldersHitboxTop

Front:
weldersHitboxFront

Side:
weldersHitboxSide

### DIFF
--- a/units/tankcon.lua
+++ b/units/tankcon.lua
@@ -16,14 +16,14 @@ return { tankcon = {
   canMove                = true,
   canPatrol              = true,
   category               = [[LAND]],
-  collisionVolumeOffsets = [[0 0 0]],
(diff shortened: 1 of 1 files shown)


GoogleFrog:
This looks better except for the use of TANK3. At one point I reduced all constructor hitboxes to 2x2 so they were more able to path through bases. I suppose Welders are reasonably maneuverable and their weapon imposes a constraint unfelt by other constructors.

PR #3797, opened 2020-02-18
Replacing previous unit_auto_reclaim_heal_assist widget

danfireman:
The current unit_auto_reclaim_heal_assist widget is buggy in that it will occasionally tell constructors that already have orders to go reclaim/etc. something nearby.

This one uses precise number of orders and order position to avoid overruling any human orders.

### DIFF
--- a/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
+++ b/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
@@ -1,87 +1,137 @@
------------------------------------
--- Author: Johan Hanssen Seferidis
---
--- Comments: Sets all idle units that are not selected to fight. That has as effect to reclaim if there is low metal
---					 , repair nearby units and assist in building if they have the possibility.
---					 If you select the unit while it is being idle the widget is not going to take effect on the selected unit.
---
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
```diff
@@ -1,87 +1,137 @@
...
-		enabled = false
-	}
+   return {
+      name         = "Auto Reclaim/Heal/Assist",
+      desc         = "Makes idle unselected builders automatically repair nearby damages units, reclaim nearby wrecks and assist nearby construction.",
+      author       = "dyth68",
+      date         = "2020",
+      license      = "PD", -- should be compatible with Spring
+      layer        = 11,
+      enabled      = true
```
GoogleFrog: I disagree.

GoogleFrog:
This widget is not enabled by default so I don't see much harm in changing it.

LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
danfireman: Oops. Missed that.

Fixed.

PR #3808, opened 2020-02-24
Adjust mex RoI to what GoogleFrog believes it is

sprunk:
100m → 50% of mex cost


GoogleFrog:
Weird. I read MEX_REFUND_SHARE = 0.5 and called it a day.

Issue #2554, opened 2017-10-09
Lua terrain smoothing (quake, tremor, etc) can produce vertical spikes


Anarchid:
1) Construct a building on a spire
2) Have some smoothing weapon hit that building and the spire
3) Smoothing works as it should, but then building terraform prevention kicks in and moves the building back without respecting pyramidism.

![image](https://user-images.githubusercontent.com/3822768/31356696-ad7f816e-ad47-11e7-94d8-b89083ebcfe1.png)

GoogleFrog:
Already known. It is good to have a ticket about it though.

sprunk:
57f4315dc457e4572017f5d6d8639c23e9da1437

GoogleFrog:
That change did not solve this.

PR #3810, opened 2020-02-25
seismic events displace and level structures

Anarchid:
Re: http://zero-k.info/Forum/Post/214373#214373

### DIFF
--- a/LuaRules/Gadgets/unit_terraform.lua
+++ b/LuaRules/Gadgets/unit_terraform.lua
@@ -163,6 +163,7 @@ local terraUnitLimit = 250 -- limit on terraunits per player
 local terraUnitLeash = 100 -- how many elmos a terraunit is allowed to roam
 
 local costMult = 1
+local buildingResistanceModifier = 0.8; -- make a modoption? 0 is fully immune
 local modOptions = Spring.GetModOptions()
 
 if modOptions.terracostmult then
@@ -185,7 +186,8 @@ local structure          	= {}
 local structureTable		= {}
 local structureCount	 	= 0
 
-local structureAreaMap      = {{}{}{}{}} -- each vertex can be shared by at most 4 structures
+local structureAreaMap      = {{}} -- a list of maps, with extra map layers added as necessary
+local numStructureAreaMaps = 1
 
 local structureLUT          = {}
 
@@ -212,6 +214,7 @@ local MAX_UPDATE_PERIOD     = 30
 local updatePeriod          = 15 -- how many frames to update
 local terraformOperations   = 0 -- tracks how many operations. Used to prevent slowdown.
 local nextUpdateCheck       = 0 -- Time at which to check performance
+local immediateDisplacements = {} -- list of units that have been deliberately displaced by terraform
 
 -- Map terraform commands given by teamID and tag.
 local fallbackCommands  = {}
@@ -383,7 +386,7 @@ local function SetTooltip(unitID, spent, estimatedCost)
 end
 
 local function IsPositionTerraformable(x, z)
-	for k=1,4 do
+	for k=1,numStructureAreaMaps do
 		if structureAreaMap[k][x] and structureAreaMap[k][x][z] then
 			return false
 		end
@@ -3000,6 +3003,18 @@ local function updateTerraform(health,id,arrayIndex,costDiff)
 	return 1
 end
 
+local function ApplyUnitDisplacement(unitID)
+	local unit = structure[unitID]
+	if unit then
+		local height = spGetGroundHeight(unit.x, unit.z)
+		if height ~= unit.h + unit.dh then
+			spLevelHeightMap(unit.minx,unit.minz,unit.maxx,unit.maxz,unit.h+unit.dh)
+			unit.h = unit.h + unit.dh
+			unit.dh = 0
+		end
+	end
+end
+
 local function DoTerraformUpdate(n, forceCompletion)
 	local i = 1
 	while i <= terraformUnitCount do
@@ -3084,6 +3099,12 @@ function gadget:GameFrame(n)
 	--if n % 300 == 0 then
 	--	GG.Terraform_RaiseWater(-20)
 	--end
+
+	for k, v in pairs(immediateDisplacements) do
+		ApplyUnitDisplacement(k);
+	end
+
+	immediateDisplacements = {}
 	
 	if n >= nextUpdateCheck then
 		updatePeriod = math.max(MIN_UPDATE_PERIOD, math.min(MAX_UPDATE_PERIOD, terraformOperations/60))
@@ -3160,16 +3181,7 @@ function gadget:GameFrame(n)
 	if struc then
 		local i = 1
 		while i <= struc.count do
-			local unit = structure[struc.unit[i]]
-			if unit then
-				local height = spGetGroundHeight(unit.x, unit.z)
-				if height ~= unit.h + unit.dh then
-					spLevelHeightMap(unit.minx,unit.minz,unit.maxx,unit.maxz,unit.h+unit.dh)
-					unit.dh = 0
-				end
-			else
-
-			end
+			ApplyUnitDisplacement(struc.unit[i]);
 			i = i + 1
 		end
 	end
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Why is this a modoption? This also looks like it creates sheer cliffs, so I'll have to do more testing.

Anarchid:
> Why is this a modoption? 

Would be a good way to test it before actually making default.

> This also looks like it creates sheer cliffs

[Absence](https://github.com/ZeroK-RTS/Zero-K/issues/2554) of this causes them. The function that blocks smoothing just does a boolean block on the vertices where structures stand on, without applying pyramid effect.

LuaRules/Gadgets/unit_terraform.lua
```diff
@@ -3148,20 +3156,22 @@ function gadget:GameFrame(n)
...
-	local struc = structureCheckFrame[n % structureCheckLoopFrames]
-	if struc then
-		local i = 1
-		while i <= struc.count do
-			local unit = structure[struc.unit[i]]
-			if unit then
-				local height = spGetGroundHeight(unit.x, unit.z)
-				if height ~= unit.h then
-					spLevelHeightMap(unit.minx,unit.minz,unit.maxx,unit.maxz,unit.h)
+	if blockSeismic then
```
GoogleFrog: Doesn't this change make all terrain deformation cause structures to move?

GoogleFrog:
This does not need to be a modoption. It can all be tested locally.

I know that the current behaviour causes sharp cliffs. This PR will cause them too.

You have made normal sources of deformation move structures too. Have you tested this thoroughly? Giving back control of structure position to the engine could have many side effects. Can you say exactly how the height of a structure is determined?

Anarchid:
The mechanical consequences can indeed be tested locally; the effect on gameplay less so.

It looks very much like Spring does the same thing as the gadget for determining where the building is: altitude under center. It is possible to create "overhangs" with larger buildings by eroding terrain under their footprint without moving them.

It also looks fairly silly when a building falls down a spire without taking any damage when a mobile would be badly harmed in such a situation. But that can be done separately.

Should i remove modoption and make this the unconfigurable default?

GoogleFrog:
You did not convince me that this PR will not cause sharp cliffs. I had to test and prove that to myself. Instead the PR does something worse.

image

This is what I get if I make IsPositionTerraformable return true and remove 'check structures for terrain deformation'. The PR looks like it allows this to happen (albeit not with Dirtbags). Based on this I think the PR is very incomplete.

The ground under structures should be flat (unless unchanged since the structure was built) and cliffs should be avoided where feasible. It seems like avoiding sharp cliffs should be feasible in all interactions between a seismic missile with a single structure.

Anarchid:
> This is what I get if I make IsPositionTerraformable return true and remove 'check structures for terrain deformation'. The PR looks like it allows this to happen (albeit not with Dirtbags). Based on this I think the PR is very incomplete.
This is what i meant by "overhangs", yes.

I'm not sure what you mean by "sharp cliffs" here though. Later context suggests "sharp cliffs" is actually "non-flat footprint"?

> The ground under structures should be flat (unless unchanged since the structure was built) and cliffs should be avoided where feasible. It seems like avoiding sharp cliffs should be feasible in all interactions between a seismic missile with a single structure.
I see two solution approaches to this, with the same cost of sampling every vertex under footprint both on each smoothing and on each snapping:

- Damage, then level to previous. This resolves the unintuitive zero damage to falling structures by only moving them after killing them.

- Level to average. 

Suggestions?

GoogleFrog:
A sharp cliff is exactly what you would expect - a 8 elmo difference in horizontal distance with a large difference in height.

GoogleFrog:
I don't see the need for damage or for non-smoothing to move structures. Do smoothing movement along the lines of leveling to average in such a way that cliffs are rarely created. Make it so that whether a smoothing effect moves structures can be configured.

Anarchid:
If only smoothing does the moving, then this can be done nearly for free:
- When smoothing vertices under structures, instead of applying the change, divide it by footprint and add result to the height field of the structure
- When checking against structures being moved by other means, the changed height will be considered and the structure will be snapped to a flat rect with the average smoothed altitude
- Sharp cliffs will be possible, but rare.
- If wanted, this displacement could be stored in a separate field and used for damage.

I'll probably try this sometime today.

Anarchid:
Turns out that won't work: `structureAreaMap` only stores how many structures affect a given vertex, not *which*.

Anarchid:
> Look up structures from positions in some other way

Added a shitty lookup table with grid cell sizes just a bit larger than the largest footprint. Probably this part should be automatic in some way.

There is some table creation, but no table deletion. There is some pairs(), but only within a single lookup-table grid-cell.

Entirely untested blindcode, therefore WIP.

GoogleFrog:
Why don't you just store unitID in structureAreaMap? I don't see how a vertex could be adjacent to more than four structures, so simply make four copies of structureAreaMap and fill them sequentially. This is an alternative to spamming tables by making each entry of structureAreaMap a table.

Anarchid:
Ooh. I didn't realise the multiple structs per vertex was for shared edges and assumed it was some sort of hyper-defensive code in case of multiple blocking structures existing in the same place, like stacked mexes/shields/winds.

If it's just edges, then yes, four copies makes sense.

LuaRules/Gadgets/unit_terraform.lua
```diff
@@ -341,6 +346,120 @@ elseif modOptions.terrarestoreonly == "1" then
...
+					numFound = numFound+1
+					found[numFound] = structID
+				end
+			end
+		end
+	end
+	return found;
+end
+
+-- this assumes only one structure is allowed to exist at the same point to avoid table creation
```
GoogleFrog: Does this assumption mean that the code is broken?

Anarchid: There's a version that works with multiple structs for that case. But the code is currently broken in more ways than that, and it makes sense to rewrite it with multiple copies like suggested in last comment.

Anarchid:
Added a version that has 4 StructureAreaMaps. Probably still broken. Will test sometime around weekend probably.

LuaRules/Gadgets/unit_terraform.lua
```diff
@@ -185,7 +185,8 @@ local structure          	= {}
 local structureTable		= {}
 local structureCount	 	= 0
 
-local structureAreaMap      = {}
+local structureAreaMap      = {{}{}{}{}} -- each vertex can be shared by at most 4 structures
```
sprunk: does this parse? shouldn't it be `{{},{},{},{}}`

Anarchid: Nope, is untested blindcode. Also good point about pathological cases.

I wonder if ad-hoc creating more maps is going to be horrible (probably not as common case will be non-degenerate and less than 4). Later-added maps are likely to also be mostly empty, and iterations that dont care for which structure blocks them are a majority that will break early.

Anarchid:
Tested and fixed. 

Some sharp edge creation is nearly inevitable due to flat footprint constraint, so it happens. I think the only way to completely eliiminate it is to bleed some smoothing around affected structures. I think that's likely too horrible (consider long tightly packed wind--chains, and oh god, wind-circles)

Made the number of map copies dynamic. `/give 100 staticstorage` boosts it to 4 as expected; abuses like repeated `/give staticstorage` on the same coordinate increase it by 1.

(Maybe it makes sense to check the biggest map and see if it's actually empty once in a while, then decrement?)

Added a table for instant-check for structure displacements in gameframe so that effects of smoothing are not delayed (it looks rediculus).

Added a 0.8 structure displacement modifier because dividing area by 16 made it accumulate too much displacement and spire buildings would often end up underground. This could serve as a modoption value to restore the old mode (with 0).

GoogleFrog:
> Added a 0.8 structure displacement modifier because dividing area by 16 made it accumulate too much displacement and spire buildings would often end up underground. This could serve as a modoption value to restore the old mode (with 0).

The fact that this is a behaviour created by your system is concerning. I don't know why a system would send structures deeper than the terrain smooth height. I have half a mind to pull it as is, use the data structures and replace the calculations to fix the issues.

PR #3814, opened 2020-02-29
Reducing Funnelweb shield HP 23000->20000 so it only needs 2 Shockley

danfireman:
Given current Funnelweb usage I think it needs another counter and can afford to lose 3000 hp on the shield.

This means that mass Funnelweb (e.g. 3 Funnels) in FFA can be countered cheaply with a single Silo of well aimed Shockleys.


Anarchid:
Yay for round numbers!

GoogleFrog:
23000 is pretty round.

GoogleFrog:
Do you have a thread or replay for this change, just so I know where to point people?

danfireman:
I thought I'd seen some FFAs degenerate into Funnelweb spam and team games where Funnels protected Tremors from enemy fire, but I'm struggling to find them.

Hold off on merging til I confirm the issue.

danfireman:
@GoogleFrog : Reckon this is good enough? https://zero-k.info/Forum/Thread/27099?postID=215392#215392

PR #3831, opened 2020-03-18
Fix area dmg

sprunk:


GoogleFrog:
https://github.com/ZeroK-RTS/Zero-K/commit/1da28260450046ecb544077b0af7cdce3c818eec

Issue #3833, opened 2020-03-19
Make Hunter torpedo bigger


GoogleFrog:
I sometimes find sea battles annoying because it is hard to see what is shooting at what. Hunter torpedoes are essentially invisible so it looks like units randomly die. It is hard to judge the effectiveness of your units while this is the case.

Maybe Urchin torpedo could be a bit larger too.

Anarchid:
Seawolf missile is quite visible - i think a CEG replacement would do more here than model scaling.

sprunk:
94b80a66ad60d0a759a84ae399566075f3d0c3db et al, can be closed?

PR #3840, opened 2020-03-25
make torps more visible

Anarchid:
Makes Urchin and Hunter torpedoes much more visible

image

As a side effect, makes it obvious that current Hunter torpedo falls through the boat. Perhaps the balance changes from https://github.com/ZeroK-RTS/Zero-K/pull/3471#issuecomment-474237821 are inevitable.

re: #3833


Anarchid:
Much less drastic physics changes were sufficient for shoe hunter.

GoogleFrog:
I'm not a fan of the white trail on ugliness, but it is quite visible. Have you done any testing at all regarding whether the startVelocity change allows Hunter to hit units consistently?

Anarchid:
I think torps can be made less ugly both on trail and explosion parts. Will do some tinkering sometime.

PR #3842, opened 2020-03-28
Nicer torp effects

Anarchid:
- Hunter/Urchin underwater explosion is a collapsing bubble with some small rising bubbles around it
- Hunter/Urchin overwater explosion has a foam ring
- Hunter/Urchin torpedo trail has bubbles underwater
- Seawolf explosion is less obviously incorrect

Will make gif later.


Anarchid:
Seawolf: https://vimeo.com/401729763
Hunter against UW target: https://vimeo.com/401729887
Hunter against surface target: https://vimeo.com/401729845

Collapsing bubble seems to spawn even for surface targets - apparently it's already deep enough to be underwater? I guess this is as much as can be done without gadgetising the thing.

PR #3846, opened 2020-04-01
Ui preset

TRONAS:
Tacticool ui preset proposal
screen00006
screen00010
screen00013
screen00014
screen00001
screen00002
screen00004
screen00005


Anarchid:
This include your qol widgets config. Even if it necessitates it, that should be in a different PR.

PR #3847, opened 2020-04-01
Tacticool ui preset proposal

TRONAS:
screen00002
screen00004
screen00005
screen00006
screen00010
screen00013
screen00014
screen00001


Anarchid:
Doesn't seem to contain any changes? The two commits just add and remove the widget. 

If this is intended to be a non-code proposal, it should be an issue. Otherwise you should add the changes :P

TRONAS:
the widget adding and removing was trial an error, i was indeed making a non code proposal

PR #3852, opened 2020-04-03
Build bar widget

TRONAS:
screen00019

### DIFF
--- a/LuaUI/Widgets/gui_buildbar.lua
+++ b/LuaUI/Widgets/gui_buildbar.lua
@@ -1,0 +1,1156 @@
+-- $Id: gui_buildbar.lua 2491 2008-07-17 13:36:51Z det $
+
+function widget:GetInfo()
+  return {
+    name      = "BuildBar",
+    desc      = "An extended BuildMenu to access the BuildOptions of factories\neverywhere on the map without selecting them before",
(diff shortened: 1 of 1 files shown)


TRONAS:
works better than the others imho

GoogleFrog:
Wow this looks like super old pre chili.

Licho1:
13 year old jK's original widget.. archeology!

sprunk:
>works better than the others imho

Explain "better". Any improvements would ideally be extracted and ported to
https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Widgets/gui_chili_facbar.lua (and facpanel, if applicable) since this one is not chili (harder to maintain, or port to develop in the future) and it would be the third version of what is basically the same widget.

TRONAS:
its transparent, the icons are bigger and opens build options when hovered

sprunk:
Good, so I think the thing to do would be to add opacity/size sliders to the existing Chili versions and a checkbox to open when hovered.

GoogleFrog:
I would not be surprised if this performs a lot better than any previous chili version of facbar. If this works then it seems worthwhile adding it.

Licho1:
Remember jK made both this and chili..

PR #3854, opened 2020-04-04
Increase Blastwing sight distance

a-nop:
Now matches Bandit.

http://zero-k.info/Forum/Post/218303#218303


sprunk:
450 might have been intentional to make it die when discovering a Lotus (unsure so ideally GF would say something).

PR #3856, opened 2020-04-04
Increase Raptor air superiority

a-nop:
Planes can be superfluid, too.

Combat speed: 50% -> 35%
Combat turn speed: 200% -> 286%
Slowdown restore delay: 250ms -> 150ms
Acceleration increased 10%

With these changes Raptors spend more time firing during each pass and are more responsive to micro, e.g. firing then retreating to limit exposure to ground fire.

Trident can still get two salvos out before the Raptor's second pass when not microed.

Raptor can still chase and shoot at Nimbus and Krow.

### DIFF
--- a/scripts/planeheavyfighter.lua
+++ b/scripts/planeheavyfighter.lua
@@ -13,7 +13,7 @@ local smokePiece = {base, engineL, engineR}
 --variables
 local gun = false
 
-local RESTORE_DELAY = 250
+local RESTORE_DELAY = 150
 local FIRE_SLOWDOWN = tonumber(UnitDef.customParams.combat_slowdown)
 
 --signals
--- a/units/planeheavyfighter.lua
+++ b/units/planeheavyfighter.lua
@@ -28,7 +28,7 @@ return { planeheavyfighter = {
     modelradius    = [[10]],
     refuelturnradius = [[120]],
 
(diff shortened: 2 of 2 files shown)


sprunk:
> [Firepluk] problem of current viper is that is dies even to nimbus swarm lol, I mean -seriously-
> [Firepluk] it dies even to BD
> [Firepluk] and to not die some extraordinary micro skills required, by default u will die

GoogleFrog:
This looks like the Quant's Rule take on Raptor balance, as it is moving towards being a floating AA turret in combat. I question whether the changes do what they say they are meant to do. Slowing down makes Raptor even weaker to flex AA. Perhaps it needs more health alongside the change.

I think there are some fundamental issues here that are unaddressed:
 - Swift burst missiles will beat constant DPS Raptor unless Raptor has a lot more range and/or DPS than it does now.
 - Swift with the burst and D is more fun to use. Could any reasonable D ability be added to Raptor?

Anarchid:
> Swift with the burst and D is more fun to use. Could any reasonable D ability be added to Raptor?

Swift's D-ability is to move itself. A logical opposite of that would be an ability to prevent others from moving. 

For example, the flying turret mode could be a D-ability, and come with extra slow damage..

PR #3869, opened 2020-04-09
Dante napalm missiles burn time 37 -> 15s

sprunk:
To make consistent with other sources of burning and to make it not doom Snipers it randomly hits all the time (such counter).

### DIFF
--- a/units/striderdante.lua
+++ b/units/striderdante.lua
@@ -209,6 +209,7 @@ return { striderdante = {
       customParams              = {
         setunitsonfire = "1",
         burnchance = "1",
+        burntime = 450,
         reaim_time = 1,
       },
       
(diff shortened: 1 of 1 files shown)


Lokisar-GH:
pls no

dahn-zk:
+ for consistency
but its not particularly powerful to be simply nerfed

Lasica:
It is very powerful and I'd argue it's one of Dante's best perks among striders - ability to safely annihilate groups of weak units from a distance.

I agree it needs some tweaking though. Why not make it 15 sec but more intensive?

sprunk:
Fire (the unit burning status effect) is always 15 DPS (for consistency).

Some other tweaks could be instated if Dante is already considered weak, for example the heatray could get range 430 → 460 (to match missiles) and DPS 950 → 1000 (to annoy GoogleFrog by using a round number).

PR #3879, opened 2020-04-12
Assault Skuttle

Oflameo:
I redesigned Skuttle to interact with both heavy units and light units better by replacing its stealth with health. This Skuttle can actually close in on the assaults and heavy units it is suppose to target, yet a raider can still disarm it without taking any damage because their range is greater than the explosion radius. Skuttle still has a chance to escape from raiders with its jump.

Assault Skuttle can combo with Placeholder by exploding at the bottom of a gravity well. Assault Skuttle can also combo with with a crowd of cheap assault units like Hermits and Knights by taking out an important target quickly.


sprunk:
Skuttle is slow, won't this just allow heavies to retreat upon seeing the Skuttle? Note that this makes Skuttle have Jack traits (tanky, slow, brutal damage but at melee range) so presumably this Skuttle will keep failing at antiheavy for the same reasons Jack does.

Oflameo:
Yes, heavies would retreat upon seeing Skuttle. The play is finding ways to use other units to either hide the Skuttle or cut of the retreat of the targeted heavies. The big difference between Assault Skuttle and Jack is that Skuttle does all of its damage at once, and Jack does its damage over time; and Skuttle is one and done, and Jack can be used over and over again.

Sgathec:
Does the factory offer good units for hiding the Skuttle? Isn't the Puppy too small?

Moderators or Placeholders could slow down units, but they die easily because they have to get close.

Oflameo:
Jumpbots have to rely on trapping units with Moderators, Placeholders, and Jugglenauts. Assault Skuttle is capable of blowing up a whole stack of units at the bottom of a gravity well created by Placeholders because the explosion's area of effect is a 90 elmo radius and the gravity well has a 70 elmo radius.

The obvious factory to use to help hide Assault Skuttle is cloakbots. Cloakbots have Conjurer's area cloak,  Iris's area cloak, but also Skuttle can be camouflaged in a group of other assaults using them as a screen because Skuttle only needs an instant to do all of its damage. Spiders are also a good factory to use with Assault Skuttle. Spiders can trap units with Widows and Venoms and give Skuttle enough time to close in. Skuttle can hide in a group of Hermits and use them as a screen until they close in on something big for the Skuttle to take out instantly.

Sgathec:
I think the modified Skuttle would call for a particular kind of units. For instance, the Dirtbag looks big enough and its low cost seems to suit shielding too. A need to make another factory would be too much in 1v1 games in my opinion.

Lokisar-GH:
hmm, I think that is a neat idea, and would fix it up a bit, what about a small speed boost?

Oflameo:
I don't want to give Assault Skuttle a speed boost until after we see it in the meta. Right now, I think it is fast enough for what it needs to do. Making another factory wouldn't be a big cost relative to the heavy units it targets.

Sgathec:
One Skuttle costs 550. 15 Dirtbags is 15 * 30 = 450, but an Iris (600) would be better. Building a new factory would mean having to add 800 metal to the overall cost. For example, 800 + 450 + 550 = 1800. That's more than three times the current total and more than a Jugglenaut.

Arch-Shaman:
`  initCloaked            = true,`

This should probably be removed.

`  description            = [[Jumping Anti-Heavy Bomb]],`

You want to edit LuaUI/Configs/lang/units.en.json (as well as the other languages) as well to reflect this.

HTMLPhoton:
[Stuff] how would skuttle close the final distances on the opponent though? dosn't jump reveal cloaked units? or is it suppose to tank it's way there?

Arch-Shaman:
> dosn't jump reveal cloaked units?

Jump does not uncloak the user (although I think it might for jump jet units and not jump legs units). Jumping near a unit does uncloak both units.

GoogleFrog:
I don't see how this unit is meant to work without at least a cloaker or a Placeholder. Every heavy unit that is faster than it should dodge the jump or walk away, taking far less than maximum damage.

Visually this unit does not look like it has 2500 health as it is small and spindly.

Oflameo:
@Arch-Shaman I can make a new commit fixing those  2 issues. I will need to start a working group to verify the changes for the other languages.

@GoogleFrog It would work similar to SSB Jigglypuff's rest. Trapping and deception are both valid methods of getting Skuttle's hitbox to overlap with the target's hurtbox.

sprunk:
What is SSB Jigglypuff's rest?

GoogleFrog:
> Trapping and deception are both valid methods of getting Skuttle's hitbox to overlap with the target's hurtbox.

I don't think so

Trapping requires the ability to control space on the map. Skuttle has none of this. Its trajectory is set once it jumps, so good play involves dodging it. The quick reactions this requires constitute fighting the UI. Its circle of danger is tiny for a trap unit and it is slower than everything. A unit that is having its movement constrained by enemy units on the 300-400 elmo scale still has the time to locally walk away from or around a Skuttle that acts of the 50 elmo scale. I'm sure you could make a Skuttle connect if you somehow completely flanked an enemy unit, but how often is that going to happen when Skuttle has no stealth?

Deception in the form of a unit model merely being difficult to spot is against Zero-K design. If a Skuttle is the most important thing in a group of enemy units then the UI should highlight the Skuttle.

Oflameo:
Placholder is in the same factory. Placeholder traps, Skuttle follows up. Skuttle with stealth doesn't close in on an enemy unit because it dies before it reaches the target.

The Deception is in the form of unit composition in a group, not the model of the unit.

Oflameo:
The commits just get stacked on the pull request. That is convent. 

Todo: units.de.json  units.pl.json  units.tr.json units.ru.json  units.zh.json

Arch-Shaman:
> What is SSB Jigglypuff's rest?

Super Smash Bros move that instakills for... reasons... by sleeping. Someone was high that day.

sprunk:
I don't see this getting through in the current form.

GoogleFrog:
Here is a link to an issue rather than a comment on a PR: https://github.com/ZeroK-RTS/Zero-K/issues/4380

PR #3881, opened 2020-04-16
Reclaim highlight

esainane:
This is largely refactored out from the Widget available at:
https://github.com/lhog/ZK-Widgets/blob/master/EcoHelper.lua

All known features are clusterised and highlighted with flashing,
purple, convex hulls laid over the viewport.

This does a pretty good job of having hulls merge and split when
you would expect it to.


LuaUI/Widgets/libs/Optics.lua
```diff
@@ -0,0 +1,251 @@
...
+-- Adapted from:
+-- https://gist.githubusercontent.com/frnsys/6014eed30c69c6722177/raw/ec32813a900348cbb3ab60205921d70c60286ce3/optics.py
+--
+local PQ = VFS.Include("LuaUI/Widgets/libs/PriorityQueue.lua")
+
+local function DistSq(p1, p2)
+	return (p1.x - p2.x)^2 + (p1.z - p2.z)^2
+end
+
+local function TableEcho(data, name, indent, tableChecked)
```
GoogleFrog, line 11: Could this be replaced with Spring.Utilities.TableEcho, or somehow merged into it and extended?

GoogleFrog:
The generic utilities that could also be used by LuaRules should be added to LuaRules/Gadgets/Include, or everything could be moved to a better directory.

GoogleFrog:
This should have an epicmenu toggle in the path Settings/Interface/Something Suitable. It could also do with some epicmenu options to control things like visibility, front size, and perhaps some thresholds. Whatever is relatively easy to add or seems necessary for accessibility.

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,581 @@
...
+	featuresUpdated = false
+	clusterMetalUpdated = false
+	benchmark:Enter("UpdateFeatures 1loop")
+	for _, fID in ipairs(Spring.GetAllFeatures()) do
+		local metal, _, energy = Spring.GetFeatureResources(fID)
+		metal = metal + energy * E2M
+
+		if (not knownFeatures[fID]) and (metal >= minFeatureMetal) then --first time seen
+			knownFeatures[fID] = {}
+			knownFeatures[fID].lastScanned = gf
```
sprunk, line 86: Could optimize the array indexing out:
```lua
local f = {}
f.lastScanned = gf
f.x = fx
-- etc
knownFeatures[fID] = f
```

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,581 @@
...
+
+	local frame=Spring.GetGameFrame()
+	color = 0.5 + 0.5 * (frame % checkFrequency - checkFrequency)/(checkFrequency - 1)
+	if color < 0 then color = 0 end
+	if color > 1 then color = 1 end
+end
+
+function widget:GameFrame(frame)
+	local frameMod = frame % checkFrequency
+	if frameMod == 0 then
```
sprunk, line 490: Could do `if frameMod ~= 0 then return end` and save an indentation level in the rest of the function.

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,581 @@
...
+		gl.PopMatrix()
+	end
+end
+
+local checkFrequency = 30
+
+local cumDt = 0
+function widget:Update(dt)
+	cumDt = cumDt + dt
+	local cx, cy, cz = Spring.GetCameraPosition()
```
sprunk, line 472: The various `Spring.foo`, `gl.foo` etc could be localized so as not to keep indexing the `Spring`, `gl` etc tables repeatedly.

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,581 @@
...
+	end
+end
+
+local drawFeatureConvexHullEdgeList
+local function DrawFeatureConvexHullEdge()
+	gl.PolygonMode(GL.FRONT_AND_BACK, GL.LINE)
+	for i = 1, #featureConvexHulls do
+		gl.PushMatrix()
+
+		gl.BeginEnd(GL.LINE_LOOP, function()
```
sprunk, line 410: AFAIK 3rd+ arguments to `BeginEnd` get forwarded to the function in the 2nd argument so it would be good if the function was not recreated every time.

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,581 @@
...
+
+local function ColorMul(scalar, actionColor)
+	return {scalar * actionColor[1], scalar * actionColor[2], scalar * actionColor[3], actionColor[4]}
+end
+
+function widget:Initialize()
+	-- This information is pretty useful as a spectator too
+	-- CheckSpecState(widgetName)
+	local curModID = string.upper(Game.modShortName or "")
+	if ( curModID ~= "ZK" ) then
```
sprunk, line 371: This check looks fine for a local version but makes little sense gameside.

LuaUI/Widgets/libs/ConvexHull.lua
```diff
@@ -0,0 +1,112 @@
...
+		end
+	end
+
+	local p = lbMostPointIndex
+	local hull = {} -- The convex hull to be returned
+
+	-- Process CCW from the left-most point to the start point
+	repeat
+		-- Find the next point q such that (p, i, q) is CCW for all i
+		q = points[p + 1] and p + 1 or 1
```
sprunk, line 46: `q` leaks to global namespace

LuaUI/Widgets/libs/ConvexHull.lua
```diff
@@ -0,0 +1,112 @@
...
+--- JARVIS MARCH
+
+--- MONOTONE CHAIN
+-- https://gist.githubusercontent.com/sixFingers/ee5c1dce72206edc5a42b3246a52ce2e/raw/b2d51e5236668e5408d24b982eec9c339dc94065/Lua%2520Convex%2520Hull
+
+-- Andrew's monotone chain convex hull algorithm
+-- https://en.wikibooks.org/wiki/Algorithm_Implementation/Geometry/Convex_hull/Monotone_chain
+-- Direct port from Javascript version
+
+function MonotoneChain(points, benchmark)
```
sprunk, line 67: Leaks to global namespace

GoogleFrog:
So is anything going to happen?

sprunk:
I can apply the cleanups myself if nothing happens and I find some mana so IMO let this hang for a while.

esainane:
Still TODO:
- Epic Menu entries
- BeginEnd optimisation

LuaUI/Configs/epicmenu_conf.lua
```diff
@@ -387,6 +387,9 @@ local pathSelectionBluryHalo = 'Settings/Interface/Selection/Blurry Halo Selecti
 	ShButton(pathSelectionPlatters, 'Toggle Team Platters', function() spSendCommands{"luaui togglewidget TeamPlatter"} end, "Puts team-coloured disk below units")
 	ShButton(pathSelectionBluryHalo, 'Toggle Blurry Halo Selections', function() spSendCommands{"luaui togglewidget Selection BlurryHalo"} end, "Places blurry halo around selected units")
 
+local pathReclaimHighlight = "Settings/Interface/Reclaim Highlight"
+	ShButton(pathReclaimHighlight, 'Toggle Reclaim Field Highlight', function() spSendCommands{"luaui togglewidget Reclaim Highlight"} end, "Draws shapes around fields of reclaim, and shows their equivalent metal value")
```
sprunk, line 392: Do a checkbox in the main widget's options instead (and unregister all callins when disabled but keep the widget toggled). See #494

GoogleFrog: Isn't this code in epicmenu_conf a standard way to make widgets disabled but accessible through epicmenu?

sprunk: It is, but it uses a button instead of a checkbox. I don't mind that much whether the checkbox is defined here or in the widget itself as long as it is a checkbox. Doing it in the widget reduces the chance of breakage when somebody forgets about the part in epicmenu_conf when touching the widget.

GoogleFrog: Being a button rather than a checkbox is standard too. It works well enough when the button is in the same path as the widget options, as then you can see the widget options appear and disappear.

sprunk: I aim to change the standard. A checkbox would also make the options appear and disappear if done properly, but is strictly superior as a standard because you can also tell whether the thing is enabled for widgets that don't have separate option sets.

FWIW this PR can keep it as a button, it's not a terribly important thing and I can change it to a checkbox later alongside the other buttons.

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,653 @@
...
+local spGetFeatureTeam = Spring.GetFeatureTeam
+local spGetGaiaTeamID = Spring.GetGaiaTeamID
+local spGetGameFrame = Spring.GetGameFrame
+local spGetGroundHeight = Spring.GetGroundHeight
+local spGetMyAllyTeamID = Spring.GetMyAllyTeamID
+local spIsGUIHidden = Spring.IsGUIHidden
+local spIsPosInLos = Spring.IsPosInLos
+local spTraceScreenRay = Spring.TraceScreenRay
+local spValidFeatureID = Spring.ValidFeatureID
+
```
sprunk, line 101: plus usage:
```suggestion
local GL_FRONT_AND_BACK = GL_FRONT_AND_BACK
local GL_FILL = GL.FILL
local GL_TRIANGLE_FAN = GL.TRIANGLE_FAN
local GL_LINE = GL.LINE
local GL_LINE_LOOP = GL.LINE_LOOP
local GL_SRC_ALPHA = GL.SRC_ALPHA
local GL_ONE_MINUS_SRC_ALPHA = GL.ONE_MINUS_SRC_ALPHA
```

LuaUI/Widgets/eco_helper_reclaim_highlight.lua
```diff
@@ -0,0 +1,653 @@
+VFS.Include("LuaRules/Configs/customcmds.h.lua")
+
+local widgetName = "Reclaim Highlight"
+
+function widget:GetInfo()
+  return {
+    name      = widgetName,
```
sprunk, line 7:
```suggestion
function widget:GetInfo()
  return {
    name      = "Reclaim Highlight",
```

PR #3886, opened 2020-04-18
Anti-bait gadget for things like Artemis

danfireman:
Redone, tis the anti-chaff gadget is now ready for review.


LuaRules/Gadgets/unit_overkill_prevention.lua
```diff
@@ -416,7 +488,10 @@ end
 
 function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOptions)
 	if (cmdID ~= CMD_PREVENT_OVERKILL) then
-		return true  -- command was not used
```
sprunk, line 419: weird nesting, how about
```Lua
if cmdID == CMD_PREVENT_OVERKILL then
    return PreventOverkillToggleCommand(unitID, cmdParams, cmdOptions)
elseif cmdID == CMD_METAL_DESIRED_TO_PREVENT_OVERKILL then
    return PreventMetalOverkillToggleCommand(unitID, unitDefID, cmdParams, cmdOptions)
else
    return true
end
``

Comment (danfireman) [reply LuaRules/Gadgets/unit_overkill_prevention.lua:419]: Sure, though pretty much everything I've written is horrible for this...

**Comment by GoogleFrog** [comment]
If it barely interacts with OKP then it does not need to be in the OKP gadget. I just thought you might need to use BlockShot and all the relevant units (should) already call an OKP function in BlockShot.

When making the new gadget make sure to look at other gadgets with gadget:AllowWeaponTarget and set the layer in a way that makes sense.

File: LuaRules/Gadgets/unit_targetting_metal_limit.lua [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+local preventChaffShootingCmdDesc = {
+	id      = CMD_MIN_METAL_TO_TARGET,
+	type    = CMDTYPE.ICON_MODE,
+	name    = "Min metal for kill.",
+	action  = 'preventchaffshootingmetallimit',
+	tooltip	= 'Enable to prevent units shooting at units which are very cheap.',
+	params 	= {0, 0, 100, 300, 1000}
+}
+
+function GG.ChaffShootingPrevention_CheckMinMetal(unitID, targetID, damage)
```
GoogleFrog, line 83: Why is this GG?

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,207 @@
...
+	end
+end
+
+function gadget:AllowWeaponTarget(unitID, targetID, attackerWeaponNum, attackerWeaponDefID, defPriority)
+	if GG.ChaffShootingPrevention_CheckMinMetal(unitID, targetID) then
+		return false, defPriority
+	end
+	return true, defPriority
+end
+
```
GoogleFrog, line 195: Should AllowWeaponTargetCheck or AllowUnitTargetRange also be used here?

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,207 @@
...
+--------------------------------------------------------------------------------
+
+function gadget:GetInfo()
+  return {
+    name      = "Unit Metal Targetting",
+    desc      = "Prevents some units from firing at units that cost less than.",
+    author    = "dyth68",
+    date      = "20 April 2020",
+    license   = "GNU GPL, v2 or later",
+    layer     = -1,
```
GoogleFrog, line 16: This gadget appears to be vetoing targets rather than modifying the target priority. As such make sure it is called before gadgets that modify target priority to avoid wasting work.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
GoogleFrog: Perhaps because you want to call GG.ChaffShootingPrevention_CheckMinMetal in Unit Priority to make such targets lower priority? Is that required with this interaction?

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,207 @@
...
+
+function GG.ChaffShootingPrevention_CheckMinMetal(unitID, targetID, damage)
+	Spring.Echo("Min metal check")
+	if not (unitID and targetID and unitMetalMin[unitID]) then
+		return false
+	end
+
+	if spValidUnitID(unitID) and spValidUnitID(targetID) then
+		local gameFrame = spGetGameFrame()
+		local unitDefIDAsKnown = CallAsTeam(Spring.GetUnitTeam(unitID), spGetUnitDefID, targetID)
```
GoogleFrog, line 91: I prefer 
`	local states = spGetUnitLosState(unitID, allyTeamID, false)
	return spGetUnitDefID(unitID), states and states.typed
`
nowdays.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,207 @@
...
+		local cmdDescID = spFindUnitCmdDesc(unitID, CMD_MIN_METAL_TO_TARGET)
+
+		local newParams = {}
+		if (cmdDescID) then
+			for k,v in pairs(MetalHandledUnitDefIDs[unitDefID]) do newParams[k] = v end
+			newParams[1] = state
+			preventChaffShootingCmdDesc.params = newParams
+			spEditUnitCmdDesc(unitID, cmdDescID, {params = preventChaffShootingCmdDesc.params})
+		end
+		Spring.SendLuaRulesMsg(minMetalMsg .. " ".. tostring(unitID) .. " " .. tostring(MetalHandledUnitDefIDs[unitDefID][state + 2]))
```
GoogleFrog, line 123: Why is a gadget sending a LuaRulesMsg?

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: Changed.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: I've had a look through the gadgets and I think it's below all the targetting priority ones. (I first looked at all gadgets/widgets with priority in the name, then all gadgets/widgets with negative priority)

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: I'm unclear on what the former does and can't find the latter in https://springrts.com/wiki/Lua:Callins

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: I don't think it's needed. Removed.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: So that people can write their own widgets that call in to the gadget and change a unit's minimum metal targetting threshhold.

I haven't split this into the UI widget and the underlying gadget yet.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
GoogleFrog: Experiment with both. See the gadget handler for AllowUnitTargetRange.

danfireman:
Ready for review @GoogleFrog

LuaRules/Gadgets/unit_targetting_metal_limit.lua
sprunk:
```suggestion
    layer     = -1, -- vetoes targets, so is before ones that just modify priority
```

LuaRules/Configs/unit_targetting_metal_limit_defs.lua
```diff
@@ -0,0 +1,23 @@
+-- Value is the default state of the command
+local metalThreshholdsByUnitDefIDs = {
+	[UnitDefNames["hoverarty"].id] = {0, 0, 100, 300, 1000},
+	[UnitDefNames["cloaksnipe"].id] = {0, 0, 100, 300, 1000},
```
sprunk, line 4: Is this supposed to be the same table for every unit? In that case could do
```suggestion
local STATES = {0, 0, 100, 300, 1000} -- first value is the default state of the command
local metalThreshholdsByUnitDefIDs = {
	[UnitDefNames["hoverarty"].id] = STATES,
	[UnitDefNames["cloaksnipe"].id] = STATES,
```
etc

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,202 @@
...
+		local cmdDescID = spFindUnitCmdDesc(unitID, CMD_MIN_METAL_TO_TARGET)
+
+		local newParams = {}
+		if (cmdDescID) then
+			for k,v in pairs(metalThreshholdsByUnitDefIDs[unitDefID]) do newParams[k] = v end
+			newParams[1] = state
+			preventChaffShootingCmdDesc.params = newParams
+			spEditUnitCmdDesc(unitID, cmdDescID, {params = preventChaffShootingCmdDesc.params})
+		end
+		Spring.SendLuaRulesMsg(minMetalMsg .. " ".. tostring(unitID) .. " " .. tostring(metalThreshholdsByUnitDefIDs[unitDefID][state + 2]))
```
sprunk, line 118: Looks suspicious, why not just call the relevant thing directly?
```suggestion
		unitMetalMin[unitID] = metalThreshholdsByUnitDefIDs[unitDefID][state + 2]
```

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,202 @@
...
+function gadget:AllowCommand_GetWantedCommand()
+	return {[CMD_MIN_METAL_TO_TARGET] = true}
+end
+
+function gadget:AllowCommand_GetWantedUnitDefID()
+	return true
+end
+
+function gadget:RecvLuaMsg (msg, playerID)
+	if msg:sub(1, #minMetalMsg) ~= minMetalMsg then
```
sprunk, line 133: I'm unsure if the `#` operator works on string length reliably. But this whole func may not be needed, see above.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
```diff
@@ -0,0 +1,202 @@
...
+	end
+end
+
+function gadget:UnitDestroyed(unitID)
+	if canHandleUnit[unitID] then
+		if unitMetalMin[unitID] then
+			unitMetalMin[unitID] = nil
+		end
+		canHandleUnit[unitID] = nil
+	end
```
sprunk, line 181: Looking the value up takes just as much time as setting it
```suggestion
	unitMetalMin[unitID] = nil
	canHandleUnit[unitID] = nil
```

LuaRules/Configs/unit_targetting_metal_limit_defs.lua
danfireman: That's due to me waffling back and forth over whether or not this should vary between units.

Eventually decided on simplicity.

Fixed now

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: Can't remember what this was about...

Changed to your suggestion

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: Removed function.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
danfireman: Doesn't that bloat canHandleUnit unnecessarily though? Taking up a (small) amount of extra memory on units that don't use metal threshhold targetting?

GoogleFrog: As far as I know `nil` takes up no space.

LuaRules/Gadgets/unit_targetting_metal_limit.lua
GoogleFrog: AllowUnitTargetRange is not a Spring callin. Spring overloads gadget:AllowWeaponTarget to do (at least) two distinct things by setting magic numbers in its parameters. I split one of the other uses off into AllowUnitTargetRange in the gadget handler.

GoogleFrog:
Whatever came of the idea of also adding a set of bad targets? Felon wants to shoot Ronin but not Razor. AA wants to shoot Raven but not Raptor, and both units cost the same. This gadget doesn't seem able to do either of these things.

I'm also wondering how to integrate it with the default UI. Maybe it should distinguish between manual commands and automatic target acquisition (don't fire at radar could have a mode where it does the same)? I'm not all that keen on adding a state toggle to the default UI, and it would be a shame for people to have to be told this exists to use it.

I will most likely look closer on the weekend.

GoogleFrog:
Just updated the name for searchability.

GoogleFrog:
A Phantom with Fight shoots Glaives.

GoogleFrog:
Which is a bad thing.

GoogleFrog:
Maybe it's not bad, but it is certainly an oversight. Solving this was the big task in Don't Fire At Radar, and I dislike my solution there. It messes with command queues a lot and possibly isn't even scalable. I was hoping for a solution here, but I guess this at least works perfectly for turrets (as far as I can tell) and units without Fight. I'll pull it on that basis, although I am going to have to rename everything so I can search it years later. I consider this type of thing 'bait', and don't think it should be tied exactly to metal.

PR #3892, opened 2020-04-23
Prevent Locust firing while crashing

a-nop:


a-nop:
Ick, this is not working yet..

a-nop:
Works now.

GoogleFrog:
Don't all the gunships fire while crashing for the rule of cool?

a-nop:
Locust is under a lot of scrutiny since it dominates on a few MM maps right now.  Free damage seemed like an ok thing to fix.  Maybe it's better to just make some numbers smaller instead.

sprunk:
I wouldn't expect firing while crashing to be a major contributing factor to dominance and I agree that this breaks the rule of cool and consistency with other GS.

sprunk:
Dominance addressed in 7e3755a88d64ddf2da8d56e87d14f13e29060e31

PR #3893, opened 2020-04-24
Tighter velocity prediction for Scorcher

a-nop:
Match Scorcher's velocity prediction to the frame rate of tactical AI.  This is specifically to prevent "overshoot" when predicting the movements of Locust and should help it keep close range against other units too.

### DIFF
--- a/LuaRules/Configs/tactical_ai_defs.lua
+++ b/LuaRules/Configs/tactical_ai_defs.lua
@@ -633,6 +633,7 @@ local behaviourConfig = {
 		swarmLeeway = 300,
 		skirmLeeway = 10,
 		stoppingDistance = 8,
+		velocityPrediction = 20,
 	},
 	
 	["hoverscout"] = {


GoogleFrog:
I don't see anything particularly magical about matching the values here. It looks like you've just reduced the prediction distance of Scorcher.

a-nop:
Tactical AI predicts the target's position `velocityPrediction` frames in the future, potentially less if we are faster than the target, [here](https://github.com/ZeroK-RTS/Zero-K/blob/49301022dd494737b11b98bddac535b1f3a8e813/LuaRules/Gadgets/unit_tactical_ai.lua#L343-L354).

Velocity is magically re-predicted every 20 frames as seen [here](https://github.com/ZeroK-RTS/Zero-K/blob/49301022dd494737b11b98bddac535b1f3a8e813/LuaRules/Gadgets/unit_tactical_ai.lua#L644).

Matching `velocityPrediction` to 20 frames seems ideal for minimizing error for units that want to stay a fixed distance from their targets, and this is already set for many raiders.  The default value of 30 yields up to 50% overprediction which is probably fine for skirmishers to keep a safe distance.

GoogleFrog:
> Tactical AI predicts the target's position velocityPrediction frames in the future, potentially less if we are faster than the target, here.
> Velocity is magically re-predicted every 20 frames as seen here.

I know. My point is that if tactical AI updated every frame instead of every 20th, then unit behaviour would be smoother but overall unchanged. Scorchers would still try to move away from units that are moving towards them.

a-nop:
Since tactical AI is updated every 20th frame and not every frame, any value for `velocityPrediction` lower than 20 potentially causes collisions or undershoot in the time between updates.  If tactical AI were updated every frame I would have set `velocityPrediction` to 1.

I see what you mean about the Scorcher never reaching its waypoint, but sometimes it can reach the waypoint, and sometimes its current position is further tangent to target velocity, so we might as well put it in a decent spot.

GoogleFrog:
>  If tactical AI were updated every frame I would have set `velocityPrediction` to 1.
If you set `velocityPrediction` to 1 then the prediction will never move Move order more than a few elmos. If this is what you think would result in good behaviour then why not set `velocityPrediction` to zero.

a-nop:
I didn't set it to zero, I set it to 20.

PR #3898, opened 2020-04-25
Set gui_attrition_counter to draggable=false

Legomenon-gh:
A quick test PR for a personal pet peeve, I don't see a strong reason to have info widgets be draggable outside of the UI tweak mode since all other default widgets cannot be dragged during play.

### DIFF
--- a/LuaUI/Widgets/gui_attrition_counter.lua
+++ b/LuaUI/Widgets/gui_attrition_counter.lua
@@ -488,7 +488,7 @@ function CreateWindow()
 		minHeight = 65,
 		maxHeight = 65,
 		minWidth = 250,
-		draggable = true,
+		draggable = false,
 		resizable = false,
 		tweakDraggable = true,
 		tweakResizable = true,


danfireman:
I like dragging them around to be out of the way of the action.

sprunk:
I like the change on ideological grounds but for some arcane reason the attrition counter is perpetually in an annoying spot which it needs to be dragged away from. Possibly this could be an advanced options, but no strong opinion either way.

Legomenon-gh:
That makes sense, I can add a widget option instead to control its draggability instead to make it easier to test the widget (and ideally fix the positioning problem one day, and remove the setting then).

Anarchid:
I don't remember why it's movable. Probably cargo cult!

Perhaps the part where "it's always in some annoying spot" can be addressed by some form of alternate layout that would flow better with the rest of the UI - perhaps, similar to resource bars.

I have tripped on accidentally moving the widget several times, so this would be good even without an option IMO.

Anarchid:
Ah, well. Klon just set it to draggable pretty much without any further explanation: 
" team game support, remade window, tracks detailed player info now "

https://github.com/ZeroK-RTS/Zero-K/commit/7535c5e79865722ca54986d295289dc029ab40ad

PR #3909, opened 2020-04-30
subs can traverse shallows, but don't get super armor

Anarchid:
Currently:

- They take 1/20 of damage from non-waterweapon sources, so that they are not killable with AoE weapons shooting at the water plane. This is inconsistent with amphs, who take full damage, submerged buildings, etc.
- ~~They can be physically thrown out of water, in which case the armor fails to go away.~~
- They can travel on water which is too shallow to hide their collision volume - Scylla, for example, has waterline 25 and minDepth 15. It will poke out if it travels at depths of -15, and will be shot by surface units. An example of this occuring is straits on Scorpio.
- Their min water depths are higher than for most other ships (excluding striders). This makes designing maps with borderline ship-navigable areas annoying.

Proposed:
- Submarines have same water depth as other ships
- Submarines have no special armor.
- Poking out of water behaviour is no longer a bug, but a feature. The player gets to decide whether to send them into shallows where they can be attacked, or not.

### DIFF
--- a/gamedata/movedefs.lua
+++ b/gamedata/movedefs.lua
@@ -228,7 +228,7 @@ local moveDefs = {
 	UBOAT3 = {
 		footprintx = 3,
 		footprintz = 3,
-		minwaterdepth = 15,
+		minwaterdepth = 5,
 		crushstrength = 150,
 		subMarine = 1,
 	},
--- a/gamedata/weapondefs_post.lua
+++ b/gamedata/weapondefs_post.lua
@@ -365,6 +365,20 @@ end
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 --
+-- Remove submarine damage modifier 
+-- TODO: also remove from defs
+
+for _, weaponDef in pairs(WeaponDefs) do
+	if weaponDef.damage then
+		if weaponDef.damage.subs and weaponDef.damage.default then
+			weaponDef.damage.subs = weaponDef.damage.default
+		end
+	end
+end
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+--
 -- Fix the canAttack tag
 --
 
--- a/units/subraider.lua
+++ b/units/subraider.lua
@@ -45,7 +45,7 @@ return { subraider = {
   turninplace            = 0,
   turnRate               = 600,
   upright                = true,
-  waterline              = 20,
+  waterline              = 55,
   workerTime             = 0,
 
   weapons                = {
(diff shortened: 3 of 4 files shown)


sprunk:
> They can be physically thrown out of water, in which case the armor fails to go away.

[Does it?](https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/subraider.lua#L37)

Anarchid:
> > They can be physically thrown out of water, in which case the armor fails to go away.
> 
> [Does it?](https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/subraider.lua#L37)

Oooh. I guess this could be used to give them armor conditional on not poking out of water, then. Except instead of SfxOccupy, it could just be a check for whether waterline - altitude > 0.

GoogleFrog:
And this just works?

If this just works then I want you to mess with the Tidal parameters and sub movement depth until they are able to move under Tidals.

Anarchid:
Yep, just works. 

I guess i'll try the tidal thing. Unsure the engine gets it though.

sprunk:
(already said on Discord, repeating to prevent tears in the rain)

> They take 1/20 of damage from non-waterweapon sources, so that they are not killable with AoE weapons shooting at the water plane. This is inconsistent with amphs, who take full damage, submerged buildings, etc.

It's not fully inconsistent because subs are perpetually at a depth just below the surface, reachable with surface weaponry AoE. Meanwhile amphs are vulnerable at a thin layer of depth just deeper than their height, and as they go deeper the amount of AoE needed to reach them increases. Consider ocean of depth 1000; a sub still lives at -20 and is reachable by everything with a bit of AoE, but a Duck lives at -1000 and can't even be nuked. Ideally if sub were to lose the special armor then it would live at a depth deeper than surface weaponry AoE. This potentially makes it grind against the oceanfloor on most maps which is slightly lame and also makes it have to attack surface ships at a weird sharp angle.

Anarchid:
Seawolves pass under tidals if given a waterline greater than 35. They also seem to path around obstacles on sea floor.

> Ideally if sub were to lose the special armor then it would live at a depth deeper than surface weaponry AoE

A direct Envoy hit over a waterline-50 seawolf deals it 150 damage out of Envoy's 600; the same Seawolf also survives a full broadside from a Shogun. 

At waterline 60, Envoy and Shogun broadside do around 30 damage to a Seawolf with a perfect hit, and Scalpel cannot damage it at all. All other sources of sea-borne AoE can just target it directly. 

Most maps designed for ZK sea play have depths around 70 for deep water areas; this is about as much depth as needed for Grizzly to become submerged.

PR #3912, opened 2020-05-02
Implement weapon lookahead mechanics

esainane:
This obsoletes and implements in gadgetspace most of what the prefire
widgets did in widgetspace: Look at nearby units, including ones out
of immediate range, and if they would be in range when a projectile
would hit, fire in that direction.

Unlike PreFire, this also causes units to aim their weapons toward that
direction in advance, provided they don't already have a better target.
Units are now that much smarter!

This effectively implements PreFire without latency issues or lua speed,
fully uses engine synced target prediction, preserves all Zero-K
target priorities, and leaves the user free to use manual set-target
without interference.

Caveats:
 - Due to an engine issue, this won't do anything for Cannon
   projectile types until mantis #6382 is resolved.
 - terve's PreFire widgets consider weapon inaccuracy when deciding
   whether to set target on the ground. In rare cases you can have
   positions where a Reaver won't fire with engine lookahead but would
   fire and sometimes hit with PreFire. Similar cases are more common
   with Stardust.

This is applied to a fairly conservative list of units right now,
taken from the list of PreFire widgets and the subset of SweepAttack
widgets where I think PreFire is a non-trivial component. This list
could easily be extended to units where swiveling to aim at units
before they are immediately in range (but not firing until they would
hit) to make units feel much smarter in a many situations.

Some possible follow-ups:
 - Raider exchanges are more even when the retreating side has their
   body aimed away in advance, but pre-aims the weapon so they can fire
   at the same time chasers begin firing. Currently chasers win even
   engagements due retreaters having to waste time either repositioning
   their body or repositioning their weapons.
 - Stingers look a lot smarter when they move to aim in advance.
   This might be due to the visual prominence of their turret, but it
   does save a lot of very real frames if care wasn't taken to build
   the stinger facing the direction of expected attacks.
   (I freely admit I haven't bothered to do this)
 - Units that receive a height range bonus (such as Crab) seem to already
   lookahead slightly as they seem to overestimate the range bonus.
   The difference doesn't seem to be as large as expected, but Crab fits
   the definition of "uses slow moving projectiles with large AoE", and
   being as smart on even ground probably helps a little.
 - Vehicles with slowly turning turrets feel smarter during early
   engagements, though nothing compensates for body turn outpacing
   turret turn when the user really wants to spin corners.
 - Cyclops broadsides become fun and (maybe) practical. TAAAAAANKS!

### DIFF
--- a/LuaRules/Configs/target_lookahead_defs.lua
+++ b/LuaRules/Configs/target_lookahead_defs.lua
@@ -2,10 +2,7 @@
 local unitArray = {}
 
 for i=1,#UnitDefs do
-	local cp = UnitDefs[i].customParams
-	if cp and cp.lookahead then
-		unitArray[i] = cp.lookahead
-	end
+	unitArray[i] = UnitDefs[i].customParams.lookahead
 end
 
 return unitArray
--- a/LuaRules/Configs/target_lookahead_defs.lua
+++ b/LuaRules/Configs/target_lookahead_defs.lua
@@ -1,0 +1,11 @@
+-- unitDefID -> lookahead
+local unitArray = {}
+
+for i=1,#UnitDefs do
+	local cp = UnitDefs[i].customParams
+	if cp and cp.lookahead then
+		unitArray[i] = cp.lookahead
+	end
+end
+
+return unitArray
--- a/LuaRules/Configs/target_lookahead_defs.lua
+++ b/LuaRules/Configs/target_lookahead_defs.lua
@@ -1,8 +1,0 @@
--- unitDefID -> lookahead
-local unitArray = {}
-
-for i=1,#UnitDefs do
-	unitArray[i] = UnitDefs[i].customParams.lookahead
-end
-
-return unitArray
--- a/LuaRules/Gadgets/unit_target_lookahead.lua
+++ b/LuaRules/Gadgets/unit_target_lookahead.lua
@@ -1,0 +1,29 @@
+if not gadgetHandler:IsSyncedCode() then
+	return
+end
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+function gadget:GetInfo()
+  return {
+    name      = "Auto-Target Lookahead",
+    desc      = "Has units with slow moving/high AoE attacks look outside of their immediate range for automatic targeting.\n"..
+    "This obsoletes the aspect of PreFire that predicts when enemies will be in range by the time a projectile would hit them.",
+    author    = "esainane",
+    date      = "2020-05-03",
+    license   = "GNU GPL, v2 or later",
+    layer     = -1,
+    enabled   = true,
+  }
+end
+local lookaheadUnitDefs = include("LuaRules/Configs/target_lookahead_defs.lua")
+
+function gadget:UnitCreated(unitID, unitDefID)
+	local lookahead = lookaheadUnitDefs[unitDefID]
+	if not lookahead then
+		return
+	end
+	for weaponIdx=1,#UnitDefs[unitDefID].weapons do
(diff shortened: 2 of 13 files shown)


LuaRules/Configs/target_lookahead_defs.lua
```diff
@@ -0,0 +1,11 @@
+-- unitDefID -> lookahead
```
sprunk, line 1: put this right in the gadget IMO, both files are tiny and Configs was for things not in customparams (some files which just read customparams and pass the table exist there but those should be abolished at some point)

LuaRules/Configs/target_lookahead_defs.lua
```diff
@@ -0,0 +1,11 @@
...
+local unitArray = {}
+
+for i=1,#UnitDefs do
+	local cp = UnitDefs[i].customParams
+	if cp and cp.lookahead then
+		unitArray[i] = cp.lookahead
+	end
```
sprunk, line 8: Can drop the `cp` check, all defs are supposed to have a valid customparams table (even if empty)
```suggestion
	unitArray[i] = UnitDefs[i].customParams.lookahead
```

LuaRules/Configs/target_lookahead_defs.lua
GoogleFrog: Agreed. This is more of a workaround for the engine failing to implement new weapondefs than a gadget.

GoogleFrog:
I don't see any point to this while the engine is broken. I can't find your engine PR.

LuaRules/Configs/target_lookahead_defs.lua
esainane: Done!

LuaRules/Gadgets/unit_target_lookahead.lua
```diff
@@ -0,0 +1,33 @@
...
+end
+
+function gadget:UnitCreated(unitID, unitDefID)
+	local lookahead = lookaheadUnitDefs[unitDefID]
+	if not lookahead then
+		return
+	end
+	for weaponIdx=1,#UnitDefs[unitDefID].weapons do
+		Spring.SetUnitWeaponState(unitID,weaponIdx,"autoTargetRangeBoost",lookahead)
+	end
```
sprunk, line 32: Could avoid indexing the `UnitDefs` table and the `#` operator by caching it:
```suggestion

local lookaheadUnitDefs = {}
local weaponCounts = {}
for i=1, #UnitDefs do
	local unitDef = UnitDefs[i]
	lookaheadUnitDefs[i] = unitDef.customParams.lookahead
	weaponCounts[i] = #unitDef.weapons -- even for non-lookaheads, this keeps the internal representation an array
end

function gadget:UnitCreated(unitID, unitDefID)
	local lookahead = lookaheadUnitDefs[unitDefID]
	if not lookahead then
		return
	end
	for weaponIdx=1, weaponCounts[unitDefID] do
		Spring.SetUnitWeaponState(unitID, weaponIdx, "autoTargetRangeBoost", lookahead)
	end
```

GoogleFrog: A yes. UnitDefs access is slow.

esainane:
Engine fixed as of 104.0.1-1499-g2d729c4761. See spring/spring@cd6a2ed and spring/spring@d49d8d3

GoogleFrog:
Check whether this breaks on engines prior to 1499.

esainane:
There's no hard breakage pre-1499. Setting the parameter operates as expected for non-cannon (and non-starburst) weapon types, and silently does nothing with cannon (or starburst) weapon types.

From 1499 onwards, setting the parameter operates as expected for all weapon types.

PR #3913, opened 2020-05-02
change Bolas icon

Anarchid:
Bolas has the Support right-arrow icon. This is bad because it is very easy to confuse with Scalpel, as the arrow's tip is located exactly on the outline of the Hover chassis sub-icon.

For extra aggravation, one of the most effective use of Bolas is *alongside* Scalpel.

image

This PR attempts to resolve this by introducing a new icon role symbol of "heavy raider", represented by right angle bracket.

image


sprunk:
Maybe change Dagger to `hoverscout` instead? Then Bolas can be regular `hoverraid`.

GoogleFrog:
The angle bracket just looks like rotated AA. The outline style is too visually dissimilar from the style of other lines.

The best approach would be making Dagger scout and Bolas raider, as with Kodachi and Blitz, but that would be locally confusing.

Anarchid:
Another approach could ve to shift the arrow tip towards the center.

GoogleFrog:
The solution is to give Dagger the scout icon and Bolas the raider icon (see Kodachi and Blitz).

Issue #3917, opened 2020-05-05
Extending Command tooltips with information about State hotkeys


danylo-dubinin:
Hi, I've been augmenting my UI with hotkeys, chaning and experimenting with them a lot to develop an optimal hotkey config for my playstyle. Because of that I sometimes might forget a new hotkey that I'm trying out. 
If it's a **State hotkey**, to recall it I open Epic Menu, click Search, type the name of the action if I remember the name, and find the hotkey in the result list.

I think command tooltips can be extended to cantain more info about configured hotkeys, because we all are sometimes in that moment when you are morphing your geo in 1 match out of 30, and you cannot remember what key you have binded to Low Misc. Priority. I think it's a not neccessary moment of stress so I propose extending the tooltips so they look like in the following examples

- A subset of states has hotkeys.
**Low** doesnt have a hotkey so it's not in the list. *Toggle* is not set, so it's not in the list as well.
![image](https://user-images.githubusercontent.com/20540273/81111284-7bb14b00-8f25-11ea-9aee-a9d83b358b21.png)
Before the change it is empty, nothing is shown at all.

- Only toggle has a hotkey.
Didn't manage to make a screenshot for this one for some reason because I'm totally noob in Windows, but for a state that is only a boolean toggle it says "Toggle: ..." in the same manner as above. If it has another hotkey for On / Off / etc states then it's also listed. 

In short, making all the hotkeys for a State command displayed in the tooltips.
I can submit a PR if you guys find something in it. Criticism is welcome.

sprunk:
>we all are sometimes in that moment when you are morphing your geo in 1 match out of 30, and you cannot remember what key you have binded to Low Misc. Priority.

>Low doesnt have a hotkey so it's not in the list.

Sounds like it would be good to make all of the states show up if any of them has a hotkey (in this example with `Low: (none)`).

sprunk:
6ac84566d15c1d170305c0d375bd3221515ac550 (#3919)

PR #3919, opened 2020-05-06
show states hotkeys in chili integral menu tooltips

danylo-dubinin:
From https://github.com/ZeroK-RTS/Zero-K/issues/3917:

Hi, I've been augmenting my UI with hotkeys, chaning and experimenting with them a lot to develop an optimal hotkey config for my playstyle. Because of that I sometimes might forget a new hotkey that I'm trying out.
If it's a State hotkey, to recall it I open Epic Menu, click Search, type the name of the action if I remember the name, and find the hotkey in the result list.

I think command tooltips can be extended to contain more info about configured hotkeys, because we all are sometimes in that moment when you are morphing your geo in 1 match out of 30, and you cannot remember what key you have binded to Low Misc. Priority. I think it's a not neccessary moment of stress so I propose extending the tooltips so they look like in the following examples

- A subset of states has hotkeys:
states

- Only toggle has a hotkey:
toggle

- States do not exist:
(no change)
short

In short, making all the hotkeys for a State command displayed in the tooltips.

Criticism is welcome.

> Sprunk:
Sounds like it would be good to make all of the states show up if any of them has a hotkey (in this example with Low: (none)).

### DIFF
--- a/LuaUI/Widgets/gui_chili_integral_menu.lua
+++ b/LuaUI/Widgets/gui_chili_integral_menu.lua
@@ -493,6 +493,7 @@ end
 --- Returns:
 --- - hotkeys: Table[state_name, value_text], nil if action cannot be found or action has no states; value_text is
 ---   "(none)" if a hotkey is not set
+--- - n: Integer - number of hotkeys that are set, nil under same conditions as for hotkeys
 local function GetHotkeysForStatesText(action_name)
 	local action = custom_cmd_actions[action_name]
 	if not action then return nil end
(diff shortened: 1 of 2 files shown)


GoogleFrog:
I am slightly against the suggestion of adding '(none)' for unset hotkeys if some hotkey is set. Most (all?) of the default hotkeys for states are just toggles, not keys that set states. I want to avoid scaring people who are looking around the UI tooltips with too much information.

For consistency with the other tooltips I propose that the state toggle hotkey be shown in parenthesis at the end of the description string. If the state toggle has hotkeys that set a particular state then make the list of hotkeys with "State Hotkeys:" at the top as shown in your screenshot. This would list all the states and the hotkeys for each, including '(none)' for states that have no hotkey.

sprunk:
> I am slightly against the suggestion of adding '(none)' for unset hotkeys if some hotkey is set. Most (all?) of the default hotkeys for states are just toggles, not keys that set states.

I agree, what I said earlier had poor wording (didn't mean to suggest Low/Normal/High appearing if just Toggle is set, just only if one of L/N/H is).

danylo-dubinin:
Agreeing your comments @GoogleFrog and @sprunk. Updated.

- State hotkeys exist, but only toggle is set:
image

- No toggle is set, some state hotkey is set:
image

- Toggle is set, some state hotkey is set:
image

LuaUI/Widgets/gui_chili_integral_menu.lua
```diff
@@ -478,27 +480,65 @@ local function RemoveAction(cmd, types)
...
-		if tooltip and hotkey then
-			tooltip = tooltip .. " (\255\0\255\0" .. hotkey .. "\008)"
+		local hotkey_for_toggle = GetHotkeyText(command.action)
+		local hotkeys_for_states, states_n = GetHotkeysForStatesText(command.action)
+		if hotkey_for_toggle then
+			tooltip = tooltip .. " (" .. GetGreenStr(hotkey_for_toggle) .. ")"
+		end
+		if hotkeys_for_states and states_n > 0 then
+			tooltip = tooltip .. SEP .. "State Hotkeys:"
+			for state_name, hotkey in pairs(hotkeys_for_states) do
```
GoogleFrog, line 539: I should have requested that this pairs that exists purely to iterate over a transient table be made to not exist.

danylo-dubinin: Is it a performance consideration?

GoogleFrog: Yes.

PR #3920, opened 2020-05-06
Iterate Kestrel

a-nop:
Add to the rest of the configs, normalize with other planes.
* Configs for CAI, target priority, overkill prevention
* Fixed-wing takeoff script
* Stat tweaks


sprunk:
Kestrel is a Raven with less ground interaction, not meant to be a multiplayer unit. The OKP stuff etc is fine though.

a-nop:
In this patch Raven is good for:
* Bombing statics.  In this capacity it has a higher cruising altitude, more health, and has less ground interaction than Kestrel.
* Flex bombing constructors, gunships, skirmishers, and other vulnerable support units like Funnelweb.
* Generally less risk because it is cheaper, and could be even cheaper than it is now taking the new rearm cost into account.

Kestrel is good for:
* Bombing mobile units.  Raven has never been great at this.  Kestrel has less risk because it doesn't have to tag its target.
* Flex anti-air, anti-static.
* Not taking as much ground fire except from dedicated AA.

Kestrel pays premium for its capability, just like Likho, and is more vulnerable to dedicated AA than any of the above because of its lower health pool.

As far as what you would build Kestrel to counter, there's at least as much overlap with Phoenix and Likho as Raven.  The splitting difference with Phoenix is that Kestrel requires less quantity to deal with heavier units.  Compared to Likho, Kestrel has much better damage:cost and health:cost but lacks AoE.

The ideal targets for this Kestrel are striders, commanders, and assault units in loose formation.  It can bomb an advancing Dante without being hit.  It can bomb an advancing Knight without being stunned.  In these cases it outperforms Likho and Phoenix (by damage:cost/rearm time) and Raven (by not dying).  It is the missing air-to-ground anti-heavy link.

sprunk:
> Bombing mobile units. Raven has never been great at this. Kestrel has less risk because it doesn't have to tag its target.
> Not taking as much ground fire except from dedicated AA.

Raven (as the "staple" bomber) interacting with ground units is by design.

>The ideal targets for this Kestrel are striders, commanders, and assault units in loose formation.

All of those were supposed to be in Raven's set of ideal targets. If it is not the case then the solution is probably to tweak Raven instead.

units/bomberstrike.lua
```diff
@@ -20,15 +20,18 @@ return { bomberstrike = {
...
-    --modelradius    = [[10]],
+    modelradius      = [[10]],
   },
 
   explodeAs           = [[GUNSHIPEX]],
   floater             = true,
   footprintX          = 3,
   footprintZ          = 3,
-  iconType            = [[bomber]],
+  iconType            = [[bomberspecial]],
```
GoogleFrog, line 34: Should be skirm.

Histidine91:
I'd like for the model to get a retexture or a full replacement; the one I made is pretty simple/cartoonish.

GoogleFrog:
900 damage for 380 cost is much too close to 800 damage for 300 cost. The choice between Raven and Kestrel presented here is one of survivability vs. localised AA for a bit of extra cost. The units have the same speed and their health only differs by 100. Kestral is essentially a generalist Raven by virtue of ignoring some of Raven's counters.

When I was thinking about a flying Scalpel I meant something of around 250 cost, 550 damage, and less health and speed. It could have less reammo time (5s) to encourage forward airpads. Raven is still great at bombing mobile units. Kestrel can be good at bombing lighter mobile units and areas covered by riots.

I have not looked at the model recently, but I do not recall it being particularly good.

sprunk:
Phoenix is also meant to fly high enough not to care much about riots, and to be good vs lighter stuff than Raven. If the area is covered with enough riots that Raven can't even reach its target then it's probably a group dense enough that Phoenix' AoE damage makes it viable as well despite low single target damage.

a-nop:
kestrel

a-nop:
kestrelicon

a-nop:
Costing 250 makes it amusing next to Sparrow, but I'll try it.

a-nop:
I do think making it lighter and crappier messes up Phoenix's role.  Would rather make Raven anti-static and have Kestrel take its place between Phoenix and Likho.

> Raven is still great at bombing mobile units.

It tends to get toasted and it looks silly trying to tag things.

a-nop:
* Cost increased to 600.
* Rearm time increased to 15 seconds.
* Speed reduced to 222 elmo/s
* Weapon damage increased to 1080 total
* Weapon range increased slightly to 400
* Weapon speed increased to 230 elmo/s
* Weapon tracking reduced to 49 deg/s
* Weapon flight time reduced to 3.2s

Now oneshots:
* Crasher
* Gremlin
* Flail
* Vandal
* Trident
* Phoenix
* Locust
* Lobster
* Riots except Ogre, Mace, Scallop, Corsair
* Cons except Convict, Welder, Mariner, stationary Conch
* Skirms except Bouy and Harpy

a-nop:
Going the other direction since airfac needs a more affordable earlygame.  I've removed Kestrel from the factory in this PR.  I leave the final balance, possible downscaling of the model, and reskinning in the capable hands of others.

GoogleFrog:
I'd need to see it in a screenshot with other planes to be sure, but the model seems to be good enough to try it out. The piece positions for the jets and wing trails seem to be off though.

The numbers are looking ok. 660 damage for 240 cost is better damage/cost than Raven, which is bad since Raven would be edged-out as an anti-heavy. Raven has the most counterplay (in the form of riot escorts) so should remain the most efficient choice for damage. Kestrel is set up to be an annoying sniper plane with limited flex-counterplay so could get away with having a lot less damage. A nice round 500 damage seems reasonable, with more survivability in the form of range and turnrate if it turns out to be too bad.

Its matchups will depend on exactly how much damage it does and what sort of flex options are good against it. Does it fly into moderator range? How good is Fencer? For that matter, how good is Gremlin? I could see this unit destroying Cloaky and Spider due to their glass-cannon nature and the way their AA works. Jumpy and Tank could be immune to it depending on how it interacts with Moderator and Ogre.

a-nop:
This Kestrel is instacooked by 2 Moderators after firing its missiles.  Jumpbots are mostly immune to airfac anyway.

Ogre can also shoot back after the missiles are fired, but two are needed to kill any Kestrels.

Knight can hit the retreating Kestrel.  Gremlin play is the usual bait and kill.

Spiders will have a hard time, but a Tarantula parked behind a Recluse has enough time to fire three missiles and almost kill the Kestrel.  This is one less missile than it takes to kill a Raven.  Redback can shoot it.

kestrelraven

a-nop:
Changing damage from 660 to 640 makes important differences.

Kestrel no longer oneshots:
* Recluse
* Mistral

No longer twoshots:
* Mace

GoogleFrog:
640 is still Raven efficiency. I think the designs of the two units would benefit from having a gap in efficiency.

a-nop:
Reduced damage, health and maneuvering but buffed speed.  I finally figured out which knobs actually tweak aircraft maneuvering.

Most riots do damage to an attacking Kestrel now.

Partial list of no longer oneshots:
* Gremlin
* Seawolf
* Rogue

Partial list of no longer twoshots:
* Scallop
* Angler
* Tarantula
* Blitz
* Harpy
* Thunderbird

Partial list of no longer threeshots:
* Ogre
* Zephyr
* Hercules

| Plane | Cost | Health (/cost) | Damage (/cost) | Speed (/cost*100) |
|--|--|--|--|--|
| Kestrel | 240 | 780 (3.25) | 540 (2.25) | 8.4 (3.5) |
| Raven | 300 | 1100 (3.67) | 800 (2.67) | 7.8 (2.6) |
| Phoenix | 360 | 900 (2.5) | Burn | 8 (2.22) |
| Likho | 2000 | 2360 (1.18) | 2000 (1.0) | 9 (0.45) |

a-nop:
Cleaned up for incremental merge, won't be finishing this anytime soon.

PR #3924, opened 2020-05-09
Improve startup commander chassis descriptions

Legomenon-gh:
Attempt to reflect 1.8.3.2 changes to commander differentiation and
remove reference to dual wielding since all commanders can do that now.

Those strings are kind of hard to keep informative and concise, so feel free to reword them some more. (I'll update https://zero-k.info/mediawiki/index.php?title=Commander too)

### DIFF
--- a/LuaUI/Widgets/init_startup_info_selector.lua
+++ b/LuaUI/Widgets/init_startup_info_selector.lua
@@ -76,9 +76,9 @@ if (vsx < 1024 or vsy < 768) then
 end
 
 local commTips = {
-	["LuaUI/Images/startup_info_selector/chassis_benzcom.png"] = "Select Guardian Chassis\nA tanky chassis with access to a wide range of weapons. Hampered by its slow speed. Can dual wield.",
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Dual wielding refers to the ability to have two normal weapons. Recon and Engineer were always able to equip a normal weapon and a special weapon.

Legomenon-gh:
Ha, I've updated the descriptions to read `can equip two normal weapons` which should be a bit clearer (though still not perfect since it's not very clear how the other chassis behaves).

Not sure how much detail should be given about the bonuses scaling but that probably belongs in the commander upgrade window rather than here (for Strike I mention levels explicitly since that bonus only kicks in at level 3)

GoogleFrog:
Normal is not particularly clear. Maybe 'basic'?

Legomenon-gh:
Yeah I reverted to "can dual wield", that's terser and clear enough in hindsight (and bringing up "basic weapons" in this screen is just more confusing than anything else)

PR #3933, opened 2020-05-18
Increase Raptor maneuvering

a-nop:
Raptor is doing better at keeping distance, but needs a little more help with disengagement and tracking fast targets.

* Acceleration increased 10%.
* Aileron increased 33%.
* Rudder increased 33%.
* Elevator increased 50%.

### DIFF
--- a/units/planeheavyfighter.lua
+++ b/units/planeheavyfighter.lua
@@ -41,9 +41,11 @@ return { planeheavyfighter = {
   iconType               = [[stealthfighter]],
   idleAutoHeal           = 5,
   idleTime               = 1800,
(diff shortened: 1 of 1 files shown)


Licho1:
quantum violation detected

GoogleFrog:
I am not so sure that it violates Quant's Rule. Swift is meant to be the fighter that can quickly reach a location and burst down targets. Raptor is meant to take more time to get there, but then act more like a bit of flying AA once it does. Making it able to turn faster degrades how heavy it looks, but I don't think it is a role problem.

I am a bit concerned that a high turn rate makes jinking at max range viable enough to quality not doing so as bad unit AI.

PR #3934, opened 2020-05-18
Don't draw unit shapes and friends for radar icons

a-nop:
Waste less widget time when controlling units while zoomed out.

### DIFF
--- a/LuaUI/Widgets/gfx_halo.lua
+++ b/LuaUI/Widgets/gfx_halo.lua
@@ -303,7 +303,7 @@ local function SetTeamColor(teamID)
 end
 
 local function DrawVisibleUnitsAll()
-  local visibleUnits = GetVisibleUnits(ALL_UNITS,nil,true)
+  local visibleUnits = GetVisibleUnits(ALL_UNITS,nil,false)
   for i=1,#visibleUnits do
     local unitID = visibleUnits[i]
     SetTeamColor(GetUnitTeam(unitID))
--- a/LuaUI/Widgets/gfx_outline_shader.lua
+++ b/LuaUI/Widgets/gfx_outline_shader.lua
@@ -77,6 +77,8 @@ local applicationShader
 local pingPongIdx = 1
 
 local shadersEnabled = Spring.Utilities.IsCurrentVersionNewerThan(104, 1243) and LuaShader.isDeferredShadingEnabled and LuaShader.GetAdvShadingActive()
+
+local unitsVisible = false
 -----------------------------------------------------------------
 -- Configuration
 -----------------------------------------------------------------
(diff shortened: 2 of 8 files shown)


GoogleFrog:
Does it? spGetVisibleUnits is highly cached so changing how it is called might make it miss the cache and cause worse performance.

sprunk:
As long as there's a different default widget which calls it with `false` it should be fine.

a-nop:
It works, I checked it in the profiler so you don't have to.

a-nop:
I can grep that for you too.

```
LuaUI/Widgets/gfx_night.lua:  local visibleUnits = GetVisibleUnits(-1, nil, false)
LuaUI/Widgets/gfx_outline_no_shader.lua:        local visibleUnits = GetVisibleUnits(ALL_UNITS,nil,false)
LuaUI/Widgets/gui_healthbars.lua:                       visibleUnits = GetVisibleUnits(-1, nil, false)    -- expensive
LuaUI/Widgets/gui_team_platter.lua:  local visUnits = spGetVisibleUnits(-1, nil, false)
LuaUI/Widgets/unit_healthbars.lua:              visibleUnits = GetVisibleUnits(-1, nil, false) --this don't need any delayed update or caching or optimization since its already done in "LUAUI/cache.lua"
LuaUI/Widgets/unit_selectionblurryhalo.lua:     local units = spGetVisibleUnits(-1, 30, false)
LuaUI/Widgets/unit_shapes.lua:                  local units = spGetVisibleUnits(-1, nil, false)
LuaUI/Widgets/unit_shapes.lua:          local units = spGetVisibleUnits(-1, nil, false)
```

a-nop:
In fact usage of GetVisibleUnits including radar icons is in the minority and could potentially be switched, how many of these do you think really need to return radar icons?

```
LuaUI/Widgets/gfx_halo.lua:  local visibleUnits = GetVisibleUnits(ALL_UNITS,nil,true)
LuaUI/Widgets/gui_team_platter.lua:  local units = spGetVisibleUnits(-1, nil, true)
LuaUI/Widgets/gui_xrayhaloselect.lua:  local units = spGetVisibleUnits(-1, nil, true)
```

sprunk:
Sounds fine to make them all use `false`.

a-nop:
With all these widgets turned on, I spawned 50 Rippers, selected them all, and zoomed out.  This branch uses about 10% less widget CPU overall.  About the same improvement when dragging a selection box over the Rippers.

GoogleFrog:
Overall widget CPU is good to check, since the profiler will be confused by the caching being more expensive than the later lookups. The additional changed widgets are not enabled by default, so the changes look fine.

PR #3941, opened 2020-05-21
fix order and performance of hotkey tooltips in integral menu

dahn-zk:
follow-up for https://github.com/ZeroK-RTS/Zero-K/pull/3919

- avoid using `pairs` to improve performance
- the refactoring happens to fix the order of the list. it is
now listed as the states are defined in customCmdTypes

| before | after |
| --- | --- |
| before | after |

### DIFF
--- a/LuaUI/Widgets/gui_chili_integral_menu.lua
+++ b/LuaUI/Widgets/gui_chili_integral_menu.lua
@@ -520,33 +520,40 @@ end
 
 --- Combines the information about the command, its state and hotkeys
 local function GetButtonTooltip(displayConfig, command, state)
-	local SEP = "\n  "
-	
-	local tooltip = displayConfig and displayConfig.stateTooltip and displayConfig.stateTooltip[state]
-	if not tooltip then
-		tooltip = (displayConfig and displayConfig.tooltip) or (command and command.tooltip)
+	local PARAGRAPH = "\n  "
+
+	local tooltip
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_chili_integral_menu.lua
```diff
@@ -528,16 +529,21 @@ local function GetButtonTooltip(displayConfig, command, state)
...
-				tooltip = tooltip .. SEP .. GetGreenStr(state_name .. ": " .. hotkey)
+	if command then
+		local action_name = command.action
+		if command.action then
+			local hotkey_for_toggle = GetHotkeyText(action_name)
+			local states, hotkeys_for_states, number_of_set_hotkeys = GetHotkeysForStatesText(action_name)
+			if hotkey_for_toggle then
+				tooltip = tooltip .. " (" .. GetGreenStr(hotkey_for_toggle) .. ")"
+			end
+			if hotkeys_for_states and number_of_set_hotkeys > 0 then
```
GoogleFrog, line 540: While we're at it, you could turn the deep if-pyramid into a short list of if not (bla and bla) then return.

dahn-zk: Updated
- depyramidize `if`s in GetButtonTooltip
- simplify logic around initial value of `tooltip`
- `SEP` is actually a `PARAGRAPH`

PR #3960, opened 2020-06-05
Small fixes to unit categorization for end graph

Legomenon-gh:
The second commit is the one I'm least sure about, I followed the tooltips but I'm happy to update this further.

One more question I have: should the Funnelweb be defense by virtue of having shields, economy by virtue of being a builder, or other by virtue of being a weird mixture?

Feel free to suggest any additional known bugs or fixes to the end graphs too!

### DIFF
--- a/LuaRules/Configs/unit_category.lua
+++ b/LuaRules/Configs/unit_category.lua
@@ -28,8 +28,8 @@ local function GetUnitCategory(unitDefID)
 	if cp.level then -- commander
 		return "other"
 	end
-	
-	if cp.ismex or cp.income_energy or (cp.pylonrange and ((tonumber(cp.pylonrange) or 0) > 400)) or ud.energyStorage > 0 then
+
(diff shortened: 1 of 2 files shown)


GoogleFrog:
Defense is meant for static things, so Funnelweb is not defense.

Check that ud.isBuilder does not have anything you would not expect in it, such as reef.

Legomenon-gh:
Ha, I didn't parse the Defense tooltip carefully enough `Value of armed structures (and shields) with range [...]`. It's a little awkward to rephrase (maybe `armed and shielded structures`) but doesn't really matter since the behavior matches the intuitive meaning of Defense.

Unit classifications of tricky units after this PR:
- Aspis: army
- Aegis: defense
- Athena: econ
- Badger: army (mines still not counted)
- Funnelweb: econ
- Reef: army
- Scylla: army
- Trinity/Antinuke: other
- All commanders: still other (including guardian with its drones)
- Radars/sonar: still other

Legomenon-gh:
Just realized my first change was actually unnecessary now, so I took that out.

PR #3980, opened 2020-06-23
Select all hotkeys for air units

thomas11111111:
hotkeys for selecting all Swift, Phoenix and Athena


Licho1:
Before doing more advanced hotkeys, we should add in-game UI to edit and see them all

GoogleFrog:
What exactly is the purpose of this PR? As far as I know this change will have no effect on a player unless they manually open their keybinds file and look inside. Any hotkeys added here should be added as buttons in the appropriate selection hotkeys menu. Perhaps menus to select all or idle units of each type should be generated.

sprunk:
I agree regarding "generated", currently the set looks fairly arbitrary. Ideally it would be in some sort of submenu because 100+ buttons is spammy.

GoogleFrog:
@thomas11111111 I was correct, this PR did nothing. Adding a None hotkey to zk_keys did nothing. See the commit above.

You should be responding to comments on PRs rather than firing and forgetting.

PR #3982, opened 2020-06-27
Send idle bombers to newly built airpad

### LINKED ISSUE #3979: Feature; unloaded planes use newly built pads
When waitting for a spot on a rearming pad planes will react to pads being built and use them for rearming.

Legomenon-gh:
Fixes #3979, looking for feedback on Lua/Spring library, and if there are major flaws with my logic (the bomber logic is pretty complicated so I suspected I missed something somewhere, especially around queued commands).

I've verified this behaves fine in normal 1v1 games but cannot vouch for edge cases in a large teamgame.

### DIFF
--- a/LuaRules/Gadgets/unit_refuel_pad_handler.lua
+++ b/LuaRules/Gadgets/unit_refuel_pad_handler.lua
@@ -588,3 +588,45 @@ function gadget:GameFrame(f)
 		UpdatePadLocations()
 	end
 end
+
+-- Search for any idle bombers to rearm at a newly built airpad
+function gadget:UnitFinished(unitID, unitDefID, teamID)
+	local unitDef = UnitDefs[unitDefID]
+	local padCount = tonumber(unitDef.customParams.pad_count)
+	if (padCount == nil or padCount == 0) then return end
+
+	local bombersToRearm
+	local idleBombers = GG.GetIdleBombersForTeam(teamID)
+	local idleBombersCount = #idleBombers
+
+	if (idleBombersCount <= padCount) then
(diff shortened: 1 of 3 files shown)


LuaRules/Gadgets/unit_refuel_pad_handler.lua
```diff
@@ -588,3 +588,45 @@ function gadget:GameFrame(f)
...
+	else
+		bombersToRearm = GetFirstNearestIdleBombers(unitID, idleBombers, padCount)
+	end
+
+	GG.RedirectBombersToAirpad(bombersToRearm, unitID)
+end
+
+function GetFirstNearestIdleBombers(airpadID, idleBombers, count)
+	local unitIDsByDistance = {}
+	for i = 1, #idleBombers do
```
Legomenon-gh, line 613: There has to be a simpler way to do this... all I want is to sort `idleBombers` by their distance and return the first X bombers, but Lua tables really don't do ordering (or mappings)

Legomenon-gh:
I could throw in https://zero-k.info/Forum/Thread/31656 too (disable AA from hitting grounded targets) if this PR looks sane and nobody else wants to work on that since I'm in that code already

GoogleFrog:
Don't do them both in the same PR. They should be handled distinctly.

Legomenon-gh:
Of course that would be a separate PR. I'm not going to touch the AA stuff though since that's out of my Spring-fu zone to do well

Legomenon-gh:
Closing this PR since I have insufficient confidence in its validity and I believe it can be implemented better

PR #3985, opened 2020-06-29
Hooked up COFC to standard camera movement hotkeys

dvr333:


LuaUI/Widgets/camera_cofc.lua
```diff
@@ -737,6 +739,34 @@ local keys = {
...
 }
+
+local function HotkeyChangeNotification()
+	keys = {}
+
+	local key = WG.crude.GetHotkeyRaw("moveleft")
+	local keycode = ToKeysyms(key and key[1])
+	keys[keycode] = 'left'
+	key_code.left = keycode;
+	local key = WG.crude.GetHotkeyRaw("moveright")
```
sprunk, line 750: shadows the previous variable, consider either a new var
```suggestion
	local key2 = WG.crude.GetHotkeyRaw("moveright")
```
or reusing the existing one:
```suggestion
	key = WG.crude.GetHotkeyRaw("moveright")
```

dvr333: Done

PR #3990, opened 2020-07-08
Remove Collision from Zenith beam

### LINKED ISSUE #3987: zenith visual beam is hide-able, thus NOT revealing position
the visual beam into the sky of zenith is not visible if you park something in its path, e.g. charon

http://zero-k.info/Battles/Detail/916987

this is a strategic advantage, hiding the position when you know this. is this intended? else fix or give the skybeam damage? -.-

Arch-Shaman:
Fixes #3987 by making the zenith beam not collide with enemies, neutrals (eg: zombies), enemies or features. Also makes the beam a water weapon so that killing it then burying it underwater and resurrecting it won't hide it.

(Also sorry for the double PR, I realized my white space was off)


sprunk:
I took the waterweapon change because it is worth doing even if the other PR is picked.

Arch-Shaman:
Both have waterweapon = true in them, which I guess is why the commit doesn't create a conflict?

sprunk:
yes

Arch-Shaman:
Closing because skyline check was chosen instead.

PR #3991, opened 2020-07-08
Alternative Zenith fix (Skyline check)

### LINKED ISSUE #3987: zenith visual beam is hide-able, thus NOT revealing position

Arch-Shaman:
Fixes #3987 by implementing a "skyline" check through GetUnitsInCylinder and GetFeaturesInCylinder. This is an alternative solution to PR #3990.

### DIFF
--- a/LuaRules/Gadgets/unit_zenithskylinecheck.lua
+++ b/LuaRules/Gadgets/unit_zenithskylinecheck.lua
@@ -3,15 +3,15 @@ if not gadgetHandler:IsSyncedCode() then
 end
 
 function gadget:GetInfo()
-  return {
-    name      = "Zenith Skyline Checker",
-    desc      = "Blocks Zenith Meteor spawn when the beam is broken",
-    author    = "Shaman",
-    date      = "July 8 2020",
-    license   = "CC-0",
-    layer     = -1,
-    enabled   = true,
-  }
+	return {
+		name    = "Zenith Skyline Checker",
+		desc    = "Blocks Zenith Meteor spawn when the beam is broken",
+		author  = "Shaman",
+		date    = "July 8 2020",
+		license = "CC-0",
+		layer   = -1,
+		enabled = true,
+	}
 end	
 
 
--- a/LuaRules/Gadgets/unit_zenithskylinecheck.lua
+++ b/LuaRules/Gadgets/unit_zenithskylinecheck.lua
@@ -20,10 +20,10 @@ local spSetUnitRulesParam = Spring.SetUnitRulesParam
 
 function gadget:Explosion(weaponDefID, px, py, pz, AttackerID, ProjectileID)
 	if weaponDefID == gravityWeaponDefID then
-		spSetUnitRulesParam(AttackerID, "isBlocked", 1)
+		spSetUnitRulesParam(AttackerID, "meteorSpawnBlocked", 1)
 	end
 end
 
 function gadget:Initialize()
-	Script.SetWatchWeapon(gravityWeaponDefID, true)
+	Script.SetWatchExplosion(gravityWeaponDefID, true)
 end
--- a/LuaRules/Gadgets/unit_zenithskylinecheck.lua
+++ b/LuaRules/Gadgets/unit_zenithskylinecheck.lua
@@ -5,10 +5,10 @@ end
 function gadget:GetInfo()
   return {
     name      = "Zenith Skyline Checker",
-    desc      = "A meteor blocking test gadget",
+    desc      = "Blocks Zenith Meteor spawn when the beam is broken",
     author    = "Shaman",
     date      = "July 8 2020",
-    license   = "GNU GPL, v2 or later",
+    license   = "CC-0",
     layer     = -1,
     enabled   = true,
   }
@@ -16,10 +16,11 @@ end
 
 
 local gravityWeaponDefID = WeaponDefNames["zenith_gravity_neg"].id
+local spSetUnitRulesParam = Spring.SetUnitRulesParam
 
 function gadget:Explosion(weaponDefID, px, py, pz, AttackerID, ProjectileID)
 	if weaponDefID == gravityWeaponDefID then
-		Spring.SetUnitRulesParam(AttackerID, "isBlocked", 1)
+		spSetUnitRulesParam(AttackerID, "isBlocked", 1)
 	end
 end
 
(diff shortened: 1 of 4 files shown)


Arch-Shaman:
Also tried Spring.GetUnitWeaponHaveFreeLineOfFire(unitID,1, ux,uy,uz,ux,9000,uz) -- this doesn't seem to provide the right results so I opted for getunitsincylinder with an additional getfeatures with 20 radius (+4 elmo from beam radius) which visually seemed to provide the correct behavior. The waterweapon addition to the beam is necessary otherwise the player can simply do UW guass nonsense with it and block the beam that way.

GoogleFrog:
I would much rather this be done registering and handling the event of the Zenith beam impacting something before fading out, as this seems like it would be computationally free in normal usage. The other PR (making the beam penetrate everything) is fine though.

Arch-Shaman:
> I would much rather this be done registering and handling the event of the Zenith beam impacting something before fading out, as this seems like it would be computationally free in normal usage.

ExplosionCreated or WatchWeapon is the closest thing I could think of. Unless there's something else I'm not seeing?

sprunk:
> ExplosionCreated or WatchWeapon is the closest thing I could think of. Unless there's something else I'm not seeing?

IIRC fading lasers don't trigger the Explosion callin, but ProjectileDestroyed should be fair game (called both on expiry and collision, but just an `y > 5000` check should be fine). WatchWeapon just enables the other callins and will need to be used regardless.

Arch-Shaman:
> IIRC fading lasers don't trigger the Explosion callin, but ProjectileDestroyed should be fair game (called both on expiry and collision, but just an y > 5000 check should be fine). WatchWeapon just enables the other callins and will need to be used regardless.

I'll check it out. Thanks for the info.

GoogleFrog:
I would start by looking at Explosion. You don't need to detect the case where the projectile fades out, so watching for ProjectileDestroyed should not be required.

Arch-Shaman:
Should be good to go barring any unforeseen things.

LuaRules/Gadgets/unit_zenithskylinecheck.lua
```diff
@@ -0,0 +1,29 @@
...
+  }
+end	
+
+
+local gravityWeaponDefID = WeaponDefNames["zenith_gravity_neg"].id
+local spSetUnitRulesParam = Spring.SetUnitRulesParam
+
+function gadget:Explosion(weaponDefID, px, py, pz, AttackerID, ProjectileID)
+	if weaponDefID == gravityWeaponDefID then
+		spSetUnitRulesParam(AttackerID, "isBlocked", 1)
```
sprunk, line 23: "isBlocked" is a bit generic, how about "meteorSpawnBlocked"

LuaRules/Gadgets/unit_zenithskylinecheck.lua
```diff
@@ -0,0 +1,29 @@
...
+	end
+end
+
+function gadget:Initialize()
+	Script.SetWatchWeapon(gravityWeaponDefID, true)
```
sprunk, line 28: IIRC there might be a more specific SetWatchWeaponExplosion

scripts/zenith.lua
```diff
@@ -191,8 +192,10 @@ local function SpawnProjectileThread()
 			RegainControlOfMeteors()
 			currentlyStunned = false
 		end
-		
-		SpawnMeteor()
+		if not spGetUnitRulesParam(unitID,"isBlocked") or spGetUnitRulesParam(unitID,"isBlocked") == 0 then -- Only spawn if there's a clear view to the sky
```
sprunk, line 195: could do `~= 1` to handle both `nil` and `0`

units/zenith.lua
```diff
@@ -105,6 +105,7 @@ return { zenith = {
       turret                  = true,
       weaponType              = [[LaserCannon]],
       weaponVelocity          = 6000,
+      waterWeapon             = true,
```
sprunk, line 108: Should be brought back to false now that the visual actually matches the logic

sprunk:
> You don't need to detect the case where the projectile fades out

Ideally the fade would be what triggers the meteor to spawn. Currently the block is just applied to the next beam (which is not a problem given the reloadtime is both not high enough for this to be noticeable and not low enough for there to be 2 beams inbound).

sprunk:
Also I just realized a minor issue: you can prevent the "meteors past the cap fall on your head" this way. IIRC that was done by design to incentivize usage (but it's a bit opaque on what is happening, and other incentives already exist, so this is mostly just the cool factor in the end).

units/zenith.lua
Arch-Shaman: I think if it's false, the check won't go through and the meteors will gather anyways. I can run a test to see if this is true though.

scripts/zenith.lua
Arch-Shaman: Good point, I'll update it.

units/zenith.lua
sprunk: If it doesn't call `Explode` on hitting water that's an engine bug. Or maybe it needs a fake water CEG defined to do that.

Arch-Shaman: It won't even fire underwater. It's similar to if you take a ravager and drop it in the sea AFAIK.

Arch-Shaman:
> Also I just realized a minor issue: you can prevent the "meteors past the cap fall on your head" this way. IIRC that was done by design to incentivize usage (but it's a bit opaque on what is happening, and other incentives already exist, so this is mostly just the cool factor in the end).

What if being blocked for some time caused the meteors to fall.

GoogleFrog:
I think being blocked should be equivalent to being turned off - the meteors instantly fall. Falling meteors are recaptured when the Zenith turns back on with the current on/off toggle so we need not worry about a plane flying over and destroying your base.

The change is essentially a bit of rule-of-cool for Raven or Krow interactions.

Arch-Shaman:
Ravens while testing were extremely finicky to get to block the beam (as was licho) but Krow is probably more reasonable interaction expectation. I'll add it to the to do.

GoogleFrog:
Are you sure this tries to fire but is blocked underwater? I don't see how that could be the case with waterWeapon = true.

Arch-Shaman:
> Are you sure this tries to fire but is blocked underwater?

With waterweapon=false, the zenith fires once (when initially spawned) then never fires again: https://streamable.com/izm9mk

PR #3997, opened 2020-07-16
GBC: Remove another crash source

esainane:
This happened whenever there was an unidentified enemy very, very
close to the location of a soon-to-be build site.

This could only really happen when:
 - you're building something on the frontline
 - you're using GBC to do so
 - you don't have clear los to the build site
 - you do have clear radar coverage of the build site

This meant it was both very rare, and typically happened in moments
where you really don't want to take time out to go bug hunting.

It's even harder to reproduce locally unless you knew what
you were looking for, since global spec control (for positioning
enemy units) was usually paired with globallos, which prevents
this condition from happening.

### DIFF
--- a/LuaUI/Widgets/unit_global_build_command.lua
+++ b/LuaUI/Widgets/unit_global_build_command.lua
@@ -1440,8 +1440,9 @@ function GiveWorkToUnit(unitID)
 			for i=1, #localUnits do -- locate the 'terraunit' if it still exists, and give a repair order for it
 				local target = localUnits[i]
 				local udid = spGetUnitDefID(target)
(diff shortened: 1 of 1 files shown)


sprunk:
Looks good. Maybe put the `GetUnitTeam == myTeam` check in front of the `unitDef.humanName` check instead though? This way `unitDef` won't be indexed for enemies and it's probably a cheaper check than string parsing as well.

esainane:
Ideally it'd save UnitDefNames.terraunit.id to a local once, then check against that every time it's needed, which is how other widgets have done it. Touching GBC is always slightly terrifying, but should be fine here.

PR #3998, opened 2020-07-16
DRAFT: GBC jobs integration

esainane:
Do not merge!

This is done partly as a sanity test and is here so it is not lost.
Hook direction will be reversed later, and everything else will be
made less bad.

The plan is to eventually show allied GBC jobs through a minimal
receiver widget that is much less intrusive than GBC itself.


sprunk:
This PR seems dead (old, conflicts, is draft). I pushed this to upstream `gbc-job-integration` branch so it doesn't get lost though.

Issue #2790, opened 2018-02-16
LuaUI code cleanup


GoogleFrog:
Here is a list of issues that gajop generated https://pastebin.com/FZ20nksq.

gajop:
I suggest you try it with `luacheck` yourself, and run on all relevant areas (+ LuaRules, Chobby, etc.)
Needs some tweaking or it tends to report a lot more false positives

PR #3999, opened 2020-07-16
Style and linter definitions

esainane:
Not comprehensive, but still serves as a decent anti-footgun.

There are a lot of tradeoffs, especially in luacheckrc.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+root = true
+
+[*]
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.lua]
+indent_style = tab
+
+[units/*.lua]
+indent_style = space
+indent_size = 2
+
+[weapons/*.tdf]
+indent_style = space
+indent_size = 2
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,112 @@
+ignore = {
+  "611", -- GF wishes to permit lines consisting entirely of whitespace
+}
+self = false
+unused_args = false
+-- unused = false -- Another false negatives, false positives conundrum
+max_line_length = false
+allow_defined_top = true -- We're going to get a lot of false negatives with this disabled, but a lot of false positives with it enabled. Pick your poison.
+stds.ext = {
+  globals = { "math.round", "math.bit_or", "table.ifind", "table.show", "table.save", "table.echo", "table.print" }
+}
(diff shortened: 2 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.lua]
+indent_style = tab
+
+[units/*.lua]
+indent_style = space
+indent_size = 2
```
sprunk, line 13: ;_;

esainane: I know D: this is definitely here to be descriptive rather than prescriptive, which is why I've had to make different rules for different paths

esainane: I think I can automate style conversion if wanted, though

.luacheckrc
```diff
@@ -0,0 +1,112 @@
...
+}
+stds.customcmds = {
+  globals = {
+    -- Automatically generated from customcmds.h.lua
+    "CMD_SET_FERRY",
+"CMD_RETREAT_ZONE",
+"CMD_SETHAVEN",
+"CMD_RESETFIRE",
+"CMD_RESETMOVE",
+"CMD_BUILDPREV		",
```
sprunk, line 31:
```suggestion
"CMD_BUILDPREV",
```

.luacheckrc
```diff
@@ -0,0 +1,112 @@
...
+stds.widgets = {
+  read_globals = { "LUAUI_DIRNAME", "KEYSYMS", "GL" },
+  globals = { "widget", "widgetHandler", "WG", "gl", "options", "options_path", "options_order" }
+}
+stds.gadgets = {
+  globals = { "gadget", "gadgetHandler", "GG" }
+}
+stds.customcmds = {
+  globals = {
+    -- Automatically generated from customcmds.h.lua
```
sprunk, line 25: Also check `customcmds.lua` (without `.h`), this one seems to be missing some constants (for example I can't see `CMD_GOO_GATHER`).

GoogleFrog:
What is this for @esainane ?

sprunk:
`.editorconfig` is so that text editors automatically apply the correct whitespace (apparently it's becoming a sort of standard to support it, https://editorconfig.org/ )

`.luacheckrc` is settings for running `luacheck` (the tool from #2790) which otherwise produces a ton of false positives.

PR #4001, opened 2020-07-17
Pure disarm/slow damage now decloaks units

### LINKED ISSUE #3932: Phandom does not decloak on disarm damage
Look at 15:40 on this cast: https://youtu.be/CyFENplD170?t=940

My Thunderbird nicked his Phantom and you can see the disarm, but it remained cloaked.

Arch-Shaman:
Fixes #3932:
* Pure disarm/slow damage now is considered damage by the decloak gadget.

### DIFF
--- a/LuaRules/Gadgets/unit_decloak_damaged.lua
+++ b/LuaRules/Gadgets/unit_decloak_damaged.lua
@@ -323,8 +323,8 @@ end
 function gadget:Initialize()
 	local cp
 	for i=1, #WeaponDefs do -- cache pure slow and disarm damage weapons
-		cp = WeaponDefs[i].customParams
-		if cp.disarmdamageonly == '1' or cp.customParams.timeslow_onlyslow == '1' then
+		cp = WeaponDefs[i].customParams or {}
+		if cp.disarmdamageonly == '1' or cp.timeslow_onlyslow == '1' then
 			forceDecloakWeapons[i] = true
 		end
 	end
--- a/LuaRules/Gadgets/unit_decloak_damaged.lua
+++ b/LuaRules/Gadgets/unit_decloak_damaged.lua
@@ -321,14 +321,12 @@ function gadget:UnitCreated(unitID, unitDefID)
 end
 
 function gadget:Initialize()
-	local cp
 	for i=1, #WeaponDefs do -- cache pure slow and disarm damage weapons
-		cp = WeaponDefs[i].customParams or {}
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Access to the WeaponDefs table is slow. The weapons that deal pure slow or disarm damage should be cached when the file is loaded.

Arch-Shaman:
Understood. Will update it momentarily.

LuaRules/Gadgets/unit_decloak_damaged.lua
```diff
@@ -320,6 +321,14 @@ function gadget:UnitCreated(unitID, unitDefID)
 end
 
 function gadget:Initialize()
+	local cp
+	for i=1, #WeaponDefs do -- cache pure slow and disarm damage weapons
+		cp = WeaponDefs[i].customParams or {}
```
sprunk, line 326: customparams is never nil
```suggestion
		cp = WeaponDefs[i].customParams
```

LuaRules/Gadgets/unit_decloak_damaged.lua
```diff
@@ -320,6 +321,14 @@ function gadget:UnitCreated(unitID, unitDefID)
...
+		if cp.disarmdamageonly == '1' or cp.timeslow_onlyslow == '1' then
+			forceDecloakWeapons[i] = true
+		end
+	end
+	cp = nil
```
sprunk, line 331: not how the garbage collector works (`cp` will get garbage collected when it goes out of scope regardless, setting it to `nil` does not make this happen any sooner)
```suggestion
```

LuaRules/Gadgets/unit_decloak_damaged.lua
```diff
@@ -320,6 +321,14 @@ function gadget:UnitCreated(unitID, unitDefID)
 end
 
 function gadget:Initialize()
+	local cp
```
sprunk, line 324: Should be local to the scope it's used in
```suggestion
```

sprunk:
The most ideal solution would be to just filter out things which do zero damage (fake bomber trigger weapon etc) in the gadget handler and drop all the `damage > 0` checks everywhere but that is a larger and more dangerous task.

sprunk:
This seems to be missing capture damage. Possibly there should be some `Spring.Utilities` function which returns weapons that do something despite zero nominal damage (see #4003).

Arch-Shaman:
I'll open a new one with #4003's utilities.

PR #4002, opened 2020-07-17
Chickens can no longer be reclaimed

### LINKED ISSUE #3809: Consider making chicken units unreclaimable
Mostly for fluff reasons, though many of them have 0 metal value anyway

Arch-Shaman:
Fixes #3809
* Sets all chicken unitdefs to `reclaimable = false`

### DIFF
--- a/units/chicken.lua
+++ b/units/chicken.lua
@@ -34,6 +34,7 @@ return { chicken = {
   noChaseCategory     = [[SHIP FLOAT SWIM TERRAFORM FIXEDWING SATELLITE GUNSHIP]],
   objectName          = [[chicken.s3o]],
   power               = 100,
+  reclaimable         = false,
   selfDestructAs      = [[NOWEAPON]],
 
   sfxtypes            = {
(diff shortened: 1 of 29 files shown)


Feugent:
Why was this necessary? This breaks chickenfactory for Zero-Wars.

Arch-Shaman:
An issue was open at the time and it seemed like an easy fix.

sprunk:
How does reclaim affect a factory?

Arch-Shaman:
I'm presuming they mean it has some broken gameplay effect?

Feugent:
It does not affect the fac itself but the ability to change the currently used chicken units in the spawning platform.
Talked with PetTurtle and he said he will fix it in Zero-Wars itself.
Idk why the change to non reclaimable was necessary though.

Arch-Shaman:
> Idk why the change to non reclaimable was necessary though.

> Mostly for fluff reasons, though many of them have 0 metal value anyway

I don't see how reclaiming a 0m unit is anything other than an area reclaim live unit work sink. Plus it looks silly.

PetTurtle:
I believe only the chickens player can reclaim the chickens, but it's currently impossible to be the chickens player.
Also zero-wars fixed chickens having 0 metal value.
```
for _, ud in pairs(UnitDefs) do
    if (ud.unitname:sub(1, 7) == "chicken") then
        ud.buildcostmetal = ud.buildtime
        ud.buildcostenergy = ud.buildtime
    end
end
```
I'm not sure why this is not the case with default chickens, but it does help show the value of the chickens your fighting.

PR #4003, opened 2020-07-17
Disarm damage/slow damage now closes solar

### LINKED ISSUE #3310: Disarm and Capture damage should close Solar
Apparently Racketeer and Thunderbird do not cause Solars to close. They should cause it to close. In general, any damage that is reduced by armour should cause Solar to close.

Arch-Shaman:
Fixes #3310.
* Adds pure disarm and slow to the solar's damage check.


GoogleFrog:
See https://github.com/ZeroK-RTS/Zero-K/pull/4001

scripts/energysolar.lua
```diff
@@ -116,7 +116,7 @@ end
```
sprunk, line 116:
```suggestion
local applyDefensiveManeuverDefs = {}
do
	local utils = Spring.Utilities
	local s = utils.IsPureSlowWeapon
	local c = utils.IsCaptureWeapon
	local d = utils.IsPureDisarmWeapon

	for i = 1, #WeaponDefs do
		applyDefensiveManeuverDefs[i] = s(i) or c(i) or d(i)
	end
end
```

scripts/energysolar.lua
```diff
@@ -116,7 +116,7 @@ end
 
 -- this happens before PreDamaged but only in 97.0+
 function script.HitByWeapon(x, z, weaponDefID, damage)
-	if damage > 0 and not (weaponDefID and noFFWeaponDefs[weaponDefID]) then
+	if (damage > 0 or Spring.Utilities.IsPureSlowWeapon(weaponDefID) or Spring.Utilities.IsCaptureWeapon(weaponDefID) or Spring.Utilities.IsPureDisarmWeapon(weaponDefID)) and not (weaponDefID and noFFWeaponDefs[weaponDefID]) then
```
sprunk, line 119:
```suggestion
	if (damage > 0 or applyDefensiveManeuverDefs[weaponDefID]) and not (weaponDefID and noFFWeaponDefs[weaponDefID]) then
```

GoogleFrog:
If we're being really finicky, then I'd localise `Spring.Utilities.IsPureSlowWeapon` since it could be called a lot. I disagree with Sprung's suggestion of duplicating a table the size of #WeaponDefs for every living Solar. noFFWeaponDefs could also be moved to this new utility file for special weapon types.

Also, `if bla then return true else return false end` can be replaced with `return (bla and true) or false`

sprunk:
> I disagree with Sprung's suggestion of duplicating a table the size of #WeaponDefs for every living Solar.

In that case how about `Spring.Utilities.LegitZeroDamageWeapon` (possibly with a better name) which could also be used for the "status damage causes decloak" #4001 gadget?

Arch-Shaman:
On it.

Arch-Shaman:
The name probably isn't the world's best name, but it was the most descriptive thing I could come up with.

LuaRules/Utilities/SpecialWeaponCaching.lua
```diff
@@ -15,25 +15,17 @@ for i = 1, #WeaponDefs do
 end
 
 function Spring.Utilities.IsPureSlowWeapon(weaponID)
-	if pureSlowWeapons[weaponID] then
-		return true
-	else
-		return false
-	end
+	return (pureSlowWeapons[weaponID] and true) or false
```
sprunk, line 18: how about just
```suggestion
	return pureSlowWeapons[weaponID]
```
(if `false` vs `nil` is a problem then set stuff to false in the table)

LuaRules/Utilities/SpecialWeaponCaching.lua
```diff
@@ -15,25 +15,17 @@ for i = 1, #WeaponDefs do
...
 function Spring.Utilities.IsPureDisarmWeapon(weaponID)
-	if pureDisarmWeapons[weaponID] then
-		return true
-	else
-		return false
-	end
+	return (pureDisarmWeapons[weaponID] and true) or false
 end
+
+function Spring.Utilities.GetWeaponHasStatusEffect(weaponID) -- Single call for checking all three
```
sprunk, line 29: `IsWeaponPureStatusEffect`? Disruptors and the scout boat have status effects so sound like they should be counted but wouldn't as unpure.

LuaRules/Utilities/SpecialWeaponCaching.lua
```diff
@@ -0,0 +1,31 @@
...
+for i = 1, #WeaponDefs do
+	if WeaponDefs[i].customParams.disarmdamageonly == '1' then
+		pureDisarmWeapons[i] = true
+	end
+	if WeaponDefs[i].customParams.timeslow_onlyslow == '1' then
+		pureSlowWeapons[i] = true
+	end
+	if WeaponDefs[i].customParams.is_capture == '1' then
+		captureWeapons[i] = true
+	end
```
sprunk, line 14:
```suggestion
	local cp = WeaponDefs[i].customParams
	pureDisarmWeapons[i] = (cp.disarmdamageonly  == '1')
	pureSlowWeapons  [i] = (cp.timeslow_onlyslow == '1')
	captureWeapons   [i] = (cp.is_capture        == '1')
```

LuaRules/Utilities/SpecialWeaponCaching.lua
Arch-Shaman: That sounds better.

LuaRules/Utilities/SpecialWeaponCaching.lua
Arch-Shaman: 1aae5a2

LuaRules/Utilities/SpecialWeaponCaching.lua
Arch-Shaman: 1aae5a2

Arch-Shaman:
#4001 update is getting pr'd after testing.

PR #4004, opened 2020-07-17
Transports no longer can pick up cargo when disarmed

### LINKED ISSUE #3949: Hercules can still pick up unit when disarmed
http://zero-k.info/Battles/Detail/901986?

about 3-4 mins in firepluk comm nap

Arch-Shaman:
Fixes #3949
* When trying to pick up a passenger, the transport will instantly drop it.

Also transports drop their cargo on death.

### DIFF
--- a/scripts/gunshiptrans.lua
+++ b/scripts/gunshiptrans.lua
@@ -214,7 +214,7 @@ function script.BeginTransport(passengerID)
 		return
 	end
 	Move(link, y_axis, -Spring.GetUnitHeight(passengerID), nil, true)
-
+	
 	if not GG.TransportAllowed(passengerID) then
 		Sleep(10)
 		unitLoaded = passengerID
--- a/scripts/gunshiptrans.lua
+++ b/scripts/gunshiptrans.lua
@@ -210,12 +210,12 @@ function script.MoveRate(curRate)
 end
 
 function script.BeginTransport(passengerID)
+	local disarmed = Spring.GetUnitRulesParam(unitID, "disarmed")
 	if loaded then
 		return
 	end
 	Move(link, y_axis, -Spring.GetUnitHeight(passengerID), nil, true)
-	
-	if not GG.TransportAllowed(passengerID) then
+	if not GG.TransportAllowed(passengerID) or (disarmed and disarmed == 1) then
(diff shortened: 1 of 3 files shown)


sprunk:
Looks a bit hacky, also this ties the code to specific transport scripts instead of being a gamewide rule. How about a gadget?
```lua
local spGetUnitRulesParam = Spring.GetUnitRulesParam
function gadget:AllowUnitTransport(transporterID, transporterUnitDefID, transporterTeam, transporteeID, transporteeUnitDefID, transporteeTeam)
    return (spGetUnitRulesParam(unitID, "disarmed") ~= 1)
end
```

scripts/gunshiptrans.lua
```diff
@@ -210,12 +210,12 @@ function script.MoveRate(curRate)
...
 
 function script.BeginTransport(passengerID)
+	local disarmed = Spring.GetUnitRulesParam(unitID, "disarmed")
 	if loaded then
 		return
 	end
 	Move(link, y_axis, -Spring.GetUnitHeight(passengerID), nil, true)
-	
-	if not GG.TransportAllowed(passengerID) then
+	if not GG.TransportAllowed(passengerID) or (disarmed and disarmed == 1) then
```
sprunk, line 218: Imo put the disarm check inside `GG.TransportAllowed` so as not to have to put it inside all transport scripts (won't be forgotten if we get a new transport, just one place to modify if we want to change the rule, etc)

Arch-Shaman: Alright, will do.

Arch-Shaman: a69fde9

scripts/transports.lua
```diff
@@ -3,6 +3,10 @@ if GG.TransportAllowed then
 end
 
 function GG.TransportAllowed(passengerID)
+	local disarmed = Spring.GetUnitRulesParam(unitID, "disarmed")
+	if (disarmed and disarmed == 1) then
+		return false
+	end
```
sprunk, line 9: operator == works correctly with `nil`, could just do
```suggestion
	if (spGetUnitRulesParam(unitID, "disarmed") == 1) then
		return false
	end
```

PR #4009, opened 2020-07-18
Seperate Cloak field from P. Cloak

### LINKED ISSUE #4005: Increase recloak delay for units without personal cloak
Units seem to decloak after 3.3 seconds. Units without personal cloak could increase this to 5 or 6 seconds. This is a simple alternative to https://github.com/ZeroK-RTS/Zero-K/issues/3972.

Arch-Shaman:
Fixes #4005
* Decloak gadget now has a separate constant for cloakfields (set to 6s).

### DIFF
--- a/LuaRules/Gadgets/unit_decloak_damaged.lua
+++ b/LuaRules/Gadgets/unit_decloak_damaged.lua
@@ -60,6 +60,7 @@ for i = 1, #WeaponDefs do
 	end
 end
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I don't think this takes into account commanders or GG.PokeDecloakUnit.

GoogleFrog:
Fixed.

PR #4013, opened 2020-07-19
DRAFT: Yet Another Player List

esainane:
Do not merge!

This is here to make it easier to request broader feedback,
and so that it is not lost.

This is long-term effort to bring the features of Deluxe to the
performance level of Crude, and perhaps implementing some
long-standing TODOs from Deluxe along the way.

This is directly based on the Crude Player List, and some fixes
may be worth backporting.


sprunk:
Looks stale. Pushed upstream to `yet-another-player-list` branch so it doesn't get lost https://github.com/ZeroK-RTS/Zero-K/tree/yet-another-player-list

Reopen if there's progress.

PR #4023, opened 2020-07-23
Revert "Deluxe Playerlist: Unique window name"

sprunk:
Reverts #3995. Closes #4017.


esainane:
This would be by far the simplest approach for now.

PR #4027, opened 2020-07-28
Fix commander's speed when walking in formation

pengvado:
It previously used the default speed of the commander's chassis, ignoring any changes due to modules.


GoogleFrog:
Accessing the ud table is slow.

PR #4028, opened 2020-07-29
Don't filter restricted projectiles for spectators

esainane:
Spectators can still switch off /specfullview to preserve the selected
player's perspective.

### DIFF
--- a/LuaUI/cache.lua
+++ b/LuaUI/cache.lua
@@ -28,6 +28,7 @@ local spGetProjectileTeamID = Spring.GetProjectileTeamID
 local spGetMyTeamID = Spring.GetMyTeamID
 local spAreTeamsAllied = Spring.AreTeamsAllied
 local spGetProjectileDefID = Spring.GetProjectileDefID
+local spGetSpectatingState = Spring.GetSpectatingState
 local filteredWeaponDefID = {}
 for wdid, wd in pairs(WeaponDefs) do
 	if wd.customParams.restrict_in_widgets then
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Why does anyone need to see restricted projectiles?

esainane:
The old nuke tracker (that constantly does GetAllProjectilesInArea over the entire map) used this to to identify the target area whenever an allied nuke is launched. It would be nice to see where nukes were going as a spectator without first having to work out which team launched it, then select that team.

sprunk:
>Why does anyone need to see restricted projectiles?
Should allies? Currently they can.

I think the proper solution for the specific case of nuke warning would be to just make nukes (and tacnukes?) call some callin (`widget:NukeLaunched` or such) for allies and fullview spectators.

esainane:
If being a spectator with full read isn't an appropriate condition, is there a better guard that could be used?

GoogleFrog:
I just wanted to know what this would be for. As far as I know nothing uses restricted or non-restricted projectiles. If you want to do something with restricted projectiles then a different approach is likely to be much better.

esainane:
I tried making a nuke indicator many months ago, which sort of stalled. In a somewhat recent teams game, I think on either castel gods or sandcastles, an allied nuke ended up killing more value from my Jack assault force than value from the part of the enemy base it was aimed at. This was an experience that was both memorable and motivating enough to try to revive the project.

The bulk of the work is already done in #4030, but from testing the same library with a (slow, expensive) widget variant, there seem to be some cases where the circle is a bit off where it should be. It'd be easier if I could sit back and watch games unfold as a spectator without worrying about missing half the cases.

Having a widget cycle through `/specteam N` in a loop sort of accomplishes this, but makes the game stutter quite a bit, as almost every addon assumes PlayerChanged is still a rare callin. Perhaps the handler could isolate the cases where Spring is invoking PlayerChanged from switching the team being spectated, and redirect those to a different callin which only widgets that care about that case need handle?

... ... okay, I can see why you might be worried that yak shaving is happening.

sprunk:
I think that despite the specific use case being better done via a new event, this PR still makes sense (fullview specs should see everything players do, and also this saves them some perf).

PR #4029, opened 2020-07-29
Optimize previous commit, "Fix commander's speed when in formation"

pengvado:
It didn't occur to me that it might be doing enough work to bother profiling the first time around, but GoogleFrog is right https://github.com/ZeroK-RTS/Zero-K/pull/4027#issuecomment-665376680. My previous commit was unnecessarily slower than the corresponding code before, which could add up to ~1 cpu-millisecond per formation-order for large armies.

### DIFF
--- a/scripts/amphbomb.lua
+++ b/scripts/amphbomb.lua
@@ -124,6 +124,8 @@ local function Killed(recentDamage, maxHealth)
 end
 
 function script.Killed(recentDamage, maxHealth)
+	-- keep the unit technically alive (but hidden) for some time so that the death
+	-- blast can tell who its owner is (so that it can do no damage to allies)
 	return GG.Script.DelayTrueDeath(unitID, unitDefID, recentDamage, maxHealth, Killed, WAVE_TIMEOUT)
 end
 


LuaUI/Widgets/cmd_customformations2.lua
```diff
@@ -442,19 +442,27 @@ local function GiveNotifyingOrderToUnit(uID, cmdID, cmdParams, cmdOpts)
...
 local function SendSetWantedMaxSpeed(alt, ctrl, meta, shift)
 	-- Move Speed (Applicable to every order)
 	local wantedSpeed = 99999 -- High enough to exceed all units speed, but not high enough to cause errors (i.e. vs math.huge)
 	if ctrl then
 		local selUnits = spGetSelectedUnits()
 		for i = 1, #selUnits do
-			local ud = UnitDefs[spGetUnitDefID(selUnits[i])]
+			local unitDefID = spGetUnitDefID(selUnits[i])
+			local ud = UnitDefs[unitDefID]
 			local uSpeed = ud and ud.speed
```
sprunk, line 458: Maybe also cache `uSpeed` while you're at it?

PR #4030, opened 2020-07-29
DRAFT: Allied missile warning indicator

esainane:
Provides accurate readings to allies, with an outer circle that gradually
closes and fades in on the area of effect as the time to impact approaches.

Could probably do with icons to draw in the middle of each circle,
so people more easily understand what it represents.


LuaRules/Gadgets/unit_nukewarn.lua
```diff
@@ -0,0 +1,140 @@
...
+
+local mapSizeX, mapSizeZ = Game.mapSizeX, Game.mapSizeZ
+
+-- These should probably be modoptions
+local doGlobalGiveFuzzyLocation = true
+-- Can be up to five times this, but biased towards being closer to the real point.
+-- Seems to be 2-3 times this far off on average.
+local fuzziness = max(1200,1500*7200/max(mapSizeX, mapSizeZ)) -- Wide enough to make it impossible to rule anything out, accurate enough to eventually be used in ~~rage-inducing~~ hilarious misreads
+
+-- NB: This *does* reveal the number of nukes launched, compare to unit_gamerules_events
```
sprunk, line 42: sounds like a nope since the lack of counting is by design

LuaRules/Utilities/StarburstPredict.lua
```diff
@@ -0,0 +1,181 @@
...
+		return dx/dist,dy/dist,dz/dist,dist
+	end
+	return dx,dy,dz,dist
+end
+local function Normalize(dx,dy,dz,dist)
+	if not dist then dist = sqrt(dx*dx+dy*dy+dz*dz) end
+	return dx/dist,dy/dist,dz/dist,dist
+end
+
+function StarburstPredict(projectileID, weaponDefID, curFrame, targetPosition)
```
sprunk, line 88: Pollutes the global namespace. Either
 * `function Spring.Utilities.StarburstPredict(...)` or
 * `local function StarburstPredict(...)` and `return StarburstPredict` at the end, or
 * put somewhere else other than `LuaRules/Utilities`

LuaRules/Gadgets/unit_nukewarn.lua
```diff
@@ -0,0 +1,130 @@
...
+		layer = 0,
+		enabled = true
+	}
+end
+
+--
+-- Common code
+--
+-- Protocol numbers and which missiles to handle are used in both synced and unsynced
+local MAGIC_CREATED   = 0
```
sprunk, line 18: I think `SendToUnsynced` is sent to all gadgets so this is a fairly poor magic. Is `ProjectileCreated` still unavailable in unsynced?

esainane: SetWatchWeapon cannot be called from unsynced:

`attempt to call field 'SetWatchWeapon' (a nil value)`

Calling SetWatchWeapon from synced, and just having the callins in unsynced, does not see the unsynced callins be called.

Adding a gadget consisting entirely of the following:

```
if gadgetHandler:IsSyncedCode() then return end

function gadget:GetInfo()
	return {
		name = "Unrelated gadget",
		desc = "Gadget completely unrelated to anyone or anything.",
		author = "esainane",
		date = "2020-07",
		license = "PD",
		layer = 0,
		enabled = true
	}
end

function gadget:RecvFromSynced(...)
  Spring.Echo('FINDME Unrelated gadget RecvFromSynced', ...)
end
```

...and firing an Eos which triggers enough synced<->unsynced communication for the impact point to appear doesn't have `FINDME` show up in infolog or console output. I'll take a look through the spring source to be sure, though.

sprunk: Check with `layer = -math.huge`, IIRC it's a form of lazy evaluation and the first match eats the event.

esainane: You are right: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/gadgets.lua#L2189

Magic strings it is, then!

LuaRules/Gadgets/unit_missilewarn.lua
```diff
@@ -0,0 +1,133 @@
...
+local trackedMissiles = include("LuaRules/Configs/tracked_missiles.lua")
+
+
+if gadgetHandler:IsSyncedCode() then
+--
+-- Synced code, only here to forward to unsynced
+--
+function gadget:Initialize()
+	for weaponDefID in pairs(trackedMissiles) do
+		Script.SetWatchWeapon(weaponDefID, true)
```
GoogleFrog, line 30: Don't use SetWatchWeapon, use the specialised version for the callins you want.

esainane: Done!

GoogleFrog:
@esainane what is left for the draft?

sprunk:
Looks obsoleted by #4933

PR #4035, opened 2020-08-05
Fix the Auto Reclaim/Heal/Assist widget

pengvado:
If you queued a manual command after an auto-generated one, it would forget that manual command.


GoogleFrog:
I don't know how to find dyth68 anywhere. I will just merge this, but have done no testing.

PR #4038, opened 2020-08-09
replace koda fire with gun

Anarchid:
I find that the current version of Kodachi flame weapon looks and feels awful, and that the fire mechanic is entirely superfluous to it given that the fire lasts for less than a second. 

This  PR replaces the kodachi weird owled-fire ball spam with an autocannon, mainly for aesthetic purposes. It also removes Kodachi's fire immunity.

I have not compensatedfor  the loss of fire dps in any way, but 4 Kodachi still seem fully able to kill 6 Scorcher with this.

### DIFF
--- a/scripts/tankraid.lua
+++ b/scripts/tankraid.lua
@@ -219,6 +219,7 @@ function script.Shot(num)
 	EmitSfx(firepoint, GG.Script.FIRE_W2)
 	Turn(firepoint, y_axis, 0)
 	--]]
+	EmitSfx(firepoint, 1024)
 	StartThread(Recoil)
 end
 
--- a/units/tankraid.lua
+++ b/units/tankraid.lua
@@ -20,7 +20,6 @@ return { tankraid = {
   corpse              = [[DEAD]],
 
   customParams        = {
-    fireproof      = [[1]],
     specialreloadtime = [[850]],
     modelradius       = [[20]],
     aimposoffset      = [[0 5 0]],
@@ -46,6 +45,15 @@ return { tankraid = {
   objectName          = [[logkoda.s3o]],
   script              = [[tankraid.lua]],
   selfDestructAs      = [[BIG_UNITEX]],
+
+  sfxtypes               = {
+    explosiongenerators = {
+      [[custom:WARMUZZLE]],
+      [[custom:emg_shells_l]],
+    },
+
+  },
+
   sightDistance       = 600,
   trackOffset         = 6,
   trackStrength       = 5,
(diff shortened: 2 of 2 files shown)


units/tankraid.lua
```diff
@@ -76,29 +84,21 @@ return { tankraid = {
...
-        area_damage_radius = 54,
-        area_damage_dps = 35,
-        area_damage_plateau_radius = 20,
-        area_damage_duration = 1.5,
-        
-        light_color = [[1.6 0.8 0.32]],
-        light_radius = 320,
+      customParams            = {
+        muzzleEffectShot = [[custom:WARMUZZLE]],
+        miscEffectShot = [[custom:DEVA_SHELLS]],
```
sprunk, line 98: Aren't those 2 just meaningful for dyncomms?

scripts/tankraid.lua
```diff
@@ -219,6 +219,7 @@ function script.Shot(num)
 	EmitSfx(firepoint, GG.Script.FIRE_W2)
 	Turn(firepoint, y_axis, 0)
 	--]]
+	EmitSfx(firepoint, 1024)
```
sprunk, line 222: what about `1025` for shells? add the shell or remove from sfxtypes

Anarchid: I'm not sure it has the emit point for shells, but that was the intention.

units/tankraid.lua
Anarchid: fair point

GoogleFrog:
I am unconvinced. You've made a speed-Reaver that files single shots instead of bursts. In this PR you have:
 - Doubled weapon velocity, essentially removing the deliberate range advantage on retreat.
 - Relatively nerfed Kodachi vs static units by removing ground fire. Test its DPS vs statics such as Mex and LLT before and after.

You have made no comment on the first and have dismissed the second without any testing.

GoogleFrog:
I've tested, tell me why I shouldn't close this PR just for it being based on a completely false premise.

> the fire mechanic is entirely superfluous to it given that the fire lasts for less than a second.

Currently Kodachi kills mexes in eight shots. The Kodachi in this PR kills mexes in 16 shots. The fire damage is half its damage against small static structures, this is far from 'entirely superfluous'!

The PR Kodachi has 100 DPS. The current Kodachi has 152 DPS over the course of killing an LLT and 114 over the course of killing a Cloakbot Factory.

The fire makes up much less of its damage output against quick raiders _by design_. It is meant to be a raider that wants to run by mexes rather than engage enemy raiders. It still requires tweaks to put it in a healthy spot within this role, but this PR essentially removes the chance to make it anything but an anti-raider raider.

Licho1:
Reversion to average.. don't do this.
The fire thing makes it unique.

Anarchid:
> I've tested, tell me why I shouldn't close this PR just for it being based on a completely false premise.

My main concern is that the firebomb spam looks and feels wrong, rather than the balance implications. If fire is absolutely necessary, i'll try to find a way to make the fire less ugly.

PR #4039, opened 2020-08-10
Chili Scrollpanel RTT 

esainane:


esainane:
Anecdata: Spectating a very large pot toward the end of the game, Nvidia-smi suggests VRAM usage goes from about 1.5G to about 2.2G when switching from no RTT to RTT by default. Framerate typically boosts by 50%.

GoogleFrog:
Two notes:
 - Spectating is a poor test because you are not frequently changing the UI.
 - The RTT disabled behaviour needs to remain functional because chobby uses latest chili.

PR #4040, opened 2020-08-13
Let the music live

### LINKED ISSUE #3883: Music widget options disappear at 0% volume
0% volume actually causes the music widget to get disabled which also takes down its options. The "use included tracks" specifically was reported missing, I haven't checked but presumably the "pause music alongside game" also goes away.

Arch-Shaman:
* Removes all calls to disable music player. 

Fixes #3883

### DIFF
--- a/LuaUI/Configs/epicmenu_conf.lua
+++ b/LuaUI/Configs/epicmenu_conf.lua
@@ -654,8 +654,8 @@ local pathAudio = 'Settings/Audio'
 			local prevValue = WG.music_volume
 			--settings.music_volume = self.value
 			WG.music_volume = self.value
-			if (prevValue > 0 and self.value <=0) then widgetHandler:DisableWidget("Music Player") end
-			if (prevValue <=0 and self.value > 0) then widgetHandler:EnableWidget("Music Player") end
+			--if (prevValue > 0 and self.value <=0) then widgetHandler:DisableWidget("Music Player") end
+			--if (prevValue <=0 and self.value > 0) then widgetHandler:EnableWidget("Music Player") end
 		end,
 		simpleMode = true,
 		everyMode = true,
--- a/LuaUI/Configs/epicmenu_conf.lua
+++ b/LuaUI/Configs/epicmenu_conf.lua
@@ -654,8 +654,6 @@ local pathAudio = 'Settings/Audio'
 			local prevValue = WG.music_volume
 			--settings.music_volume = self.value
 			WG.music_volume = self.value
-			--if (prevValue > 0 and self.value <=0) then widgetHandler:DisableWidget("Music Player") end
-			--if (prevValue <=0 and self.value > 0) then widgetHandler:EnableWidget("Music Player") end
 		end,
 		simpleMode = true,
 		everyMode = true,
--- a/LuaUI/Widgets/gui_epicmenu.lua
+++ b/LuaUI/Widgets/gui_epicmenu.lua
@@ -2320,16 +2320,16 @@ local function GetMainPanel(parent, width, height)
 							end
 							settings.music_volume = self.value
 							WG.music_volume = self.value
-							if (self.prevValue > 0 and self.value <= 0) then
-								widgetHandler:DisableWidget("Music Player")
-							end
-							if (self.prevValue <= 0 and self.value > 0) then
+							--if (self.prevValue > 0 and self.value <= 0) then
+								--widgetHandler:DisableWidget("Music Player")
+							--end
+							--if (self.prevValue <= 0 and self.value > 0) then
 								-- Disable first in case widget is already enabled.
 								-- This is required for it to notice the volume
 								-- change from 0 in some cases.
-								widgetHandler:DisableWidget("Music Player")
-								widgetHandler:EnableWidget("Music Player")
-							end
+								--widgetHandler:DisableWidget("Music Player")
+								--widgetHandler:EnableWidget("Music Player")
+							--end
 							self.prevValue = self.value
 						end
 					},
--- a/LuaUI/Widgets/gui_epicmenu.lua
+++ b/LuaUI/Widgets/gui_epicmenu.lua
@@ -2320,16 +2320,6 @@ local function GetMainPanel(parent, width, height)
 							end
 							settings.music_volume = self.value
 							WG.music_volume = self.value
-							--if (self.prevValue > 0 and self.value <= 0) then
-								--widgetHandler:DisableWidget("Music Player")
-							--end
-							--if (self.prevValue <= 0 and self.value > 0) then
-								-- Disable first in case widget is already enabled.
-								-- This is required for it to notice the volume
-								-- change from 0 in some cases.
-								--widgetHandler:DisableWidget("Music Player")
-								--widgetHandler:EnableWidget("Music Player")
-							--end
 							self.prevValue = self.value
 						end
 					},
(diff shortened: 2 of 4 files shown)


Arch-Shaman:
I noticed that the music player now only "mutes" the music and you may sometimes hear silence for a few moments. Seems fine though as it does immediately pick up back where it left off on mute. Is this desirable behavior?

GoogleFrog:
You have removed special handling for zero volume without adding any new handling. I dislike having a widget play sounds at zero volume. Does doing so break anything?

Arch-Shaman:
It doesn't seem to break anything. I've been playing some test games and nothing unexpected happens. I'll add removal of the music stream at 0% volume real quick.

LuaUI/Widgets/snd_music.lua
```diff
@@ -310,18 +303,19 @@ function widget:GameStart()
...
+		end
 	end
 	
 	--Spring.Echo("Track: " .. newTrack)
-	newTrackWait = 0
 end
 
 -- Safety of a heisenbug
-function widget:GameFrame()
+--[[function widget:GameFrame() -- this looks safe to remove. In testing removing this changed nothing
```
sprunk, line 315: see the comment above the func

LuaUI/Widgets/snd_music.lua
```diff
@@ -310,18 +303,19 @@ function widget:GameStart()
 		gameStarted = true
 		previousTrackType = musicType
 		musicType = "peace"
-		StartTrack()
+		if select(1,Spring.GetSoundStreamTime()) > 0 then -- if there's a briefing track playing, stop it and start peace track.
```
sprunk, line 306: select(1, x) probably isnt needed here

LuaUI/Widgets/snd_music.lua
```diff
@@ -288,19 +271,29 @@ function widget:Update(dt)
...
+		if not musicPaused and totalTime > 0 and paused and options.pausemusic.value then -- game got paused with the pausemusic option enabled, so pause the music stream.
 			Spring.PauseSoundStream()
-			wasPaused = paused
+			musicPaused = true
+		end
+		if musicPaused and paused and not options.pausemusic.value then -- user disabled pausemusic option, unpause the music.
+			Spring.PauseSoundStream()
+			musicPaused = false
+		end
+		if musicPaused and paused == false then -- resume playing the current track
```
sprunk, line 294: two ifs could be combined: `musicPaused and (not paused or not options.pausemusic.value)`

LuaUI/Widgets/snd_music.lua
```diff
@@ -288,19 +271,29 @@ function widget:Update(dt)
...
+			Spring.PauseSoundStream()
+			musicPaused = false
```
sprunk, line 296: shouldn't this bool be `true`? if you call `PauseSoundStream` then music is paused

sprunk, line 296: `musicPaused` being `false` after calling `PauseSoundStream` sounds like either the value should be `true` or the variable is misnamed

LuaUI/Widgets/snd_music.lua
```diff
@@ -185,9 +170,9 @@ local function SetPeaceThreshold(num)
 end
 
 function widget:Update(dt)
-	if gameOver then
+	--[[if gameOver then -- gameOver is never actually changed within the code and can be safely removed.
```
sprunk, line 173: the intent looks still valid; consider replacing this with `widgetHandler:RemoveCallIn('Update')` in `widget:GameOver`

LuaUI/Widgets/snd_music.lua
Arch-Shaman: This is because PauseSoundStream actually toggles whether or not the soundstream is paused. There is no UnpauseSoundStream.

LuaUI/Widgets/snd_music.lua
Arch-Shaman: I've updated the comment with additional info then.

LuaUI/Widgets/snd_music.lua
Arch-Shaman: Good catch.
What about `musicPaused and not (paused or options.pausemusic.value)` to consolidate both nots?

LuaUI/Widgets/snd_music.lua
Arch-Shaman: c746694

LuaUI/Widgets/snd_music.lua
Arch-Shaman: c746694

LuaUI/Widgets/snd_music.lua
Arch-Shaman: c746694

Arch-Shaman:
"Complaint" is probably the wrong word. I just lack the correct term.

LuaUI/Widgets/snd_music.lua
Arch-Shaman: I've changed it with my idea. How does it look now?

LuaUI/Configs/epicmenu_conf.lua
```diff
@@ -654,8 +654,8 @@ local pathAudio = 'Settings/Audio'
 			local prevValue = WG.music_volume
 			--settings.music_volume = self.value
 			WG.music_volume = self.value
-			if (prevValue > 0 and self.value <=0) then widgetHandler:DisableWidget("Music Player") end
-			if (prevValue <=0 and self.value > 0) then widgetHandler:EnableWidget("Music Player") end
+			--if (prevValue > 0 and self.value <=0) then widgetHandler:DisableWidget("Music Player") end
```
GoogleFrog, line 657: The nonsense of this file is so dense that I would rather delete unused lines then retain them.

LuaUI/Widgets/gui_epicmenu.lua
```diff
@@ -2320,16 +2320,16 @@ local function GetMainPanel(parent, width, height)
 							end
 							settings.music_volume = self.value
 							WG.music_volume = self.value
-							if (self.prevValue > 0 and self.value <= 0) then
-								widgetHandler:DisableWidget("Music Player")
-							end
-							if (self.prevValue <= 0 and self.value > 0) then
+							--if (self.prevValue > 0 and self.value <= 0) then
```
GoogleFrog, line 2323: Same as above. The nonsense of this file is so dense that I would rather delete unused lines then retain them.

Arch-Shaman: One trimming coming up then.

Arch-Shaman: 808bfcc

Arch-Shaman: better now after 808bfcc ?

GoogleFrog: My comments were on epicmenu and its configs.

Arch-Shaman: ah. whoops.

Arch-Shaman: bf36395

sprunk:
What about the people who currently have volume set to 0%? Without the code that re-enables the music widget when increasing to 1%+ they will be stuck with it disabled.

Arch-Shaman:
Alwaysstart = true

GoogleFrog:
Rename the widget to reset its defaults.

PR #4041, opened 2020-08-13
Sonic Archer

palimpsestate:


GoogleFrog:
Is the fake sonic really required? It has the same parameters as the real sonic and the real sonic is already blocked from firing while underwater in the LUS.

I see that the sonic weapon is a Cannon, but I don't see any runtime modifications to its range to take the height bonus of floating into account.

GoogleFrog:
Why isn't weaponVelocity = 700 to match Siren?  I prefer to aim for consistency within weapon types, especially if the weapon type is rare. Using weaponVelocity = 300 opens up Archer to Kodachi-levels of prefire, which seems bad for a floating unit.

palimpsestate:
The weapon velocity was set to 300 specifically to match with the speed of the scallop weapon, as I wanted to start with a reasonably close facsimile of it.  It does make the projectile a bit dodgeable.  I don't have any reason to specifically stick with 300 though and have no moral objection to 700 velocity.  I did do some testing vs ducks and didn't have that specific problem, but it's true that I was having results where it would be difficult to defeat large numbers of ducks.

The fake sonic weapon was used because I could not figure out how to get archer to not fire underwater without it.  I saw there were a few other ways that it looked like preventing uw firing from working were done, but the original archer also had a fake weapon so it seemed usable for the job.

Anarchid:
> I see that the sonic weapon is a Cannon, but I don't see any runtime modifications to its range to take the height bonus of floating into account.

Conceptually sonic weapons feel somewhat different from cannons in ZK (the game) even if they're Cannons in Spring Engine (and initially Sonic Blaster used to be a LaserCannon).

Afaict setting [heightBoostFactor](https://springrts.com/wiki/Gamedev:WeaponDefs#heightBoostFactor) = 0 in the weapondef will make zero-gravity cannons behave like a LaserCannon for targetting purposes.

palimpsestate:
Hmm, I'll try playing around with that and see if I like it better.  The sonic weapon initially was a direct ripoff of the siren sonic cannon, so it didn't think making too many changes to the way that the weapon worked would be a good idea.  Sonic weapon as a kind of laser kind of makes sense and is at least worth trying.

Anarchid:
FWIW, Siren, as well as all Gauss units, could probably also benefit from this. This sounds like a thing that could do away with that annoying gauss overshoot bug.

sprunk:
I think Siren shouldn't be a laser cannon because sea is 3D and laser cannon is spherical so it would be difficult to tell how much actual range you have against ocean floor targets (and in deep enough sea, stuff directly below you could be out of range). Cannon also gains a bit of range but this is both less pronounced at default settings and more controllable. Ditto for gauss guns, especially due to Dagger.

If cylindrical lasercannon exists then I'll agree it would be superior, but ATM it does not (see https://springrts.com/mantis/view.php?id=6342 ).

palimpsestate:
I'm currently using the laser-style weapon Anarchid suggested.  It's hard to notice too much difference in most of the testing I've been doing so I can't say whether I've grown attached to it or not.  

Currently I feel that the performance is going mostly in an intended direction... e.g., archer is handling daggers, ducks, and sea wolves to a reasonable capacity.  They are a bit more awkward on land, they tend to not do much better than cost vs things like bandits or scorchers but beat glaives pretty handily.  They do not trade evenly against other riots on land or in sea.  This is mostly all okay with me conceptually, as the concept is for scallop to fill those roles that archer can't.  And it gives the amph bots a more real riot/raider that can attack into single lotus (not cost effectively still but at least it doesn't die) and can take on urchins quite nicely.

palimpsestate:
With the last change I think the new archer is probably in a good spot to have others give it a go.  I do not plan to change anything else without further feedback at this point.

GoogleFrog:
Archer cannot use heightBoostFactor because Siren does not use heightBoostFactor. Change Siren if you really want to use heightBoostFactor, the weapon type only exists on one other unit so far so its behaviour is not set in stone. What I don't want is weapon types becoming inconsistent for no reason. If the Archer weapon is to be a sonic gun then it should inherit the properties of a sonic gun, with variation to AoE, range and damage shown by modifications to the size or look of the projectile/explosion.

I think sonic gun should use the default Cannon range increase because it provides bonus range against underwater units. This seems desirable as units that use sonic guns will tend to be designed to be good against underwater targets.

I'll have time to take a closer look tomorrow.

> FWIW, Siren, as well as all Gauss units, could probably also benefit from this. This sounds like a thing that could do away with that annoying gauss overshoot bug.

This was solved a while ago. Gauss aims and fires spherically now.

palimpsestate:
I don't have a problem with that and will go ahead and remove the heightboostfactor.  However, if we are talking about inconsistency, the Siren's weapon is also a cannon and I directly ripped the archer's weapon from the siren before tweaking aoe, damage and effect sizes.  If I changed the archer's projectile to get a range boost vs underwater units then wouldn't you need the siren to also get that?

GoogleFrog:
heightBoostFactor defaults to -1, so no.

GoogleFrog:
image

image

This PR seems to break all cegs.

palimpsestate:
Uh oh, there I go accidentally deleting braces again.

PR #4046, opened 2020-08-16
Unit Marker Improvements

Arch-Shaman:
Adds a delay (5s) between each unit's ping. This fixes the ping spam on los edge.
Removes pings after 10s.

### DIFF
--- a/LuaUI/Widgets/unit_marker.lua
+++ b/LuaUI/Widgets/unit_marker.lua
@@ -1,6 +1,6 @@
 function widget:GetInfo() return {
 	name    = "Unit Marker Zero-K",
-	desc    = "[v1.3.10] Marks spotted buildings of interest and commander corpse.",
+	desc    = "[v1.3.11] Marks spotted buildings of interest and commander corpse.",
 	author  = "Sprung",
 	date    = "2015-04-11",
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Won't the removal remove other markers and make gaps in drawings?

Technically the 5s delay is unitID hax and it will still spam for armies of spotted units. How about a distance based frequency limit?

Arch-Shaman:
Yes it will unfortunately remove placer markers and screw with drawings. However the benefit is that we can de-clutter a mass of labels when big scythe or Athena armies uncloak. I think this benefit outweighs the drawback, though if there was a way to remove local pings that would be great.

Mostly this is an improvement in big team games where mass decloaks aren't uncommon.

I take it you want to store the uncloak position and frame of this position then check on decloak how far it was from it? Perhaps an ordered table of coordinates followed by a ping like " 12 scythes spotted" on next frame?

GoogleFrog:
Grouping things together sounds ideal. I primarily want to avoid adding unitID hax - tracking information about hidden units.

sprunk:
Looks stale, is anything going to happen?

GoogleFrog:
It looks ok but is still a draft.

PR #4045, opened 2020-08-16
WIP: Commshare maintenance

Arch-Shaman:
Fixes #4025 (all three parts):
* Ctrl-A Ctrl-D resign immunity to squads (would force resign them)
* Fixes bug where merging, disconnecting then inviting your old squad would give nanoframes,etc to the old team.

Fixes #4026 
* See above

Improvements on #3884:
* Hopefully fixed unit given indicator

Fixes #3764 (duplicate of #4026)
Fixes #1853
* Adds support for commshare.

TODO:
- [ ] Small scale testing.
- [ ] Additional bug fixes?

PR #4047, opened 2020-08-17
License Update

Arch-Shaman:
All Shaman code gets PD license instead of CC-0 as per https://github.com/ZeroK-RTS/Zero-K/pull/4045#discussion_r471228670


sprunk (approved):
Looks good to me but I'll leave the merge to GF since he seems to be more competent at license-adjacent matters.

PR #4051, opened 2020-08-18
Submarine particle effects

Anarchid:
Submarines:
- Smoke when damaged and surfaced; sizzle when damaged and submerged
- Generate wakes when surfaced
- No longer do a noop HitByWapon

### DIFF
--- a/scripts/subraider.lua
+++ b/scripts/subraider.lua
@@ -3,6 +3,22 @@
 local wake = piece "wake"
 local base = piece "base"
 local firepoint = piece "firepoint"
+local smokePiece = {base}
+local moving = false
+local criticalHeight = 0
+
+local function MoveScript()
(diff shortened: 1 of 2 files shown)


GoogleFrog:
I don't want underwater sizzle because it can give away their location.

Anarchid:
True but unsure how consequential. 

The sizzle effect is very small at submarine depths, and i doubt non-cloaked sub stealth is very important in general as a special rule that only affects Seawolf and Scylla particle generation.

I can make it not do the sizzle, of course.

sprunk:
Wakes and smoke are obsoleted by the change that made subs not surface, and GF doesn't want sizzling.

That leaves just the no-op HitByWeapon removal but removing that special damage should be more thorough. See #4339

PR #4052, opened 2020-08-18
Softban AutoLLT

Arch-Shaman:
Auto LLT or disco lotus is a troublesome public widget that can end up making games literally unplayable with 3+ sec latency after even just a handful of lotuses. This pr seeks to softban it (see requiring users to load it twice) to help discourage casual users. It comes with an authorative warning that use may result in bans (due to its antisocial network use). A better version of the widget could be publicized that would be unaffected by this ban.

### DIFF
--- a/LuaUI/Widgets/api_widget_disabler.lua
+++ b/LuaUI/Widgets/api_widget_disabler.lua
@@ -26,11 +26,13 @@ local spEcho = Spring.Echo
 local badwidgets = {
 	[1] = "*Mearth Location Tags*1.0", -- Map: Mearth_v4.
 	[2] = "Metalspot Finder (map)", -- oktogon v3
+	[3] = "Auto LLT", -- troublesome public widget
 }
 
 local reason = {
 	[1] = "Causes black ground on some graphics cards, possible copyright issues.",
 	[2] = "breaks everything mex-related",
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Does the reason actually occur in chat? The text probably needs bikesheding as you're expecting moderators to enforce what it says.

Arch-Shaman:
I can make it print it in chat. Currently it dumps it unhelpfully into the info log. Which is bad news bears for would be users

Having games where _everyone_ is catching up falls safely under "modactionable" though ban might be too strong as afaik mods have local widget ban punishments.  Most of the time its simply ignorance of how network hostile autollt is and most people arent bad faith acting with it. 

Im open to suggestions on text to be displayed, as always.

Anarchid:
I don't know of any punishment effect that suppresses local widgets.

Arch-Shaman:
Iirc it was part of the update that added disable local widgets.

Anarchid:
It doesn't seem to exist at the moment.

GoogleFrog:
There is no punishment effect that blocks local widgets. If there is part of such an effect (which iirc there isn't) then it only exists as game side support. What does exist is a modoption that disables local widgets.

Arch-Shaman:
Idea: instead of softbanning, why not make terve886's sweep fire / autollt widgets into a gadget and completely kill the widget that way?

sprunk:
1) doesn't completely kill the widget because people may still think their version is somehow superior (or they just forget to disable)
2) I hear disco llt is actually pretty crappy at its supposed job of decloaking and I'm not sure it's worth increasing the state icon clutter for everybody (and also having everybody pay the perf cost, forever) just to work around some widget.

PR #4056, opened 2020-08-22
GBC Refactoring

esainane:
More to follow.

### DIFF
--- a/LuaUI/Widgets/gbc/Pathfinding.lua
+++ b/LuaUI/Widgets/gbc/Pathfinding.lua
@@ -1,0 +1,161 @@
+
+-- luacheck: read globals WG
+-- provides: read globals UpdateOneWorkerPathing UpdateOneJobPathing CleanPathing
+
+-- Global Build Command/Pathfinding: Responsible keeping track of which units
+-- can reach which jobs and vice versa.
+-- Gives a lot of false positives: Spring.RequestPath currently seems overly
+-- generous, and will often provide a path if it thinks the unit can get
+-- closer without necessarily being able to reach the target.
+-- XXX: State is currently still tightly coupled with core GBC state.
+
+local Echo               = Spring.Echo
+local spGetUnitDefID     = Spring.GetUnitDefID
+local spGetUnitPosition  = Spring.GetUnitPosition
+local spIsUnitAllied     = Spring.IsUnitAllied
+local spValidUnitID      = Spring.ValidUnitID
+local spUnitIsDead       = Spring.GetUnitIsDead
+local spRequestPath      = Spring.RequestPath
+
+local CMD_REPAIR         = CMD.REPAIR
+local CMD_RECLAIM        = CMD.RECLAIM
+
+local abs                = math.abs
+local sqrt               = math.sqrt
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+-- Helper Functions ------------------------------------------------------------
+--[[
+HOW THIS WORKS:
+	-- EXTERNAL API --
+	UpdateOneWorkerPathing()
+		Caches pathing info for one worker for every job in buildQueue. This is called
+		whenever a new unit enters our group, and adds the hash of any job that cannot be reached
+		and/or performed to 'includedBuilders[unitID].unreachable'. Does not do anything for commands targetting
+		units (ie repair, reclaim), since they may move and invalidate cached pathing info.
+	UpdateOneJobPathing()
+		Caches pathing info for one job for every worker in includedBuilders. This is called whenever a new
+		job is added to the queue, and does basically the same thing as UpdateOneWorkerPathing().
+	CleanPathing()
+		Performs garbage collection for unreachable caches, removing jobs that are no longer on the queue.
+	-- INTERNAL FUNCTIONS --
+	CanBuildThis()
+		Determines whether a given worker can perform a given job or not.
+	IsTargetReachable()
+		Checks pathing between a unit and destination, to determine if a worker
+		can reach a given build site.
+--]]
+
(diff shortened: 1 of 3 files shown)


LuaUI/Widgets/gbc/Pathfinding.lua
```diff
@@ -0,0 +1,161 @@
...
+local function CanBuildThis(cmdID, unitID)
+	local unitDefID = spGetUnitDefID(unitID)
+	local unitDef = UnitDefs[unitDefID]
+	if cmdID < 0 then -- for build jobs
+		local bcmd = abs(cmdID) -- abs the command ID to get the unitDefID that it refers to
+		for _, options in ipairs(unitDef.buildOptions) do
+			if ( options == bcmd ) then -- check whether our unit can build it
+				return true
+			end
+		end
```
sprunk, line 60: `ipairs` is slower (and probably a bit more opaque to ~~lua noobs~~ aspiring modders)
```suggestion
		local bo = unitDef.buildOptions
		for i = 1, #bo do
			if bo[i] == bcmd then
				return true
			end
		end
```

LuaUI/Widgets/gbc/Pathfinding.lua
```diff
@@ -0,0 +1,161 @@
...
+		Determines whether a given worker can perform a given job or not.
+	IsTargetReachable()
+		Checks pathing between a unit and destination, to determine if a worker
+		can reach a given build site.
+--]]
+
+--This function tells us if a unit can perform the job in question.
+local function CanBuildThis(cmdID, unitID)
+	local unitDefID = spGetUnitDefID(unitID)
+	local unitDef = UnitDefs[unitDefID]
```
sprunk, line 53: `UnitDefs` access is comparatively slow, maybe could be cached? Not necessarily the whole `buildOptions`, possibly just
```lua
{
  bo = def.buildOptions,
  canRez = def.canResurrect,
}
```

LuaUI/Widgets/gbc/Pathfinding.lua
```diff
@@ -0,0 +1,161 @@
...
+	if cmdID < 0 then -- for build jobs
+		local bcmd = abs(cmdID) -- abs the command ID to get the unitDefID that it refers to
```
sprunk, line 55: We know `< 0` so we can use the cheaper operator instead of invoking a function
```suggestion
		local bcmd = -cmdID
```

LuaUI/Widgets/gbc/Pathfinding.lua
```diff
@@ -0,0 +1,161 @@
...
+	local moveID = UnitDefs[unitDefID].moveDef.id -- unit pathing type
+	if moveID then -- air units have no moveID, and we don't need to calculate pathing for them.
+		local path = spRequestPath( moveID,ox,oy,oz,tx,ty,tz, 10)
+		if path then
+			local waypoints = path:GetPathWayPoints()
+			local finalCoord = waypoints[#waypoints]
+			if finalCoord then -- unknown why sometimes NIL
+				local dx, dz = finalCoord[1]-tx, finalCoord[3]-tz
+				local dist = sqrt(dx*dx + dz*dz)
+				if dist < buildDist + 40 then -- is within radius?
```
sprunk, line 84: Magic 40

LuaUI/Widgets/gbc/Pathfinding.lua
```diff
@@ -0,0 +1,161 @@
...
+-- This function process result of Spring.PathRequest() to say whether target is reachable or not
+local function IsTargetReachable(unitID, tx,ty,tz)
+	local ox, oy, oz = spGetUnitPosition(unitID)  -- unit location
+	local unitDefID = spGetUnitDefID(unitID)
+	local buildDist = UnitDefs[unitDefID].buildDistance -- build range
+	local moveID = UnitDefs[unitDefID].moveDef.id -- unit pathing type
+	if moveID then -- air units have no moveID, and we don't need to calculate pathing for them.
+		local path = spRequestPath( moveID,ox,oy,oz,tx,ty,tz, 10)
+		if path then
+			local waypoints = path:GetPathWayPoints()
```
sprunk, line 79: Pyramid code, how about early returns?
```lua
	if not moveID then -- air units have no moveID, and we don't need to calculate pathing for them.
		return true
	end

	local path = spRequestPath( moveID,ox,oy,oz,tx,ty,tz, 10)
	if not path then
		return true
	end

	local waypoints = path:GetPathWayPoints()
	...
```

LuaUI/Widgets/gbc/Rendering.lua
```diff
@@ -0,0 +1,346 @@
...
+
+local statusColor         = {1.0, 1.0, 1.0, 0.85}
+local queueColor          = {1.0, 1.0, 1.0, 0.9}
+
+
+-- Zero-K specific icons for drawing repair/reclaim/resurrect, customize if porting!
+local rec_icon = "LuaUI/Images/commands/Bold/reclaim.png"
+local rep_icon = "LuaUI/Images/commands/Bold/repair.png"
+local res_icon = "LuaUI/Images/commands/Bold/resurrect.png"
+local rec_color = {0.6, 0.0, 1.0, 1.0}
```
sprunk, line 47: I think these can be read directly from the command somehow (but it's a pie-in-the-sky triviality, let it stay)

GoogleFrog:
@esainane as I wrong to assume that "more to follow" means that this is WIP?

sprunk:
Apparently more already followed compared to the initial set of PR'd commits
image

GoogleFrog:
Well I'm not up to reading GBC, so if it has been working I guess we just merge it?

PR #4060, opened 2020-08-25
Get rid of error spam

Arch-Shaman:
Removes all debug echos. Ill take care of the rest in 4-6 days.

### DIFF
--- a/LuaRules/Gadgets/game_share_mode.lua
+++ b/LuaRules/Gadgets/game_share_mode.lua
@@ -125,12 +125,6 @@ local function GetSquadSize(teamID)
 	return #spGetPlayerList(teamID, true)
 end
 
-local function DebugEcho(str)
-	if debug then
-		spEcho(str)
-	end
-end
-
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
No idea how to unrevert stuff from web interface on mobile so... >.<

GoogleFrog:
I don't think this would fix the errors.

Arch-Shaman:
Hm. Well I'm 4-6+ days without a test environment. Ill look into the best I can with the environment I have (mobile phone.. web interface, crappy cell reception).

Arch-Shaman:
Yeah I'm still out of commission right now. New apartment doesn't have ethernet ran around the unit so had to get a wireless card. I lack a chair as well at the moment. Eta for solution is in 2-4 days.

Thought i would give an update here.

Arch-Shaman:
see #4088. Dunno how to fix the conflicts, but I figured out what was causing it.

PR #4062, opened 2020-08-25
If a building can't be underwater, don't try to terraform-lower it below water-level.

pengvado:
The attempt would fail, of course, but only after wasting metal on the too-deep pit.

### DIFF
--- a/LuaUI/Widgets/gui_persistent_build_height.lua
+++ b/LuaUI/Widgets/gui_persistent_build_height.lua
@@ -196,6 +196,8 @@ local function SendCommand()
 			end
 		end
 	end
+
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_persistent_build_height.lua
```diff
@@ -196,6 +196,8 @@ local function SendCommand()
 			end
 		end
 	end
+
+	pointY = math.max(pointY, -UnitDefs[buildingPlacementID].maxWaterDepth)
```
sprunk, line 200: Could be cached so as to avoid the `UnitDefs` access, the `maxWaterDepth` indexing and the `-` operator each time.

Issue #3100, opened 2018-05-27
Pick and apply new sound contributions


Histidine91:
A number of our users on Discord have contributed sounds which may be useful. Partial collection here: [Google Drive](https://drive.google.com/drive/folders/1ZsxAvly1qm_58fKhcAv7AV7DLNr2Obrq?usp=sharing)

Some suggestions:

- stringstorm/MixedWithNapalm.mp3: Kodachi, Firewalker, etc. hit sound
- Apelsinsaft/big_begrtha_gun_fire.wav: Something firing, unsure what (dgun?)
- Apelsinsaft/lrpc_hit.wav: Bertha or some other suitably high-impact artillery weapon
- Ozmium/sythe_sound2.wav: Scythe
- Ozmium/mouse_click.wav: Lobby GUI button clicked

Licho1:
Napalm +1
Bertha - sounds more like railgun to me (likewise the hit), but they are very nice sounds.
Scythe +1

Anarchid:
I don't like the napalm sounds, but +1 to Bertha and ambivalent on Scythe.

Licho1:
Mininuke has been suggested for tacnuke explosion

Issue #3037, opened 2018-05-08
Sound improvements


GoogleFrog:
The most general task would be to make a bunch selection and reply sounds to replace some low quality ones and increase diversity. These would be similar to the current ones and covers things like robotic whirs, vehicle noises and plane whooshes.

The following sounds are missing
- button clicking and menu transitions for the main menu

The following sounds are low quality or unfitting
- Blastwing explosion
- Shogun/Cerberus impact
- Spitter fire and impact
- Disruptor beams
- Matchmaker match found
- Many selection or order received sounds

The following sounds are duplicates so should be replaced, possibly with something similar.
- Buoy and Halberd fire and impact.
- Cyclops impact
- Seawolf fire and impact
- Scallop torpedoes fire and impact
- Racketeer launch
- Siren fire and impact
- Many selection or order received sounds

See https://github.com/ZeroK-RTS/Zero-K/issues/2592 for a related ticket.

Kingstad:
Figured i ought to write down as I go what sounds I have been working on so people know:
**Cyclops** shot, impact. **Badger** shot, impact. **Trinity** shot, impact. **Phoenix** impact. **Blastwing**

Anarchid:
Siren has been [dealt](https://github.com/ZeroK-RTS/Zero-K/commit/935278503c40cae256fd18be948a7deab6174f9d) with, it seems.

angerbirb:
I made some samples for the blastwing and shogun.

GoogleFrog:
See https://github.com/ZeroK-RTS/Chobby/pull/765

Issue #2592, opened 2017-10-30
Sounds


katastropheZK:
Working on sounds, would like to replace sounds that are used multiple times (or pitched versions of another). Is there a list of what sounds are used when to get an overview or will i have to browse all those LUA-files manually? Because i am simply not able to write a script...

sprunk:
There is no easily accessible list yet but I can generate one.

katastropheZK:
That would be very very useful. A big thank you in advance.

Anarchid:
Big Bertha and the Cyclops (ex-Goliath) have a very similar sound.
Shogun (battleship) and Cerberus (static artillery) have the same sound (and the same weapon, so that's not too bad).
Firewalker and Wolverine have pretty much the same sound. It was pitch-adjusted recently to not be exactly the same, but it's still close.
Afair, Impaler and Racketeer have the same sound. This is also the sound used by Detri cruise missile sidearm.

Anarchid:
I ran a script. This is the output:

```
Sound weapon/missile/rocket_hit used 7 times in 7 units: 
		amphaa
		chicken_digger
		chicken_leaper
		gunshipaa
		planefighter
		shieldaa
		spideraa

Sound weapon/missile/missile_fire7 used 6 times in 6 units: 
		amphaa
		gunshipaa
		planefighter
		shieldaa
		shipscout
		spideraa


Sound explosion/ex_med6 used 6 times in 5 units: 
		ampharty
		grebe
		gunshipkrow
		striderantiheavy
		tankraid


Sound weapon/cannon/cannon_fire3 used 5 times in 4 units: 
		ampharty
		grebe
		jumpskirm
		tankraid


Sound explosion/wet/ex_underwater used 8 times in 7 units: 
		ampharty
		amphraid
		bomberdive
		bomberprec
		grebe
		hoverdepthcharge
		turrettorp


Sound weapon/laser/heavy_laser3 used 4 times in 4 units: 
		amphassault
		amphassault
		striderdetriment
		turretheavylaser


Sound weapon/laser/small_laser_fire used 4 times in 4 units: 
		amphfloater
		amphfloater
		spiderscout
		vehscout


Sound weapon/laser/small_laser_fire3 used 4 times in 4 units: 
		amphfloater
		amphfloater
		hoverassault
		planefighter


Sound weapon/laser/laser_burn8 used 4 times in 4 units: 
		amphimpulse
		amphimpulse
		spideranarchid
		turretlaser


[...]

PR #4065, opened 2020-08-26
apply apelsinsaft bertha fire sound

Anarchid:
I've built Cyclops two times in the past week.

Both times, my allies started screaming in horror that enemy has a Bertha.

Well then. This file has been in the repository for two years and i like it better.

#3100 #3037 #2592


GoogleFrog:
It doesn't sound like Bertha to me at all.

PR #4066, opened 2020-08-26
deduplicate Badger fire sound

Anarchid:
This gives the Badger a weapon fire sound that
- Sounds like a mortar
- Does not sound exactly like Firewalker


Anarchid:
On some more pondering i decide this one should have been a commit rather than a PR.

GoogleFrog:
I disagree:
 - origin.txt was not updated.
 - Such a spammy sound should not have two distinct components. Persisting for a long time makes game audio messy.
 - When I test ingame it sounded broken, but actually the sound was being played correctly.

Anarchid:
Fair enough, i'll try to address these when i get back to a system that can run Spring.

What do you mean by "sounded broken"? My local test seemed quite ok.

GoogleFrog:
The first part of the sound just sounds a lot like a short spurt of white noise.

PR #4080, opened 2020-09-02
Badger sound deduplication v2

Anarchid:
- no second component
- also includes metallic Claw impact sound
- origins.txt updated


GoogleFrog:
Seems ok.

PR #4085, opened 2020-09-04
Remove the single biggest source of GC pressure

esainane:
...often by more than an order of magnitude.

Can it really be this simple?


sprunk:
> author  = "very_bad_soldier / versus666",
>date    = "October 21, 2007 / September 08, 2010",
Likely.

sprunk (approved):

PR #4087, opened 2020-09-04
Mex Placement Handler Optimizations

esainane:


LuaUI/Widgets/cmd_mex_placement.lua
```diff
@@ -623,9 +627,36 @@ function widget:Initialize()
...
+	if isSpectating and isFullView ~= wasFullView then
+		wasFullView = isFullView
+		return true
+	end
+	return false
+end
+
 local firstUpdate = true
-function widget:Update()
+local cumDt = 0
```
sprunk, line 652: lewd

LuaUI/Widgets/cmd_mex_placement.lua
```diff
@@ -650,6 +679,25 @@ function widget:Update()
...
+			debounceCamUpdate = nil
+		end
+	else
+		local cx, cy, cz = Spring.GetCameraDirection()
+		local newCamDir = ((math.atan2(cx, cz) / math.pi) + 1) * 180
+		if newCamDir ~= camDir then
+			camDir = newCamDir
+			gl.DeleteList(incomeLabelList)
+			incomeLabelList = glCreateList(DrawIncomeLabels)
+			debounceCamUpdate = .1
```
sprunk, line 694: Github's syntax highlighter doesn't color this like a number and I'm scared (also this style is not used anywhere else)
```suggestion
			debounceCamUpdate = 0.1
```

LuaUI/Widgets/cmd_mex_placement.lua
```diff
@@ -623,9 +627,36 @@ function widget:Initialize()
 	end
 end
 
+local wasFullView
+
+local function CheckNeedsRecalculating()
+	if not WG.metalSpots then
```
sprunk, line 633: Is there ever anything to draw when `not WG.metalSpots`? Maybe this check could be in `Initialize` and do `gadgetHandler:RemoveCallIn('Update')` if so.

esainane: There's some logic in `Update` around `IntegrateMetal` when `not WG.metalSpots` that I don't entirely follow. `DrawWorld` and `DrawWorldPreUnit` do seem to be completely guarded by the condition though, so it might be useful to `widgetHandler:RemoveCallIn` those.

PR #4058, opened 2020-08-24
Revert "Commshare maintenance"

GoogleFrog:
Reverts ZeroK-RTS/Zero-K#4045

PR #4088, opened 2020-09-06
Commshare fix

Arch-Shaman:
Fixes the error spam associated with #4058:
- This was caused by spectators triggering the debug echo with a typo in it (should have been targetID instead of playerID)


Arch-Shaman:
To do:
- Local testing

GoogleFrog:
I have not had another look yet, but there is a good way to make detailed debug information that can be run on stable releases. Do '/luarules debugpri' with cheats enabled and search for debugpri among the gadgets to see how to do it.

Arch-Shaman:
I searched through the gadgets found the following:
https://github.com/ZeroK-RTS/Zero-K/blob/0dd56a39cb4ebaf3ea3bb6da2faa48e1760a363e/LuaRules/Gadgets/unit_priority.lua#L718

leads to: https://github.com/ZeroK-RTS/Zero-K/blob/4d15340f4af3e3012a26639b9ec99090076202e0/LuaRules/gadgets.lua#L486

Something new to keep in mind. Thanks.

PR #4091, opened 2020-09-07
Disallow sharing to AFK/dead teams

### LINKED ISSUE #4068: Disable sharing to inactive teams
https://zero-k.info/Forum/Thread/31958

Arch-Shaman:
Fixes #4068

### DIFF
--- a/LuaRules/Gadgets/share_control.lua
+++ b/LuaRules/Gadgets/share_control.lua
@@ -17,11 +17,7 @@ local gaiaTeamID = Spring.GetGaiaTeamID()
 
 local function IsTeamAfk(teamID)
 	local _, shares = GG.Lagmonitor.GetResourceShares()
-	if shares == 0 then
-		return true
-	else
-		return false
-	end
+	return shares == 0
 end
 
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/share_control.lua
```diff
@@ -14,6 +14,16 @@ local spAreTeamsAllied = Spring.AreTeamsAllied
 local spIsCheatingEnabled = Spring.IsCheatingEnabled
 local spGetUnitRulesParam = Spring.GetUnitRulesParam
 local gaiaTeamID = Spring.GetGaiaTeamID()
+
+local function IsTeamAfk(teamID)
+	local _, shares = GG.Lagmonitor.GetResourceShares()
+	if shares == 0 then
```
GoogleFrog, line 20: If bla then return true else return false is ugly.

Arch-Shaman: if shares == 0 then
return true
end

if IsTeamAfk(newteamID) then return false end

?

PR #4096, opened 2020-09-08
Actually fix this check.

Arch-Shaman:
Shares is actually a table of teamIDs that are never nil. Apparently we assumed it wasn't broken because it never complained.


Arch-Shaman:
This affects commshare as well (already updated). Always thought it was weird having a function with a teamID param but not being used anywhere.

PR #4098, opened 2020-09-08
Fix tab playerlist displaying <1000 elo players as highest rank.

DeinFreund:


sprunk:
Feel free to push trivial fixes directly

Issue #3778, opened 2019-12-30
Merlin no longer able to fire after game load


Histidine91:
https://steamcommunity.com/app/334920/discussions/1/2659871169526243980/

Sharkgamedev:
I was able to reproduce this - no errors were logged to the console
However, it seems that ALL merlins lose the ability to fire after a load in a singleplayer game - no special cases needed,

Steps to reproduce
1. Start a singleplayer skirmish
2. Use /cheat /give 1 striderarty to spawn a merlin (Tested with naturally building a merlin as well)
3. Save game
4. Load the game and the merlin will no longer be able to fire 

I'll look into it.

PR #4099, opened 2020-09-09
Merlin cob animation rework.

Sharkgamedev:
Upgraded the merlin from cob to lua to fix: #3778


Sharkgamedev:
Reverted commits were because I messed up when originally setting up this branch.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+
+--MAKE SURE FIRE POINTS REMAIN CONSECUTIVEnot 
+
+local  bMoving, bAiming, isReady, gun_1, gun_1_side, Static_Var_3
+
+-- Signal definitions
+local SIG_AIM = 4
+
+
+--SmokeUnit(healthpercent, Sleep(time, smoketype)
```
GoogleFrog, line 73: May as well delete smokeUnit

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+function RestoreAfterDelay()
+	Sleep(3000)
+	Turn(torso , y_axis, 0, math.rad(90.000000))
+	Turn(launchers , x_axis, 0, math.rad(45.000000))
+	WaitForTurn(torso, y_axis)
+	WaitForTurn(launchers, x_axis)
+	bAiming = false
+end
+
+function script.AimFromWeapon1()
```
GoogleFrog, line 424: I rather have AimFromWeapon(num) rather than AimFromWeapon1, with corresponding changes for the other functions, for conistency and searchability.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+		gun_1 = gun_1 + 1
+	end
+	gun_1_side = not gun_1_side
+end
+
+function script.QueryWeapon1() 
+    local piecenum
+	if gun_1 >= 0 and gun_1 < 20 then 
+		if gun_1_side then
+			piecenum = (rfire1 + gun_1)
```
GoogleFrog, line 444: This seems to rely on the internal order of piece indexes within the model, which is bad.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+		else 
+			piecenum = (lfire1 + gun_1)
+		end
+	else 
+		piecenum = launchers
+    end
+    return piecenum
```
GoogleFrog, line 451: Don't set 'local piecenum' earlier and then return piecenum. Just return when the piecenum would be set.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+end
+
+function script.AimWeapon1(heading, pitch)
+	Signal( SIG_AIM)
+	SetSignalMask( SIG_AIM)
+	bAiming = true
+	while not isReady do
```
GoogleFrog, line 458: not isReady seems to always be true

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+		Sleep(100)
+	end
+	Turn(torso , y_axis, heading, math.rad(90.000000) )
+	Turn(launchers , x_axis, 0 - pitch, math.rad(45.000000) )
```
GoogleFrog, line 462: -pitch rather than 0 - pitch. Also you don't need trailing 0s.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+    bMoving = true    
+    --StartThread(walk)
+end
+
+function script.StopMoving()
+    bMoving = false
+	--StartThread(RestoreLegs)
+end
+
+local function SweetSpot(piecenum)
```
GoogleFrog, line 399: SweetSpot does nothing.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+			Sleep( 100)
+		end
+	end
+end
+
+function script.Create()
+
+	Hide(rfire1)
+	Hide(lfire1)
+	bMoving = false
```
GoogleFrog, line 376: Initialisation of local variables may as well be at the top of the file.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+	end
+	Turn( ltoer , x_axis, 0, math.rad(100.021978) )
+	Move( pelvis , y_axis, 4 , 2 )
+	Turn( pelvis , z_axis, math.rad(-(3.5)), math.rad(8) )
+	Sleep( 2)
+end
+
+function walklegs()
+
+		if  bMoving == true  then
```
GoogleFrog, line 216: Whitespace.

scripts/striderarty.lua
```diff
@@ -0,0 +1,505 @@
...
+		Turn( rtoer , x_axis, math.rad(-22.5016484), math.rad(100.021978) )
+		Sleep( 360)
+	end
+	Turn( ltoer , x_axis, 0, math.rad(100.021978) )
+	Move( pelvis , y_axis, 4 , 2 )
+	Turn( pelvis , z_axis, math.rad(-(3.5)), math.rad(8) )
+	Sleep( 2)
+end
+
+function walklegs()
```
GoogleFrog, line 214: functions should be local unless they have to not be local.

GoogleFrog:
Converting a bos script is a bit of a trial by fire because the converter outputs poorly formatted code 

I made some minor comments. The largest is probably replacing the hax of adding a number to a piece with a lookup table. Here are some stretch goals:
- The whitespace is inconsistent. I wrote this recently: https://zero-k.info/mediawiki/index.php?title=Zero-K:Developing#Style_Guide
- Remove MotionControl. Many bos scripts used MotionControl, possibly due to a limitation of bos, but most lua scripts removed it when they were converted. MotionControl is a redundant extra thread when the unit is idle and makes the code more complicated than it needs to be. In lua scripts we prefer to use signal masks on a single walk thread that is created each time a unit starts moving. See shieldraid for a simple example and cloakraid for a more complicated one. The animation timings and positions can be copied from the bos->lua conversion tool and the difference between aiming and not aiming can be handled by some if statements. I would do this if I were converting the file, but almost any lua is better than the equivalent bos so I'm not super worried.
- If I were redoing this from scratch I would make one piece that emits the rockets and just move the piece between some preset positions. I wouldn't do this now though.

scripts/striderarty.lua
Sharkgamedev: Are you suggesting I put the pieces in an array and iterate through that?

Sharkgamedev:
I did my best to reformat and fix all the whitespace in the code, let me know if I missed anything. (I'm 99% sure I got everything)
I also incorporated everything you suggested in the code comments except for one thing which I commented on, I did remove motion control as well and am instead using just the one walk thread as you suggested.
However something is slightly messed up with the walk cycle that I don't think was there before, I spent about an hour trying to get it fixed but couldn't quite figure it out, it's not too noticeable but essentially you can tell where the walk cycle loops now.

scripts/striderarty.lua
```diff
@@ -0,0 +1,361 @@
...
+			Move(pelvis, y_axis, 4, 2)
+			Turn(pelvis, z_axis, math.rad(3.5), math.rad(8))
+			Sleep(2)
+		else
+			Turn(lupleg, x_axis, math.rad(20.016484), math.rad(50.010989))
+			Turn(rupleg, x_axis, math.rad(-20.016484), math.rad(50.010989))
+			Turn(lfoot, x_axis, math.rad(-15.016484), math.rad(70.016484))
+			Turn(rfoot, x_axis, math.rad(5), math.rad(50.010989))
+			Turn(rleg, x_axis, math.rad(-10), math.rad(70.016484))
+			Turn(torso, x_axis, math.rad(-1), math.rad(5))
```
GoogleFrog, line 157: This block of animations is identical to the bAiming case, except that torso doesn't turn. I expect that something similar will happen in the other blocks. Instead of checking for bAiming at the start of the loop you should check within the loop and skip the animations that interfere with aiming if the unit is aiming. Otherwise it takes a whole loop for the walk animation to change state.

scripts/striderarty.lua
GoogleFrog: Yes.

scripts/striderarty.lua
Sharkgamedev: Good point, fixed!

GoogleFrog:
The walk cycle being different could be caused by Sleep(2). I don't know how Sleep(2) is executed because the scripts run at 30hz. Sleeping in anything but multiples of 33ms does not make sense, and the interpretation may differ between lua and cob.

Sharkgamedev:
> The walk cycle being different could be caused by Sleep(2). I don't know how Sleep(2) is executed because the scripts run at 30hz. Sleeping in anything but multiples of 33ms does not make sense, and the interpretation may differ between lua and cob.

Yeah I removed sleep(2) anyway but that didn't fix it, the problem is that it seems to repeat legs like:
Left > Right
Left > RIght
Right > Left
Right > Left
etc....

Sharkgamedev:
Hmmm... I think I've fixed it. After comparing to the bos script I noticed there were a few extras -'s and upon removing them the 'loop' got hugely less noticeable, there's still some weird slight bit there but it's really hard to notice it, even when I was looking I missed it a few times and I think that was there originally.

GoogleFrog:
reaim_time = 8, -- COB in the weapondef can now be removed, as lua supports smoother reaim (I think the default is 3 frames).

Sharkgamedev:
> reaim_time = 8, -- COB in the weapondef can now be removed, as lua supports smoother reaim (I think the default is 3 frames).

Done!

GoogleFrog:
Do you want to work on making the walk animation work properly?

Sharkgamedev:
> Do you want to work on making the walk animation work properly?

I think I've pretty much gotten it after changing those numbers.

PR #4125, opened 2020-09-23
Make unit_auto_patrol_nanos smarter

rtwfroody:
1. Idle caretakers will be set to area repair if metal is high, area reclaim
      when metal is low, and patrol otherwise. (Unless patrol_idle_nanos option is
      false.)
   2. For each caretaker under this widget's control, re-evaluate the behavior based
		on the economy every 20 seconds. (Controlled by checkInterval.)
   3. For each caretaker, never issue a command more than once every 5 seconds.
      (Controlled by settleInterval.)
   4. When a user issues a stop command, this behavior is inhibited, until a
      different command issued by the user completes. (Unless stop_disables option
      is false.)
   5. When a user issues some other kind of command, the widget ignores the unit
      until it becomes idle again.

 Limitations:
   1. Area reclaim does not reclaim energy sources. We'd have to call
      GetFeaturesInRectangle() to improve this.
   2. Area repair either assists in build (using both metal and energy) or
      repairs damage (using only energy). Since assisting a factory is the most
      common use for caretakers, we tell caretakers to reclaim if metal is low.
      We'd have to call GetUnitsInCylinder() to do better.
   3. When the widget chooses repair only, there's a 0.5 second delay between
      a unit becoming idle and being told to repair. This is to give the factory
      that's being assisted time to start the next unit. It is not perfect, and the
      factory will be assisted less than if a patrol was issued. This could be
      improved by reducing the delay, and implementing a fancier incremental
      back-off than the simple 0.5s - 5s that is currently implemented.

See discussion at https://zero-k.info/Forum/Thread/32014?page=1

### DIFF
--- a/LuaUI/Widgets/unit_auto_patrol_nanos.lua
+++ b/LuaUI/Widgets/unit_auto_patrol_nanos.lua
@@ -139,8 +139,12 @@ local function RandomInterval(interval)
 	return math.floor(interval + (math.random() * interval * 0.2 - 0.1))
 end
 
-local function Log(msg)
-	spEcho("[uapn] " .. msg)
+local function Log(...)
+	local msg = "[uapn] "
+	for i = 1, select('#', ...) do
+		msg = msg .. tostring(select(i, ...))
+	end
+	spEcho(msg)
 end
 
 local function IsImmobileBuilder(ud)
@@ -159,8 +163,8 @@ local function DecideCommands()
 	local metal, metalStorage, metalPull, metalIncome =
 			spGetTeamResources(spGetMyTeamID(), "metal")
 	metalStorage = metalStorage - HIDDEN_STORAGE
-	-- Log("metal=" .. metal .. "; storage=" .. metalStorage .. "; pull=" ..
-	-- 		metalPull .. "; income=" .. metalIncome)
+	Log("metal=", metal, "; storage=", metalStorage, "; pull=",
+			metalPull, "; income=", metalIncome)
 	local energy, energyStorage, energyPull, energyIncome =
 			spGetTeamResources(spGetMyTeamID(), "energy")
 	energyStorage = energyStorage - HIDDEN_STORAGE
@@ -177,7 +181,7 @@ local function DecideCommands()
 		-- reclaiming when your storage is full.
 		get_metal = futureMetal < metalStorage and
 				metal < metalStorage * 0.75
-		use_metal = futureMetal > 0 and metal > 0
+		use_metal = futureMetal > 0 and metal > 1
 	end
 
 	if energyStorage < 1 then
@@ -187,13 +191,11 @@ local function DecideCommands()
 		local futureEnergy = energy + checkInterval * (energyIncome - energyPull) / FPS
 		get_energy = futureEnergy < energyStorage and
 				energy < energyStorage * 0.9
-		use_energy = futureEnergy > 0 and energy > 0
+		use_energy = futureEnergy > 0 and energy > energyStorage * 0.1
 	end
 
-	--Log("get_metal=" .. tostring(get_metal) .. ", " ..
-	--	"use_metal=" .. tostring(use_metal) .. ", " ..
-	--	"get_energy=" .. tostring(get_energy) .. ", " ..
-	--	"use_energy=" .. tostring(use_energy))
+	Log("get_metal=", get_metal, ", ", "use_metal=", use_metal, ", ",
+		"get_energy=", get_energy, ", ", "use_energy=", use_energy)
 
 	local commands = {}
 
@@ -272,19 +274,13 @@ end
 local remove_shift_internal = math.bit_inv(CMD_OPT_SHIFT + CMD_OPT_INTERNAL)
 local function IssuedCausesCurrent(issued, currentID, currentOpt,
 		currentParam1, currentParam2, currentParam3, currentParam4, currentParam5)
---	Log("compare")
---	Log("    issued : " .. issued[1] .. "(" ..
---		tostring(issued[2][1]) .. ", " ..
---		tostring(issued[2][2]) .. ", " ..
---		tostring(issued[2][3]) .. ", " ..
---		tostring(issued[2][4]) .. ", " ..
---		tostring(issued[2][5]) .. ") " .. issued[3] .. " (" .. issued[4] .. ")")
---	Log("    current: " .. tostring(currentID) .. "(" ..
---		tostring(currentParam1) .. ", " ..
---		tostring(currentParam2) .. ", " ..
---		tostring(currentParam3) .. ", " ..
---		tostring(currentParam4) .. ", " ..
---		tostring(currentParam5) .. ") " .. tostring(currentOpt))
+	Log("compare")
+	Log(" issued : ", issued[1], "(", issued[2][1], ", ", issued[2][2], ", ",
+		issued[2][3], ", ", issued[2][4], ", ", issued[2][5], ") ",
+		issued[3], " (", issued[4], ")")
+	Log(" current: ", currentID, "(", currentParam1, ", ", currentParam2,
+		", ", currentParam3, ", ", currentParam4, ", ", currentParam5, ") ",
+		currentOpt)
 
 	if not currentID or not issued then
 		return false
@@ -298,7 +294,7 @@ local function IssuedCausesCurrent(issued, currentID, currentOpt,
 			issued[2][4] == currentParam4 and
 			issued[2][5] == currentParam5 and
 			issued[3] == currentOpt then
-		--Log("    -> equal")
+		Log("    -> equal")
 		return true
 	end
 
@@ -313,7 +309,7 @@ local function IssuedCausesCurrent(issued, currentID, currentOpt,
 			issued[2][3] == currentParam4 and
 			issued[2][4] == currentParam5 and
 			issued[3] == math.bit_and(currentOpt, remove_shift_internal) then
-		--Log("    -> repair/reclaim")
+		Log("    -> repair/reclaim")
 		return true
 	end
 
@@ -324,19 +320,19 @@ local function IssuedCausesCurrent(issued, currentID, currentOpt,
 		if (currentID == CMD_REPAIR or currentID == CMD_RECLAIM) and
 				currentOpt == CMD_OPT_INTERNAL and
 				currentParam5 ~= nil then
-			--Log("    -> patrol, repair")
+			Log("    -> patrol, repair")
 			return true
 		elseif currentID == CMD_PATROL then
-			--Log("    -> patrol")
+			Log("    -> patrol")
 			return true
 		elseif currentID == CMD_FIGHT and
 				currentOpt == CMD_OPT_INTERNAL then
-			--Log("    -> patrol, fight")
+			Log("    -> patrol, fight")
 			return true
 		end
 	end
 
-	--Log("    -> false")
+	Log("    -> false")
 	return false
 end
 
@@ -358,12 +354,9 @@ local function SetupUnit(unitID)
 	local currentID, currentOpt, _, currentParam1,
 			currentParam2, currentParam3, currentParam4, currentParam5 =
 			spGetUnitCurrentCommand(unitID)
---	Log(unitID .. "; currently executing " .. tostring(currentID) .. "(" ..
---		tostring(currentParam1) .. ", " ..
---		tostring(currentParam2) .. ", " ..
---		tostring(currentParam3) .. ", " ..
---		tostring(currentParam4) .. ", " ..
---		tostring(currentParam5) .. ") " .. tostring(currentOpt))
+	Log(unitID, "; currently executing ", currentID, "(", currentParam1,
+		", ", currentParam2, ", ", currentParam3, ", ", currentParam4, ", ",
+		currentParam5, ") ", currentOpt)
 
 	if currentID then
 		if IssuedCausesCurrent(cmds[1], currentID, currentOpt,
@@ -371,7 +364,7 @@ local function SetupUnit(unitID)
 				currentParam4, currentParam5) then
 			-- The unit is doing something that could be caused by the top command
 			-- we were going to issue. That's good enough.
-			--Log("Unit is doing good work. Don't touch it.")
+			Log("Unit is doing good work. Don't touch it.")
 			return
 		end
 
@@ -390,13 +383,13 @@ local function SetupUnit(unitID)
 		if not isIssued then
 			-- Unit is doing something we never asked for. Must have been commanded
 			-- by a user.
---			Log(unitID .. " was commanded " .. tostring(currentID) .. "(" ..
---				tostring(currentParam1) .. ", " ..
---				tostring(currentParam2) .. ", " ..
---				tostring(currentParam3) .. ", " ..
---				tostring(currentParam4) .. ", " ..
---				tostring(currentParam5) .. ") " .. tostring(currentOpt))
-			Log("Ignore unit " .. unitID .. " until it becomes idle.")
+			Log(unitID, " was commanded ", currentID, "(",
+				currentParam1, ", ",
+				currentParam2, ", ",
+				currentParam3, ", ",
+				currentParam4, ", ",
+				currentParam5, ") ", currentOpt)
+			Log("Ignore unit ", unitID, " until it becomes idle.")
 			trackedUnits[unitID] = nil
 			return
 		end
@@ -405,20 +398,14 @@ local function SetupUnit(unitID)
 	for i, cmd in ipairs(cmds) do
 		if i == 1 then
 			spGiveOrderToUnit(unitID, cmd[1], cmd[2], cmd[3])
---			Log(unitID .. "; give order " .. cmd[1] .. "(" ..
---				tostring(cmd[2][1]) .. ", " ..
---				tostring(cmd[2][2]) .. ", " ..
---				tostring(cmd[2][3]) .. ", " ..
---				tostring(cmd[2][4]) .. ", " ..
---				tostring(cmd[2][5]) .. ") " .. cmd[3] .. " (" .. cmd[4] .. ")")
+			Log(unitID, "; give order ", cmd[1], "(", cmd[2][1], ", ",
+					cmd[2][2], ", ", cmd[2][3], ", ", cmd[2][4], ", ",
+					cmd[2][5], ") ", cmd[3], " (", cmd[4], ")")
 		else
 			spGiveOrderToUnit(unitID, cmd[1], cmd[2], cmd[3] + CMD_OPT_SHIFT)
---			Log(unitID .. "; queue order " .. cmd[1] .. "(" ..
---				tostring(cmd[2][1]) .. ", " ..
---				tostring(cmd[2][2]) .. ", " ..
---				tostring(cmd[2][3]) .. ", " ..
---				tostring(cmd[2][4]) .. ", " ..
---				tostring(cmd[2][5]) .. ") " .. cmd[3] .. " (" .. cmd[4] .. ")")
+			Log(unitID, "; queue order ", cmd[1], "(", cmd[2][1], ", ",
+					cmd[2][2], ", ", cmd[2][3], ", ", cmd[2][4], ", ",
+					cmd[2][5], ") ", cmd[3], " (", cmd[4], ")")
 		end
 	end
 	trackedUnit.settleFrame = currentFrame + RandomInterval(settleInterval)
@@ -521,12 +508,12 @@ function widget:UnitCommand(unitID, unitDefID, unitTeam, cmdID, cmdParams, cmdOp
 	if stopHalts then
 		if cmdID == CMD_STOP then
 			if not stoppedUnit[unitID] then
-				Log("Ignore unit " .. unitID .. " until it is given a command.")
+				Log("Ignore unit ", unitID, " until it is given a command.")
 			end
 			stoppedUnit[unitID] = true
 		elseif stoppedUnit[unitID] then
 			if stoppedUnit[unitID] ~= nil then
-				Log("Pay attention to unit " .. unitID .. " again.")
+				Log("Pay attention to unit ", unitID, " again.")
 			end
 			stoppedUnit[unitID] = nil
 		end
@@ -553,7 +540,7 @@ function widget:UnitIdle(unitID, unitDefID, unitTeam)
 
 	If the command was ordered with SHIFT it would get appended after the patrol. ]]
 
-	--Log("UnitIdle(" .. unitID .. "):")
+	Log("UnitIdle(", unitID, "):")
 	--TableEcho(trackedUnits[unitID], "- ")
 
 	trackedUnits[unitID] = trackedUnits[unitID] or
@@ -584,8 +571,8 @@ function widget:GameFrame(frame)
 			local unitID = entry[2]
 
 			if trackedUnits[unitID] then
-				--Log(unitID .. "; " .. frame .. " >= ".. entry[1] ..
-				--		"; checkFrame=" .. tostring(trackedUnits[unitID].checkFrame))
+				Log(unitID, "; ", frame, " >= ".. entry[1],
+						"; checkFrame=", trackedUnits[unitID].checkFrame)
 
 				if entry[1] == trackedUnits[unitID].checkFrame then
 					-- Otherwise, we queued this unit multiple times and this is not
--- a/LuaUI/Widgets/unit_auto_patrol_nanos.lua
+++ b/LuaUI/Widgets/unit_auto_patrol_nanos.lua
@@ -12,8 +12,8 @@
 --      includes repair units), area reclaim metal, area reclaim energy, or patrol,
 --      depending on available resources.
 --   2. For each caretaker under this widget's control, re-evaluate the behavior based
---		on the economy every 20 seconds. (Controlled by checkInterval.)
---   3. For each caretaker, never issue a command more than once every 5 seconds.
+--		on the economy every 10 seconds. (Controlled by checkInterval.)
+--   3. For each caretaker, never issue a command more than once every 2.5 seconds.
 --      (Controlled by settleInterval.)
 --   4. When a user issues a stop command, this behavior is inhibited, until a
 --      different command issued by the user completes. (Unless stop_disables option
@@ -86,6 +86,8 @@ local mapCenterZ = Game.mapSizeZ / 2
 --------------------------------------------------------------------------------
 -- Constants
 
+local FPS = 30
+
 local PATROL = 1
 local RECLAIM_METAL = 2
 local RECLAIM_ENERGY = 3
@@ -94,12 +96,12 @@ local BUILD_ASSIST = 5
 
 -- Check that a unit is still doing the right thing every `checkInterval`
 -- frames.
-local checkInterval = 300
+local checkInterval = 10 * FPS
 -- Don't issue a new command if less than `settleInterval` frames have passed,
 -- even if the unit became idle. This is a simple rate limiter on issuing
 -- commands in case there are caretakers with nothing to do for their current
 -- state.
-local settleInterval = 75
+local settleInterval = 2.5 * FPS
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
@@ -164,11 +166,12 @@ local function DecideCommands()
 
 	local metal, metalStorage, metalPull, metalIncome =
 			spGetTeamResources(spGetMyTeamID(), "metal")
-	metal = metal + metalIncome - metalPull
+	metal = metal + checkInterval * (metalIncome - metalPull) / FPS
 	metalStorage = metalStorage - HIDDEN_STORAGE
+	Log("metal=" .. metal .. "; storage=" .. metalStorage .. "; pull=" .. metalPull .. "; income=" .. metalIncome)
 	local energy, energyStorage, energyPull, energyIncome =
 			spGetTeamResources(spGetMyTeamID(), "energy")
-	energy = energy + energyIncome - energyPull
+	energy = energy + checkInterval * (energyIncome - energyPull) / FPS
 	energyStorage = energyStorage - HIDDEN_STORAGE
 
 	local slop = 5
--- a/LuaUI/Widgets/unit_auto_patrol_nanos.lua
+++ b/LuaUI/Widgets/unit_auto_patrol_nanos.lua
@@ -11,6 +11,9 @@
 --
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
+VFS.Include("LuaRules/Configs/constants.lua")
+
+local logfile
 
 function widget:GetInfo()
   return {
@@ -26,19 +29,25 @@ end
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
--- Speeups
+-- Speedups
 
+local CMD_INSERT        = CMD.INSERT
 local CMD_MOVE_STATE    = CMD.MOVE_STATE
 local CMD_PATROL        = CMD.PATROL
+local CMD_RECLAIM       = CMD.RECLAIM
+local CMD_REPAIR        = CMD.REPAIR
 local CMD_STOP          = CMD.STOP
 local spGetGameFrame    = Spring.GetGameFrame
 local spGetMyTeamID     = Spring.GetMyTeamID
 local spGetTeamUnits    = Spring.GetTeamUnits
 local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
+local spGetCommandQueue = Spring.GetCommandQueue
 local spGetUnitDefID    = Spring.GetUnitDefID
 local spGetUnitPosition = Spring.GetUnitPosition
 local spGiveOrderToUnit = Spring.GiveOrderToUnit
 local spGetGameRulesParam = Spring.GetGameRulesParam
+local spEcho            = Spring.Echo
+local spGetTeamResources = Spring.GetTeamResources
 
 local abs = math.abs
 
@@ -47,11 +56,27 @@ local mapCenterZ = Game.mapSizeZ / 2
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
+-- Constants
+
+-- Check that a unit is still doing the right thing every `checkInterval`
+-- seconds.
+local checkInterval = 20
+-- Don't issue a new command if less than `settleInterval` seconds have passed,
+-- even if the unit became idle.
+local settleInterval = 5
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
 -- Variables
 
 local stoppedUnit = {}
 local enableIdleNanos = true
 local stopHalts = true
+-- Map from unitID -> { command, commandArgs, checkTime }
+local trackedUnits = {}
+-- The current time, in seconds (I think)
+local time = 0
+local nextCheck
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I think you can get around the limitations:

1. If you hold Ctrl while area reclaiming, then energy-only features are ignored.
2. If you hold Ctrl while area repairing, then units under construction are ignored.

This implies that there is some combination of cmdOpts that can be set to achieve these behaviours. Note that Ctrl is not necessarily the option you're looking for because ZK disagrees with the engine regarding what is a good, unmodded, default. Also the engine uses meta (space) for things while ZK always uses it to insert commands. There is a widget or gadget that intercepts these orders and messes with the cmdOpts to translate the correct behaviour from the ZK UI to what the engine expects. You could find that widget/gadget and see what you need to solve 1 and 2, or just try stuff and see what happens.

Licho1:
When selecting what to do, caretakers should observe construction priority. 
For example if some construction is high priority next to low priority, it should not assist the low priority one.

Few more notes. I believe the logic should be:

1) if you are not stalling metal (stalling = being at 0 or at reserved metal (ctrl click on bar)) -> do assist of HIGH priority constructions

2) if you are not excessing metal AND not stalling energy -> do repair of nearby damaged units (if unit is under attack, repair the most damaged to avoid destruction, if not under attack repair the most healthy mobile first to let you send unit back to front)

3) do build assist in priority order

You also need to check more often than once per 5s

GoogleFrog:
I think the PR looks like a good first step.

Arch-Shaman:
I'd have to agree. Solid work. There's some styling problems here, but I can take care of that if you want.

rtwfroody:
@GoogleFrog, I implemented your suggesting about using control/meta keys to tweak reclaim and repair behavior. This is much better. Thanks!

@Licho1, that sounds like a good improvement, but I don't know how to achieve that without looking at all possible construction in the area. This might have a too high performance impact. @GoogleFrog has warned me away from doing much of that, but I don't know how much is too much.

@Arch-Shaman, if there's a style guide I can probably go through and fix my code to match. Or if you want to teach me by commenting on the PR that's also fine.

Sharkgamedev:
> if there's a style guide I can probably go through and fix my code to match. Or if you want to teach me by commenting on the PR that's also fine.
There is a style guide! https://zero-k.info/mediawiki/index.php?title=Zero-K:Developing#Style_Guide

Licho1:
You can burn some performance in widget and if you do it only like once per second sits perfectly fine.. you could also do some sort of clustering. But there are engine functions which will give you nearby units relatively cheaply using the map system.

Arch-Shaman:
> if there's a style guide I can probably go through and fix my code to match. Or if you want to teach me by commenting on the PR that's also fine.

I'd be happy to review the pr for style errors. I'll do that tonight. I would prefer if the task did not demotivate you over trivial matters, which is why I offered to take care of it for you. It's mostly just a bunch of trivial things like `commas needing spaces` or `arithmetic needs spaces` or `tabification`.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,299 @@ end
...
+-- is just some of them reclaiming.
+local function RandomInterval(interval)
+	return interval / 2 + math.random() * interval
+end
+
+local function Log(msg)
+	spEcho("[uapn] " .. msg)
+end
+
+local function LogTable(table, prefix)
```
sprunk, line 119: Use `Spring.Utilities.TableEcho`

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,299 @@ end
...
+		if type(value) == "table" then
+			Log(prefix .. tostring(key) .. ":")
+			LogTable(value, prefix .. "  ")
+		else
+			Log(prefix .. tostring(key) .. ": " .. tostring(value))
+		end
+	end
+end
+
+local function TableEqual(a, b)
```
GoogleFrog, line 135: The tables sent to TableEqual are indexed from 1 to n, so pairs is not required.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: Thanks! Is this documented anywhere, or do you have to dig through the sources to find it (which I just did)?

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: So you think I should use ipairs, which presumably performs better? I'm pretty new at Lua.

rtwfroody:
@GoogleFrog, can you comment on performance implications? People have suggested both here and at https://zero-k.info/Forum/Thread/32014?page=1 that 20s is too long. Obviously things work better if those intervals are reduced, but I don't know how to evaluate what a reasonable impact is.

GoogleFrog:
image

My caretakers has no commands at full metal and energy. This seems wrong. If I start a unit in the factory then they get a repair command for it, where did that command come from? Then they get a patrol command and lose the patrol command at full metal and energy.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
GoogleFrog: We tend to use for i = 1, #thing do.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+			return false
+		end
+	end
+
+	return true
+end
+
+local function UpdateNextCheck()
+	nextCheck = nil
+	for _, info in pairs(trackedUnits) do
```
GoogleFrog, line 139: Should be an IterableMap.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
 --------------------------------------------------------------------------------
 -- Variables
 
 local stoppedUnit = {}
 local enableIdleNanos = true
 local stopHalts = true
+-- Map from unitID -> { checkTime, settleTime, command }
+local trackedUnits = {}
+-- The current time, in seconds (I think)
+local time = 0
```
GoogleFrog, line 103: 'time' means something in lua and is highlighted by notepad++ so I would prefer a different variable name.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+				energy < energyStorage * 0.9
+		use_energy = energy > slop and energy > energyStorage * 0.1
+	end
+
+	--Log("get_metal=" .. tostring(get_metal) .. ", " ..
+	--	"use_metal=" .. tostring(use_metal) .. ", " ..
+	--	"get_energy=" .. tostring(get_energy) .. ", " ..
+	--	"use_energy=" .. tostring(use_energy))
+
+	local reclaim_metal = {CMD_RECLAIM, {x, y, z, buildDistance}, 0, "reclaim metal"}
```
GoogleFrog, line 207: I think these strings are causing the command to not draw. The engine will not draw commands with the wrong number of parameters. I don't even know what the engine does if you pass a string to a command, so you should avoid doing it.

Adding a redundant parameter is a nice way to make the engine not draw the queue, but if we were to do that then I'd want some other way to indicate that the caretaker is busy and not idle. I prefer the approach of giving a massive area command.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
GoogleFrog: That said, the more relevant thing to consider is whether you need to check the equality of two arbitrary tables at all. Surely the widget knows more about the form of the tables it expects.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
 
-local function SetupUnit(unitID)
-	-- set immobile builders (nanotowers?) to the ROAM movestate,
-	-- and give them a PATROL order (does not matter where, afaict)
-	local cmdID = spGetUnitCurrentCommand(unitID)
-	if cmdID and cmdID ~= CMD_PATROL then
-		return
+local resourceCacheInterval = 1
+local resourceCache = { updated=nil }
+local function DecideCommands(x, y, z, buildDistance)
```
GoogleFrog, line 153: It looks like this function could be split into one that makes a global decision about the mode that caretakers should switch to, and another that implements that decision for each caretaker as it is updated.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -186,20 +437,44 @@ function widget:UnitIdle(unitID, unitDefID, unitTeam)
...
+			time + delta)
+	if nextCheck == nil then
+		nextCheck = time + delta
+	else
+		nextCheck = min(nextCheck, trackedUnits[unitID].checkTime)
+	end
+	--TableEcho(trackedUnits[unitID], "+ ")
 end
 
 function widget:Update(dt)
```
GoogleFrog, line 462: My intuition is saying that this widget would be simpler and more reliable if GameFrame were used instead of Update. With GameFrame you could simply partially iterate over your list of Caretakers at a rate that will achieve the desired individual update frequency. As it is you seem to be doing pairs over a map of caretakers twice every Update.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+	-- Patrolling doesn't do anything if you target the current location of
+	-- the unit. Point patrol towards map center.
+	local vx = mapCenterX - x
+	local vz = mapCenterZ - z
+	local patrol = {CMD_PATROL, {x + vx*25/abs(vx), y, z + vz*25/abs(vz)}, 0, "patrol"}
+
+	local commands = {}
+
+	if get_metal and use_metal and use_energy then
+		table.insert(commands, patrol)
```
GoogleFrog, line 221: I dislike table.insert. If you're putting something at the end of a list do commands[#commands + 1] = bla

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
-		x = x + vx*25/abs(vx)
-		z = z + vz*25/abs(vz)
+		local unitDefID = spGetUnitDefID(unitID)
+		local buildDistance = UnitDefs[unitDefID].buildDistance
+		trackedUnits[unitID] = trackedUnits[unitID] or {}
+		trackedUnits[unitID].checkTime = time + RandomInterval(checkInterval)
+		local cmds = DecideCommands(x, y, z, buildDistance)
+		TableEcho(cmds, "cmds: ")
+
+		local commandQueue = spGetCommandQueue(unitID, -1)
```
GoogleFrog, line 266: You don't need the whole command queue. Getting the command queue as a table spams temporary tables into memory, so Spring.GetUnitCurrentCommand(unitID, n) was added and applied a few years ago to cut down on table creation. You should be able to use Spring.GetUnitCurrentCommand on the first few commands to check what you need, failing early if you can.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -144,22 +389,29 @@ function widget:UnitCreated(unitID, unitDefID, unitTeam)
 	end
 
 	SetupUnit(unitID)
+	UpdateNextCheck()
 end
 
+-- Called whenever a unit gets a command from any source, including this script!
 function widget:UnitCommand(unitID, unitDefID, unitTeam, cmdID, cmdParams, cmdOptions)
```
GoogleFrog, line 396: Checking for OPT_INTERNAL in cmdOptions should be sufficient to make the handling go completely dormant, removing much of the need to check for user commands in the check loop.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+						if current.id == cmd[1] and
+								TableEqual(current.params, cmd[2]) then
+							foundIssuedCommand = true
+						end
+					end
+				end
+			end
+		end
+
+		if AllTrue(foundCommand) then
```
GoogleFrog, line 304: You know that foundCommand is a list, so you don't need pairs.

GoogleFrog:
You should be able to get away with a check interval of around 1 second. Here is the profiler:
image

I did '/give 100 static con', '/give energsingu', '/give energsingu',  '/give 100 energysolar_dead' and '/luarules damage' to give the Caretakers a bunch of things to reclaim and repair and the performance wasn't good.However, then I set checkInterval = 100 and found that even when all the units were idle the widget would jump up to 3% cpu usage and drop back down to 0.5% periodically over a few seconds.

image

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+
+	--Log("get_metal=" .. tostring(get_metal) .. ", " ..
+	--	"use_metal=" .. tostring(use_metal) .. ", " ..
+	--	"get_energy=" .. tostring(get_energy) .. ", " ..
+	--	"use_energy=" .. tostring(use_energy))
+
+	local reclaim_metal = {CMD_RECLAIM, {x, y, z, buildDistance}, 0, "reclaim metal"}
+	local reclaim_energy = {CMD_RECLAIM, {x, y, z, buildDistance}, CMD_OPT_CTRL, "reclaim energy"}
+	local repair_units = {CMD_REPAIR, {x, y, z, buildDistance}, CMD_OPT_META, "repair units"}
+	local build_assist = {CMD_REPAIR, {x, y, z, buildDistance}, 0, "build assist"}
```
GoogleFrog, line 210: You're creating a lot of tables and not using them.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,61 +47,284 @@ end
...
+local function AllTrue(table)
+	for _, v in pairs(table) do
+		if not v then
+			return false
+		end
+	end
+	return true
+end
+
+local function SetupUnit(unitID)
```
GoogleFrog, line 256: Caretakers don't move, so unit position may as well be saved.

GoogleFrog:
I did some testing and wrote about IterableMap here https://github.com/ZeroK-RTS/Zero-K/issues/4132

GoogleFrog:
This probably needs a bit of a restructure, which makes my piecemeal comments not so useful. Here is a summary.
 - Iterating over things is always slow. Ideally units would be stored in a single list. This may not be possible if you want to handle UnitIdle and UnitCommand, so use IterableMap. Don't iterate over arbitrary command queues.
 - Replace widget:Update with widget:GameFrame. This will let you predictably iterate through all the Caretakers at a known rate. 
 - Creating tables only to throw them away later is slow.
 - Send the engine acceptable combinations of parameters, otherwise it won't draw the commands.
 - Don't grab the entire order queue to check a units state. You should be able to do something like this: https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_tactical_ai.lua#L165. It isn't pretty but it works.
 - You should be able to save a units state rather than checking it all the time. Some combination of UnitIdle and UnitCommand should be able to tell you whether the player or the widget wishes to handle the caretaker. As long as the widget is handling the caretaker you know what state it is in because you set the state. 
 - Put the global economy decision in GameFrame, independent of Caretaker control. You should be able to sync this to the frame at which the economy is updated in mex_overdrive for reduced control latency.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: I'll change TableEqual to use the for loop you suggested.

What's being compare are arguments for commands. While it so happens that all the commands currently generated by DecideCommands() take the same argument form, I don't want to write code that assumes that because it makes the code very fragile. Since TableEqual only gets called if the actual command IDs match, I don't think the performance overhead of a for loop (compared to hardcoding the indices) is significant.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: These strings are not passed to GiveOrderToUnit() at all. They're just there so I can print a better log message when the command is issued.

rtwfroody:
> My caretakers has no commands at full metal and energy. This seems wrong. If I start a unit in the factory then they get a repair command for it, where did that command come from? Then they get a patrol command and lose the patrol command at full metal and energy.

What's happening in this case is that you have lots of metal and energy. A patrol command is unsuitable because it would reclaim metal. The widget issues area build assist followed by area repair unit commands. Since there is nothing to actually do, those completed quickly and the queue is empty. Every 5s this will keep happening. If a caretaker does need to assist in something, it will take up to 5s before that happens. (We could easily reduce the 5s, of course, and issue commands more often.)

GoogleFrog:
Issue a permanent area repair command. There is some opt that does it.

Licho1:
permanent area repair is by default alt +

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: What do you think that function should return?

The simple solution would be `return {get_metal, use_metal, get_energy, use_energy}` except the logic further down also needs to know whether metal > energy. So I could add metal and energy, or just the result of the comparison. My intuition is to use a table where these things are named, so there's no mess with indices, although I could create constants for them. But based on your other comments it seems like the only data structure that lua offers is really slow.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: I can take a look at GameFrame. Does that mean it's better to change logic that uses seconds to use frames instead? I would have thought that people who get poor frame rates should not necessarily get less responsive caretakers.

The iteration only happens if time > nextCheck (which is a simple, imperfect hack to help performance), although that gets more common as there are more caretakers. Is there a priority queue I can use? That's the real solution to the iteration.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: Incredibly, that takes about 75% of the time of table.insert. I'll change it. (Although more and more I wish we were using a compiled language so that we don't have to worry about these semantic differences for performance's sake.)

LuaUI/Widgets/unit_auto_patrol_nanos.lua
GoogleFrog: Update is called every draw frame. GameFrame is called every simulation frame.

I think there is a priority queue somewhere, but it isn't required. Partially iterating over the list is simple and fine.

rtwfroody:
> You should be able to save a units state rather than checking it all the time. Some combination of UnitIdle and UnitCommand should be able to tell you whether the player or the widget wishes to handle the caretaker. As long as the widget is handling the caretaker you know what state it is in because you set the state.

That's not true because the widget does not know the result of its area commands. They may take forever to complete, or complete instantly if there's nothing to be done.

Let's say you have lots of energy but no metal. The widget will queue repair units, then reclaim metal. Depending on whether there are units to repair in the area, and whether there is metal to reclaim, the caretaker can have 0, 1, or 2 issued commands in its queue (plus internal ones). But the widget could track which commands are issued, and after finding the first issued command infer what the rest of the queue looks like.

> Issue a permanent area repair command. There is some opt that does it.

That sounds like it will never get on to the next command. In the example above, the caretaker would never reclaim metal even if there is nothing to repair.

> However, then I set checkInterval = 100 and found that even when all the units were idle the widget would jump up to 3% cpu usage and drop back down to 0.5% periodically over a few seconds.

This is probably due to the widget iterating over all 100 caretakers, trying to find the one (or potentially multiple) which need attention.


I now have a much better idea of what's expensive and not, as well as how widgets work. I'll take some time to go over the widget and see what all I can incorporate to make it faster. Thanks for the feedback.

GoogleFrog:
> > You should be able to save a units state rather than checking it all the time. Some combination of UnitIdle and UnitCommand should be able to tell you whether the player or the widget wishes to handle the caretaker. As long as the widget is handling the caretaker you know what state it is in because you set the state.
> 
> That's not true because the widget does not know the result of its area commands. They may take forever to complete, or complete instantly if there's nothing to be done.
>
>Let's say you have lots of energy but no metal...

In this case the state is "Repair but then move on to reclaim". With this information you just need to check whether the first command in its queue is CMD.REPAIR and then you know the entire state. This is much simpler and cheaper then comparing the entire command queue to an arbitrary set of commands that it was issued. If the state is just "Repair then do nothing" you don't even need to check the command queue at all. My previous dot point was about cutting down on how deeply you query the queue.

rtwfroody:
@GoogleFrog, you're right. Thanks. I'm mostly done with a rewrite that is more efficient than what I had, but I'm not spending a ton of time on it. I'll probably have something ready for review again in a week or so.

rtwfroody:
This version performs much better, is a little more responsive, and in general works great.

But... I don't think it works with comshare at all. Can somebody confirm that in comsharing, if some other player issues a command, the widget on my computer does not get the UnitCommand() callback for that command?

(Also, playing just now, I need to tweak the get metal logic. It was reclaiming metal even when it should not have.)

rtwfroody:
I think it was supposed to be reclaiming metal then... The situation was that metal storage was full, metal income was X, and metal use was -X - some extra. Is it true that in that case it's still good to reclaim metal, or is it better to wait for the storage itself not to be full?

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -81,14 +89,29 @@ local mapCenterZ = Game.mapSizeZ / 2
 --------------------------------------------------------------------------------
 -- Constants
 
+local FPS = 30
```
sprunk, line 92:
```suggestion
local FPS = Game.gameSpeed
```

rtwfroody:
I think this is it! Performance seems good. (<0.5% and 50KB/s in 100 caretaker idle test).

Should work with comshare (although stop won't work right, and AFAICT never did). Makes caretakers less dumb.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -53,35 +118,310 @@ local stoppedUnit = {}
...
+		end
+	end
+
+	decisionCache.commands = commands
+
+	return commands
+end
+
+local function MakeCommands(decidedCommands, unitID)
+	if trackedUnits[unitID].commandTables == nil then
```
GoogleFrog, line 230: This sort of thing is usually
`	if not trackedUnits[unitID].commandTables then`
Using `== nil` implies you may be distinguishing between false and nil, which is generally confusing,

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -53,35 +118,310 @@ local stoppedUnit = {}
...
+end
 
 local function IsImmobileBuilder(ud)
 	return(ud and ud.isBuilder and not ud.canMove and not ud.isFactory)
 end
 
+local decisionCache = { updated=nil }
+local function DecideCommands()
+	if decisionCache.updated ~= nil and
+			decisionCache.updated + decisionCacheInterval > currentFrame then
```
moreginger, line 153: Micro-optimization, but I'd guess it's slightly faster to add the interval when you assign updated (so rename to 'updateAfterFrame' or something).

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -53,35 +118,310 @@ local stoppedUnit = {}
...
+				currentParam1, currentParam2, currentParam3,
+				currentParam4, currentParam5) then
+			-- The unit is doing something that could be caused by the top command
+			-- we were going to issue. That's good enough.
+			--Log("Unit is doing good work. Don't touch it.")
+			return
+		end
+
+		local isIssued = false
+		if trackedUnits[unitID].commands then
```
moreginger, line 378: For performance, is it worth assigning `trackedUnits[unitID]` to a local rather than doing it several times?

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: Theoretically that helps. I couldn't measure the difference, but I changed it. (Are there profiling tools besides looking at the continuous output of the profiler widget? I'd like to see e.g. CPU time used in the last 30 seconds.)

LuaUI/Widgets/unit_auto_patrol_nanos.lua
rtwfroody: Same as above.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
moreginger: I don't know, I was just making suggestions from prior Lua encounters (a decent compiler / JIT would be able to make these optimizations automatically). @GoogleFrog can you suggest any tools for profiling?

GoogleFrog:
It looks well structured, but I'm getting bad behaviours.
image
Here I built a Caretaker and sat idle until my resources filled up. Then I build a factory and started making Glaives, assisting the factory with my commander. The Caretaker alternates between reclaiming the nearby trees and assisting the factory. Also, the way it assists with area repair seems to introduce a lag time between Glaives.

GoogleFrog:
Here the Caretaker is letting my metal get high
image
It waits until I have 45 metal before applying a burst of assist.

GoogleFrog:
The widget seems to mistake its own commands for user-issued commands sometimes.

GoogleFrog:
Make the code more readable. Use Spring.Utilities.TableEcho instead of making multi-line logs. Send multiple arguments to Spring.Echo and note that `tostring` is redundant. Use Spring.Echo instead of essentially renaming it to Log. Don't split if ... then across multiple lines until each line is longer than around 140 characters.
```
--	Log("compare")
--	Log("    issued : " .. issued[1] .. "(" ..
--		tostring(issued[2][1]) .. ", " ..
--		tostring(issued[2][2]) .. ", " ..
--		tostring(issued[2][3]) .. ", " ..
--		tostring(issued[2][4]) .. ", " ..
--		tostring(issued[2][5]) .. ") " .. issued[3] .. " (" .. issued[4] .. ")")
--	Log("    current: " .. tostring(currentID) .. "(" ..
--		tostring(currentParam1) .. ", " ..
--		tostring(currentParam2) .. ", " ..
--		tostring(currentParam3) .. ", " ..
--		tostring(currentParam4) .. ", " ..
--		tostring(currentParam5) .. ") " .. tostring(currentOpt))

	if not currentID or not issued then
		return false
	end

	-- Commands that were exactly what we issued.
	if currentID == issued[1] and
			issued[2][1] == currentParam1 and
			issued[2][2] == currentParam2 and
			issued[2][3] == currentParam3 and
			issued[2][4] == currentParam4 and
			issued[2][5] == currentParam5 and
			issued[3] == currentOpt then
		--Log("    -> equal")
		return true
	end
```

rtwfroody:
> Here I built a Caretaker and sat idle until my resources filled up.

Roughly how long did it sit idle for? An idle caretaker should get reissued commands every 1.25--3.75 seconds. (That spread is unnecessarily high. I'll reduce it.)

> Then I build a factory and started making Glaives, assisting the factory with my commander. The Caretaker alternates between reclaiming the nearby trees and assisting the factory. Also, the way it assists with area repair seems to introduce a lag time between Glaives.

That's probably the same issue. When a caretaker becomes idle, the widget always waits 0.5 seconds (idleWaitInterval) before issuing more commands. This is specifically for the case where it's assisting a factory, to give the just-built unit time to walk out of the factory. That number could be tweaked, but I suspect there is no single number that's good for glaives to grizzlies. Possible fixes are to reduce settleInterval, or for the first half second or whatever that a caretaker is idle, issue commands quickly. These are all performance trade-offs, and I don't know how to decide them. What do you think?

> Here the Caretaker is letting my metal get high
> It waits until I have 45 metal before applying a burst of assist.

The current rule is:
```use_metal = futureMetal > 0 and metal > metalStorage * 0.1```
(where futureMetal is the amount of metal you're predicted to have in 10s). This leads to two cases where the caretaker isn't using all metal:
1. You have less than 10% metal. I'll change this to `metal > 0`.
2. The futureMetal computation just looks at current metal use, ignoring what caretakers are actually doing, so it could lead to oscillations where they start/stop assisting. The more caretakers you have, the less of a problem this is, and I don't think it will ever lead to excessing metal. The reason for this check is if you have multiple caretakers, that can both assist and reclaim, you might want some assisting and some reclaiming, instead of all or nothing.

> The widget seems to mistake its own commands for user-issued commands sometimes.

I thought I had that all sussed out. Do you know what it was doing when this happened?

> Make the code more readable.
> Use Spring.Utilities.TableEcho instead of making multi-line logs.

Funny enough, I use the multi-line log statements in the code so that the data in the log file ends up on a single line. That makes for much more readable log files, which is more important to me than concise code.

> Send multiple arguments to Spring.Echo and note that tostring is redundant.

If the variable might be nil, then tostring is required when using .. operator. But maybe not if it's a separate argument to Spring.Echo. I'll take a look.

> Use Spring.Echo instead of essentially renaming it to Log.

Alright. I used Log so I could have a prefix for each line that I can easily grep for, but in practice that's not as helpful as I'd thought.

> Don't split if ... then across multiple lines until each line is longer than around 140 characters.

Just if/then, or do you want all code to be formatted for 140 columns?

**Comment by rtwfroody** [comment]
> I'll change this to metal > 0.

I don't like the behavior that leads to. With one caretaker, factory, and some metal, it spends too much time helping the factory when there's not really enough metal to do that.

This is yet another artifact of not checking often enough. Do you have some guidance on what kind of performance is reasonable, so I can find the number that balances performance with checking as often as possible?

**Comment by thomas11111111** [comment]
make a slider or crank it at max in a test game

**Comment by rtwfroody** [comment]
> Then I build a factory and started making Glaives, assisting the factory with my commander. The Caretaker alternates between reclaiming the nearby trees and assisting the factory.

I looked into this some more. What happens here is that the widget decided to queue area build, area reclaim energy. As long as the caretaker is not idle, the widget will ignore the caretaker until checkInterval has passed. So the caretaker assists in building the glaive. When that is done there is nothing else to assist on, and the area build command completes. Now the area reclaim energy command is being executed. Evidently there was something to reclaim, so the caretaker did that. When checkInterval is up, the widget will see that the caretaker is not doing the first command (area build assist), so reissues the sequence.

If, however, there is nothing to reclaim, then the caretaker will become idle as soon as the glaive has built. Then the widget waits idleWaitInterval before reissuing the commands. What's more, the widget will also never issue commands more often than once per settleInterval. Between those two things, you end up with the problem below:

> Also, the way it assists with area repair seems to introduce a lag time between Glaives.

I'm still waiting on some measurable performance goals. With those, I can spend some time adjusting the numbers to reach the desired performance while having the caretakers perform as good as possible in a variety of situations.

**Comment by GoogleFrog** [comment]
> Roughly how long did it sit idle for? An idle caretaker should get reissued commands every 1.25--3.75 seconds. (That spread is unnecessarily high. I'll reduce it.)

I sat idle, not the Caretaker.

> I'm still waiting on some measurable performance goals. With those, I can spend some time adjusting the numbers to reach the desired performance while having the caretakers perform as good as possible in a variety of situations.

I would hope that improvements can be made to Caretaker behaviour without a noticeable impact on performance. I don't have a particularly good feel for where the draw the line, perhaps aim for 0.1% at 100 Caretakers?

> I don't like the behavior that leads to. With one caretaker, factory, and some metal, it spends too much time helping the factory when there's not really enough metal to do that.

Trying to spend a bit more metal than you have is a desirable state to be in. The Caretaker AI should do as least as well as what we have now. It shouldn't let players float metal since even 50 metal means that you are wasting almost a full Glaive.

A bunch of Caretakers next to a factory, with nothing else nearby, should be constantly assisting the factory. Even if it is making Fleas. If this requires a very short check interval then perhaps something is needed in addition to check intervals.

There is a tradeoff to be made between trying to implement optimal behaviour and confusing players as their Caretakers switching between tasks rapidly and cause economic fluctuation. I think it should be rare for a Caretaker to go idle due to its AI as there is no such thing as pulling too much metal. If there is something nearby to assist or repair then I think the Caretaker should be working on it. If there is something to reclaim then a Caretaker should reclaim it, unless it is metal and the team has no free metal storage. Always adding a patrol command to the end of most Caretaker queues should achieve this.

**Comment by GoogleFrog** [comment]
How about processing pulls such that the widget sees the maximum of the pulls over the last few seconds. This would remove the control shock when a unit exits a factory.

**Comment by rtwfroody** [comment]
This version addresses, I think, all feedback, except for one:
When a unit is assisting a factory and there is sufficient available metal/energy storage, then the unit will spend some time reclaiming metal/energy after the factory has finished building something, before being told to assist the factory again. This does not happen in a "normal" economy where you are limited by metal and have an energy surplus. As a workaround you can give a caretaker a patrol order.

I'm not sure how to fix this without trying to figure out whether there is a nearby factory with a long enough queue.

At times (maybe more than half the time) the widget uses more than .1% of CPU for 100 caretakers. To reduce the CPU usage, simply increase checkInterval. If there's some specific way to measure usage aside from watching the the widget profiler and guessing what the average is, I can take a look at that.

**Comment by rtwfroody** [comment]
I just did a little experiment regarding caretakers getting distracted. To my surprise, the new algorithm pumped out glaives faster than the old one even though it sometimes gets distracted reclaiming energy. The reason is that on the map (lowland crossing) there is also metal lying around, and "patrol" gets distracted reclaiming metal, and doesn't switch back until it's reclaimed an entire rock. The new algorithm gets distracted but only for a second or two before switching back to production.

Then I tried this on SpringMountainDelta, which only has energy reclaim near the starting point and no rocks. Surprisingly, the new algorithm never got distracted reclaiming energy. It did reclaim energy towards the end, when production was metal stalled, and that did not affect build speed at all. The only way I can explain this is if the area build assist command never "completes." But if so, I'm not sure how to explain what GoogleFrog saw initially. But overall I feel pretty good about where the new algorithm is at.

Data, and a graph: https://docs.google.com/spreadsheets/d/1V69-RY_3iPTfhHnPjki_t97JQ1wu698MIOha6eNEu08/edit?usp=sharing

**Comment by GoogleFrog** [comment]
That's sounding good. I should be taking a closer look later this week.

**Comment by rtwfroody** [comment]
Still, at 4 to 5 minutes in https://zero-k.info/Battles/Detail/995008, you can see my caretakers reclaim energy when they could be assisting building a little bit faster instead.

File: LuaUI/Widgets/unit_auto_patrol_nanos.lua [hunk 18]
```diff
@@ -26,25 +49,58 @@ end
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
--- Speeups
+-- Speedups
 
-local CMD_MOVE_STATE    = CMD.MOVE_STATE
-local CMD_PATROL        = CMD.PATROL
+local CMD_PATROL        = CMD.PATROL		-- 15
```
GoogleFrog, line 54: Whitespace.

LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -26,25 +49,58 @@ end
...
 local spGetMyTeamID     = Spring.GetMyTeamID
 local spGetTeamUnits    = Spring.GetTeamUnits
-local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
 local spGetUnitDefID    = Spring.GetUnitDefID
 local spGetUnitPosition = Spring.GetUnitPosition
 local spGiveOrderToUnit = Spring.GiveOrderToUnit
 local spGetGameRulesParam = Spring.GetGameRulesParam
+local spEcho            = Spring.Echo
+local spGetTeamResources = Spring.GetTeamResources
+local spValidUnitID		= Spring.ValidUnitID
```
GoogleFrog, line 73: Whitespace too.

GoogleFrog:
image
- I would like the teal and purple circles to be removed for automatic behaviour, since they are going to be spammed. Perhaps you can give it a massive radius or a 5h parameter to trick the engine into not drawing it.
- Caretakers still switch back and forth between tasks as a group, such as here where I have +10 base income. Ideally each Caretaker would switch progressively until a steady state is reached. I suppose it is fine as-is though. The patrol behaviour doesn't do better.

rtwfroody:
I've addressed whitespace and circles in #4209.

Caretakers do tend to switch as a group sometimes. It does tend to even out if you're building something that takes a long time, but that would be rare. I think the problem is that when a caretaker is issued a command, it does not instantly start doing that command, so even if we look at caretakers on separate frames, we still don't have exact knowledge of what is happening. There are probably small tweaks that could make this better. Maybe a solution is to reassign all caretakers at the same time, instead of doing it on a per-caretaker basis. Something to ponder for future improvements.

GoogleFrog:
Caretakers not instantly taking new actions is unsolvable, due to latency, and the widget should be highly resilient to it.

PR #4131, opened 2020-09-27
make badger have oomph again

Anarchid:
This reapplies the version of the Badger fire sound from "dedup v1", which i find much better in practice, tested by firing 3 Firewalkers, 10 Badgers, and a Tremor in unison.

- The "two-component" issue seems absent in practice, and in any case Tremor does it worse, and spams much harder.
- origins.txt does not need to be changed because that's also a Kingstad sound.
- It sounds way less like a rifle shot or some smack, and much more like a mortar launch.
- It benefits from the increased weapon fire sound volume from the commit that fixes the current version, which i hope fixes the "random patch of white noise" issue (but it doesn't seem to happen to me either using laptop speakers, studio headphones, or bluetooth headphones).


GoogleFrog:
When I try it ingame it just sounds like a clicky-slap or a cartoony balloon pop. The low frequencies are drowned out by the high peak at the start and the rest of the game.  It sticks out from the rest of the sounds. The two components are not heard.

image

A is rarely heard and B should simply be removed for technical reasons.

Here is a clip: https://www.youtube.com/watch?v=1mtknm8nY3c

I have edited the sound to demonstrate what it sounds like in practise: [badger_fire_ingame_approx.zip](https://github.com/ZeroK-RTS/Zero-K/files/5295884/badger_fire_ingame_approx.zip)

Anarchid:
> The low frequencies are drowned out by the high peak at the start and the rest of the game 
Hm, i can distinctly hear them in your video.

>  It sticks out from the rest of the sounds
What doesn't?

> B should simply be removed for technical reasons.
Agreed and inflicted.

> I have edited the sound to demonstrate what it sounds like in practise: badger_fire_ingame_approx.zip
0_o

sprunk:
image
As far as I can hear the GF edit looks like this as far as I can hear. The vertical line is a cutoff and the arc is some sort of scaling to zero to make it soft.

Anarchid:
Fair enough. Comparing it with the current sound now that it's louder, i think the current is sufficiently ok. If this is what it sounds like to people who are not me, that's quite bad.

GoogleFrog:
I could hear A in the right conditions if I paid attention, but compared to the rest of the battle it was drowned out.

Issue #3790, opened 2020-02-14
Add a way to mark allied airpads unusable


sprunk:
See title. People sometimes place airpads in places where you wouldn't want your planes to be. Reef especially has an airpad sidearm that doesn't go together well with its combat proximity.

sprunk:
Some UI proposals.

1) Separate global command which just tags airpads (or untags if already tagged). Simple but causes a bit of a clutter since it only is useful when you're a planes player in multi.

2) Use the retreat command to tag. Reduces clutter but the double use might be confusing and is hard to convey.

3) Make planes require the airpad to be in a retreat zone to be eligible for plane landing; make retreat zones containing an airpad ineligible for land units; make static airpads place a retreat zone by default (optionally? can be toggled userside). This unifies the usage to remove the problem from 2, but overwhelms newbs by placing some confusing circles and the ruleset becomes a bit opaque. Reef becomes only really usable when stationary and requires some retreat zone placement micro since retreat zones don't move alongside it, this may well be a good thing.

4) Same as 3 but also make retreat zones unit-targetable to solve Reef. This has the downside of requiring more mana but a commander or a funnel probably make decent retreat zones.

Sharkgamedev:
@sprunk 
I like these ideas, although
Not saying that I think this is a superior option or anything but I disagree about 1 adding clutter, it's just one pretty small button in the top left of the screen, where there's already some free space IIRC,

sprunk:
The reason there's some free space there is because we've been refraining from successively adding one pretty small button at a time.

Sharkgamedev:
Yeah, that's fair.

GoogleFrog:
How about giving air units an 'avoid airpad' command that can be targeted on an airpad to toggle whether all air units avoid a particular airpad. This uses as much of the command API as possible. Airpads that are to be avoided could have an overhead icon when rearming units are selected (and optionally on shift, like the movestate icons).

Sharkgamedev:
Do you mean something that would show up under the orders tab on an air unit or it would be in with the unit retreat at % priority and other controls area? Either way I like the idea a lot.

sprunk:
Orders tab, not the toggleable states.

GoogleFrog:
Under the orders tab, if it is to be targeted command.

Sharkgamedev:
I really like the idea of having it in the orders tab.

PR #4136, opened 2020-09-30
Mark allied airpad's unusable 

Sharkgamedev:
Fixes: #3790
Everything seems good here, adds a command on all bomber units that allows you to exclude any airpad (Airfac, airpad, reef) and prevent planes from landing on said thing. 

Stability: From every test I've done so far it's good: but I'm going to do a larger-scale test soon to make sure.
Testing with Commshare has been limited so far so I'm going to test that more extensively soon as well.

PS.
There is a custom cursor for the command now as well, I just need to update that gif. Also not in that gif is the icon having more of an offset on the airpad to look a bit nicer.

DEMO:
GithubGifv4

Todo before finished:
- [x] My current method of widget > gadget communication is not up to snuff, so that needs to be fixed
- [x] Display an icon above any air pads that have been marked unusable
- [x] Styling (Sort of I suppose? I'm probably gonna have to revisit this one)
- [x] Make exclusions run per team rather than for everyone
- [x] Make icons per team rather than global.
- [x] Make command not interrupt current queue
- [x] Custom command icon and cursor.
- [x] Use cmdParams instead of Spring.GetMouseState() in the widget to get mouse pos.
- [x] Correctly implement in integral menu culling. I'm thinking hidden by default since this is more of an advanced thing?


LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,836 @@
...
+				  bestPadBookFactor = bookFactor
+			  end
+		  end
+	  end
+	  
+	  Spring.Echo("Find best airpad at coords run")
+	  return bestPadID
+  end
+  
+  local function tablelength(T)
```
GoogleFrog, line 312: I dislike this function and it appears to be unused,

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,836 @@
...
+		  if bookFactor < bestPadBookFactor then
+			  local ux, uy, uz = Spring.GetUnitPosition(airpadID)
+			  if (ux-x)^2 + (uz-z)^2 < r^2 then
+				  bestPadID = airpadID
+				  bestPadBookFactor = bookFactor
+			  end
+		  end
+	  end
+	  
+	  Spring.Echo("Find best airpad at coords run")
```
GoogleFrog, line 308: Errant Echo.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,34 @@
...
+	end
+end
+
+function widget:CommandNotify(cmdID, cmdParams, cmdOptions)
+    if cmdID == CMD_EXCLUDEAIRPAD then
+        findUnitUnderCursour(cmdParams)
+        return true,true
+    end
+    return false -- command not used
+end
```
GoogleFrog, line 34: Needs a newline.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,34 @@
...
+-------------------------------------------------------------------------------------
+
+function widget:GetInfo()
+    return {
+      name      = "Airpad Excluder",
+      desc      = "Handler for exclude airpads gadget.",
+      author    = "SharkGameDev",
+      date      = "2020-09-26",
+      license   = "PD",
+      handler   = true,
```
GoogleFrog, line 11: I don't see how handler is required.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,34 @@
...
+      layer     = 0,
+      enabled   = true,
+    }
+end
+
+VFS.Include("LuaRules/Configs/customcmds.h.lua")
+
+local function findUnitUnderCursour (cmdParams) 
+	--local mouseX,mouseY,mouseZ = cmdParams[1], cmdParams[2], cmdParams[3]
+	local mouseX, mouseY = Spring.GetMouseState()
```
GoogleFrog, line 21: You shouldn't have to get mouse state. Use the appropriate type of command (perhaps ICON_UNIT) and read the parameters.

LuaUI/Configs/integral_menu_culling.lua
```diff
@@ -19,6 +19,7 @@ local configList = {
 	{cmdID = CMD_FIND_PAD              , default = true, name = "Resupply"},
 	{cmdID = CMD.RESURRECT             , default = true, name = "Resurrect"},
 	{cmdID = CMD.UNLOAD_UNITS          , default = true, name = "Unload"},
+	--{cmdId = CMD.EXCLUDEAIRPAD		   , default = true, name = "Exclude"},
```
GoogleFrog, line 22: I don't think this command is important enough to not list in the command culling options. It should appear here.

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,836 @@
...
+	  local msg_table = Spring.Utilities.ExplodeString('|', msg)
+
+	  if msg_table[1] ~= 'addExclusion' then
+		  return --Not a messege for us
+	  end
+
+	  local padID = tonumber(msg_table[2])
+
+	 --Spring.Echo('unit id: ' .. padID)
+
```
GoogleFrog, line 528: padID can be any number sent from luaUI, so spGetUnitDefID(padID) could easily be nil, leading to an error.

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,836 @@
...
+  }
+  
+ local cmdEXCLUDEAIRPAD = {
+	 	id 		= CMD_EXCLUDEAIRPAD,
+		type    = CMDTYPE.ICON_MAP,
+		tooltip = 'Excludes an airpad from the running.',
+		cursor  = 'Repair',
+		action  = 'excludeairpad',
+		params  = { },
+		texture = 'LuaUI/Images/commands/states/divebomb_shield.png',
```
GoogleFrog, line 158: We like to set texture luaUI-side in the integral menu configs. I'm not sure why but it would be worth maintaining that consistency.

LuaUI/Configs/integral_menu_culling.lua
GoogleFrog: Hiding by default sounds ok.

GoogleFrog:
As far as I can tell pad exclusion is not per-team, and not even per-allyTeam. It looks like this command could be used to prevent enemy planes using enemy pads. It should be per-team.

Use the overhead icons framework to display pad exclusion. It is probably best to set a unitrulesparam for whether a pad is excluded by a team, then extend the gadget_icons widget slightly to check for teamID-based unitrulesparams.

LuaUI/Widgets/cmd_airpad_excluder.lua
Sharkgamedev: I was under the assumption that I needed the widget to do unsynced things like the raycast. I'll also need it to display the icon above any excluded pads right?

LuaRules/Gadgets/unit_bomber_command.lua
Sharkgamedev: Should I do the same thing for tooltip?

Sharkgamedev:
Rather than respond to everything separately I'll just do everything in one large chunk as well as a - with their progress:
Right now exclusions are not per-team 
Thanks for the tips on the overhead icons!
Yeah I don't remember what I put table length in there for - removed
Sorry about the errant echos, those were from when I was trying to understand the code - removed
"I don't see how handler is required"
I was under the assumption that I needed the widget to do unsynced things like the raycast. I'll also need it to display the icon above any excluded pads right?
Yeah I was using spring.getmousestate just to make things easier for testing and had planned on replacing it. 
- wip
I was getting a weird error when I had it in integral menu culling, I'll look into it more later. - on hold
Good point, I added a check to make sure padID was not nil. - done
Ok, I'll move the texture over to integral menu config - I'm going to use the one I showed in discord for now  (Option 3 protruding wing and nose) - wip
Thanks for all the feedback!

GoogleFrog:
> I was under the assumption that I needed the widget to do unsynced things like the raycast. I'll also need -it to display the icon above any excluded pads right?

Try it and see. I'm pretty sure handler doesn't do either of these things.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,36 @@
...
+end
+
+VFS.Include("LuaRules/Configs/customcmds.h.lua")
+
+local function findUnitUnderCursour (cmdParams) 
+    local mouseX,mouseY = cmdParams[1], cmdParams[2]
+    Spring.Echo("cmd x: " .. mouseX .. " y: " .. mouseY)
+    local bmouseX, bmouseY = Spring.GetMouseState()
+    Spring.Echo("spring x: " .. bmouseX .. " y: " .. bmouseY)
+	local type, id = Spring.TraceScreenRay(mouseX, mouseY, false)
```
GoogleFrog, line 24: Aren't the params meant to be a unitID now?

GoogleFrog:
Are you still doing commits? Some conversations marked as resolved don't seem to be resolved. Exclusion still seems to be global rather than per-team.

Sharkgamedev:
> Are you still doing commits? Some conversations marked as resolved don't seem to be resolved. Exclusion still seems to be global rather than per-team.

Weird, I committed and pushed a new version just the other day not sure what happened. Exclusions are global right now. Your right - with the unitid being passed through from the command I can get rid of the widget now.

LuaUI/Widgets/cmd_airpad_excluder.lua
Sharkgamedev: If it had committed like it was supposed to it would be that way, something must be up with my git, I'll look into it.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,27 @@
...
+      layer     = 0,
+      enabled   = true,
+    }
+end
+
+VFS.Include("LuaRules/Configs/customcmds.h.lua")
+
+function widget:CommandNotify(cmdID, cmdParams, cmdOptions)
+    if cmdID == CMD_EXCLUDEAIRPAD then
+        if type == "unit" then
```
sprunk, line 21: `type` doesn't seem to be a variable that exists.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,27 @@
...
+            Spring.SendLuaRulesMsg('addExclusion|' .. cmdParams[1])
```
sprunk, line 22: "Add exclusion" is fairly vague (keep in mind every gadget receives this message and decides whether it wants it). How about "addAirpadExclusion" or something along that line?

LuaUI/Widgets/cmd_airpad_excluder.lua
sprunk: No, handler is broadly for manipulating other widgets. Check `WG.icons` or something like that to access the icons API.

LuaUI/Widgets/cmd_airpad_excluder.lua
```diff
@@ -0,0 +1,27 @@
...
+      author    = "SharkGameDev",
+      date      = "2020-09-26",
+      license   = "PD",
+      handler   = true,
+      layer     = 0,
+      enabled   = true,
+    }
+end
+
+VFS.Include("LuaRules/Configs/customcmds.h.lua")
```
sprunk, line 17: Nit: no need to pollute the global namespace with a ton of `CMD_FOO` constants from the `.h` version, use `customcmds.lua`:
```suggestion
local CMD_EXCLUDEAIRPAD = VFS.Include("LuaRules/Configs/customcmds.lua").EXCLUDEAIRPAD
```

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,830 @@
...
-    desc      = "Handles aircraft repair/rearm",
-    author    = "xponen, KingRaptor, GoogleFrog",
-    date      = "20 April 2014, 25 Feb 2011",
-    license   = "GNU LGPL, v2.1 or later",
-    layer     = 0,
-    enabled   = true  --  loaded by default?
+	return {
+	  name      = "Aircraft Command",
+	  desc      = "Handles aircraft repair/rearm",
+	  author    = "xponen, KingRaptor, GoogleFrog",
```
sprunk, line 19: Mixed tabs and spaces (was 4 spaces, now it's tab and two spaces)

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,830 @@
...
-		local bookFactor = airpadsData[airpadID].reservations.count / airpadsData[airpadID].cap
-		if bookFactor < bestPadBookFactor then
-			local ux, uy, uz = Spring.GetUnitPosition(airpadID)
-			if (ux-x)^2 + (uz-z)^2 < r^2 then
-				bestPadID = airpadID
-				bestPadBookFactor = bookFactor
-			end
-		end
-	end
+	  local padID = tonumber(msg_table[2])
```
sprunk, line 519: What if somebody sends the message `addExclusion|-1`? What if somebody sends `addExclusion|pwned`?

LuaUI/Widgets/cmd_airpad_excluder.lua
Sharkgamedev: Noticed that and removed it in an earlier commit that had yet to be pushed.

LuaUI/Widgets/cmd_airpad_excluder.lua
Sharkgamedev: Widget has been removed entirely in latest commit.

LuaUI/Widgets/cmd_airpad_excluder.lua
Sharkgamedev: Should've pushed my last commit sooner I guess, anyway resolved with the removal of the widget entirely.

LuaRules/Gadgets/unit_bomber_command.lua
Sharkgamedev: Fixed by removing the need to send messeges by just moving the command logic over to the gadget, I will add in a check to make sure a number is in fact a unit as well.

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,830 @@
...
-	for airpadID,airpadDefID in pairs(airpadsPerAllyteam[allyTeam]) do
-		if spGetUnitIsDead(airpadID) or (not airpadsData[airpadID]) then --rare case. Can happen if airpad is built & die in same frame
-			if (not airpadsData[airpadID]) then
-				Spring.Echo("Warning: airpadsData for " .. airpadID .. " is NIL")
-			end
-			airpadsPerAllyteam[allyTeam][airpadID] = nil
+	  --Check if the unit is an airpad and if it already exists
+	  if airpadDefs[spGetUnitDefID(padID)] then
+		if not excludedPads[padID] then
+		  excludedPads[padID] = true
```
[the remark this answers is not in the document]
Sharkgamedev: Yeah that's the next big thing I need to do...

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -13,786 +13,830 @@
...
+	  id      = CMD_REARM,
+	  name    = "Rearm",
+	  action  = "rearm",
+	  cursor  = 'Repair',
+	  type    = CMDTYPE.ICON_UNIT,
+	  tooltip = "Select an airpad to return to for rearm",
+	  hidden	= true,
+  }
+  
+ local cmdEXCLUDEAIRPAD = {
```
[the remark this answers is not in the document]
Sharkgamedev: Renamed!

GoogleFrog:
To make exclusions per team then you need some table of the form exclusion[teamID][unitID].

GoogleFrog:
What is the yellow dot in the gif above, and shouldn't the command have a cursor?

Sharkgamedev:
@GoogleFrog I was wondering what would be a good cursor for this? The yellow dot may just be me having forgotten to turn of the show clicks option in my gif recorder. And thanks for the guidance on making it per team!

GoogleFrog:
Some cursors are just a little white arrow in the top left and the command picture. Perhaps you could take the arrow and make a cursor.

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -511,26 +512,25 @@ function gadget:GetInfo()
...
+  local function addOrRemoveExclusion(padID, teamID)
+	  if padID == nil or not Spring.ValidUnitID(padID)then --Also check if it's a valid unit
 		return
 	  end
 
-	  local pInfo = Spring.GetPlayerInfo()
-	  local teamID = pInfo[3]
-
-	  Spring.Echo(teamID)
+	  if not Spring.GetUnitTeam(padID) == teamID then
```
sprunk, line 520: You're looking for "allyteam" here. A "team" usually corresponds to 1 player unless they do a squad merge - think of it as the same colour; a merged squad has all units (say) green so they're the same teamID, your allies have blue units so they are a different teamID, but still the same allyTeam

sprunk: Also `if not x == y` → `if x ~= y`

LuaUI/Widgets/api_gadget_icons.lua
```diff
@@ -214,6 +221,24 @@ function SetIcons()
 			end
 		end
 
+		local padExcluded = excludedPadUnitDef[unitDefID] and Spring.GetUnitRulesParam(unitID, "padExcluded")
```
sprunk, line 224: This should probably be done via widget icons and not gadget icons because a unit rules param does not discriminate between teams

LuaRules/Gadgets/unit_bomber_command.lua
Sharkgamedev: I wouldn't want to exclude it for the whole ally team though right? The whole point of this is so that you can prevent your planes from landing on your allies airpads, but not prevent you allies planes from landing on said airpads?

LuaUI/Widgets/api_gadget_icons.lua
GoogleFrog: Whether a pad is excluded is determined by gadgets, not widgets, and remembering the state that was sent wouldn't survive luaui reload. Why not just append a teamID to padExcluded?

sprunk: Sounds good

LuaRules/Gadgets/unit_bomber_command.lua
sprunk: Yes, that is the point, but this function is about adding airpads to the exclusion (so should work on the whole "ally team") and not about deciding whose planes actually obey that exclusion (which should be just the "team" that sent it).

```suggestion
	  local allyTeamID = select(6, Spring.GetTeamInfo(teamID, false))
	  if Spring.GetUnitAllyTeam(padID) ~= allyTeamID then
```

Sharkgamedev: Ah ok, that makes sense now!

LuaUI/Widgets/api_gadget_icons.lua
Sharkgamedev: Do you mean setting the unitrulesparam value to the teamID that wants to see it?

sprunk: No, incorporating the teamID into the param key like this:
```suggestion
		local padExcluded = excludedPadUnitDef[unitDefID] and Spring.GetUnitRulesParam(unitID, "padExcluded" .. Spring.GetMyTeamID())
```

LuaRules/Gadgets/unit_bomber_command.lua
```diff
@@ -10,789 +10,840 @@
...
+	  local allyTeamID = select(6, Spring.GetTeamInfo(teamID, false))
+	  if Spring.GetUnitAllyTeam(padID) ~= allyTeamID then
+	     return
+	  end
+
+	  --Check if the unit is an airpad and if it already exists
+	  if airpadDefs[spGetUnitDefID(padID)] then
+		if not excludedPads[teamID][padID] then
+		  excludedPads[teamID][padID] = true
+		  Spring.SetUnitRulesParam(padID, "padExcluded", 1)
```
sprunk, line 530:
```suggestion
		  Spring.SetUnitRulesParam(padID, "padExcluded" .. teamID, 1)
```

GoogleFrog:
This should be squashed as I am running into unreasonable problems when trying to checkout or pull it.

GoogleFrog:
Hmm I guess I can do it directly, but it requires merging. I'll do it to squash.

Sharkgamedev:
Do you still need it to be squashed?

PR #4138, opened 2020-10-07
Add dvorak grid option

floomby:
Added the option for dvorak in the grid selection to remove the need for setting a custom grid. Additionally to facilitate the change special character support was added for the predefined grids.


LuaUI/Widgets/gui_chili_integral_menu.lua
```diff
@@ -192,6 +193,7 @@ options = {
 			{name = 'QWERTY (standard)',key = 'qwerty', hotkey = nil},
 			{name = 'QWERTZ (central Europe)', key = 'qwertz', hotkey = nil},
 			{name = 'AZERTY (France)', key = 'azerty', hotkey = nil},
+			{name = 'DVORAK (standard)', key = 'dvorak', hotkey = nil},
```
sprunk, line 196:
```suggestion
			{name = 'Dvorak (standard)', key = 'dvorak', hotkey = nil},
```

LuaUI/Widgets/gui_chili_keyboardmenu.lua
```diff
@@ -154,6 +154,7 @@ options = {
 			{key='qwerty', name='QWERTY', },
 			{key='qwertz', name='QWERTZ', },
 			{key='azerty', name='AZERTY', },
+			{key='dvorak', name='DVORAK', },
```
sprunk, line 157:
```suggestion
			{key='dvorak', name='Dvorak', },
```

sprunk (commented):
Looks fine

LuaUI/Widgets/gui_chili_keyboardmenu.lua
floomby: Your right that all caps is probably not what is wanted. Do you prefer that I amend the original commit, or just commit the suggestions from the github ui and let you squash when you merge?

sprunk: I don't mind either way.

sprunk:
Thanks!

PR #4141, opened 2020-10-12
Changed airpad model and texture

darloth:
Now uses atlas texture, proper tex2 and normals, looks better.

Shot of it working in action via mutator: image showing new airpad


GoogleFrog:
It should have a new aoplane and buildpic. There is some gadget that makes buildpics, I think you can find it by searching LuaRules for 'buildicon'.

There is a spring tool that builds aoplanes for s3o. Maybe you'll have to convert the model https://springrts.com/phpbb/viewtopic.php?t=24006

sprunk:
The mesh has tall antennas but its height seems defined not to include them (which e.g. causes clipping with unit icons):
image

Touching the model is probably a good opportunity to fix that (either by removing the antennae, or increasing the height).

sprunk:
See https://github.com/ZeroK-RTS/Zero-K/blob/master/Objects3d/Lordmuffe_Pene_dead.dae.lua#L3 for an example of how to edit height for dae.

sprunk:
Another minor nitpick: this is a good opportunity to rename the model file `staticrearm.dae`.

units/staticrearm.lua
```diff
@@ -1,4 +1,4 @@
-return { staticrearm = {
+unitDef = {
```
sprunk, line 1: Why change the `return { staticrearm = { ... } }` surrounding the file into `unitDef = { ... } `return lowerkeys({ staticrearm = unitDef })`?

darloth: Two reasons, both of which were sadly mistakes!
1. The base mod you could download had the bottom structure with the lowerkeys.
2. The top structure crashed the game. (But only in the older version that locally selected mods used by default, retrieved with the depends={[[rapid:zk:stable]] string in modinfo.lua)

Unfortunately, when used on the latest version, the bottom version crashes the game.  It's probably a separate issue that I don't think either of these should just -crash- because of casing errors when the spring development wiki specifically says things should be lowercased... See [https://springrts.com/wiki/Gamedev:UnitDefs#Details](https://springrts.com/wiki/Gamedev:UnitDefs#Details) for that suggestion.

I will fix these so they work in the latest ZK version, but this is why I did it this way and thought it was ok.

Anarchid:
probably also make sense to rm `airpad.s3o`

darloth:
Updated for proper version, tested in dev mode with full sources, airpad.s3o already removed (check commit https://github.com/ZeroK-RTS/Zero-K/pull/4141/commits/64ba2bf684f07d39053a6ae33d7da6a562035a71) so I think (hope) it's done properly this time.

Anarchid:
Right, yes, i didn't see the deletion diff.

GoogleFrog:
> error: Your local changes to the following files would be overwritten by checkout:
Objects3d/airpad.s3o
Please commit your changes or stash them before you switch branches.
Aborting

Is this to do with your git fight? I guess I'd better just pull then test.

GoogleFrog:
Pulling this made pulling to my local version impossible.

PR #4147, opened 2020-10-16
Detriment Rework

TechAUmNu:
This pull request attempts to turn the detriment into the powerful assault strider it should be. 

- Can now jump. Jump is the biggest change and adds a way to engage with shock and awe. Players must be careful though as it will cause some damage when taking off and significant damage to units when landing. The reload time is high so it can either be used to engage deep into the backline and take out eco / superweapons, or to escape from being overwhelmed (but not both).

- Snitch cannon. Replacement for the rocket launcher, which fires a volley of 12 snitches in a high arc. Travel time is significant to allow units to move out the way. Mainly designed for weakening defensive porc lines.

- Improved DPS. The Gauss cannons have been buffed to 1250dps each. Now appropriately melts most units, still struggles against large numbers of raiders.

- HP and regen buff. HP increased to 100000 and regen to 120/s. Lets it actually tank at the front of an army without immediately exploding and destroying said army! Also lets it tank a few more ultimatum shots.

- Radar and sonar added. Lets it work a better alone.

- Explosion nerfed. This is more of a buff than a nerf as generally the nuclear explosion would take out most of your own army. Reducing the severity of the explosion allows the other weapons to be more powerful.

- Improved animations. Gauss animations now work properly, walking doesn't leave a bizzare trail and instead leaves a terraformed path. Walking now causes minor damage to units that go directly under the feet. (You can also see it coming)

I have probably missed something!

Terrible video showing some of this: https://www.youtube.com/watch?v=sf09G398OdE
Please try out the changes and let me know what you think!


Licho1:
Hah, well the jump and snitches look fun, but for 20k this would be some crazy OP unit breaking the rule that bigger units are not more cost effective than small units.

Sgathec:
Sorry, I don't like the beginning of the jump because the rocket engine visuals don't look coherent. The jump could also be improved by lowering the trajectory considerably. I wish a jet could be seen almost until the end as well.

TechAUmNu:
> Hah, well the jump and snitches look fun, but for 20k this would be some crazy OP unit breaking the rule that bigger units are not more cost effective than small units.

glaves, ulti, minotaur beats it easily... It really struggles with groups of units surrounding it as the turn rate is slow. How can you make that argument against drp or zenith? On larger maps with long games, how are you supposed to reach those superweapons? 

Fair enough it might need to increase cost a little to say 25k and some stuff definitely needs nerfed. but its nowhere on the same level of OP as the other supers.

TechAUmNu:
> Sorry, I don't like the beginning of the jump because the rocket engine visuals don't look coherent. It could also be improved by lowering the trajectory considerably. I wish a set of jets could be seen until the jump is ending as well.

Lowering the trajectory would make the impact very out of place. Its supposed to be boosting up into the air then freefalling back down, which is why there is no landing burn. (The jump code doesn't appear to let you do a freefall though)

What do you mean by not coherent? In rockets there is a distinct section of pure flame before any smoke appears. If smoke appears immediately it just looks wrong. Maybe it could be slightly thinner at the start you mean, at the transition from flame to smoke?

Here is video just showing jump from the side: https://youtu.be/hbIjhpneKPw?t=16
Alt Text

Sgathec:
Yes, I suppose it should preferably burn units while landing without causing the shock. The visuals weren't coherent since I saw two "jets" that looked separate a few seconds after the Detriment started rising.

GoogleFrog:
Walking
- The changes here implicitly make Detriment less stealthy. That isn't necessarily bad, but the ability to hide a Det is one of the advantages that it has over the superweapons in FFA.
- The step sound and global ground decal makes a moving Detriment technically as visible as an activated Starlight.
- I'm not so sure about the terraform. It seems very minor and there is no way to make the Det stamp on the spot. I expect that the terraform would make Det stuck more often than it does anything useful. I think some sort of terraform-assault ability should be on a different unit.

image
The unit gives off damage or a poinpointable effect while cloaked, so it should be decloaked momentarily each time it steps.

image
It is possible to get the Det stuck through its own terraforming and then have the walk animation jitter, spamming out damage, sounds and particles. if the footstep is to be linked to damage then the footstep needs to be more mechanically consistent than usual footsteps.

image
I was unable to make the footstep deal damage in normal situations. Furthermore, it would seem to me that the footstep damage would harm allies a lot more than it harms enemies because enemies can avoid swarming around Dets feet fairly easily, while this is not usually expected of allies.

GoogleFrog:
Minor:
- Give the Jump UI an AoE indicator like the one that appears for Jugglenaut.
- Jump commands for jumplegs units are cancelled if they are midair instead of queued. This should be fixed separately at some point.

GoogleFrog:
Jumping
- There is too much smoke on launch and impact. The launch smoke spreads in a way that clashes with the scale of the game (the smoke makes the space of the game look smaller). Players need to see what is going on in the relevant areas.
- Give the Jump UI an AoE indicator like the one that appears for Jugglenaut.
- I dislike how the jump has the visuals of jumpjets and the mechanics of jumplegs.
- Aiming and firing can interrupt parts of the jump animation. If Det is meant to need its arms in a particular arrangement to jump then it should be unable to do anything else with its arms. I'm unsure whether the animation should override aiming though.
- The pre-jump animation doesn't do enough to make it look like the Det has acknowledged the command.
- Minor: Jump commands for jumplegs units are cancelled if they are midair instead of queued. This should be fixed separately at some point.

image
- Perhaps cegs could come out of the seams around its back spines and around its feet.
- It could lock its legs and hunker down to some extent, rather than moving its arms.
- Jumpers don't need to rotate to the direction they are jumping. Det would probably look better if it rotated, in which case it should do a walking on the spot stamp-stamp as it turns (and then probably locks). If there is rotation then it should add to the time required to jump.
- Since jump is so important and has such a long reload time, I think it should be easy to see whether Det can jump based on its model. At the moment I'm thinking that its spines could fully retract into its back when it uses jump, and the cap could move down to hide the interior light. As it nears its recharge the spines should move back out until they are fully extended, and the caps should reveal the interior light.

image
The impulse on the launch is ridiculous.

image
The impulse on landing is similarly ridiculous. Perhaps it is more justified.

image
The turning to jump angle animation looks unnatural, and it doesn't turn to the correct angle.

GoogleFrog:
Combat
- I'm not sure that very much actually beats Det for cost.
- 24 Minotaur lose to a retreating no-ability Det, leaving it at 43% health.
- 9 Cyclops do even worse vs a retreating no-ability Det, but they do slow it fairly quickly. Perhaps a combination would win,
- Ultimatum kills Det in 5 well-connected shots, which takes 6s. Det kills Ulti much faster than that.
- 27 Nimbus beats Det, but depends on the non-existence of AA.
- I think aggressive Det jump will be an auto-win in teams unless the other team has significantly invested in an anti-Det air force.
- I don't think it is possible to kill a defensive Det short of a well coordinated multi-Ulti strike.

Weapons
- I think the reduced number of standard weapons (four down to just one) is an unnecessary reduction of coolness.
- Det is a lot weaker to terraform without its smoothing missile and I think the AA lasers may as well exist.
- The face laser seemed to work and could exist just to be cool.

GoogleFrog:
Update: I realised that the foot stamp is cratering, not smoothing. This is not going to make the terrain more passable.

GoogleFrog:
The jump and manualfire irks me in a particular way and I think I've pinned it on their arcs. Both abilities push the 'playable space' of the map higher for relatively local abilities.
image
At this zoom level you can take in the effects of Tremor and Impaler. There is also a mid-air Det above this area of the map, but it is too high to see.

image
At this zoom level you can comfortably see Impaler shooting at max range and control the tactics of a battle. The game hasn't turned into icon wars. I don't want to force people to zoom out further than this to manage a battle. If battle-relevant projectiles or jumping arc higher than this zoom level then people will have to zoom out further and it will 'break' the scale-feel of the game. There is some argument that Det could be at the upper end of the scale, but I don't think it should break it.

image
Most tacnukes reach about as high as Impaler. Nonsense with Placeholder and/or Lobster caps out at this level. As far as I recall, the only things that can happen above Impaler level are:
- Shockley.
- Newton stuff.
- Nuke.
- Superweapons.

All of these things are global, apart from Shockley. If you're dodging a nuke or superweapon then zooming out to do so is acceptable. The Det bombs are much more local than Shockley and the jump ability may end up having to be more local. In any case, Shockley tends to be targeted at structures (although not always) and is less impactful than a Det landing.

image
Here are the bombs.

GoogleFrog:
The jump launch ceg smoke seems to jitter up and down.

Anarchid:
> I don't think it is possible to kill a defensive Det short of a well coordinated multi-Ulti strike.

Ultimatum cannot aggress against a Det, because the Det is faster. Detriment must walk itself into a vulnerable position and engage its guns elsewhere to become vulnerable.

GoogleFrog:
I didn't say "aggress", I said kill. Provided your jump is up I doubt a static lone Det can be killed by anything but a well coordinated multi-Ulti strike.

Also I thought of a few more things.
- I like the singu explosion on the current Det because it makes it a risk to your army and makes suiciding it into defenses a real choice. Det having a drawback when working with your army is working as intended.
- The jump range should probably be at most half of what it currently is. This would deal with the high-arc issue, and I think the extreme jump would make Det essentially an autowin in teamgames.

Licho1:
In any case, I really don't want ZK to become tech rush game, det must not be more efficient per cost than reasonable set's of counter units. And that set must not contain just the ultimatum, because then it's a hard counter.

Detriment with such capabilities looks like something that should be in 40k price range, it's already quite cost efficient unit even if you look at raw stats HP/cost etc.

TechAUmNu:
- Regen 120 -> 60, Speed 1.6->1.2 (original speed), Turnrate 482->400 
- Death explosion added back
- The Gauss now are burst again with much reduced dps, fire slower and more damage than original ones but dps only 5% increased. 
Old: 200d, 3x0.2s burst, 1.2s reload, 600 range  -- 666 combined dps
New: 500d, 3x 0.6s burst, 2.5s reload, 550 range -- 697 combined dps
- Face laser and AA added back.
Face laser damage increased,
Old: 600d, 3x beams, 6s reload, 600 range  -- 300 dps
New: 1000d, 3x beams, 6s reload, 700 range -- 500 dps
- [ ] Stop face laser firing out of sensible arc. Currently it can fire backwards through the head which makes killing with an ultimatum from behind rather hard,
- Bomb launcher replaced with missile launcher (inaccurate burst). 
I really wanted to have something using starburstlauncher, so it had a more sensible trajectory. However I can't find any way to make starburst anything but pinpoint accurate? spray angle, accuracy, etc, don't change where individual projectiles in a burst land.

- Walk stomp removed and instead just fixed the existing track to have the correct gate. 
- [ ] Make new track decal to fit the actual size of feet
- [ ] Fix walking animation breaking when messed about with newtons etc 

I have toned down the jump a lot based on your feedback. 

- [x] Range 3000 -> 1200, Height 1500 -> 700
- [x] Smoke is now not ridiculous
- [x] AOE indicator on landing, removed takeoff explosion
- [x] Landing impulse turned down to a sensible level
- [x] Landing damage reduced x3, it now kills light units it lands directly on and pushes surrounding units slightly. Rather than vaporising giant armies by pushing units at warp speed through them.
- [ ] Fix turn animation and add hunker before jump
- [ ] Stop firing when during jump
- [ ] Add some animation to show when jump is available

GoogleFrog:
- Your first gauss configuration seemed ok, but the new one could be better. I will have to check it.
- If you can make stomp work technically, then I think it could be at least worthwhile to have Det make a stompy dirt CEG appear when it lands. This CEG would have to be LOS restricted. It would imply that Det cannot be cloaked while walking, but I think that could be a fine mechanic to add.
- Maintaining a StarburstLauncher pattern in the presence of 3D terrain is tricky as the missiles may spread out or miss depending on how they are given their targets. Do the missiles fly out to hit a horizontal line in space, or do they fly out to hit a straight XZ line that is snapped to the terrain? The second option strikes me as a bit too artificial while the first may be hard to use. You can experiment by setting the projectile target when it launches.
- If you want wobbly missiles then perhaps do something like Merlin or Paladin.

Wobbly missiles may be a bit overused among striders. It looks like the Det back gun is supposed to be able to ride a rail onto the Det's shoulder and turn into some sort of direct-fire weapon. Perhaps the manualfire could come from a shoulder-mounted gun, and when manualfire is unavailable the gun would ride the rail to position itself on the back.
- Does the weapon need to be highly dodgeable?
- I've been thinking some sort of AoE inaccurate burst of laser fire could work, probably with terrain smoothing.

Licho1:
Terrain smoothing laser sounds fun, but is it right for det to have so big versatility to avoid terra traps?

TechAUmNu:
Have added animations for the head laser (pops up to fire) and the shoulder cannon now rotates round to fire.

There are a bunch of weapons that I was messing around with, haven't really decided on what looks best. If you want to try them they go on weapon 5.
Weapons: 
LAZER - the crater laser from starlight
LASERSPRAY - very silly laser 
OBLITERATOR - deletes one unit
BOMB_LAUNCHER - giant shotgun
FLECHETTE - giant shotgun

GoogleFrog:
I'll take a look soon, but there is a lot going on so this may not make it into next stable.

izirayd:
In style, the detriment should have huge jet turbines on its back, similar to a jump com

GoogleFrog:
I'll apply some polish.

Issue #3515, opened 2019-04-09
Stinger model exposes inside of mesh


DeinFreund:
It's possible to see into the back of the Stinger's barrel, exposing the backside of some faces.

![picture](https://i.imgur.com/QQPIFsE.jpg)

PR #4150, opened 2020-10-19
Updated Stinger model and texture mapping

darloth:
Much fancier looking barrel, far fewer mesh errors and no floating geometry with backfaces showing.

Old: old stinger barrel

New: new stinger


sprunk:
Fixes #3515 ?

darloth:
Yes, it does fix #3515 .

GoogleFrog:
I'm disappointed because I asked for the colvol to be unchanged, and it clearly isn't. This Stinger is immune to torpedo weapons.

GoogleFrog:
image

image

GoogleFrog:
image

image

GoogleFrog:
The new model looks better close up, but the old one has clearer shapes and lines at playing zoom levels. The fins on the barrel and neck make these shapes blur into adjacent areas.

darloth:
Thanks, that feedback is useful.  I'll try and produce before/after shots like that in future as well, I can see the value in that now you've demonstrated it.  Can easily remove the fins on the neck, and at playing zoom yes they're not as good as I hoped.  Could remove the ones on the barrel but I think they look ok even at play zoom.  I'll make them thicker but more obvious and perhaps teamcolour them?

The collision volume was not 'changed' but has probably moved up a bit.  I'll make sure it gets moved back down.  I've been putting the midpos in the middle and setting the radius to half the width because that's what the spring docs say they ought to be, which is apparently wrong by comparison to most of the other ZK structures - I will record this on the wiki somewhere.

Sorry for not testing an amphibious building against torpedoes, I should have caught that.  I'll write that into a list of things to test as well, I think - it's easy to forget, but it needs to not be.

GoogleFrog:
Midpos is used for many things and the spring docs cannot be right about how to make the tradeoff between the effects of different midpos positions.

sprunk:
Any advances?

darloth:
Family crisis followed by me completely forgetting about it, sorry.

If people are still interested I'll pick it up this week or next, or if I haven't done it by end of next week then you can assume I'm too busy to ever generate anything useful and close it?

GoogleFrog:
Sounds good.

darloth:
Reproduced the bug in a development copy, fixed the visual issues I think (screenshots will come after I've finished fiddling with midpos and got the torpedoes actually hitting it).

In short, progress is being made.

darloth:
'height' is ignored, far as I can tell.  'midpos' adjusts the collision volume origin but it's always based on the derived size of the model, there seems to be no way to actually change the collision volume. 'midpoint' (which was what was there for some reason) is completely ignored and was probably just me making a stupid mistake.

darloth:
Screenshots pending later today / tomorrow, then it should be ready to merge.  Right now it -might- be, but I recommend leaving for screenshots and having a look yourself due to edits and of course the previous problems.  Think I've fixed them now though.

darloth:
Original model collision volume: orig side view

New model collision volume: new side view
(Might be very slightly lower down by an elmo or two at most, but torpedoes certainly hit this one.  Is the same size as the original, but placement was done manually by eye)

Original vs Top view comparison:
anim top view

I believe this is now ready to merge, but saying that, I thought so last time too -_-

darloth:
If someone could check I merged master into this branch properly that would also be good - I relied on github saying "I can do this for you".  Thought I ought to because it was two years behind so I wanted to test it on latest.  Seems to work, but Faraday looks weird as you can probably see from the screenshot.  I haven't touched Faraday though so I don't think this was me, and I didn't do anything of note to the textures either, so I've no idea what might have affected it.

GoogleFrog:
image

Broken offset but otherwise seems to work.

GoogleFrog:
image

After merging it now aims incorrectly. What.

PR #4153, opened 2020-10-22
Revert "Revert "Changed airpad model and texture""

GoogleFrog:
Reverts ZeroK-RTS/Zero-K#4152


GoogleFrog:
image

image

The colvol is terrible here but I guess I'll fix it.

PR #4162, opened 2020-10-31
Claymore no longer fires at targets in the air

### LINKED ISSUE #4158: Claymore ground-fires at units that are flying through the air
Also check whether Claymore prediction is broken.

Arch-Shaman:
Fixes #4158

I'm not sure this is the right route to take. I got no idea where the ground fire code is, but this seems to stop the behavior.


GoogleFrog:
Is this using base position or mid position? The leeway of 15 seems large for base position.

Arch-Shaman:
15 during testing was enough  to differentiate between things that get tossed up a little and things that are presently flying. I can test 5 and 10 too if you'd like.

As per base position of midposition, it's base position according to the wiki. (We're using GetUnitPosition which defaults to base position)

Arch-Shaman:
Looking at the other stuff within the script, 5 seems to probably be a better leeway.

PR #4172, opened 2020-11-06
Add stardust overshoot AI

### LINKED ISSUE #4083: Consider Stardust overshoot AI
Currently Stardust can tickle enemies on the edge of its range by force firing nearby, causing burnblow AoE to deal some damage. Stardust should have this ability automated and its range reduced by the appropriate amount (so that Ronin can easily walk into range and fire).

Arch-Shaman:
Implements/fixes #4083.

TODO:
* [ ] Fix blocking issues with unit implementation.
* [ ] Possibly use settarget gadget to implement single weapon use for dante and desolator's UW glancing. (Current function uses all weapons, which may not be desirable for some units)

Unit implementation and configuration:
* [ ] crab -- velocity gate: 1.8
* [ ] shogun
* [ ] banisher
* [ ] venom
* [ ] Dante's flame rockets?
* [ ] Reaver

### DIFF
--- a/LuaRules/Gadgets/unit_overshoot.lua
+++ b/LuaRules/Gadgets/unit_overshoot.lua
@@ -16,12 +16,11 @@ end
 
 local overshootdefs = {} -- unitdefs to watch
 local overwatch = {} -- units that have the ai enabled.
-local watchlist = {} -- list of all the unitdefs this has
-local updaterate = 5
-local terrainupdaterate = 45/updaterate
+local unitstocheck = {}
+local updaterate = 5 -- the rate we update units.
+local terrainupdaterate = 12 -- update range checks every 12 updates. EG: update rate of 5 means we update ranges every 60 frames.
 
 local spSetUnitTarget = Spring.SetUnitTarget
-local spGetUnitWeaponTestTarget = Spring.GetUnitWeaponTestTarget
 local spGetCommandQueue = Spring.GetCommandQueue
 local spGetGroundHeight = Spring.GetGroundHeight
 local spGetGroundOrigHeight = Spring.GetGroundOrigHeight
@@ -32,6 +31,9 @@ local spGetUnitDefID = Spring.GetUnitDefID
 local spGetTeamList = Spring.GetTeamList
 local spGetTeamUnitsByDefs = Spring.GetTeamUnitsByDefs
 local spUtilitiesGetEffectiveWeaponRange = Spring.Utilities.GetEffectiveWeaponRange
+local spGetUnitVelocity = Spring.GetUnitVelocity
+local spGetUnitSeparation = Spring.GetUnitSeparation
+local spGetUnitWeaponTestRange = Spring.GetUnitWeaponTestRange
 local EMPTY = {}
 
 for i = 1, #UnitDefs do
@@ -42,20 +44,25 @@ for i = 1, #UnitDefs do
 			local WeaponDefID = weapons[w].weaponDef
 			local WeaponDef = WeaponDefs[WeaponDefID]
 			local cp = WeaponDef.customParams
-			if cp and cp.overshoot then -- NOTE: overshoot field is a multiplier to unit range. Use 1 for things that aren't like stardust. 
+			if cp and cp.overshoot then
 				c = c + 1
 				if overshootdefs[i] == nil then
 					overshootdefs[i] = {}
-					watchlist[#watchlist + 1] = i
 				end
-				overshootdefs[i][c] = {id = w, aoe = WeaponDef.damageAreaOfEffect, bonus = tonumber(cp.overshoot), range = WeaponDef.range}
+				local bonus = 0
+				if WeaponDef.selfExplode then
+					bonus = WeaponDef.projectilespeed * 2 -- this is an attempt at computing the max travel of a bullet. This is the travel distance within 4 frames.
+				end
+				if cp.overshoot_override then
+					bonus = tonumber(cp.overshoot_override)
+				end
+				overshootdefs[i][c] = {id = w, aoe = WeaponDef.damageAreaOfEffect, bonus = bonus, range = WeaponDef.range, velocity = WeaponDef.projectilespeed}
+				if cp.overshoot_maxvel then
+					overshootdefs[i][c].maxvel = tonumber(cp.overshoot_maxvel)
+				end
 			end
 		end
 	end
-end
-
-local function GetDistance(x, z, x2, z2)
-	return math.sqrt(((x2 - x) * (x2 - x)) + ((z2 - z) * (z2 - z)))
 end
 
 local function GetLowestHeightOnCircle(x, z, radius, points)
@@ -77,10 +84,25 @@ local function GetLowestHeightOnCircle(x, z, radius, points)
 	return lowest
 end
 
+local function RemoveUnitFromList(unitID)
+	if overwatch[unitID] == nil then
+		return
+	end
+	local index = overwatch[unitID].index
+	if index == #unitstocheck then
+		unitstocheck[index] = nil
+	else
+		table.remove(unitstocheck, index)
+	end
+	overwatch[unitID] = nil
+end
+
 local function GetFirePoint(radius, x, z, targetx, targetz)
-	local angle = Spring.Utilities.Vector.Angle(targetx - x, targetz - z)
-	local tx = x + (radius * math.cos(angle))
-	local tz = z + (radius * math.sin(angle))
+	local dx = targetx - x
+	local dz = targetz - z
+	local mag = 1 / math.sqrt((dx * dx) + (dz * dz))
+	local tx = x + (radius * dx * mag)
+	local tz = z + (radius * dz * mag)
 	return tx, tz
 end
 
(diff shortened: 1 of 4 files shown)


LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+		for id, data in pairs(overwatch) do
+			local unitdefID = data.def
+			if #spGetCommandQueue(id, 1) == 0 then -- has no active command, so feel free to engage. Note this check is mostly so that users can override the AI.
+				for weapon, _ in pairs(overshootdefs[unitdefID]) do
+					UpdateUnitTarget(id, unitdefID, weapon)
+				end
+			end
+		end
+	end
+end
```
GoogleFrog, line 143: Newline

GoogleFrog:
This clashes with SetTarget, so probably only works for static units. This seems fine though.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+	angle = -angle + math.rad(270) -- I hate trig in spring. Don't ask me how this works, it just does.
+	--local angle = Spring.Utilities.Vector.Angle(x - targetx, z - targetz) -- this doesn't seem to work, gives wrong angle.
+	local tx = x + (radius * math.sin(angle))
+	local tz = z + (radius * math.cos(angle))
+	return tx, tz
+end
+
+local function AttackPosition(unitID, x, y, z, weaponID)
+	local testresult = spGetUnitWeaponTestTarget(unitID, weaponID, x, y, z)
+	if spGetUnitWeaponTestTarget(unitID, weaponID, x, y, z) then -- probably needs some friendly fire test
```
GoogleFrog, line 85: This isn't required. SetUnitTarget is at a higher level than fire tests. It causes the unit to aim and fire at the target location, doing all the usual aiming and line of fire tests. The only think you have to worry about with SetUnitTarget is overriding better targets that the unit has acquired for itself.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+		end
+	end
+	if lowest < 0 then -- don't bother with UW stuff.
+		lowest = 0
+	end
+	return lowest
+end
+
+
+local function GetFirePoint(radius, x, z, targetx, targetz)
```
GoogleFrog, line 74: The maths in this function is nonsense, that is why 'random' things are working or not working.
- You have swapped x and z in atan2.
- You are finding the angle from the enemy to the stardust, not the other way around.
- You have swapped the canonical use of sin and cos.

The combination of these mistakes cancels out, but it is going to confuse whoever has to look at this function.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+	spSetUnitTarget(unitID, nil)
+	overwatch[unitID].engaged = false
+	spGiveOrderToUnit(unitID, CMD.STOP, EMPTY, EMPTY)
+end
+
+local function UpdateUnitTarget(unitID, unitDefID, weaponID)
+	local x2, y2, z2 = spGetUnitPosition(unitID)
+	local overshootdef = overshootdefs[unitDefID][weaponID]
+	local originalrange = overshootdef.range
+	local oy = math.min(spGetGroundOrigHeight(x2, z2), GetLowestHeightOnCircle(x2, z2, originalrange, 9)) -- rudimentary check for cliffs and stuff.
```
GoogleFrog, line 101: The terrain doesn't change very fast so you don't need to do this every update.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+function gadget:UnitReverseBuilt(unitID, unitDefID, unitTeam)
+	overwatch[unitID] = nil
+end
+
+function gadget:UnitDestroyed(unitID, unitDefID, unitTeam, attackerID, attackerDefID, attackerTeam)
+	overwatch[unitID] = nil
+end
+
+function gadget:GameFrame(f)
+	if f%updaterate == 0 then
```
GoogleFrog, line 133: There should be no pairs here.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,143 @@
...
+		for id, data in pairs(overwatch) do
+			local unitdefID = data.def
+			if #spGetCommandQueue(id, 1) == 0 then -- has no active command, so feel free to engage. Note this check is mostly so that users can override the AI.
+				for weapon, _ in pairs(overshootdefs[unitdefID]) do
```
GoogleFrog, line 137: This use of pairs is particularly bad as this is a static configuration table.

Arch-Shaman: Use indexes instead?
[1] = {weaponID = 2,},
[2] = {weaponID = 6},
etc?

LuaRules/Gadgets/unit_overshoot.lua
Arch-Shaman: How would I dislodge the pairs here? I need to iterate over a list of unitIDs. I suppose tracking an ordered list might work..

LuaRules/Gadgets/unit_overshoot.lua
GoogleFrog: If you just want a weaponID you don't even need a list of tables. Just make a list of numbers.

units/turretriot.lua
```diff
@@ -77,6 +77,7 @@ return { turretriot = {
         light_color = [[0.8 0.76 0.38]],
         light_radius = 180,
         proximity_priority = 5, -- Don't use this unless required as it causes O(N^2) seperation checks per slow update.
+		overshoot = 1.33,
```
GoogleFrog, line 80: Whitespace.

LuaRules/Gadgets/unit_overshoot.lua
Arch-Shaman: Ah I thought it was dumbfire.

Arch-Shaman:
> This clashes with SetTarget, so probably only works for static units. This seems fine though.

How can I expand this to not clash with SetTarget? I would like to increase the robustness here. Do I have to listen to it somehow?

GoogleFrog:
I recently made a GG function for Ronin to use for shooting at Felon shields. I'm not that happy with how it works within the set target gadget, but you could just use that GG here as well.

GoogleFrog:
This is the commit: https://github.com/ZeroK-RTS/Zero-K/commit/0829f8d553d48d0c8ae8eeb6671b77a9d04253cf

I made it so that units with SetTarget could simply have their target set by the set target gadget, instead of removing the ground target at a later point. I also didn't want to add any extra unit tracking/timing to the tactical AI gadget. The GG essentially says "aim here for a certain number of frames".

Arch-Shaman:
Useful. I'll update the code. Probably have it target the position for an update loop's length. The config is now done btw. I tested values and found that the sweet spot for ronin v stardust is within the range of [370 - 380). The burnblow adds somewhere between 19% and 20% range. These two values work pretty well.

How do you feel about Faraday having Overshoot AI?

Arch-Shaman:
> Useful. I'll update the code.

During testing with [[Rapid://ZK-TEST]], I got spammed with errors, so I don't feel safe commiting such a change.

> [f=0000780] Error: [LuaRules::RunCallInTraceback] error=2 (LUA_ERRRUN) callin=GameFrame trace=[Internal Lua error: Call failure] [string "LuaRules/Gadgets/unit_overshootai.lua"]:88: attempt to call field 'SetTemporaryPosTarget' (a nil value)
stack traceback:
	[string "LuaRules/Gadgets/unit_overshootai.lua"]:88: in function 'AttackPosition'
	[string "LuaRules/Gadgets/unit_overshootai.lua"]:114: in function 'UpdateUnitTarget'
	[string "LuaRules/Gadgets/unit_overshootai.lua"]:153: in function 'GameFrame'
	[string "LuaRules/gadgets.lua"]:957: in function <[string "LuaRules/gadgets.lua"]:955>
	(tail call): ?

Idk if zk:test has been updated or not and I'm dead tired right now, so I'll work on it tomorrow. Let me know if there's anything else that needs revising in the current form.

GoogleFrog:
Don't use [[Rapid://ZK-TEST]] to test.

Faraday reload is too long and tickling enemies with EMP doesn't seem that useful, so I would leave it out for now.

Arch-Shaman:
I need to test the configuration one last time but the system is functional.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,248 @@
...
+
+function gadget:GameFrame(f)
+	if f%updaterate == 0 then
+		for i = 1, #unitstocheck do
+			local unitID = unitstocheck[i]
+			local data = overwatch[unitID]
+			if data then
+				local unitDefID = data.def
+				local weapons = data.weapons
+				if #spGetCommandQueue(unitID, 1) == 0 then
```
GoogleFrog, line 232: This is bad because it makes a table. There is some argument for spGetCommandQueue that just returns the number of commands (it might be 0 in the second argument, I forget). Use that.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,248 @@
...
+
+local function RemoveUnitFromList(unitID)
+	if overwatch[unitID] == nil then
+		return
+	end
+	local index = overwatch[unitID].index
+	if index == #unitstocheck then
+		unitstocheck[index] = nil
+	else
+		table.remove(unitstocheck, index)
```
GoogleFrog, line 95: I would be very surprised if table.remove actually worked here. Where is the reindexing? Just user IterableMap instead of reimplementing this here.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,248 @@
...
+	if overwatch[unitID].engaged and (not enemy or distance <= actualRange or (enemy and distance > actualEffectiveRange)) then
+		--Spring.Echo("STOP")
+		spGiveOrderToUnit(unitID, CMD.STOP, EMPTY, 0)
+		overwatch[unitID].engaged = false
+	end
+end
+
+function gadget:UnitFinished(unitID, unitDefID, unitTeam)
+	if overshootdefs[unitDefID] then
+		overwatch[unitID] = {def = unitDefID, engaged = false, weapons = {}, updatecount = 0, index = #unitstocheck + 1}
```
GoogleFrog, line 207: This should be in a function called AddUnit or similar.

LuaRules/Gadgets/unit_overshoot.lua
```diff
@@ -0,0 +1,248 @@
...
+			lowest = groundy
+		end
+	end
+	if lowest < 0 then -- don't bother with UW stuff.
+		lowest = 0
+	end
+	return lowest
+end
+
+local function RemoveUnitFromList(unitID)
```
GoogleFrog, line 87: Stylistically this function should be just above the callins, as it isn't used by anything else and it is nice to keep different parts of the gadget separate (unit tracking and unit handling).

GoogleFrog:
Is this still a draft?

Arch-Shaman:
Yes, there's that bug with it I haven't figured out yet.

Arch-Shaman:
Booted it up today to steal code from, turns out all of a sudden it seems to be working fine (at least for statics)

GoogleFrog:
Stardust might not even want to overshoot anymore, but this is a thing that Crab is balanced around (vs Stinger).

Arch-Shaman:
Right, things have changed.

GoogleFrog:
I possibly just made this obsolete.

https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_ward_fire.lua

PR #4179, opened 2020-11-07
Gauss gains sonar

### LINKED ISSUE #3244: Make Gauss buildable on sea
Gauss now has a cool floatation pillar (d860b491122cd832bdff5eb1504540ce64e03c83) but it is wasted as it cannot be built on sea (only appears if wreck is sunk and rezzed).

Arch-Shaman:
Gauss:
- Sonar 0 -> 664
~~- Now buildable on water (up to 305 depth)~~


sprunk:
I slightly dislike the arbitrary (from a player PoV) 305 limit because there isn't any decent UI which says what's going on when you can't build the Gauss sometimes. A model-side fix would be to use the scaling hax (56ec443e8ecb1b802d4fe1c9bd3bda135d068903, [demo](https://www.youtube.com/watch?v=wCfmtbODQ5I)) to just stretch the pillar vertically past 305 if need be. In practice most games don't have 305 depth though so this is relatively minor.

GoogleFrog:
Mex has a visual limit but no technical limit. I think making Gauss do the same would be fine.

On the PR as a whole, I'm not so sure this is a good idea. Gauss could make sea very porcy. When gauss was given the ability to penetrate water (for the benefit of Dagger) we thought about Gauss and decided to keep it land-only for this reason. I don't think the lack of a model prohibited it from floating.

If it were to not float, then probably shouldn't have the model.

sprunk:
How about making it have sonar but no buildability on water? It's a bid awkward though.

Anarchid:
Gauss is frequently enough built on water with a terraformed pillar, for its utility at shooting both surface and submerged targets and not dying to two shots from Envoy.

As such, it makes sense for it to have sonar.

Arch-Shaman:
I think at least sonar would help tremendously in usability and consistency:
* Most other things that live in sea have sonar.
* All other things except raven (maybe detri?) that can attack UW have sonar.
* Needing urchins nearby that die to 2 envoy shots is awkward and janky.

Birdulon:
>All other things except raven (maybe detri?) that can attack UW have sonar.

I was going to comment Ultimatum but I just checked and Ultimatum and Detriment both appear to have sonar already.
Sea porc feels pretty weird with having to balance out anti-sub/surface/air in a way that doesn't really show up on land because of all the flex AA. It would be nice to have a sturdy alternative to Urchins at least.

sprunk:
Doesn't seem to be going anywhere and it's a trivial 1-liner so we can reproduce it if need be, not worth cluttering up the PR list.

PR #4188, opened 2020-11-12
Add missing new line to chicken_starter.bos

Arch-Shaman:


GoogleFrog:
Ok, but generally bos is read-only.

sprunk:
Recompiled COB where!!!

Issue #3832, opened 2020-03-19
Fix Pheonix CEG on sea impact.


GoogleFrog:

psimyn:
setting `underwater = true` appears to fix, I think some of the explosions are too far underwater

PR #4190, opened 2020-11-12
Fix Pheonix CEG on sea impact

psimyn:
Fixes https://github.com/ZeroK-RTS/Zero-K/issues/3832

maybe this has terrible side effects, I don't know much about CEGs

before:  The shock waves were still visible above water, but not the fire

image

after: 

image

### DIFF
--- a/effects/napalm.lua
+++ b/effects/napalm.lua
@@ -32,6 +32,7 @@ local cegs = {
       count              = 6,
       ground             = true,
       water              = true,
+      underwater         = true,
       properties = {
         delay              = 0,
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
Looking good. 👍

psimyn:
The shockwaves do, the orange explosions are only above water

underwater: false

above water
image

underwater
image



underwater: true

above water
image

under water
image

GoogleFrog:
I expect that there are more cegs to check for underwater=true.

PR #4193, opened 2020-11-15
Add mutelite

### LINKED ISSUE #4192: Spec chat mute that ignores moderators
Perhaps the teams host should have a spec chat mute, as some of the spec chat is antagonistic and if a spectator says anything even slightly cheaty or tilty it probably isn't worth retaining..A modoption for spec chat mute exists, but I think we would need to make a version of it that lets moderators say things (so that they can moderate games). We could also consider adding a moderator-lite type of player flag that singles out players as particularly helpful. This essentially gives them the ability to help with technical issues as a spectator, or even to warn players off doing something abusive.

Arch-Shaman:
* Mutes all spectators except moderators and helpful players.
* Tag is 'cp.helpful'.

* Needs testing.

Implements/fixes #4192

### DIFF
--- a/units/amphsupport.lua
+++ b/units/amphsupport.lua
@@ -1,6 +1,6 @@
 return { amphsupport = {
-  unitname               = [[Bulkhead]],
-  name                   = [[amphsupport]],
+  unitname               = [[amphsupport]],
+  name                   = [[Bulkhead]],
   description            = [[Deployable Amphibious Fire Support (must stop to fire)]],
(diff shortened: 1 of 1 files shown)


sprunk:
* nobody is going to set this modoption manually, so I think any exceptions should just be incorporated into the default
 * admins (and room bosses) can already use `!say` to communicate, I think it's good that you need to be explicit around breaking the mute
 * `helpful` needs infra support

PR #4204, opened 2020-11-22
Better airfac colvol

### LINKED ISSUE #4036: Airfac airpad inconsistency
Airpad which comes with the factory is all weird and gross. Unlike with normal pads, It's difficult to reliably deal damage to the landed bomber (doesn't take dmg from Swift laser, largely blocks Raptor dmg from straight above).

IndeIrae:
Adds an actual non-default collision to airfac. Allows Swifts, Raptors and other magical creatures to reliably attack reloading bombers.
Fixes #4036.

Old:
Screenshot_3
Screenshot_4

New:
99915711-5ac01400-2d16-11eb-957c-a38ce9e5ce70
99915717-5dbb0480-2d16-11eb-8567-38540d257b3b


GoogleFrog:
Why did you make it so much larger than the geometry? Couldn't it look like this:
image
image

PR #4205, opened 2020-11-23
BlackStarV2 ffa boxes

thomas11111111:


thomas11111111:
forgot startpos

sprunk:
I took a look and there are some issues.

1) Crude hand-drawn (apparently) circles
image
2) The file name has to match map name, `BlackStar_v2` (mostly the underscore, possibly the `v` may also have to be lowercase)
3) A config for v1 already exists and could be reused via `VFS.Include`

PR #4209, opened 2020-11-26
Few cleanups for unit_auto_patrol_nanos

rtwfroody:
Follow up on feedback to https://github.com/ZeroK-RTS/Zero-K/pull/4125.

### DIFF
--- a/LuaUI/Widgets/unit_auto_patrol_nanos.lua
+++ b/LuaUI/Widgets/unit_auto_patrol_nanos.lua
@@ -323,7 +323,10 @@ local function unitNew(unitID)
 	local vz = mapCenterZ - z
 	commandTables[PATROL] = {CMD_PATROL, {x + vx*25/abs(vx), y, z + vz*25/abs(vz)}, 0, "patrol"}
 
-	local area = {x, y, z, UnitDefs[unitDefID].buildDistance}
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/unit_auto_patrol_nanos.lua
```diff
@@ -323,7 +323,10 @@ local function unitNew(unitID)
 	local vz = mapCenterZ - z
 	commandTables[PATROL] = {CMD_PATROL, {x + vx*25/abs(vx), y, z + vz*25/abs(vz)}, 0, "patrol"}
 
-	local area = {x, y, z, UnitDefs[unitDefID].buildDistance}
+	-- Instead of using a range of UnitDefs[unitDefID].buildDistance, we make
+	-- the range giant so that when you hold down shift you don't see circles
+	-- around every caretaker.
+	local area = {x, y, z, 50000}
```
sprunk, line 329: Won't this cause caretakers to needlessly process units out of range? What happens if a unit moves out of build range, is the order dropped correctly with the larger circle?

No opinion on hiding circles (they may well help see what a caretaker is doing when there aren't too many of them).

GoogleFrog: I tested this at some point and Caretakers were fairly good at dropping out of range orders. I'll have to check again though.

PR #4213, opened 2020-12-05
Make radarColor2 not add to losColor.

Birdulon:
Enables players to have a two-tone fog of war setup where radar coverage shows as lighter fog, without radar overbrightening areas in LOS.

Exaggerated example for clarity:
Green radar

Intended example:
Light white-ish radar

I couldn't think of a use-case for having radar color add to LOS color, but if that behaviour is desired I'll think up some way to add a toggle option.
Additionally, minimum fog brightness could probably stand to be lowered some more.

### DIFF
--- a/shaders/GLSL/infoLOS.lua
+++ b/shaders/GLSL/infoLOS.lua
@@ -49,19 +49,20 @@ return {
 		#define getTexel texture2D
 	#endif
 		void main() {
-			gl_FragColor  = vec4(0.0);
-			//gl_FragColor  = alwaysColor;
-			vec2 radarJammer = getTexel(tex2, texCoord).rg;
-			gl_FragColor += jamColor * radarJammer.g;
-			gl_FragColor += radarColor2 * step(0.8, radarJammer.r) * radarJammer.r;
-			gl_FragColor.rgb = fract(gl_FragColor.rgb);
-			
+			vec2 radarJammer = getTexel(tex2, texCoord).rg;  // r = radar, g = jamming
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Isn't this a problem when the edge of your radar range intersects your LOS, with the default settings?

Birdulon:
Default Settings after latest commit:


There's still soft fringing on the edge of LOS but that seems to be the same in master.

Example settings after latest commit:


As an aside, it seems difficult to get clear jammer appearance with the current blending. Adding configurable alpha or an outline color for that seems like an engine-side task though so I won't pursue that for now.

Birdulon:
Any other issues to resolve?

I played with turning off the random fringing effect on LOS (present on master) but it looks worse without it due to the low resolution of the LOS and air LOS map textures.

GoogleFrog:
The green band on the interior edge is entirely gone. This is a problem. Old:

image

New:

image

Birdulon:
Aha, I'll look into it, thanks.

Birdulon:
Should have it this time!

As an aside: I tried to break it off into `infoLOS.glsl` for ease of syntax highlighting etc., but couldn't get `VFS.ReadFile` to work with it. I tried `VFS.ReadFile("shaders/GLSL/infoLOS.glsl", VFS.RAW_FIRST)`, and without the second argument, but both seemed to result in no fragment shader being loaded.

sprunk:
Seems to work as advertised. Thanks!

Screenshots for default and the newly-possible two-tone setup
image
image

PR #4235, opened 2020-12-14
Make shield units have activatewhenbuilt

Arch-Shaman:
Thanks, @sprunk


sprunk (changes requested):
lacks commentary and overrides explicit values in the def

Issue #4226, opened 2020-12-11
Smart Nanos stuck on reclaim for a while when excessing metal.


GoogleFrog:
![image](https://user-images.githubusercontent.com/1318249/101852441-bfea8500-3bb1-11eb-9220-4f14e6fc777d.png)

In this battle: http://zero-k.info/Battles/Detail/1006569

rtwfroody:
I've sometimes seen this, but generally when I look at the excess graph later it turns out I didn't excess metal. There might be a real bug here, but it seems like the top center widget does something that means it doesn't quite reflect what's going on in the moment.

thomas11111111:
Long game but it's shows smart nanos is a ''hermetic'' environment.
The interesting part is after i queued a fortress and went afk, upon returning metal storage was full and caretakers looked stuck, they were connecting with factories but no metal was flowing. 
https://zero-k.info/Battles/Detail/1008206

GoogleFrog:
Were you testing in a teamgame? It is hard to actually excess in team games.

rtwfroody:
Some things I noticed while watching the first 2/3rds or so of this replay:

1. In this video, the metal widget in the top shows income of 30-something, and pull of -130, yet the metal inventory is going up. That doesn't make sense to me. http://casualhacker.net/zero-k-caretaker/economy-2020-12-14_18.50.02.mp4
2. At some point the factory, which was on low priority, just stopped building. http://casualhacker.net/zero-k-caretaker/factory-2020-12-14_18.51.21.mp4
3. Then it built, but slowly. http://casualhacker.net/zero-k-caretaker/building-again-2020-12-14_18.53.35.mp4
4. Later it stopped building again. http://casualhacker.net/zero-k-caretaker/not-building-2020-12-14_19.51.01.mp4

I think the building might have stopped because Thomas was repeatedly building phantoms, and eventually they sucked up so much energy that the factory and caretakers (which were on low priority) didn't have any to build.

GoogleFrog:
1. Energy stall. Energy is pulling 120 more than income and is at the reserve point.
2. Construction costs energy and there was less energy than the reserve point, so all construction that isn't high priority halts.
3. Same as 1.
4. Same as 3.

rtwfroody:
[...]

PR #4239, opened 2020-12-17
Fix unit_auto_patrol_nanos_v2 bugs

rtwfroody:
This helps in cases where metal is full but you're stalled on energy.
Before this change the widget would decide that reclaiming metal is a
good thing to do in that case.

Addresses one discovery I made when investigating https://github.com/ZeroK-RTS/Zero-K/issues/4226, but does not solve the core issue.


sprunk (approved):
lgtm

rtwfroody:
What do I need to do to move this PR along? They definitely improve the widget, and it would be nice to get them in for whenever the next update is.

PR #4251, opened 2020-12-21
Replace walk animation for Dirtbag

psimyn:
Replace dirtbag animation. Stops feet clipping underground so much, and matches leg speed better with movement speed.

Uses output from https://github.com/Beherith/Skeletor_S3O converted to lua

https://user-images.githubusercontent.com/1731922/102774441-1a1fed00-43df-11eb-92a7-9e03ecbf1efe.mp4


scripts/shieldscout.lua
```diff
@@ -170,25 +128,194 @@ end
 -----------------------------
 -- Walking
 
-local function Stopping()
-	Signal(SIG_Walk)
```
psimyn, line 174: `Signal` was moved to `script.StartMoving`

lhog:
Was this ported to Lua by hand or did you improve the original script to output Lua directly?

Beherith:
Great work! There is one thing I might recommend (it might not have been emphasized enough in the documentation) , on page 5 of the skeletor s3o manual: 
Another method of keeping the feet always oriented correctly, and not pointing everywhere is by setting Bone Constraints -> IK -> Rotation checkbox. 

Also, I saw you did a PR to skeletor to do lua output, why did you delete the PR?

psimyn:
@lhog I started converting the script to generate lua. Current progress is in https://github.com/psimyn/Skeletor_S3O/pull/1/files

@Beherith I had meant to make the PR on my fork. It was just hacking over the top of the BOS output (was mostly a find/replace syntax, plus some cleanup). Thankyou for the doc and tutorial video!

psimyn:
> "Another method of keeping the feet always oriented correctly, and not pointing everywhere is by setting Bone Constraints -> IK -> Rotation checkbox. "

I thought I had done this, but might have been in a different model. This one just had limb stiffness. I am still largely randomly clicking on things to try to figure out what they do

Beherith:
Excellent, the doc really covers every option that has relevance in animating units for spring :)

scripts/shieldscout.lua
GoogleFrog: I prefer signals to be at the start of functions that set a signal mask, so it is clear to the reader that only one copy of the thread can run at a time and it is harder to mess up.

GoogleFrog:
Looks good. Here are some comments.
 - I will try to run a benchmark on this, just in case updating every four frames has an impact. For all I know removing the WaitForTurns from the previous animation acts as an improvement. Dirtbag is spammy so this seems good to do, and I assume the aim is to do this sort of thing for more units.
 - As far as I can recall the variable `bMoving` is a COBism that doesn't do anything here. The thread is terminated by `Signal(SIG_WALK)` so `bMoving` is always true during `Walk()`.
 - Perhaps a moving variable could be useful to avoiding calling `StartThread(Walk)` if the unit is already walking. As it is currently, you are trusting the engine to not call StartMoving too often, which has not been something to trust in the past.
 - It concerns me that this animation system is essentially write-only, that is, that the LUS is no longer its own editable source file. Could you PR the sources to the artwork repository, with instructions for how to run them? It isn't vital at the moment, but if we go down this path then we'll be left with a bunch of uneditable animations. They will break with a speed buff/nerf or when something else comes up.
 - I think you should go a bit further and take slow damage into account, since this PR is about polish. This may be tricky to do perfectly due to the integer nature of your keyframes, so perhaps the speed modifier should increase by integer frames. See what looks good. Speed is controlled by the unitRulesParam `totalMoveSpeedChange`, take a look at how Dante works. This method may be replaced, see #4252.
 - Check how the head bob interacts with aiming and firing. You may need to fix the head while the unit has a target.

scripts/shieldscout.lua
psimyn: 👍 wasn't sure if there was some kind of subtle difference between the two positions

psimyn:
thanks @GoogleFrog! That was exactly kind of info I was after.

> I will try to run a benchmark on this, just in case updating every four frames has an impact. For all I know removing the WaitForTurns from the previous animation acts as an improvement. Dirtbag is spammy so this seems good to do

this would be great. Is benchmark using http://zero-k.info/wiki/Benchmarker ? I'll have a look into setting up some benchmarks as well so I can more easily test this for other anims

> and I assume the aim is to do this sort of thing for more units.

Yep, this was the plan. I have some WIP scripts for Rogue and Outlaw, working up to fixing some comms

> It concerns me that this animation system is essentially write-only, that is, that the LUS is no longer its own editable source file.... if we go down this path then we'll be left with a bunch of uneditable animations. They will break with a speed buff/nerf or when something else comes up.

yeah I didn't love that it's just a blob of generated numbers. Was thinking of some ways to clean up the output (i.e mirror a lot of the positions in x axis). but will start with instructions for regenerating and link to repo with blend files.

> I think you should go a bit further and take slow damage into account, since this PR is about polish. This may be tricky to do perfectly due to the integer nature of your keyframes, so perhaps the speed modifier should increase by integer frames. See what looks good. Speed is controlled by the unitRulesParam totalMoveSpeedChange, take a look at how Dante works. 

hadn't thought of slow damage, I will have a look into this. There is also a `variableSpeed` option in skeletor exporter that might do something similar(?).

> Check how the head bob interacts with aiming and firing. You may need to fix the head while the unit has a target.

Can it still have head bob in z-axis, or would it interfere with aiming?

Beherith:
So, we have encountered all of these in BAR, and heres how we handle them:
1. VariableSpeed is a must, especially for kbots that move at different speeds, the framework is designed for, use it! It uses integer division to always get perfect turns, and allows for changing unit speed later on by adjusting 1 number in BOS.
2. We commit the .blend files to artwork repository (if you dont include textures, they are actually quite small)
3. Head aiming: theres a reason we use a separate aimx1 (and aimy1 where needed) empty intermediate pieces in our models, so the aim wont interfere with pretty walks, though on some units, there is more control. 
4. According to my benchmarks (on armcom.bos vs armcom.lua in BAR), LUS is fully comparable to BOS, there is at most a 25-33% difference (favoring COB). I measured a 1% additional sim load for every 100 peewees walking (they walk fast, many keyframes per sec, many animated joints)
5. Slow damage (see variable speed above)
6. bMoving may be redundant, but I found it to be sometimes needed due to COB call-scripts/start-scripts only being executed next frame. Remember to SIG_TERM both in startmoving and stopmoving (sometimes both can get called in the same frame!)
7. walk animations were already pretty much write-only, save for garbage ones.

psimyn:
> only being executed next frame. Remember to SIG_TERM both in startmoving and stopmoving (sometimes both can get called in the same frame!)

this is a COB-only thing or could happen in LUS as well? 

I enabled variable speed, and added this to the main walk loop. Slight variation on `GetSpeedMod` that some units have

```
animFramesPerKeyframe / (Spring.GetUnitRulesParam(unitID, "totalMoveSpeedChange") or 1)
```

> We commit the .blend files to artwork repository (if you dont include textures, they are actually quite small)

Currently I'm using https://github.com/psimyn/zk-blender - can move it to https://github.com/ZeroK-RTS/Zero-K-Artwork.
Can I easily readd the textures when working on it locally?

psimyn:
Without checking `bMoving` inside Walk function they can run on the spot if stuck on terrain. Will readd it

https://user-images.githubusercontent.com/1731922/102890328-5d965c00-44b0-11eb-8a55-0dc8f3fc1d2c.mp4

GoogleFrog:
> this would be great. Is benchmark using http://zero-k.info/wiki/Benchmarker ? I'll have a look into setting up some benchmarks as well so I can more easily test this for other anims
No, that page is completely outdated. The current benchmarker just runs some replays a number of times. It can be setup with `benchmarks/config.lua`.

> Head aiming: theres a reason we use a separate aimx1 (and aimy1 where needed) empty intermediate pieces in our models, so the aim wont interfere with pretty walks, though on some units, there is more control.
I have taken to doing this as well. It just means that PRs will need to edit the models in most cases.

> Without checking `bMoving` inside Walk function they can run on the spot if stuck on terrain. Will readd it.
I can make stuck units call StartMoving every frame without calling StopMoving, so `bMoving` wouldn't help there. Even if `bMoving` is solving one edge case, it would be nice to solve it in a way that looks less like a hack.
image

> walk animations were already pretty much write-only, save for garbage ones.
In your experience. I think I would prefer animations of the quality and implementation of Dante over slightly better animations that have an uneditable implementation.

Beherith:
It takes less time to edit in blender, do a reexport and copy paste it back than to even figure out what is wrong with a script-only animation. 

Sure, I'd prefer a Lamborghini too, this Prius is all we got. Still better than the Trabant.

GoogleFrog:
> It takes less time, effort, and most importantly skill, to edit in blender, do a reexport and copy paste it back than to even figure out what is wrong with manual script-only animation.

That may be the case, but tools bitrot and knowledge is lost. These are the main concerns I want addressed. Note that I haven't actually said "make animations like the Dante animation" instead of using the method in this PR. This is for the reasons you outline, it takes more time and skill from the average starting point.

scripts/shieldscout.lua
```diff
@@ -170,25 +128,184 @@ end
...
+local function StopWalking()
+	Signal(SIG_WALK)
+	SetSignalMask(SIG_WALK)
+	animSpeed = 10; -- tune restore speed here, higher values are slower restore speeds	
+	Move(base, y_axis, 0, 311.170149 / animSpeed)
+	Move(base, z_axis, 0, 301.740754 / animSpeed)
+	Move(pelvis, y_axis, 0, 311.170149 / animSpeed)
+	Move(rfoot, y_axis, 0, 100.174990 / animSpeed)
+	Move(rfoot, z_axis, 0, 70.951353 / animSpeed)
+	Turn(box, x_axis, 0, math.rad(902.129385) / animSpeed)
```
sprunk, line 285: It would be good if the tool generated `15.7451` instead of `math.rad(902.1293)` because the values are not human-friendly anyway and using radians directly would save some overhead on indexing the global table with `math`, indexing the math table with `rad`, and calling the function. For this PR it's fine but if this method gets more popular it could add up.

scripts/shieldscout.lua
```diff
@@ -170,25 +128,184 @@ end
...
-	Turn(leftLeg.thigh, x_axis, 0, math.rad(200))
-	Turn(leftLeg.calf, x_axis, 0, math.rad(200))
-	Turn(leftLeg.foot, x_axis, 0, math.rad(200))
+local function GetSpeedMod()
+	local animFramesPerKeyframe = 4
+	return animFramesPerKeyframe / (Spring.GetUnitRulesParam(unitID, "totalMoveSpeedChange") or 1)
+end
+
+local animSpeed = GetSpeedMod()
+local walking = false
```
sprunk, line 137: Could use a comment to make sure the recent findings don't become tears in the rain
```suggestion
local walking = false -- required because StartMoving is spammed when stuck on spires, see #4251
```

scripts/shieldscout.lua
Beherith: fully agree, lus being in rad was stupid anyway, C should convert from deg to rad

psimyn:
@GoogleFrog would a cleaner export (structured more like Dante) be preferred, or that's still write-only since it's being generated? I could tidy up the output a bit to make it more human readable.

Otherwise could move the generated animations to separate script files that get imported (I think?), so at least the generated vs human-editable distinction is clearer.

Mainly I wanted to minimise steps between editing in Blender and previewing in Spring. Still figuring how this will work when there starts being more complex moves (`if aiming` or other conditionals)

scripts/shieldscout.lua
psimyn: will attempt to fix in this PR, just need to dig a bit more through the export script

scripts/shieldscout.lua
psimyn: left a comment that it's to stop StartMoving getting spammed. The spire specific stuff seems related to jump, happens after the non somersault jump if unit was walking before (if jump is queued up outside of range so unit has to walk then jump)

GoogleFrog:
Multiply by `animSpeed` rather than dividing by it. In general, divison is makes code harder to read since now I have to trust that GetSpeedMod() never returns zero. In this case the division is worse because it is essentially just multiplication by `1/(1/ Spring.GetUnitRulesParam(unitID, "totalMoveSpeedChange"))`. Also, the name `animSpeed` doesn't make sense in your code because I would expect an increase in speed to make the animation faster, but actually it reduces it.

psimyn:
good point, will also rename to `speedMod` or `speedMult`. It still can't return zero (since the `Sleep` will divide), but much less division

Beherith:
A strong factor behind integer division was the fact that it is required to ensure that transitions between keyframes remain smooth. 
Thus you need to ensure that the delta change of the piece / move rate results in an integer number of frames.

GoogleFrog:
> A strong factor behind integer division was the fact that it is required to ensure that transitions between keyframes remain smooth.

Thanks, this explains some of the otherwise weird preference for division over multiplication in the code. This explains the presence of things like `Sleep((33 / speedMult) - 1)`. The code wouldn't have worked though because Lua doesn't have the distinct concept of integers.

`GetSpeedMult()` should return two values, the speed multiplier for turn/move and the sleep time.

PR #4256, opened 2020-12-23
cmd_state_reverse_toggle: Permit local config

esainane:


LuaUI/Widgets/cmd_state_reverse_toggle.lua
```diff
@@ -34,7 +34,8 @@ local STANDARD_OPTS = {
 --------------------------------------------------------------------------------
 -- Configuration
 
-local stateTypes, gadgetReverse, specialHandling = VFS.Include(LUAUI_DIRNAME .. "Configs/stateTypes.lua")
+local VFSMODE = VFS.RAW_FIRST
```
sprunk, line 37: What's the main use case? I assume the below but it would be good to make it explicit.
```suggestion
local VFSMODE = VFS.RAW_FIRST -- so that users' own commands can also have reversing
```

GoogleFrog, line 37: Why not just do VFS.RAW_FIRST in the Include directly?

esainane: This was the sister PR to https://github.com/ZeroK-RTS/Zero-K/pull/4255. Intended to make similar configuration possible if any further oversights happened, as well as if anyone wants to add their own commands.

PR #4267, opened 2021-01-18
Update Hacksaw firing points

Sharkgamedev:
Fixes the firing points on hacksaw so missiles no longer come out of the center of it as well as changing hacksaw to .dae since s3o export is deprecated. Adds a laser pointer to it (For FW) which is hidden by unit script. I left it in in case anyone wanted to use it in the future.


GoogleFrog:
I still prefer s3o to dae as spring had (and probably has) slightly better support for s3o. It also fits better into other workflows. Idk if I want to switch to dae for a small change.

Sharkgamedev:
Please note current version is very much non functional. Not sure what's going on but for some reason obj2s3o causes ZK to look for arm.tiff as a texture file which doesn't exist (And causes spring to fail to load the model) and is also by chance not the right texture file...

Sharkgamedev:
Potentially obj2s3o converter is working and I was just being a little bit stupid...

Sharkgamedev:
I'm not sure how to go about this as it seems when I export from blender to s3o it loses the empty pieces eg. FiringPoints and base. Anyone know a good way to get to obj while retaining that information?

GoogleFrog:
I use Upspring to add empty pieces.

Sharkgamedev:
> I use Upspring to add empty pieces.
Ah, thanks!

PR #4269, opened 2021-01-19
Added disarm award

floomby:
I added a disarm award. It works like the slow award by measuring the damage done, rather than measuring if the unit is actually disarmed. (I don't even know which would be preferable tbh.) The website probably needs the award information too, but I don't know anything about that.


LuaRules/Gadgets/unit_boolean_disable.lua
```diff
@@ -213,6 +213,13 @@ function gadget:UnitPreDamaged(unitID, unitDefID, unitTeam, damage, paralyzer,
 	if disarmWeapons[weaponDefID] then
 		local def = disarmWeapons[weaponDefID]
 		addParalysisDamageToUnit(unitID, damage*def.damageMult, def.disarmTimer, overstunDamageMult[weaponDefID])
+		
+		if GG.Awards and GG.Awards.AddAwardPoints then
+			local ud = UnitDefs[unitDefID]
```
GoogleFrog, line 218: I think you should be caching `ud.health * ud.metalCost` as UnitDefs access is slow. There are examples of gadgets iterating over UnitDefs near the start of the file to pick out the relevant values.

sprunk: Note that metal cost and health are not always the values from defs (because of dynamic commanders) so consider `Spring.GetUnitHealth` and `Spring.Utilities.GetUnitCost` for those.

floomby: Is the performance on `Spring.GetUnitHealth` and `Spring.Utilities.GetUnitCost` pretty good? Also since this is probably the right way to do it then it is messed up in the slow award since I basically took this code from that award.

PR #4270, opened 2021-01-20
Razor Rework/Update

Sharkgamedev:
Makes Razor look a bit better, instead of it being a hexagon trying it's hardest to be a circle it is now a circle, I've also touched up the doors to make their edges a little better as well as the inside circle. Finally, under the guidance of ShadowFury I was able to improve the metallic/smoothness map on it as well. Note: If ao plate is not strong enough set decal size to 7, I just wasn't sure if that was too strong so I did 6. There are also a few new unittextures which were leftover from me messing around trying different things, I figured I'd leave them in case they could be used in the future. It also got a new buildpic for the slight difference. Death clone has not been updated however as I didn't see the need, I can do it sometime if it is more important.
Tris count increase 2x, from 776 > 1554
Not sure about colvols, this is how they look at the moment:
image
Before
image
After
image
Before
image
After
image


GoogleFrog:
buildingGroundDecalSize should not be edited as it is derived from footprint. Ideally you would regenerate the ground decal with the aoplane generator. I think it can be found in tools.

Sharkgamedev:
> buildingGroundDecalSize should not be edited as it is derived from footprint. Ideally you would regenerate the ground decal with the aoplane generator. I think it can be found in tools.

Aoplane generator seems to be a bit finicky, is using Blender to generate it not ok? I had done it with blender before at 256 x 256 but I guess that was lower res than the old one so that's why it didn't fit? I could rebake it with a higher res in Blender or just scale it in Gimp.

Sharkgamedev:
> buildingGroundDecalSize should not be edited as it is derived from footprint. Ideally you would regenerate the ground decal with the aoplane generator. I think it can be found in tools.

Done!

sprunk (changes requested):
1) What are these files for? If they aren't used they probably belong to the artwork repo.
 * unittextures/razor_shiny_metal.dds
 * unittextures/razor_shiny_metal_lesssmooth.dds
 * unittextures/razor_shiny_metal_smooth.dds

2) The unitpic should have "floating" (surface/land) but it has "amphibious" (land/underwater) background. This sounds like an issue with the generator, I'll check it over the weekend.

3) A minor thing but the new AO is 340 kB while the old one was 20 kB, looks a bit suspicious. It would probably be good to compress it.

Sharkgamedev:
@sprunk The aoplate generator gave me infinite trouble so I'm not surprised that something like that happened... I'll see what I can do about compression in gimp.

Those extra files were just different iterations of the texture, my mistake for leaving them in. 

I remember making some modifications to the buildpic generator locally. That could have contributed to the problem, Ive since reverted the changes but I can't remember if I generated the build pic before or after that. I'll try regenerating it now and see if that resolves the issue.

GoogleFrog:
The base is certainly rounder, which is good, but the texture overall seems darker and messier. The top of the old Razor fairly crisp and the ring is consistently white. The PR Razor has a bunch of weird triangle artifacts and is generally quite dark.

image

Sharkgamedev:
> The base is certainly rounder, which is good, but the texture overall seems darker and messier. The top of the old Razor fairly crisp and the ring is consistently white. The PR Razor has a bunch of weird triangle artifacts and is generally quite dark.
> 
> 
> 
> image
> 
> 

Hmn looks like their may be some incorrect normals on that. The body is darker to because it's more metallic, I can definitely turn that down around the ring to improve contrast.

sprunk:
I checked the icon generator and while my potato PC fails to actually render units there, it draws backgrounds and Razor has the correct "floating" (water/land) background.

Try building an icon using the unmodified model and see if it works. If yes then that would point to some issue with the new model. Maybe it has too huge of a model radius (extending too far underwater)?

sprunk:
Any advances?

sprunk:
Annual ping for progress (I see that there was a thumbs-up the last time so I assume something is slowly happening).

PR #4284, opened 2021-01-31
Fixes a consistent bug with state recall failing first try

dvr333:
Now only has an inconsistent bug when doing state recall while scrolling the camera, or right after finishing scrolling the camera while an early part of the smooth interpolation is happening. Works fine when state recall happens at the tail end of the smooth interpolation


dvr333:
I can't figure out what's causing the remaining issue, and honestly I am done trying. The only other weird thing I noticed is that when scrolling while holding MMB, if you hold MMB, hit a camera recall key while holding MMB, and then move the mouse, the camera location jumps back to where it was before hitting the camera recall key. I have no idea why this is happening, as the code doesn't look like it's holding onto any particular camera state in widget:MousePress

PR #4298, opened 2021-02-15
Reducing order spam for auto reclaim heal assist widget 

danfireman:
Also spreading those orders over more frames

### DIFF
--- a/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
+++ b/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
@@ -11,7 +11,7 @@ function widget:GetInfo()
 end
 
 local myteam = Spring.GetMyTeamID()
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
```diff
@@ -76,21 +76,19 @@ end
...
 	end
 end
 
-function widget:GameFrame(n)
-	if (n%UPDATE_FRAME==0 and n > 30) then
-		for _,Con in pairs(ConStack) do
-			Con:handle()
-		end
+function widget:GameFrame(n) 
+	for _,Con in pairs(ConStack[n%UPDATE_FRAME]) do
```
GoogleFrog, line 90: This is a bad use of pairs.

PR #4301, opened 2021-02-18
Rate limit Widget use of GiveOrder*

esainane:
This adds a system which poisons Spring.GiveOrder*, storing a sliding
window of command volume over time. The rate limit is provisionally set to
110c/s over a 15 second average, and can be bike shedded later.
This is to reflect the normal nature of command volume being bursty,
and sustained regular volume being troublesome.

All functions within a Widget affect the same counter.

Widgets that legitimately create large volumes of commands can be
whitelisted. The whitelist is currently in cawidgets.lua, but could
be moved to its own file if need be.

The current whitelist is certainly incomplete. In particular, area
command splitting should be whitelisted.

As a side effect, each poisoned widget receives its own local copy of
the Spring table, and modifications made to this table no longer
affect other Spring instances. Changes made to the Spring table
before widget creation, such as in cache.lua, or in Spring.Utilities,
are still shared and reflected across all widgets.

The expectation is that BLOCK_MODE=2 is the norm in a multiplayer setting.
BLOCK_MODE defaults to 1 in the absence of any instructions from infra to
permit destructive local testing, while still sending warnings about any
excessive command volume when it occurs.

### DIFF
--- a/LuaUI/cache.lua
+++ b/LuaUI/cache.lua
@@ -1,5 +1,7 @@
 -- Poisoning for Spring.* functions (caching, filtering, providing back compat)
 
+-- luacheck: globals currentGameFrame
+
 if not Spring.IsUserWriting then
 	Spring.IsUserWriting = function()
 		return false
@@ -110,13 +112,13 @@ end
 function Spring.GetVisibleUnits(teamID, radius, Icons)
 	local index = buildIndex(teamID, radius, Icons)
 
-	local currentFrame = Spring.GetGameFrame() -- frame is necessary (invalidates visibility; units can die or disappear outta LoS)
+	-- frame is necessary (invalidates visibility; units can die or disappear outta LoS)
 	local now = Spring.GetTimer() -- frame is not sufficient (eg. you can move the screen while game is paused)
 
 	local visible = visibleUnits[index]
 	if visible then
 		local diff = Spring.DiffTimers(now, visible.time)
-		if diff < 0.05 and currentFrame == visible.frame then
+		if diff < 0.05 and currentGameFrame == visible.frame then
 			return visible.units
 		end
 	else
@@ -126,7 +128,7 @@ function Spring.GetVisibleUnits(teamID, radius, Icons)
 
 	local ret = GetVisibleUnits(teamID, radius, Icons)
 	visible.units = ret
-	visible.frame = currentFrame
+	visible.frame = currentGameFrame
 	visible.time = now
 
 	return ret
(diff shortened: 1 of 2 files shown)


GoogleFrog:
What is the performance like? Maybe disable the throttling for luauis that don't have local widgets enabled, to avoid impacting the bystanders if there is a cost.

esainane:
Theoretically, there are no temporary tables, no additional callouts (thanks to the use of currentGameFrame as an upvalue), and the overwhelming majority of bookkeeping is skipped with an early and cheap condition that only activates once per second.

Experimentally, I could not create any observable performance decrease. This held even when calibrating with huge groups of units to avoid false positives on Widgets that can send briefly send large quantities of commands, primarily Scout Dodge.

Ironically, this was a fair bit of optimization for functions that are now emphatically not meant to be called often.

GoogleFrog:
Sounds good.

LuaUI/cache.lua
```diff
@@ -152,3 +154,125 @@ function Spring.SetCameraTarget(x, y, z, transTime)
...
+function PoisonWidget(widget, widgetName)
+	if BLOCK_MODE == 0 then return end
+	-- All GiveOrder* functions use the same circle buffer to count commands for throttling.
+	-- Tracks calls made over the last CIRCLE_BUFFER_SIZE seconds, grouped by second.
+	local lastWindow = math.floor(currentGameFrame / FRAMES_PER_SECOND)
+	local lastFrame = currentGameFrame
+	local currentWindowCalls = 0
+	local circleBuffer = {}
+	local circleBufferIndex = 1
+	for i=1,CIRCLE_BUFFER_SIZE do
```
GoogleFrog, line 193: I dislike this whitespace.

LuaUI/cawidgets.lua
```diff
@@ -99,6 +101,11 @@ VFSMODE = VFSMODE or VFS.ZIP
 
 local detailLevel = Spring.GetConfigInt("widgetDetailLevel", 3)
 
+local widgetPoisonWhitelist = {
```
GoogleFrog, line 104: This should be rate limit whitelist, since others poisons exist.

GoogleFrog:
How about this:
 - Automatically whitelist all modside widgets (the widget list knows which widgets are modside and which are local, so this must be possible).
 - Make cheat mode ignore the filter.
 - Default to blocking (maybe even disabling) widgets that hit the limit.
 - Add a little more leeway to the limit.

LuaUI/cache.lua
esainane: As in, there should be a line before the `for`, or?

Arch-Shaman:
```lua
for i = 1, CIRCLE_BUFFER_SIZE
```

esainane:
I'm not sure that modside widgets are safe to the point that they should all be excluded from a virtually-free rate check? It's possibly worth running with BLOCK_MODE left at 1 for a while, so we can see if anything complains without needing to worry about enforcement.

Is there a callin for when cheats are enabled? It would be nice to avoid callouts if possible.

The limit has been given a lot more leeway (exceed 110aps over 15s) than the one originally discussed and tested with (exceed 80aps over 10s)

GoogleFrog:
> I'm not sure that modside widgets are safe to the point that they should all be excluded from a virtually-free rate check? It's possibly worth running with BLOCK_MODE left at 1 for a while, so we can see if anything complains without needing to worry about enforcement.

Gameside widgets reside at a lower customisation level so I don't want to do anything that breaks them. If they break then they should be fixed gameside. Detection without blocking achieves this.

> Is there a callin for when cheats are enabled? It would be nice to avoid callouts if possible.

Not that I know of. The engine would know. You could cache it every second or so.

PR #4311, opened 2021-02-24
One shiny !!NEW!! turretimpulse LUS

Arch-Shaman:
I don't know how to convert the explosion tags so I improvised. It works mostly like the old one from what I can see.


Arch-Shaman:
TODO: squash this

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+
+local function WobbleCrystals()
+	local mult = -1
+	while true do
+		for i = 1, #crystals do
+			if i == 3 then
+				mult = 1
+			elseif i == 4 then
+				mult = -1
+			end
```
sprunk, line 44: looks wrong, should be two of each with each sign but here only i=3 is positive

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+				mult = 1
+			end
+			if i%2 == 0 then
+				Move(crystals[i], y_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
+			else
+				Move(crystals[i], x_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
+			end
+		end
+		WaitForMove(crystals[4], x_axis)
+		Sleep(100)
```
sprunk, line 65: no sleep in the original

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+			elseif i == 4 then
+				mult = 1
+			end
+			if i%2 == 0 then
+				Move(crystals[i], y_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
+			else
+				Move(crystals[i], x_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
+			end
+		end
+		WaitForMove(crystals[4], x_axis)
```
sprunk, line 64: 4%2==0 so crystal[4] should be moving along Y axis but you're waiting on X

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+	end
+end
+
+function script.AimFromWeapon(num)
+	return center
+end
+
+function script.AimWeapon(num, heading, pitch)
+	Turn(turret, y_axis, heading, TURRET_AIM_SPEED)
+	Turn(turret, x_axis, -1 * pitch, TURRET_AIM_SPEED)
```
sprunk, line 99:
```suggestion
	Turn(turret, x_axis, -pitch, TURRET_AIM_SPEED)
```

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+
+local smokePiece = {base}
+
+local function WobbleTurret()
+	while true do
+		Move(turret, y_axis, WOBBLE_DIST, WOBBLE_SPEED)
+		WaitForMove(turret, y_axis)
+		Move(turret, y_axis, -1 * WOBBLE_DIST, WOBBLE_SPEED)
+		WaitForMove(turret, y_axis)
+		Sleep(100)
```
sprunk, line 32: no sleep in the original
```suggestion
		Move(turret, y_axis, -WOBBLE_DIST, WOBBLE_SPEED)
		WaitForMove(turret, y_axis)
```

scripts/turretimpulse.lua
```diff
@@ -0,0 +1,132 @@
...
+local turret = piece('turret')
+local ring = piece('ring')
+local center = piece('center')
+local firepoint = piece('firepoint')
+local crystals = {}
+local state = true
+crystals[1] = piece('crystal1')
+crystals[2] = piece('crystal2')
+crystals[3] = piece('crystal3')
+crystals[4] = piece('crystal4')
```
sprunk, line 22:
```suggestion
local crystals = {piece('crystal1', 'crystal2', 'crystal3', 'crystal4')}
local state = true
```

scripts/turretimpulse.lua
Arch-Shaman: This may cause hangs. I received a hang during initial testing when a newton was created (may or may not be false attribution) so I included it to be safe.

scripts/turretimpulse.lua
Arch-Shaman: Good catch. Fixed locally. Will update after round 2 of testing.

scripts/turretimpulse.lua
Arch-Shaman:
```lua
local function WobbleCrystals()
	local mult = -1
	while true do
		for i = 1, #crystals do
			if i == 3 then
				mult = mult * -1
			end
			if i%2 == 0 then
				Move(crystals[i], y_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
			else
				Move(crystals[i], x_axis, mult * WOBBLE_DIST, WOBBLE_SPEED)
			end
		end
		WaitForMove(crystals[4], x_axis)
	end
end
```
?

Arch-Shaman:
Everything looks okay now?

sprunk:
The code looks okay, I'll test today or tomorrow.

Arch-Shaman:
There's a version of it in FW. It looks normal to me. If not I'll make changes no problem.
Anyways, I'm happy to convert more when I have the time and will to change them over.

PR #4314, opened 2021-02-26
allow custom widget options icon

Pxtl:
Allow widgets to declare their own options icon.

image

### DIFF
--- a/LuaUI/Widgets/gui_epicmenu.lua
+++ b/LuaUI/Widgets/gui_epicmenu.lua
@@ -1285,7 +1285,11 @@ local function IntegrateWidget(w, addoptions, index)
 	
 	local wname = w.whInfo.name
 	local defaultpath =  w.options_path or ('Settings/Misc/' .. wname)
-	
+
+	-- add the widget's icon to the subMenuIcons table if there isn't one in that spot already
+	if w.whInfo.icon and not confdata.subMenuIcons[defaultpath] then
+		confdata.subMenuIcons[defaultpath] = w.whInfo.icon
+	end
 	
 	--[[
 	--If a widget disables itself in widget:Initialize it will run the removewidget before the insertwidget is complete. this fix doesn't work


Pxtl:
Wasn't sure if the icon should be in GetInfo or in the main body like the options, though.

GoogleFrog:
Probably in the main body, like options, for consistency. Options have a path parameter that widgets can use to put icons in different paths, eg a widget could make a submenu of Hotkeys for hotkeys it adds, and have a submenu in Settings for some configuration. We're trying to be good about putting everything you might want to hotkey in the hotkeys menu while keeping it clear of settings.

For an icon system to handle multiple paths it would have to look something like this:
```
options_icon = {
	['Settings/Graphics/Lighting'] = 'LuaUI/ImagesmyLightIcon.png',
}
options_path = 'Settings/Graphics/Lighting'
options_order = etc...
```

I'd prefer the full solution, but just doing a single options_icon is ok too.

Pxtl:
Will do.  Wouldn't it be cleaner just to put the icon on the option itself?

```lua
options_path = 'Settings/Interface/Player Centroid Marker'
options_icon = "luaui/widgets/player-centroid-marker/main-icon.png"
options_order ={...yadda...}
options = {
	enable = {
		name = "Enable player centroid markers",
		type = 'bool',
		value = false,
		desc = "Toggle the player-name markers over groups of units",
		OnChange = function() ReInitialize() end,
		advanced = false,
		icon = "luaui/widgets/player-centroid-marker/toggle-marker.png"
	},
	...
```

While it's less flexible than simply allowing a widget to stuff icons wherever it likes into the tree of icons, I think it's easier and more legible this way.

GoogleFrog:
Submenus are automatically inferred from paths, not directly from options menus. You wouldn't be able to add an icon to a submenu button with an option. Also, options don't have icons at the moment.

Pxtl:
> 
> 
> Submenus are automatically inferred from paths, not directly from options menus. You wouldn't be able to add an icon to a submenu button with an option. Also, options don't have icons at the moment.

Ahh, that makes sense, I thought you were saying the settings themselves needed icons, but no, they're going deep into various other trees and intermediate nodes on that tree may not be already defined, so we need to be able to insert icons into various points on the tree.

I was just trying to come up with a way to avoid the widget-developer from repeating themselves, but yeah I can see what you mean.

sprunk:
Any news?

PR #4323, opened 2021-03-05
Siren no longer stays at missile range during attack move

Arch-Shaman:
https://youtu.be/Jlp6oqee0ps

It will however stay at max range if given an explicit attack command. Whether or not this is useful is up to you to decide.


GoogleFrog:
This fix feels unstable under engine changes.

PR #4329, opened 2021-03-07
Gauss Animation update. can shoot down cliffs.

SteelBlueZK:


SteelBlueZK:
Make sure I didn't mess anything up, OK? Also if Anarchid could submit a new model for gauss, such that the aim piece is not attached to belt but it's attached to something that does not rotate I would love it.

SteelBlueZK:
SharkGameDev fixed the model the way I want it.
There exists a problem with the CEG that is the muzzle flash.

Unpack/pack time has slightly changed, and almost all animations are tied to a variable unpackspeed.

GoogleFrog:
Gauss closes and then reopens when told to shoot to the right, and then down a cliff. This is concerning.

image

image

image

I like to avoid changing the position of the piece returns from `script.AimFromWeapon` because doing so tends to cause close/open bugs or aiming loops. I poked around a bit and found one.

image

This Gauss is stuck in a loop of opening/closing. It never fires.

GoogleFrog:
As far as I recall the only problem with stable Gauss was that it would aim and fire when it had no line of fire, causing the project to bounce off into the sky. The first thing I would try is take stable Gauss and use the pitch argument in AimWeapon to cause the animation to peek over the edge of its base. Then when it thinks it can fire, it actually can.

SteelBlueZK:
https://user-images.githubusercontent.com/8924181/115645795-ffa74300-a2ee-11eb-973e-79aa0353de88.mp4

"feature" as follows: enemy can be fired at, but gauss will only automatically target enemy if gun is pointing in that direction. If gauss is facing opposite direction or closed then gauss will not aim at enemy. Fortunately, it is very rare that a unit will just appear within this pseudo blind spot in gameplay.

Metalhead gauss will never make a return with this script update, 100% money back guarantee. Also ships with buttery smooth animations.

I regard this as an upgrade over stable gauss, it should have the entire angle of fire of stable gauss and then some. It's aimpiece no longer moves(again) removing some edge cases of jitter.

PR #4331, opened 2021-03-07
Wind

### LINKED ISSUE #4330: Wind does not update without windgens
Noticeable now that there is a widget for it.  

It should be at least updated on frame 0 (players can force an update by nanoframing a windgen), or more likely in pregame so that you can plan your buildorder around it?

finestep:
Fixes #4330
Pregame metaphysics are unknown territory to me

### DIFF
--- a/LuaRules/Gadgets/unit_windmill_control.lua
+++ b/LuaRules/Gadgets/unit_windmill_control.lua
@@ -227,7 +227,7 @@ function gadget:Initialize()
 	Spring.SetGameRulesParam("WindHeading", 0)
 	
 	GG.WindStrength = scaleWind(strength)
-	Spring.SetGameRulesParam("WindStrength", scaleWind(strength))
+	Spring.SetGameRulesParam("WindStrength", GG.WindStrength)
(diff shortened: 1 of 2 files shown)


LuaRules/Gadgets/unit_windmill_control.lua
```diff
@@ -109,22 +109,26 @@ end
 -------------------------------------------------------------------------------------
 -------------------------------------------------------------------------------------
 
+local function scaleWind(str) 
+	return (strength - windMin)/windRange
```
sprunk, line 113:
```suggestion
local function scaleWind(str) 
	return (str - windMin)/windRange
```

GoogleFrog:
Does the value in Initialize actually mean anything? Eg it is a bit misleading to report a number pregame only for it to change instantly.

finestep:
As far as I can tell, there should be no discontinuity from Initialize() to GameFrame(0) in wind strength.
Regardless Initialize() does not seem to be the appropriate place to do this since wind widget still shows 0% in pregame for me

GoogleFrog:
I doubt resource bar does any updates prior to frame 0. To initialise the wind display correctly you would need to add a display update to something like widget:Initialize.

PR #4340, opened 2021-03-10
swift forbidden to fire when landed

Anarchid:
blindcode from web, trivial but untested

### DIFF
--- a/scripts/planefighter.lua
+++ b/scripts/planefighter.lua
@@ -104,7 +104,7 @@ function script.Activate ()
 end
 
 function script.Deactivate ()
-	active=false
+	active = false
 end
 
 function script.QueryWeapon(num)
--- a/scripts/planefighter.lua
+++ b/scripts/planefighter.lua
@@ -6,6 +6,8 @@ local base, flare1, flare2, nozzle1, nozzle2, missile, rgun, lgun, rwing, lwing,
 
 local smokePiece = {base, rwing, lwing}
 
+local active = true
+
 --variables
 local shotCycle = 0
 local flare = {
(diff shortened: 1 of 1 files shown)


sprunk:
It's a cool micro trick, any replay where this is a problem?

Anarchid:
http://zero-k.info/Battles/Detail/1070256

Note that afaict this will only make it fire when completely landed, not when land*ing*. So the trick is still there, just a bit less impactful (and less jank, since they commonly refuse to fire when fully landed anyway).

sprunk:
https://streamable.com/wcg9ag

Many mistakes apparently:
 * the pyros wouldn't engage, this would the
 * clumped pyros damage each other (incl. when dying) and make it possible for swifts to work as a clump too (for concentrated fire)
 * where's the allied air player? The world wonders.
 * after 0:20 the swifts spend almost all the time in the transitional landing phase and not actually landed so this PR won't help much

Anarchid:
> Would airborne Swifts have fared much worse?

Most of the firing here was done by landing or lifting Swifts. The one incident in which they were completely landed, most of them also failed to shoot. So *this* case would be mostly unaffected.

Anarchid:
Did some testing, works precisely like assumed:
- Fully landed Swifts refuse to shoot
- Landing or climbing Swifts shoot

It's still possible to extract quite a lot of extra DPS out of them.

GoogleFrog:
> Did some testing, works precisely like assumed:
- Fully landed Swifts refuse to shoot
- Landing or climbing Swifts shoot

This doesn't seem to affect the micro in the clip.

Anarchid:
> This seems to barely affect the micro in the clip.

Yes

> Perhaps fighters want AI that lets them track their targets speed instead.

Yes.

sprunk:
This PR won't help the issue at hand and I have seen little evidence that the issue at hand is much more than salt.

PR #4344, opened 2021-03-13
Make antinuke float

Arch-Shaman:
So when its resurrected it isn't gimped.

### DIFF
--- a/units/staticantinuke.lua
+++ b/units/staticantinuke.lua
@@ -12,7 +12,7 @@ return { staticantinuke = {
   buildingGroundDecalSizeY      = 6,
   buildingGroundDecalType       = [[antinuke_decal.dds]],
   buildPic                      = [[staticantinuke.png]],
-  category                      = [[SINK]],
+  category                      = [[FLOAT]],
   collisionVolumeOffsets        = [[0 0 0]],
   collisionVolumeScales         = [[70 55 110]],
   collisionVolumeType           = [[box]],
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
Note: this does not make it buildable on water (though that's a single line change, if desired). It merely stops players from being fucked over by a resurrected antinuke being unable to fire while underwater.

sprunk:
>being fucked over by a resurrected

The solution is to remove rez.

Arch-Shaman:
Sure but what about /give?

GoogleFrog:
Stun it when it is underwater? Simply have it check water in aimWeapon? I want to do whatever stops an UW antinuke letting a nuke go through a functional antinuke.

As with Gauss, it can't float on rez because that would imply that it should be buildable on rez.

Arch-Shaman:
The simplest approach then would be to give it a waterweapon or sublaunched weapon. However, now you have an emp immune anti. Stunning it underwater is an idea, though it adds a special case imo

GoogleFrog:
To be clear, I don't want antinuke to work from underwater, regardless of how it ends up there.

sprunk:
I think that's already the case. The issue this PR is tackling is that rezzing an anti that would end up underwater is pointless and a trap.

GoogleFrog:
I do not know whether underwater anti's simply don't work or actively harm the ability of a team to intercept nukes. To me that is the open problem in this area.

sprunk:
I've verified that underwater anti simply doesn't work and won't harm the team (i.e. other antis will reliably take over if they share coverage).

If GF doesn't want anti to float then I think this PR is moot. Keep in mind sunken antis can still be salvaged via seismic.

Issue #4143, opened 2020-10-14
Add optional includes to aid modding


GoogleFrog:
For example, unitsdefs_posts could include a file that mutators can include so that modders can use unitsdefs_posts without overriding the base ZK version (risking mod breakage if ZK updates unitdefs_posts). There are probably more things to do, but I cannot think of them at the moment.

PR #4349, opened 2021-03-16
Add support for injecting/changing custom cmds for modders/mappers

Arch-Shaman:
* Have not tested yet.
* Assists implementation of #4143

### DIFF
--- a/LuaRules/Configs/customcmds.lua
+++ b/LuaRules/Configs/customcmds.lua
@@ -125,14 +125,14 @@ WRECK =         	36734
 RESTOREBOMB = 		39735
 ]]--
 
-if VFS.FileExists("LuaRules/Configs/customcmds_mod.lua") then
-	local inject = VFS.Include("LuaRules/Configs/customcmds_mod.lua")
+if VFS.FileExists("LuaRules/Configs/customcmds_mod.lua", VFS.MOD) then
+	local inject = VFS.Include("LuaRules/Configs/customcmds_mod.lua", nil, VFS.MOD)
 	for id, num in pairs(inject) do
 		cmds[id] = num
 	end
 end
-if VFS.FileExists("LuaRules/Configs/customcmds_map.lua") then
-	local inject = VFS.Include("LuaRules/Configs/customcmds_map.lua")
+if VFS.FileExists("LuaRules/Configs/customcmds_map.lua", VFS.MAP) then
+	local inject = VFS.Include("LuaRules/Configs/customcmds_map.lua", nil, VFS.MAP)
 	for id, num in pairs(inject) do
 		cmds[id] = num
 	end
--- a/LuaRules/Configs/customcmds.lua
+++ b/LuaRules/Configs/customcmds.lua
@@ -124,4 +124,18 @@ SCRAMBLE =      	35128
 WRECK =         	36734
 RESTOREBOMB = 		39735
 ]]--
+
+if VFS.FileExists("LuaRules/Configs/customcmds_mod.lua") then
+	local inject = VFS.Include("LuaRules/Configs/customcmds_mod.lua")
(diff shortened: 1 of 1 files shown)


sprunk:
Hiding stuff here may make copying gadgets across mods harder (because now you need to know to also look into the modside customcmds. Why not just include your custom file directly in the target gadget so there's a fat `VFS.Include("even_more_cmds.lua", ...)` letting people know what's going on?

Arch-Shaman:
I basically did my interpretation of this: https://discordapp.com/channels/278805140708786177/278805140708786177/817619216253059237

sprunk:
What is a broken customcmd? It's just a number.

GoogleFrog:
A broken customcmd occurs when the game has a gadget that registers a `nil` cmdID. This occurs when a mod overrides customcmds.lua such that the number set in the game no longer exists.

LuaRules/Configs/customcmds.lua
```diff
@@ -124,4 +124,18 @@ SCRAMBLE =      	35128
 WRECK =         	36734
 RESTOREBOMB = 		39735
 ]]--
+
+if VFS.FileExists("LuaRules/Configs/customcmds_mod.lua") then
+	local inject = VFS.Include("LuaRules/Configs/customcmds_mod.lua")
```
sprunk, line 129: VFS mode strictness
```suggestion
if VFS.FileExists("LuaRules/Configs/customcmds_mod.lua", VFS.MOD) then
	local inject = VFS.Include("LuaRules/Configs/customcmds_mod.lua", nil, VFS.MOD)
```

LuaRules/Configs/customcmds.lua
```diff
@@ -124,4 +124,18 @@ SCRAMBLE =      	35128
...
+	for id, num in pairs(inject) do
+		cmds[id] = num
+	end
+end
+if VFS.FileExists("LuaRules/Configs/customcmds_map.lua") then
+	local inject = VFS.Include("LuaRules/Configs/customcmds_map.lua")
```
sprunk, line 135: VFS mode strictness
```suggestion
if VFS.FileExists("LuaRules/Configs/customcmds_map.lua", VFS.MAP) then
	local inject = VFS.Include("LuaRules/Configs/customcmds_map.lua", nil, VFS.MAP)
```

Arch-Shaman:
What is "VFS strictness" btw?

sprunk:
In this case it means a mod can't overwrite the map injects and vice versa, by accident or otherwise.

sprunk:
I still think this is just perpetrating a cargo cult.

 * your original problem was overriding customcmds. Anybody else who does this won't actually get any benefit because their file is already overridden. They may be somewhat dissuaded from overriding the file if they see the `Include()` at the bottom but that can also be achieved via comments.
 * why do you want to inject stuff to customcmds? No existing widget/gadget will attempt to read `CMD_SWEEP_FIRE` (or whatever) from the file so it doesn't matter for them if you do this. Your own wupgets, the ones that do want to read `CMD_SWEEP_FIRE` etc., can read it directly from the modside file.
 * currently the file is pure Lua without any library calls. Adding `VFS` calls imposes extra requirements (for example you can no longer include it with an empty environment, potentially breaking some user widgets which didn't want their environment polluted with a ton of random globals).
 * currently the file is also just a single table of values. If you combine multiple tables it would be ideal to also introduce some sort of handling for when values from multiple tables clash (instead of just silently overriding as in this PR).

As far as I can tell the only benefit of this approach is that instead of
```
VFS.Include("customcmds.lua")
VFS.Include("customcmds_FW.lua")
```
you can be lazy and write just
```
VFS.Include("customcmds.lua")
```
which just obscures where values are coming from and makes it harder to reuse in another mod.

PR #4350, opened 2021-03-16
Retreat improvements

Arch-Shaman:
Retreat command now takes into account capture damage and shield health.

Untested. (Different from FW)

### DIFF
--- a/LuaRules/Gadgets/cmd_retreat.lua
+++ b/LuaRules/Gadgets/cmd_retreat.lua
@@ -63,6 +63,8 @@ local spGetUnitRulesParam 	= Spring.GetUnitRulesParam
 local spSetUnitRulesParam 	= Spring.SetUnitRulesParam
 local spFindUnitCmdDesc 	= Spring.FindUnitCmdDesc
 local spGetUnitIsStunned 	= Spring.GetUnitIsStunned
+local spGetUnitDefID            = Spring.GetUnitDefID
+local spGetUnitShieldState      = Spring.GetUnitShieldState
 
 local CMD_INSERT = CMD.INSERT
 local CMD_REMOVE = CMD.REMOVE
@@ -100,6 +102,18 @@ for unitDefID, unitDef in pairs(UnitDefs) do
 	local movetype = Spring.Utilities.getMovetype(unitDef)
 	if movetype == 1 and not Spring.Utilities.tobool(unitDef.customParams.cantuseairpads) then
 		gunshipDefs[unitDefID] = true
+	end
+end
+
+for i = 1, #UnitDefs do
+	local def = UnitDefs[i]
+	local weapons = def.weapons
+	for w = 1, #weapons do
+		local wep = weapons[w].weaponDef
+		if WeaponDefs[wep].shieldPower and WeaponDefs[wep].shieldPower > 0 and shields[i] == nil then
+			shields[i] = WeaponDefs[wep].shieldPower
+			--Spring.Echo("Added shield retreat to " .. i .. " ( has " .. tostring(WeaponDefs[wep].shieldPower) .. ")")
+		end
 	end
 end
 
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/cmd_retreat.lua
```diff
@@ -103,6 +105,18 @@ for unitDefID, unitDef in pairs(UnitDefs) do
 	end
 end
 
+for i = 1, #UnitDefs do
+	local def = UnitDefs[i]
+	local weapons = def.weapons
+	for w = 1, #weapons do
+		local wep = weapons[w].weaponDef
+		if WeaponDefs[wep].shieldPower and WeaponDefs[wep].shieldPower > 0 and shields[i] == nil then
+			shields[i] = WeaponDefs[wep].shieldPower
```
sprunk, line 114: Microoptimisation:
```suggestion
		local shieldPower = WeaponDefs[wep].shieldPower
		if shieldPower and shieldPower > 0 and shields[i] == nil then
			shields[i] = shieldPower
```

LuaRules/Gadgets/cmd_retreat.lua
```diff
@@ -375,14 +389,21 @@ local function CheckSetWantRetreat(unitID)
 	if not retreatState[unitID] or retreatState[unitID] == 0 then
 		return
 	end
-	
+	local shieldratio
+	local unitdef = spGetUnitDefID(unitID)
+	if shields[unitdef] then
+		local _, currentcharge = spGetUnitShieldState(unitID)
+		shieldratio = currentcharge / shields[unitdef]
```
sprunk, line 396: Name (it's not the unit def, it's the ID) and microoptimisation:
```suggestion
	local unitDefID = spGetUnitDefID(unitID)
	local maxShield = shields[unitDefID]
	if maxShield then
		local _, currentcharge = spGetUnitShieldState(unitID)
		shieldratio = currentcharge / maxShield
```

GoogleFrog:
So now shielded units will retreat on with drained shields even at full health?

Arch-Shaman:
Yes, but they'll return to the fight after fully shielded. Alternatively I can make their unretreat condition being > math.min(threshold + threshold*0.1, 1)

sprunk:
I'm not sure if it's good for area shields to leave and lose access to smaller shields recharging it while also stripping them of its protection. Unlike health, shields regenerate in combat and extend to other units. If the threshold for unretreat is near the threshold for retreat then they just oscillate.

Capture causing retreat sounds okay at a glance. There is some worry because currently the only source of capture is Domi who has very high DPS but high reload. Check some common units that Domi likes to own (e.g. Knight, Ogre, Minotaur) on fight move vs a Domi or 2 and see if they can actually retreat before getting captured at various retreat levels.

sprunk:
Adding more handled cases to retreat raises the question of what retreat is supposed to be. The main choice seems to be between:
 * being a way to avoid APM sink into non-decisions. If you have a blob of units you no longer have to select wounded things one by one, send back to base, and go back to commanding the whole blob. Attention is still required so that you can make decisions about the trade-offs that status effects bring. This is the current state of affairs.
 * avoiding an APM sink, but also being a way to automatically make the decision regarding status effect trade-offs in favour of keeping the individual unit alive.

I am in favour of the first approach (minimize Retreat decision-making). This is not to say these improvements shouldn't be made but I think they belong to Fight with Unit AI, not to Retreat. This is because Fight Unit AI is also a safety thing (autoskirm etc) but is just easier to override if you decide you want an aggressive choice instead (just rightclick a Move command; Retreat is annoying and prevents control because it deselects the unit). For example:
 * disarmed units at >3s timer could just blanket skirm everything armed (asspulled timer value that makes sure they loiter in the back, but return into combat range by the time it dissipates).
 * Minotaur (and similar units) could skirm Domi when capture reaches some threshold such that it will comfortably get outside a Domi's range to avoid getting leashed but still move in to pound her when it is safe.

GoogleFrog:
I often drive disarmed units into enemy positions because I know many of them will survive for the 5s it takes the disarm to wear off.

I think a bunch of hidden state toggles may be the best way to go.

Arch-Shaman:
I've been looking into unit AI more lately. I'll see what I can do about moving this off to unitAI instead.

sprunk:
Reopen if there are advances.

PR #4351, opened 2021-03-16
Fix Clear Factory Queue from not stopping production on wait

Arch-Shaman:
Though now wait isn't re-added (which is acceptable losses, tbh)

See: http://zero-k.info/Battles/Detail/1072832

### DIFF
--- a/LuaRules/Gadgets/cmd_factory_stop_production.lua
+++ b/LuaRules/Gadgets/cmd_factory_stop_production.lua
@@ -29,6 +29,8 @@ local spInsertUnitCmdDesc  = Spring.InsertUnitCmdDesc
 
 local CMD_OPT_CTRL = CMD.OPT_CTRL
 local CMD_REMOVE = CMD.REMOVE
+local CMD_WAIT = CMD.WAIT
+local EMPTY = {}
 
 include("LuaRules/Configs/customcmds.h.lua")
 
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/cmd_factory_stop_production.lua
```diff
@@ -76,6 +78,8 @@ function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOpt
 	for i = 1, #commands do
 		spGiveOrderToUnit(unitID, CMD_REMOVE, commands[i].tag, CMD_OPT_CTRL)
 	end
+	spGiveOrderToUnit(unitID, CMD_WAIT, EMPTY, CMD.OPT_CTRL)
```
sprunk, line 81: There might be some `GG.WaitWait` for this sort of thing. What is the `OPT_CTRL` for?

LuaRules/Gadgets/cmd_factory_stop_production.lua
```diff
@@ -29,6 +29,8 @@ local spInsertUnitCmdDesc  = Spring.InsertUnitCmdDesc
 
 local CMD_OPT_CTRL = CMD.OPT_CTRL
 local CMD_REMOVE = CMD.REMOVE
+local CMD_WAIT = CMD.WAIT
+local EMPTY = {}
```
GoogleFrog, line 33: This seems redundant for how often it is used.

LuaRules/Gadgets/cmd_factory_stop_production.lua
Arch-Shaman: Unsure if it's necessary. I just copied what the remove portion was doing (without understanding what it does). I just know it works. I'll fish around for that global.

Arch-Shaman: Turns out CTRL and 0 do the same thing, switched to 0. (See #ZKDev)

Arch-Shaman:
Solution in action:
https://youtu.be/ZhrhxJsMTVI

(See 0 label)

PR #4352, opened 2021-03-16
Increase Silo sound limit

### LINKED ISSUE #4074: Missile Silo not playing launch sound
A well known issue of Missilo Silo being a sleazy bag of misfortune. Sometimes, especially during consecutive launches, Silo doesn't play a launch sound, which is sorta crucial for the opposing team. This seems to happen to **all** players on the server (nobody hears the sound when the bug occurs).

Arch-Shaman:
Sets the limit of silo launches to 30 (which should not be reached in any normal scenario)
Fixes #4074


sprunk:
The lua part looks okay but I'm deaf not blind so GF should probably look at the sound file.

Arch-Shaman:
I basically sliced the silence off the tail. There's about 1s-ish of silence in the tail.

Before:
image

After:
image

Arch-Shaman:
Turns out there's just really soft sound there. I'll revert it.

PR #4363, opened 2021-03-24
Decrease nimbus combat range a bit

### LINKED ISSUE #4337: Nimbuses sometimes get a stroke when firing at Lotus
Now that I've seen that one twice as well, - Nimbus sometimes stops firing after getting an attack order. Seemingly this only happens when targeting Lotus, at least from the games I've witnessed.
https://zero-k.info/Battles/Detail/1070209, ~20% runtime. Middline Nimbus gets a force-attack order on Lotus and doesn't do much of anything (doesn't fire a single shot actually).
https://zero-k.info/Battles/Detail/1055590, ~85% runtime. The attack moved Nimbus at NW corner switches target for a Lotus and stops attacking after a few shots.

Arch-Shaman:
Potentially fixes #4337 (tested, seems to work?)
* Decreases combat range by 20 (makes nimbus on flat ground get into range properly)

### DIFF
--- a/units/gunshipheavyskirm.lua
+++ b/units/gunshipheavyskirm.lua
@@ -82,7 +82,7 @@ return { gunshipheavyskirm = {
       craterMult              = 0.3,
 
       customparams = {
-        combatrange = 650,
+        combatrange = 630,
         light_camera_height = 2000,
         light_color = [[0.9 0.84 0.45]],
         light_ground_height = 120,


sprunk:
Check if this is still enough to keep it safe from things around that range (HLT, picket)

Arch-Shaman:
Will do.

Arch-Shaman:
HLT is safe.
Picket is also safe.

Issue #2548, opened 2017-10-07
Area reclaim for Puppy


GoogleFrog:
An area reclaim command would be useful for Puppy. A hack would be to make it a constructor that can only reclaim wreckages, not units or other features. I don't know whether a constructor is detected as a constructor with 0 build power. It could have positive build power and always be blocked in allowBuildStep. Alternately, it could actually have build power and in allowBuildStep it would use the income to charge its internal storage. A downside of this solution is that Fight in wreckage strewn areas would be annoying.

The other solution is some sort of lua command.

mtroyka:
I think a lua command would be pretty easy for this. "Get consumable wrecks within circle, send puppy to nearest wreck, repeat periodically".

sprunk:
>It could have positive build power and always be blocked in allowBuildStep. Alternately, it could actually have build power and in allowBuildStep it would use the income to charge its internal storage. 

Sounds like a great use of engine `harvest storage`.

GoogleFrog:
Sure, if harvest storage works.

PR #4368, opened 2021-04-01
[WIP] Add an automated grey goo command

Arch-Shaman:
This adds a command for grey goo units that will:
* When issued on a wreck the unit will move to it then consume it before performing the next command. (This is useful if that Dante wreck or whatever is nearby and you want to reclaim the puppies or something before moving on to some other task)
* When issued as an area command, the unit will move to and consume all wrecks in an area (starting with nearby ones then working to the next closest one).

The goal of this pr is to automate grey goo units in a way users can set up a command and forget about the task.

TODO:
- [ ] Add custom icon (currently uses reclaim I think? May not be necessary.)
- [ ] Grey goo units spawned from hosts that are performing CMD_GREYGOO inherit the cmd and add a guard order afterwards.
- [ ] Stop performing cmd when the greygoo state is toggled to off?
- [ ] Add demo video.
- [ ] Better command name. "Reclaim (Grey Goo)" may be misleading?

### DIFF
--- a/LuaRules/Gadgets/cmd_greygoo.lua
+++ b/LuaRules/Gadgets/cmd_greygoo.lua
@@ -1,0 +1,203 @@
+function gadget:GetInfo()
+	return {
+		name      = "Area Grey Goo Handler",
+		desc      = "Units will consume all wreckage in an area",
+		author    = "Shaman",
+		date      = "April 1, 2021",
+		license   = "CC-0",
+		layer     = 5,
+		enabled   = true,
+	}
+end
+
+if (not gadgetHandler:IsSyncedCode()) then
+	return
+end
+local IterableMap = VFS.Include("LuaRules/Gadgets/Include/IterableMap.lua")
+
+local handled = IterableMap.New()
+local wantedUnitDefs = {}
+local featurecache = {}
+
+local areaGreyGooDesc = {
+	id      = CMD_GREYGOO,
+	type    = CMDTYPE.ICON_UNIT_FEATURE_OR_AREA,
+	name    = 'Reclaim (Grey Goo)', -- TODO: better name. Marketing was out today.
+	action  = 'reclaim', -- this may break things for modders with cons that grey goo!
+	tooltip	= 'Marks an area or wreckage for grey goo.',
+}
+
+-- Speed ups --
+local spGetUnitPosition = Spring.GetUnitPosition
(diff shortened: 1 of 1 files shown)


sprunk:
#2548

Arch-Shaman:
Demo: https://youtu.be/DuX04atLlwU

sprunk:
ALT should make a persistent area command.
Wrecks being rezzed should be ignored for area command (but not single-target command).

Arch-Shaman:
It can't be ignored if the puppies run into range of them on accident or if it's next to an unprotected wreck.

sprunk:
Yes but it shouldn't be actively moved to.

LuaRules/Gadgets/cmd_greygoo.lua
```diff
@@ -0,0 +1,230 @@
...
+				return true, true
+			end
+			local persistent = cmdOptions.alt and #cmdParams > 1
+			IterableMap.Add(handled, unitID, {def = unitDefID, done = false, params = cmdParams, goal = -9999, updates = 0, persistent = persistent}) -- we found a new unit!
+		end
+		return true, false -- we're still not done here.
+	end
+end
+
+function gadget:GameFrame(f)
```
GoogleFrog, line 171: I would think that the bulk of the command handling could happen in CommandFallback.

LuaUI/Configs/integral_menu_commands.lua
```diff
@@ -43,6 +43,7 @@ local cmdPosDef = {
 	[CMD_AREA_MEX]      = {pos = 7, priority = 1},
 	[CMD.REPAIR]        = {pos = 7, priority = 2},
 	[CMD.RECLAIM]       = {pos = 7, priority = 3},
+	[CMD_GREYGOO]       = {pos = 7, priority = 3},
```
GoogleFrog, line 46: Don't duplicate values here as it could cause random swapping.

LuaRules/Gadgets/cmd_greygoo.lua
Arch-Shaman: CommandFallback calls once every long update from what I understand. I'll test it out.

LuaUI/Configs/integral_menu_commands.lua
Arch-Shaman: 8059d87a2

PR #4373, opened 2021-04-03
Apply Inculta/Alien Desert startboxes to Sertagatta and Sertaleina.

Aquanim:
Not sure if this is actually a good idea for Sertaleina (Alien Desert), though.


GoogleFrog:
Alien Desert could be improved from the simple corners, but this is as good as the current situation.

Issue #2149, opened 2017-04-10
Plopping a factory makes com queue "stuck"


let4be:
http://zero-k.info/Battles/Detail/453920

I plopped a factory and com did not proceed with the queue trying to smooth ground beneath the factory... I had to manually abort this nonsense and requeue everything
(look for Firepluk)

GoogleFrog:
This is easier to solve than the general bug. When fac plop is used get the builderID of the factory and remove the factory build order from the queue of the builder. It will fix basically all cases.

Issue #2108, opened 2017-03-22
Investigate some terraform bug


GoogleFrog:
Balgias: often having trouble with units trying to start on a terraforming order only to have them never actually start building it, then it doesn't cancel because ally nanos are trying as well, forcing me to mass select them and ctrl-d

let4be:
I often encounter another weird terra related bug: I make pre-game queue(fac, mexes, solars, etc)
com plops the factory and starts "building" tho it's ready and can produce units, it and refuses to continue going on with the queue unless I cancel the queue and manually tell com what to do next

GoogleFrog:
That is not related to this issue. Make a new one.

let4be:
I suppose this can be this built-in into engine "terraform before building can be placed",
for facs it can plop factory and then start smooth ground beneath it, at least it looks like this

sprunk:
Nothing actionable here.

PR #4384, opened 2021-04-11
Further clean up of start_unit_setup

Arch-Shaman:
* Removed polling code: This is some remnant from the pre-92.0 days or something. In modern days, it's pretty much pointless to do this as you get it for free from the engine (my guess is it was to stop command locking up or something). 
* Removed the unused variable "gaiaally"

### DIFF
--- a/LuaRules/Gadgets/start_unit_setup.lua
+++ b/LuaRules/Gadgets/start_unit_setup.lua
@@ -32,12 +32,10 @@ local setAiStartPos = (modOptions.setaispawns == "1")
 local CAMPAIGN_SPAWN_DEBUG = (Spring.GetModOptions().campaign_spawn_debug == "1")
 
 local gaiateam = Spring.GetGaiaTeamID()
-local gaiaally = select(6, spGetTeamInfo(gaiateam, false))
 
 local SAVE_FILE = "Gadgets/start_unit_setup.lua"
 
 local fixedStartPos = (modOptions.fixedstartpos == "1")
-local ordersToRemove
 
 local storageUnits = {
 	{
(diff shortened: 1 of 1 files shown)


sprunk:
I have doubts about the CMD.REMOVE thing. It may appear to work correctly on a straightforward test but IIRC the original issue (#2149) was rare.

Try plopping facs in weird spots (uneven terrain, water, terraform in progress by another con?) or other weird circumstances (disarmed commander? change torso turn speed to something low in script and plop behind the comm's back? multiple orders further in the queue, including a second factory in the comm's immediate build radius?) and see if it works correctly without the removal.

Arch-Shaman:
I'll try it out right now.

FWIW, I've had a version of this change live for a few weeks now in FW and haven't seen anything adverse happen. This thing seems like one of those hard to overcome measures.

Case 1: Smooth -> facplop (Seems fine)
image

Case 2: Disarmed commander (actually completes immediately... interesting! This could be fixed by moving facplop to buildstep if it is in fact a bug)
image

Case 3: Another in range factory (Seems fine too)

The current changes are live in future wars and I have not heard of anything regressive going on, but I lack the foot traffic to make a call on this. I could adapt a similar system I use in the current facplop code in FW to the fix and negate some of this. Problem is doing it immediately causes recursion errors occasionally (this is how I took this risk anyways).

Arch-Shaman:
More cleanup: the entire "faction" stuff can be cleaned up as nothing interacts with the system (as far as I can tell) or sends the message. AFAIK is just synonym with customcomm.

sprunk:
The command removal may have been to fix this: https://github.com/ZeroK-RTS/Zero-K/issues/2108#issuecomment-289261012

GoogleFrog:
Disarmed units should not be able to create nanoframes. I recall adding a hax to factory plates to make them unable to do it.

sprunk:
I accepted the unquestionably good part but I am too scared to merge the rest because the original issue was some sort of rare edge case so will not likely be caught by testing.

PR #4385, opened 2021-04-12
Cycle AFKers through avaliable start positions

Arch-Shaman:
Modified GetMiddleOfBox to cycle through startpoints for afk players by tracking the number of calls each allyteam makes and indexing the startpoint associated with that number (cycling back after hitting the maximum start points).

### DIFF
--- a/LuaRules/Gadgets/start_unit_setup.lua
+++ b/LuaRules/Gadgets/start_unit_setup.lua
@@ -31,6 +31,7 @@ local setAiStartPos = (modOptions.setaispawns == "1")
 local CAMPAIGN_SPAWN_DEBUG = (Spring.GetModOptions().campaign_spawn_debug == "1")
 
 local gaiateam = Spring.GetGaiaTeamID()
+local allyTeamAFKers = {}
 
 local SAVE_FILE = "Gadgets/start_unit_setup.lua"
 
@@ -236,7 +237,7 @@ local function GetFacingDirection(x, z, teamID)
 			or ((z>Game.mapSizeZ/2) and "north" or "south")
 end
 
-local function getMiddleOfStartBox(teamID)
+local function GetRecommendedStartPosition(teamID, n) -- allyteams can have multiple start positions, go and get the nth one.
 	local x = Game.mapSizeX / 2
 	local z = Game.mapSizeZ / 2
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
`getMiddleOfStartBox(teamID)` strikes me as a functional function. I do not like the addition of these side effects. Also why is the function returning `local startpos = startposList[1]`? The names of things in this file suggest that its purpose has drifted.

Arch-Shaman:
getMiddleOfStartBox is a misleading function name that was never updated during feature addition. 

The actual effect of it is that it returns a starting position for a player (see changes). This **used** to be the middle of the start box, however, after the addition of sprungboxes (circa 2017?), it now looks to see if a map has lua startboxes then fetches the first recommended start position and dumps players there, and as a fallback, it returns the middle of the map (what?),

Arch-Shaman:
Done. I've changed it to GetRecommendedStartPosition since startposList is actually a list of recommended start positions that mappers/us can make for maps.

Arch-Shaman:
Reporting test success in FW: https://github.com/Arch-Shaman/ZK-Futurewars-Mod/commit/388b2ba968ce1fdabb7ae384bff76f170ef680fa

Some notes:
* Players can get forced into multiple fronts with this. (though team games can be exited if they run into significant amounts of afkers)
* A potential improvement could be checking for non-afk players in recommended start positions so that teams aren't completely fucked over by one person going afk and droping in index 1 (where there's like 2-3 players already).

sprunk:
Single afker dropping at index 1 is the current behaviour so this is not a regression.

Arch-Shaman:
I mean like if there's a player near index 1's position we could push them to 2 so as to spread players out a bit more.

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -254,15 +255,16 @@ local function GetFacingDirection(x, z, teamID)
 			or ((z>Game.mapSizeZ/2) and "north" or "south")
 end
 
-local function getMiddleOfStartBox(teamID)
+local function GetRecommendedStartPosition(teamID, n) -- N is the number of times an ally team has called this
```
GoogleFrog, line 258: I think this comment is wrong and still implies a non-functional paradigm. Just say that an allyTeamID can have more than one recommended startpos, and that GetRecommendedStartPosition(teamID, n) returns the (n % number of startpos)th one.

Arch-Shaman: Fixed via 32a40a8

PR #4389, opened 2021-04-19
Misc unit_auto_patrol_nanos_v2 improvements.

### LINKED ISSUE #4226: Smart Nanos stuck on reclaim for a while when excessing metal.

rtwfroody:
* Do less work when viewing a replay.
* Add some more debug info (only when the widget is edited to set
  debug=true).
* Fix my understanding of reclaim commands. There is no "reclaim
  energy," but only "reclaim all." I think this fixes #4226.
* Tolerate spGetUnitResources() returning nil, which happened to me
  once.

I've been using this for 2 weeks with no issues.


sprunk (approved):
Looks good

rtwfroody:
Bump.

GoogleFrog:
I'll mark it for the next PR update. I think there will be a few small handicap tweaks for now so don't want to add potential bugs for now.

rtwfroody:
Thank you.

PR #4401, opened 2021-05-18
Allow moderators and game hosts to bypass specmute option

### LINKED ISSUE #4192: Spec chat mute that ignores moderators

Arch-Shaman:
Moderators, game hosts, and players with the TBI `helpful` custom param may bypass spectator mute modoption. This is to make it convenient for enforcement/information providing.

Implement/fixes #4192.

### DIFF
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -1146,11 +1146,7 @@ local function GetPlayerAbilityToSpecchat(playerID)
 	local boss = cp["room_boss"] or "0"
 	local admin = cp.admin or "0"
 	local helpful = cp.helpful or "0"
-	if boss == "1" or admin == "1" or helpful == "1" then
-		return true
-	else
-		return false
-	end
+	return boss == "1" or admin == "1" or helpful == "1"
 end
 
 
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -1140,6 +1140,20 @@ local function FindLowestIndex(t, i, layer)
 	return 1
 end
 
+local function GetPlayerAbilityToSpecchat(playerID)
+	-- this function controls when players can bypass the specmute modoption.
+	local cp = select(10, Spring.GetPlayerInfo(playerID))
+	local boss = cp["room_boss"] or "0"
+	local admin = cp.admin or "0"
+	local helpful = cp.helpful or "0"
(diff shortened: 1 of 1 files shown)


sprunk:
Similar feedback as #4193:
 * admins (and room bosses) can already use `!hostsay` to communicate, I think it's good that you need to be explicit around breaking the mute
 * `helpful` needs infra support

Arch-Shaman:
Problems with hostsay:
* Very inconvenient for long conversations (eg: explaining something / moderation)
* Easy to miss (the text is very dark and on some maps it's flat out ignored)

I'm not sure how specmute (CP) works with it. Someone will need to specmute one of my test accounts for it.

LuaUI/cawidgets.lua
```diff
@@ -1440,14 +1450,17 @@ function widgetHandler:AddConsoleLine(msg, priority)
 		local newMsg = { text = msg, priority = priority }
 		MessageProcessor:ProcessConsoleLine(newMsg) --chat_preprocess.lua
 		if newMsg.msgtype ~= 'other' and newMsg.msgtype ~= 'autohost' and newMsg.msgtype ~= 'userinfo' and newMsg.msgtype ~= 'game_message' then
+			local playerID_msg = newMsg.player and newMsg.player.id --retrieve playerID from message.
```
Licho1, line 1453: I think the ability to speak should be written in startscript instead of evaluated in the game itself. 
That way we would have record of what the value was and easier way to tweak logic or extend it with more data.
(For example turney managers etc)

Arch-Shaman: Is there a location I can read on startscripts to understand this statement better? It does not seem to be on the wiki that I can see.

sprunk: Gameside it's just player customkeys (basically the `cp.helpful` you already have).

sprunk:
This won't go anywhere without infra support.

PR #4405, opened 2021-05-21
Fix Stronghold team names

Arch-Shaman:


GoogleFrog:
I dislike that startbox tables are zero indexed, but I guess that is the world we live in.

sprunk:
I made them 0-based so that they correspond to allyteams without any transformation. We could probably transform them to 1-based fairly easily (any map-side config that supplies box 0 would just have them all shifted by 1).

PR #4408, opened 2021-05-24
Add unit sounds for subscout

Arch-Shaman:


sprunk:
Last time you wanted to keep random chaff on the grounds of being used by a mod though? a9ff1afc8efbc3224ec0d27857c04ba14846d7ad

Is the rule "bite the bullet and maintain the stuff that somebody already uses but don't increase that set"?

GoogleFrog:
I wanted to keep that piece of random chaff because its removal was actively breaking mods.Ideally we wouldn't have anything like it.

PR #4412, opened 2021-05-25
Retreat now removes tactical AI commands

### LINKED ISSUE #4411: Retreat AI gets confused by tactical AI commands
It would seem that retreat command confuses the move issued by tactical ai (eg: with skirmees) as a human input move.

http://zero-k.info/Battles/Detail/1122907 here I had some tsunamis which skirm the enemy. Several times their move orders (from fight command) are kept as its retreating. Since FW retreat is essentially stock with #4350 merged, the issue likely exists in bg as well.

Arch-Shaman:
Fixes #4411.

Retreat AI would keep the internally generated tactical AI commands in the queue when inserting the move command. This lead to units moving back to whereever the unit AI wanted to move tactically before reengaging fight command. Note that this should not remove user created move commands.

### DIFF
--- a/LuaRules/Gadgets/cmd_retreat.lua
+++ b/LuaRules/Gadgets/cmd_retreat.lua
@@ -210,7 +210,8 @@ local function FixQueue(unitID) -- this gets rid of internally generated command
 	local queue = spGetCommandQueue(unitID, 3)
 	for i = 1, #queue do
 		local command = queue[i]
-		if command.options.internal and (command.id == CMD.MOVE or command.id == CMD_RAW_MOVE or command.id == CMD.FIGHT) then
+		local cmdID = command.id
+		if command.options.internal and (cmdID == CMD.MOVE or cmdID == CMD_RAW_MOVE) then
 			GG.recursion_GiveOrderToUnit = true
 			local tag = command.tag
 			spGiveOrderToUnit(unitID, CMD_REMOVE, {tag}, 0)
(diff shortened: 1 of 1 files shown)


sprunk:
A potential worry: a unit with an internal move order will go back to where it started the retreat from (circle on this pic) and then continue on the path the player initially had in mind, but with the internal command removed it will go from the retreat point straight to the destination (via gigaporc on this pic).
image

sprunk:
Another thing: it would be good to check for other types of lingering internal commands (Fight spawned from Patrol? Attack spawned from some engine idle behaviour?) and handle them as well.

Arch-Shaman:
Patch demo can be found here: http://zero-k.info/Battles/Detail/1123039 (there's some moments when the retreat behavior kind of screws itself over due to pathing into enemy territory during retreat, but otherwise it felt pretty nice)

GoogleFrog:
Why not GG with tactical AI for the removal rather than guessing the format of its commands? It would be pretty easy to make a GG that tells tactical AI to cancel the commands it has given, since it remembers where they are and cancels them during normal function.

Is the intention here to remove all internal move commands (which can come from other places) or to only hit those added by tactical AI?

Arch-Shaman:
Mm that would work. The intention is for all internal move commands to be cleared. Moving units is dangerous IMO. Move command units do not fight back if they encounter enemy units. They'll happily run into Jacks or other things. In our test version we also removed the idle return move and replaced it with a fight command.

LuaRules/Gadgets/cmd_retreat.lua
```diff
@@ -206,6 +206,19 @@ GG.Retreat_ToggleHaven = ToggleHaven
 ----- Retreat Handling
 ----------------------------
 
+local function FixQueue(unitID) -- this gets rid of internally generated commands by tactical ai and keeps user commands.
+	local queue = spGetCommandQueue(unitID, 3)
```
GoogleFrog, line 210: The 3 here seems fragile.

Issue #4229, opened 2020-12-12
Potential fire at radar recursion


GoogleFrog:
https://github.com/ZeroK-RTS/CrashReports/issues/36241

[f=0018377] Error: [LuaRules::RunCallInTraceback] error=2 (LUA_ERRRUN) callin=AllowWeaponTarget trace=[Internal Lua error: Call failure] [string "LuaRules/Gadgets/unit_dont_fire_at_radar.lu..."]:104: [GiveOrderToUnit] recursion not permitted
stack traceback:
	[C]: in function 'spGiveOrderToUnit'
	[string "LuaRules/Gadgets/unit_dont_fire_at_radar.lu..."]:104: in function 'AllowWeaponTarget'
	[string "LuaRules/gadgets.lua"]:1334: in function <[string "LuaRules/gadgets.lua"]:1310>
	(tail call): ?
	[C]: in function 'spGiveOrderToUnit'
	[string "LuaRules/Utilities/ClampPosition.lua"]:57: in function 'GiveClampedOrderToUnit'
	[string "LuaRules/Gadgets/unit_tactical_ai.lua"]:987: in function 'DoUnitUpdate'
	[string "LuaRules/Gadgets/unit_tactical_ai.lua"]:1017: in function 'UpdateUnits'
	[string "LuaRules/Gadgets/unit_tactical_ai.lua"]:1032: in function 'GameFrame'
	[string "LuaRules/gadgets.lua"]:957: in function <[string "LuaRules/gadgets.lua"]:955>
	(tail call): ?

GoogleFrog:
This is a tricky one. It looks like issuing a FIGHT command can cause AllowWeaponTarget to trigger instantly, and thus prevents more commands being issued in AllowWeaponTarget.

sprunk:
Engine now allows some recursion.

PR #4413, opened 2021-05-25
Blind attempt fix at recursion

Arch-Shaman:
Saw this global set in other examples, so I think it may be a solution. I'm not entirely sure. Plus this stuff is hard to track to see if it fixes stuff. See commentary here: https://discordapp.com/channels/278805140708786177/278908415756206080/845882330300284938


Arch-Shaman:
Attempts to address https://github.com/ZeroK-RTS/Zero-K/issues/4229

GoogleFrog:
I don't see why this would do anything to that crash report. I think what you need is a recursion check that prevents don't fire at radar issuing the command if it is within another order, as indicated by GG.recursion_GiveOrderToUnit.

sprunk:
I think the blindfix to that issue would rather look like 6fad995f12f9445e11ff26a329e9a3a3f0793ead

sprunk:
Not the correct fix, as mentioned above.

PR #4416, opened 2021-05-27
Jumpers now use their jump on cooldown during retreats

Arch-Shaman:


sprunk:
Sumo shouldn't use it on cooldown, only the first time. Ditto for units whose regular movement speed is faster (this may be none of them at the moment, probably calculate this in posts?). Det should probably not use it even for the first time.

scripts/JumpRetreat.lua
```diff
@@ -3,22 +3,31 @@ local CMD_JUMP = Spring.Utilities.CMD.JUMP
 local GiveClampedOrderToUnit = Spring.Utilities.GiveClampedOrderToUnit
 local jumpRange = tonumber(UnitDefs[unitDefID].customParams.jump_range)
 
-local firstTime = true
```
sprunk, line 6: What happens if you remove `firstTime` instead?

scripts/JumpRetreat.lua
```diff
@@ -3,22 +3,31 @@ local CMD_JUMP = Spring.Utilities.CMD.JUMP
...
+	local reload, disarmed
+	while retreating do
+		reload = Spring.GetUnitRulesParam(unitID, "jumpReload") or 1
+		disarmed = (Spring.GetUnitRulesParam(unitID, "disarmed") or 0) == 1
+		if reload >= 1 and not disarmed then
 			local ux, uy, uz = Spring.GetUnitPosition(unitID)
 			local moveDistance = math.sqrt((ux - hx)^2 + (uz - hz)^2)
 			local disScale = jumpRange/moveDistance*0.95
 			local cx, cy, cz = ux + disScale*(hx - ux), hy, uz + disScale*(hz - uz)
 			GiveClampedOrderToUnit(unitID, CMD.INSERT, { 0, CMD_JUMP, CMD.OPT_INTERNAL, cx, cy, cz}, CMD.OPT_ALT)
```
sprunk, line 18: This could use a `Sleep(nominalReload)` as an optimisation (read from customparams).

scripts/JumpRetreat.lua
Arch-Shaman: I'll take a look. As far as I know the way retreat interacts with it is that the function is called when the unit is told to start retreating and that is it. My guess is that it would try jumping every time it is called to retreat. Uncertain. I'll rig up a test version with it. RetreatFunction is definitely not a thread though so it wouldn't work as we think it would.

scripts/JumpRetreat.lua
Arch-Shaman: Perhaps make it Sleep for the nominal reload after giving the order. If I understand how the code is working this would then just ask if it is disarmed or reloading still afterwards (see: slowed).

Arch-Shaman:
This seems to be creating OOM problems in FW?

[f=0015130] [unit_script.lua] Error: 6
[f=0015130] Error: [LuaRules::RunCallInTraceback] error=4 (LUA_ERRMEM) callin=GameFrame trace=[Internal Lua error: Call failure] not enough memory
[f=0015130] Error: [LuaRules::RunCallInTraceback] error=4 (LUA_ERRMEM) callin=GameFrame trace=[Internal Lua error: Call failure] not enough memory
[f=0015130] Error: [LuaUI::RunCallInTraceback] error=4 (LUA_ERRMEM) callin=GameFrame trace=[Internal Lua error: Call failure] not enough memory

Sharkgamedev:
Are you sure that's not related to hoverboard?

GoogleFrog:
Why?
 - Jump is only used because it probably reduces incoming damage and avoids scrums blocking retreat.
 - If simply using jump off cooldown increases strategic speed then jump should be slower.
 - I think having the option to manually stop retreating and jump is valuable.
 - Doesn't jumping every time increase the risk of getting stuck?

sprunk:
This is about increasing strategic speed. Jump would have to be slower than walking if it was to be unviable to use on cooldown when retreating, which would be fairly sad because then you lose the option to use it to engage against some unit types.

scripts/JumpRetreat.lua
Arch-Shaman: I've found it to do nothing but jump once.

sprunk:
See 3rd and 4th bulletpoints in GF's last post for why this would be a bad default behavior and the 2nd for why it is a bad behaviour in general.

PR #4420, opened 2021-06-12
all cegs with the ground flag set also have voidground

Anarchid:
This fixes the issue of no CEGs showing up on maps like Core Platform and Violet Rampart.

image

### DIFF
--- a/gamedata/explosions.lua
+++ b/gamedata/explosions.lua
@@ -16,7 +16,7 @@ end
 for edName, eDef in pairs(explosionDefs) do
 	for fxName, fxDef in pairs(eDef) do
 		if(type(fxDef) == 'table') then
-			if(fxDef.ground) then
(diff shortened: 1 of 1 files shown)


gamedata/explosions.lua
```diff
@@ -12,4 +12,15 @@ local explosionDefs = {}
 for i = 1, #files do
 	suCopyTable(vfsInclude(files[i]), false, explosionDefs)
 end
+
+for edName, eDef in pairs(explosionDefs) do
+	for fxName, fxDef in pairs(eDef) do
+		if(type(fxDef) == 'table') then
+			if(fxDef.ground) then
```
sprunk, line 19: What if `voidground` is already set?
```suggestion
			if fxDef.ground and fxDef.voidground == nil then
```

PR #4427, opened 2021-06-15
⚙️ Mex circle matches grid radius

sprunk:
image

```diff
+ helps place early solars efficiently because you know the grid without even queuing up the mex
+ looks better when they match (maybe that's just my autism speaking)
- the border may make it harder to tell where the grid ends on very THICC ubermexes
- some maps seem to have the spots tight enough for the bigger circle to look a bit more odd? probably applies both ways though
```

### DIFF
--- a/LuaUI/Widgets/cmd_mex_placement.lua
+++ b/LuaUI/Widgets/cmd_mex_placement.lua
@@ -907,10 +907,10 @@ local function calcMainMexDrawList()
 
 		glColor(0,0,0,0.7)
 		glLineWidth(width*2.4)
-		glDrawGroundCircle(x, 1, z, 40, 21)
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I think the ground circle being that big looks odd. The grid circle doesn't have to be that large though.

Licho1:
Its relatively minor change 40-50 and its only when placing.. i would enable it personally

sprunk:
Mex circles are not only when placing, by default they are always visible:
https://github.com/ZeroK-RTS/Zero-K/blob/ef48ebb2d4f59627b865abbe5584bef6e07b67cb/LuaUI/Widgets/gui_showeco_action.lua#L117-L120

GoogleFrog:
I will take a look later.

PR #4434, opened 2021-06-26
Override delta siege islands 8 way

Arch-Shaman:
Replaces all mexea with 3.1. (Blind edit)
See: http://zero-k.info/Maps/Detail/7314


sprunk:
Pay attention, 3.1 is a minority of mexes on the map.
image

PR #4449, opened 2021-07-09
easyFacing draws arrow for Starlight

Anarchid:
image


GoogleFrog:
I thought about doing this too. I don't like how the size of this if clause is O(N) in the number of units that will be thrown into it.

PR #4460, opened 2021-07-26
Scallop Role Fix

adams164:
Remove Anti-Sub tag from Scallop. It can have it back when it can kill a Seawolf.

### DIFF
--- a/units/amphriot.lua
+++ b/units/amphriot.lua
@@ -1,7 +1,7 @@
 return { amphriot = {
   unitname               = [[amphriot]],
   name                   = [[Scallop]],
-  description            = [[Amphibious Riot Bot (Anti-Sub)]],
+  description            = [[Amphibious Riot Bot]],
   acceleration           = 0.54,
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
You'll need to edit LuaUI/Configs/units_*.json as well.

sprunk:
Thanks, I took care of the translation files bd46429cdd0c800c3d8647db2bff0dc074a4fc79

GoogleFrog:
Hang on. Duck also says (Anti-Sub). It doesn't seem to mean that Duck kills subs. From context it seems to mean that it can shoot at underwater things from underwater.

Reverting for inconsistency. Perhaps use better words or maybe all the words can be removed. Also maybe bring the other units that can shoot submerged units into the (Anti-Sub) label.

adams164:
Duck kills seawolf for cost, scallop does not. Claymore also has the Anti-sub role and is not an underwater unit, but kills seawolf for cost.

Issue #3900, opened 2020-04-27
Default engine Lightning rendering may look better than what we have


GoogleFrog:
BAR has better looking lighting and they seem to be using fewer textures and less lua to get it. The engine seems to have some sort of procedurally generated random lightning line function with it. Test it out on our Lightning.

Anarchid:
Default engine lightning has a very small drift to its shape that is hardwired:

https://github.com/spring/spring/blob/1e514010589653392595a347d8eb3d0e1b97852d/rts/Sim/Projectiles/WeaponProjectiles/LightningProjectile.cpp#L37

This drift does not multiply with beam thickness, so thicker beams perceptually move less. 

ZK has very thick lightning beams.

GoogleFrog:
So?

Anarchid:
Youn can only get *marginally* better engine lightning by having it much less thick. Think Gnat.

sprunk:
I think a lot of the thicc might be in our texture so you wouldn't necessarily have to reduce nominal def thickness.

Anarchid:
A lot of our current thicc is that the texture ZK uses has two bolts, and engine always renders two bolts, so we get four bolts in ZK for lightnings, and things like Gnat only get two.

The def thicc is the thing that has to be reduced for the effect to become visible. Current ZK thicc is 10; The drift is barely noticeable with thicc 3 anyway.

IMO better to just replace it with a shader.

sprunk:
WIP branch https://github.com/ZeroK-RTS/Zero-K/tree/lightning_poc

Issues with the lightning itself:
* wiggle is unconfigurable (width, number of sections).
* the beam is uniformly wide (incl. wiggle). It should converge at the ends.
* moving sideways moves the origin point of the lightning but not the spherical CEG generated at tip of the barrel.
* is a mechanics change as far as gamplay.

Others:
* yellow EMP (Venom, Faraday, Widow) is not ported. It doesn't feel like lightning though so personally I feel like it could stay as-is. Apparently it could stay a discharge but have particle beam style fade to differentiate it. Venom/Faraday have fast reloadtimes so the mechanics change would be far more pronounced, and Widow probably wants the reliability of a single shot.
* burst discharges (Thunderbird, Scorpion ult) are unported. Check if they can even be.

GoogleFrog:
Burst discharges should at least look like a static version of the continual lightning, which then fades out. Actually, a non-gameplay change could be to make everything into fading engine lightning.

PR #4465, opened 2021-08-01
BAR style lightning beam example

sprunk:
JUST the beam itself: no CEG, projectile lights may be screwed up as well (my potato doesn't have them so IDK how it looks).
Applied to `tankheavyraider` Blitz (ZK texture + BAR wiggly beam) and `cloakassault` (vanilla engine texture + BAR wiggly beam).


GoogleFrog:
Just looking at the numbers I'm expecting this to be a nerf vs Glaive.

Anarchid:
> Just looking at the numbers I'm expecting this to be a nerf vs Glaive.

That sounds healthy for cloakassault.

sprunk:
The burst seems to track jinking targets (while maintaining spray angle) so jinking doesn't seem to do as much as it ought to, plus after 4 frames the Glaive gets stunned anyway. The burst also means that overkill can go through and affect multiple Glaives (like eg. Lance does), which does not help pure Knight (because it has perfect damage to kill a Glaive) but may help mixed armies.

Some other effects:
- doing worse at dealing EMP damage, because the earlier instances are dealt to a healthier unit. For example LLT no longer gets stunned on the first shot because `600/(785-230)` > 1 but `60/(785-1*23) + 60/(785-2*23) + ... + 60/(785-10*23)` < 1, though of course the beam could also be made shorter or the damage compensated.
- does worse at piercing shields
- does better at maintaining stun (if EMP ticks down during the burst, some of the later damage will make up for it)

And other thoughts:
 - longer burst means jinky units can just get grazed instead of taking a full hit, but in a large swarm somebody else is likely to take the rest of the hit. This is somewhat similar to inaccuracy. Maybe a KR balance sort of change would be to remove the inaccuracy in compensation.
 - if it's not worth making even minute changes to balance for a visual upgrade then the current behaviour could also be maintained (have the entire damage happen in the 1st frame, the others are just visual). This is doable with Lua, though ideally the engine would just get a weapondef tag that says "reshuffle the lightning wiggle every frame" because that is fundamentally what the burst does.

GoogleFrog:
So much is in the works already that I expect to wait for after the next stable.

sprunk:
Sure, it's just eye candy.

sprunk:
Linking #3900.

sprunk:
Included in #4975

PR #4469, opened 2021-08-11
Add resurrection protection

Arch-Shaman:
Adds a gadget to stop constructors from reclaiming ongoing resurrection projects.

### DIFF
--- a/LuaRules/Gadgets/game_resurrection_protection.lua
+++ b/LuaRules/Gadgets/game_resurrection_protection.lua
@@ -1,0 +1,53 @@
+if not (gadgetHandler:IsSyncedCode()) then
+	return
+end
+
+function gadget:GetInfo()
+	return {
+		name      = "Resurrection Protection",
+		desc      = "",
+		author    = "Shaman",
+		date      = "11 Elokuu 2021",
+		license   = "CC-0",
+		layer     = 5,
+		enabled   = true,
+	}
+end
+
+local IterableMap = VFS.Include("LuaRules/Gadgets/Include/IterableMap.lua")
+local protectedFeatures = {}
+local features = IterableMap.New()
+local newtable = {timer = 10, cancer = true} -- needed otherwise kaurapuurolainen will devour us.
+
+local spGetTeamInfo = Spring.GetTeamInfo
+
+function gadget:AllowFeatureBuildStep(builderID, builderTeam, featureID, featureDefID, part) -- part seems to be some sort of reclaim speed.
+	if part >= 0 then
+		if protectedFeatures[featureID] == nil then
+			IterableMap.Add(features, featureID, newtable)
+			protectedFeatures[featureID] = true
+		end
+		return true
+	else
+		return protectedFeatures[featureID] == nil
+	end
+end
+
+function gadget:FeatureDestroyed(featureID)
+	if protectedFeatures[featureID] then
+		protectedFeatures[featureID] = nil
+		IterableMap.Remove(features, featureID)
+	end
+end
+
+function gadget:GameFrame(f)
+	if f%5 == 2 then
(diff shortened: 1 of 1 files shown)


mojjj:
nice!

sprunk:
I don't think rez should block reclaim (or vice versa for that matter) via game mechanics.

sprunk:
Regardless of the above, this also blocks reclaimers from different allyteams.

Arch-Shaman:
Allyteam vs Allyteam will self resolve thus additional code was not included. There's almost never a situation where this would occur in a real game where athenas couldn't be killed/interrupted by either a flash LLT or by making fleas. The block lasts for 1/3rd second so you either finish the resurrect or it becomes free.

sprunk:
> There's almost never a situation where this would occur in a real game where athenas couldn't be killed/interrupted by either a flash LLT or by making fleas.

This misses the situation where you let the enemy Athena live and leech resources off it, which would no longer be possible.

> The block lasts for 1/3rd second so you either finish the resurrect or it becomes free.

It doesn't become free because the block gets reapplied immediately after it expires and a reclaimer (who may have been there first) has no way to prevent it (either from being first applied, or reapplied every 1/3 s).

GoogleFrog:
This gadget could still be the basis for UI. For example, reclaim on something being resurrected by an ally could only be possible with a non-right click single target reclaim command.

sprunk:
That already happens engine-wise to an extent. Area-reclaim is not supposed to touch wrecks currently being rezzed and vice versa so it'd mostly be a matter of blocking right-click.

GoogleFrog:
Then we agree. I don't see the difference between "this happens to an extent" - what you're saying - and "this does not happen exactly this way" - what I am saying.

How about this:
 - Loop over every constructor with a period of 20 seconds.
 - Exit if the next command is not a reclaim or resurrect command
 - Exit if the target of the command has no resurrect progress.
 - Resurrect - Put a per allyTeam block on the wreckage via table[allyTeamID][featureID] = thisGameFrame + 20*30.
 - Reclaim - Remove the command if conditions are met, such as the command having the right flag set.

sprunk:
Sounds okay. Here's an alternative one to just block right-click a priori. This one has the advantage that the cursor says what is going to happen and doesn't loop over all cons, but the disadvantage that a rezzer can't interrupt a reclaimer (which people seem to want).

 * `widget:DefaultCommand`.
 * Exit if the default command is not already reclaim.
 * Exit if there is somehow no feature under the cursor despite the default being reclaim.
 * Exit if the feature under the cursor has zero resurrect progress.
 * Do a cylinder check around the feature in <Athena build radius> elmos and check if anybody is resurrecting the wreck. If so, return CMD_RAW_MOVE, or maybe return CMD_RECLAIM_BLOCKED which would be a new fake command that does nothing and which has a cursor like this image

GoogleFrog:
The problem I am solving is fights in teamgames where player A is playing Athena-simulator and player B doesn't have the spare attention to even know why player A is yelling at them. To this end:
 - It doesn't matter if there are large delays involved. People playing Athena-simulator will probably accept a reclaim-resurrect fight that only lasts 10 seconds (if they notice).
  - Preexisting reclaim should be cancelled so player B does not have to pay attention.
 
Perhaps to rephrase the goal, if player A and B are fighting over a wreck, then I want to be sure that player B is consciously aware that they are fighting. Issuing a target-reclaim order with the left mouse button feels like something that is closer to ensuring this awareness.

I guess the Athena-simulator meta could move on to people putting a small amount of reclaim on a wreck to set it aside for reclaim, but I guess we will see that when we get to it. Perhaps area reclaim would latch back onto the wreck once nothing is resetting its resurrect counter.

PR #4480, opened 2021-08-28
Making Reaver UnitAI jink like it's supposed to

danfireman:


LuaRules/Configs/tactical_ai_defs.lua
```diff
@@ -980,13 +980,15 @@ local behaviourConfig = {
 	{
 		name = "cloakriot",
 		skirms = riotRangeSkirmieeArray,
-		--swarms = {},
+		swarms = {lowRangeSwarmieeArray},
```
GoogleFrog, line 983: I very much doubt this change was tested.

danfireman: It was, that's how I figured out the numbers to use (chosen to dodge almost everything while minimally harming speed).
Reavers now effective at dodging Rogue and Ronin, can dodge Bouy and Recluse to a great extent and can give Minotaur and Hermit a hard time.

It's been tested against pretty much everything that can miss.
Thanks to Reaver's very fast turn and acceleration speed it can actually dodge almost everything a Glaive can.

(Saw an attack-moving Reaver can take off half a Crab's hp before dying.)

PR #4482, opened 2021-08-30
Impaler deploy and visual fixes.

TechAUmNu:
Makes Impaler have to deploy before firing, based on Tremor script.
Fix up firing animation.
Fix wheels acceleration looking wrong.

Comparison of changes: 
https://youtu.be/ZF2gMYNKFWE

Thread:
http://zero-k.info/Forum/Thread/34102?Search=&User=&grorder=&grdesc=False&grpage=2

### DIFF
--- a/scripts/vehheavyarty.lua
+++ b/scripts/vehheavyarty.lua
@@ -14,111 +14,154 @@ local SIG_AIM = 2
 local SIG_MOVE = 1
 
 local RESTORE_DELAY = 5000
-local LOAD_DELAY = 4800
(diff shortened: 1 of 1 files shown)


sprunk:
True fixes that don't affect the unit's design/behaviour would ideally be a separate commit.

GoogleFrog:
As mentioned on the thread, I think 50% speed while the rocket area is open could be a better compromise between making it easier to catch and making it tedious to use.

GoogleFrog:
I guess this is getting pulled when I have the energy to make the further changes.

PR #4486, opened 2021-08-30
Projectile Dodge AI

PetTurtle:
unit tactical ai dodges projectiles in proj_targets_config.lua when on attack move, idle, or on idle return


LuaRules/Configs/proj_targets_config.lua
```diff
@@ -0,0 +1,52 @@
...
+  dynamic = data.wobble ~= 0 or data.dance ~= 0 or data.tracks
+  config[data.id] = {
+    aoe = data.damageAreaOfEffect,
+    wType = data.type,
+    trajectoryHeight = data.trajectoryHeight,
+    dynamic = dynamic,
+    range = data.range,
+    selfExplode = data.selfExplode,
+  }
+  spSetWatchWeapon(data.id, true)
```
sprunk, line 49: I slightly dislike doing this in a config file that otherwise looks like just a listing.

LuaRules/Gadgets/proj_dodge.lua
```diff
@@ -0,0 +1,285 @@
+function gadget:GetInfo()
+  return {
+    name = "Projectile Dodge",
+    desc = "calculates projectile dodge vector for units",
+    author = "petturtle",
+    date = "2021",
+    layer = -4,
```
sprunk, line 7: Why -4?

LuaRules/Gadgets/proj_targets.lua
```diff
@@ -0,0 +1,206 @@
+function gadget:GetInfo()
+  return {
+    name = "Projectile Target Map",
+    desc = "tracks projectiles targets",
+    author = "petturtle",
+    date = "2021",
+    layer = -5,
```
sprunk, line 7: Why -5 (presumably to be earlier than the -4 above)?

LuaRules/Utilities/vector2.lua
```diff
@@ -0,0 +1,136 @@
+local sqrt = math.sqrt
+local cos = math.cos
+local sin = math.sin
+local atan2 = math.atan2
+
+local Vector2 = {}
```
sprunk, line 6: There's already ` LuaRules/Utilities/vector.lua`, would be good not to duplicate them

LuaRules/Utilities/quad_tree.lua
```diff
@@ -0,0 +1,154 @@
+local Rect = VFS.Include("LuaRules/Utilities/rect.lua")
+
```
sprunk, line 2: There's `LuaRules/Utilities/quadField.lua` already, sounds like there's some overlap?

LuaRules/Configs/proj_targets_config.lua
```diff
@@ -0,0 +1,52 @@
...
+    jumparty_napalm_sprayer = true,
+    shipskirm_rocket = true,
+    shiparty_plasma = true,
+    turretgauss_gauss = true,
+    turretheavy_plasma = true,
+    turretantiheavy_ata = true,
+    turretheavylaser_laser = true,
+    staticarty_plasma = true,
+    staticheavyarty_plasma = true,
+}
```
sprunk, line 33: Many weapons are missing. What about, say, tacnuke?

LuaRules/Utilities/quad_tree.lua
PetTurtle: I am not familiar with quad fields, looking at the code they don't seem to be related.

LuaRules/Gadgets/proj_targets.lua
PetTurtle: proj_dodge.lua is called before proj_targets.lua to localize [GG.ProjTargets.Query, GG.ProjTargets.GetData](https://github.com/ZeroK-RTS/Zero-K/pull/4486/files#diff-cd045d3ce90f17ef5b5653fa1ddd1c95b0dd3cfabb8488d0e093235dd33873c0R231).
The localization is probably unnecessary.

LuaRules/Utilities/vector2.lua
PetTurtle: I like the read ability of my vector class, but switching can be done

LuaRules/Configs/proj_targets_config.lua
PetTurtle: I forgot about the silo missiles, but missiles that track units like scalpel or duck make it hard to predict where the projectile will hit and the unit that is targeted will need custom behavior like, do I retreat or take the hit that is not in the scope of these gadgets.

LuaRules/Gadgets/event_delay.lua
```diff
@@ -0,0 +1,48 @@
+function gadget:GetInfo()
```
GoogleFrog, line 1: Whitespace for this whole file.

LuaRules/Gadgets/proj_dodge.lua
```diff
@@ -0,0 +1,285 @@
+function gadget:GetInfo()
```
GoogleFrog, line 1: Whitespace for this file too.

LuaRules/Gadgets/unit_tactical_ai.lua
```diff
@@ -1327,4 +1358,4 @@ function gadget:UnitDestroyed(unitID, unitDefID, unitTeam)
 	if externallyHandledUnit[unitID] then
 		externallyHandledUnit[unitID] = nil
 	end
-end
+end
```
GoogleFrog: newline

LuaRules/Utilities/vector2.lua
GoogleFrog: I fear metatables so idk at this point.

PetTurtle:
The [mod page](https://zero-k.info/Mods/Detail/14) is up to date

GoogleFrog:
My apologies, the comments I left were pretty insubstantial. I'll try to find time on the weekend.

GoogleFrog:
image
I observed an issue with units being pushed out of position. Here is a Reaver with no LOS on any enemy units, so no other tactial AI behaviour is confounding the result. I moved it to position A and then fired at it with two Slings. It moved around dodging the projectiles and tended south. Then I removed the Slings and waited. The Reaver remained at position B.

Idle units that do any sort of movement AI are supposed to return to the location the player originally told them to stand when they are finished. Ideally in this case the Reaver would have a bias for dodging towards A to reduce the amount it can be pushed away.

PetTurtle:
This is intender due to units like firewalker, without it the units would dodge the projectile then just walk into the fire.

sprunk:
Possibly ground fires (and Placeholder) should just make everything behave as if there was a projectile coming that way and avoid that spot for a longer time.

PetTurtle:
I am kind of worried the dodge cmd broke something or is not implemented well / correctly, but in a game vs ai no errors / weird behavior that I noticed occurred.

GoogleFrog:
Inserting the dodge move command as soon as the unit moves out of the way makes it behave too frenetically. Fire a Sling at a Glaive, the Glaive will move out of the way, circle around and return to where it was standing, then move out of the way again before the Sling projectile arrives. This is too much running around crazily, it makes the game and behaviour noisy.

Look at what Glaive does when a Reaver approaches it then walks away:
image
The Glaive receives move orders to push it to the right distance and then, provided the Reaver doesn't move away, the Glaive stands at an appropriate range with no orders. If the Reaver moves too far the Glaive decides enough is enough and dives in. if the Reaver moves away then the Glaive sits at maximum range for a moment, then returns to its idle location.

Dodge should do something similar, and probably dodge slightly later against long flight time projectiles such as Sling. Perhaps the return delay could be modified for DoT weapons such as Firewalker so that the unit stands outside the area for the duration, then returns.

GoogleFrog:
Lots of wiggling then eight Glaives die.
image
image

GoogleFrog:
Since you've added the dodge move command you could even add a parameter that has it removed on a certain frame rather than at a certain distance. Then the gadget could give dodge move commands to units as a way of having them dodge, and just set the timeout. This has a high chance of being very messy in terms of spamming the command queue though, so my guess is that it wouldn't work.

GoogleFrog:
How do these Glaives know where the tacnuke is aiming prior to it turning around? Each StarburstLauncher needs some grace period from prediction.

LuaRules/Gadgets/unit_tactical_ai.lua
```diff
@@ -979,6 +984,29 @@ local function DoUnitUpdate(unitID, frame, slowUpdate)
...
+					spGiveOrderToUnit(unitID, CMD_REMOVE, {cmdTag}, 0 )
+				else
+					spGiveOrderToUnit(unitID, CMD_INSERT, {0, CMD_RAW_MOVE, CMD_OPT_INTERNAL, cx, uy, cz }, CMD.OPT_ALT )
+				end
+				unitData.cx, unitData.cy, unitData.cz = cx, uy, cz
+				unitData.receivedOrder = true
+			end
+			return
+		end
+	end
```
GoogleFrog, line 1009: I would look at moving this code somewhere else. Tactical AI already has a concept of idle. It knows exactly when a unit is idle and where it should return to after it does whatever behaviour it was doing while idle (attacking nearby stuff, moving away from riots, etc...). I would expect projectile dodge to just be another behaviour that is handled by the idle system. Instead, this code appears to be a parallel implementation of idle detection and handling.

GoogleFrog:
I noticed something weird about this situation and can't find a reason for it in the code: https://user-images.githubusercontent.com/1318249/132115667-5347c62d-dd72-40f6-aa3c-f2a106d238bd.png

Sometimes the push-leash return command would be Fight, while other times it would be Dodge Move.

PetTurtle:
All things mentioned above fixed. I am unhappy with how vehicles dodge, this [paper](https://gamma.cs.unc.edu/ORCA/) provides a solution to account for unit direction, and optimizes to get to target position as fast as possible. However, before I spend time implementing it I would like to benchmark this in real games to make sure I am not wasting my time.

sprunk:
I looked at this and I think I haven't found any (new and major) issues.

GoogleFrog:
All the testing, design and rebalancing this requires looks like it will be too much for me to do as as one chunk. For one thing, I suspect that units shouldn't have perfect knowledge of projectile speeds and velocities, but I won't know until I have time to dive into it.

To edge forwards into something deployed and tested, how about a full implementation for Impaler? This seems like the least controversial application and it barely affects 1v1, but it would still generate testing data and feedback.

sprunk:
Isn't that just a matter of commenting out most entries in the newly created ` LuaRules/Configs/proj_targets_config.lua`? It looks like weapons have to be tagged as dodgeable manually.

GoogleFrog:
Idk, I wasn't able to do much more than think up an approach and write the comment.

PetTurtle:
Commenting out everything but impaler from config will work. I have been slowly rewriting this pr to be more general adding the ability to easily add other things to dodge such as static fire, turrets a long with smarter dodging.

GoogleFrog:
Idle return behaviour is broken.
image
- Move a unit to location A.
- Tell the enemies to shoot at it.
- Move the unit to position B while it is still avoiding the projectiles.
- The unit returns to position A after reaching B.
This is not reliable. It depends a bit on distances and where the projectiles are. But it is fairly easy to reproduce with a bit of playing around.

LuaRules/Utilities/vector.lua
```diff
@@ -35,6 +40,14 @@ local function AbsVal(x, y, z)
 	end
 end
 
+local function Negative(v)
+	return {-v[1], -v[2]}
+end
+
+local function Mag(v)
```
GoogleFrog, line 47: This just reimplements AbsVal.

LuaRules/Utilities/vector.lua
```diff
@@ -97,6 +114,36 @@ local function Normal(v1, v2)
 	return Subtract(v1, projection), projection
 end
 
+function AngleTo(v1, v2)
+	return atan2(v1[1]*v2[2] - v1[2]*v2[1], v1[1]*v2[1] + v1[2]*v2[2]);
+end
+
+local function DirectionTo(v1, v2)
```
GoogleFrog, line 121: I would expect this to return a unit vector from its name. As is, this just reimplements Subtract.

LuaRules/Gadgets/unit_tactical_ai.lua
```diff
@@ -1312,6 +1365,9 @@ function gadget:UnitCreated(unitID, unitDefID, unitTeam, builderID)
 		active = false,
 		receivedOrder = false,
 		allyTeam = spGetUnitAllyTeam(unitID),
+		wasIdle = true,
+		idleX = ux,
+		idleZ = uz,
```
GoogleFrog, line 1370: This may be good for testing, but it doesn't make any sense for the game. Units are not idle in their factory. They should not return to inside the factory after chasing. Perhaps this works with a check for whether the unit is built upon creation (this would catch resurrect).

LuaRules/Gadgets/unit_tactical_ai.lua
```diff
@@ -1110,6 +1159,10 @@ local function AddIdleUnit(unitID, unitDefID)
 		Spring.Utilities.UnitEcho(unitID, "Idle " .. unitID)
 		Spring.Echo("=== Unit Idle", unitID, " ===")
 	end
+
+	if unitData.idleDodge then
```
GoogleFrog, line 1163: Returning early here breaks the fragile state structure of whether a unit is idle or not. If AddIdleUnit is not handled now then it will never be handled unless some explicit deferred handling exists, it is an event, not a state.

GoogleFrog:
image
An idle unit can be pushed a long way by constant fire. This is bad.

GoogleFrog:
I think the colour of the command stands out too much.

There should be some unit states to act as configuration. How about (default bold):
- Idle Dodge: Always, **Not on Hold Pos**, Never
- Move Order Dodge: Always, Not on Hold Pos, **Never**

LuaUI/Configs/zk_keys.lua
```diff
@@ -99,6 +100,7 @@ return {
 		{	"selectmissiles",	"Any+d"},
 		{	"reclaim",	"e",},
 		{	"fight",	"a",},
+		{	"rawdodgemove",	"a",},
```
GoogleFrog, line 103: This hotkey does not work. I don't recall this being the correct way to add multiple hotkeys. Regardless, I think it should be removed.

LuaUI/Configs/zk_keys.lua
```diff
@@ -85,6 +85,7 @@ return {
 		{	"clearmapmarks", "Alt+`"},
 		{	"buildfacing dec",	{"Any+delete", "]","Shift+]",}, },
 		{	"attack",	"f",},
+		{	"rawdodgemove",	"f",},
```
GoogleFrog, line 88: I think cycling commands are a bit too trap-like to use it here. Dodge probably doesn't need a default hotkey.

LuaUI/Configs/integral_menu_culling.lua
```diff
@@ -10,6 +10,7 @@ local configList = {
 	{cmdID = CMD_JUMP                  , default = true, name = "Jump"},
 	{cmdID = CMD.LOAD_UNITS            , default = true, name = "Load"},
 	{cmdID = CMD_RAW_MOVE              , default = true, name = "Move"},
+	{cmdID = CMD_RAW_DODGE_MOVE        , default = true, name = "Move"},
```
GoogleFrog, line 13: Do not call the command "Move". That will break stuff.

GoogleFrog:
Missing
image
image
image

LuaRules/Gadgets/unit_tactical_ai.lua
```diff
@@ -815,6 +821,123 @@ local function DoAiLessIdleCheck(unitID, behaviour, unitData, frame, enemy, enem
...
+	local nearbyEnemy = spGetUnitNearestEnemy(unitID, behaviour.leashAgressRange, true) or false
+	
+	if doDebug then
+		Spring.Echo("New Idle", unitData.idleX, unitData.idleZ, nearbyEnemy)
+		Spring.MarkerAddPoint(unitData.idleX, 0, unitData.idleZ, "IG")
+	end
+	-- Only update idle pos with commands not issued as part of returning to idle pos
+	if not unitData.returnCommandPos then
+		local x, _, z = Spring.GetUnitPosition(unitID)
+		unitData.idleX = x
```
GoogleFrog, line 862: Check whether (idleX, idleZ) is set, and within 32 or so elmos of (x, z). If they are close together then do not change (idleX, idleZ). This will prevent the remaining cases of waypoint drift.

This issue is in master, but is rare enough to not show up unless the unit is repeatedly sent away. Projectile dodge makes this much more likely than before.

GoogleFrog: I recall fixing this in master.

jklw:
> I think the colour of the command stands out too much.
> 
> There should be some unit states to act as configuration. How about (default bold):
> 
> * Idle Dodge: Always, **Not on Hold Pos**, Never
> * Move Order Dodge: Always, Not on Hold Pos, **Never**

Is the idea to have these two new states *instead* of the new "raw dodge move" command and just use normal move for dodging, or to have the second state, when enabled, cause raw dodge move commands to be issued instead of normal move commands on right click etc. (and the first state to cause raw dodge move commands to be auto-issued to idle units)?

GoogleFrog:
Neither.
- Raw dodge should exist. It is at least required internally.
- Which commands people issue should not be modified by these states.
- The move dodge state should change how the dodge gadget handles the unit when it has move orders.

PR #4516, opened 2021-09-15
Add empty base piece to the models of: Glaive, Iris, Conjurer, Gremlin, Scythe. Fixes #4515

### LINKED ISSUE #4515: Walking animation breaks for some Cloakbots when unit is resized
Walking animation scripts for some of the Cloakbots, particularly:
- Scythe
- Conjurer
- Gremlin
- Iris
- Glaive (sometimes)

break when the unit is enlarged, for example using dbg_size_changer.lua gadget.
When this happens these units don't move legs when walking, instead they "float" with their legs spread.
Enable dbg_size_changer.lua gadget to see this.

Although it is an issue mostly for mods that change unit size, it signals some problem with these unitscripts.
And I think I have seen Scythes not moving legs occasionally in teamgames, so it seems the issue can appear even in base ZK with no resized units, just rarely.

CarabusX:
Fixes #4515 
Tested and works.


GoogleFrog:
@CarabusX did you test whether harder to notice parameters were changed? I'm thinking of model radius but there may be others.

CarabusX:
@GoogleFrog I was careful to not change or recalculate modelRadius and modelHeight while editing .s3o. Basically I edited the s3o's with minimal clicks of Load Model->Click only what is necessary->Save Model.

As for modelRadius and modelHeight possibly automatically changing because new piece was added, it should happen as I put the base piece at the same height as where "hips" piece is. (basically moving y-offset from hips to base, because Zeus model which I used as example of empty base piece, had it done this way)

PR #4520, opened 2021-09-16
Potentially fix subtacmissile's weapon spawning into shore

Arch-Shaman:
The turn occurred too soon and the missile would spawn at its zero position instead of the upward position. This caused it to spawn inside terrain potentially when the sub was clipping into the surrounding shoreline.

### DIFF
--- a/scripts/subtacmissile.lua
+++ b/scripts/subtacmissile.lua
@@ -13,7 +13,7 @@ function script.QueryWeapon()
 end
 
 function script.AimFromWeapon()
-	return aimpoint
+	return missile
 end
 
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
Before:
image

image
image


After:
image
image

CarabusX:
Wouldn't simply:
Turn (missile, x_axis, 0, math.rad(30))
work?
(the same as in RestoreAfterDelay)

Arch-Shaman:
It could work out. I'm on a bit of a break right now though for the next few days (feeling a bit burned out). I'll check it out then.

PR #4524, opened 2021-09-23
Improve logic for setting default myGravity on ballistic weapons.

CarabusX:
The new logic handles all cases where (unwanted) default map gravity would be used.

### DIFF
--- a/gamedata/weapondefs_post.lua
+++ b/gamedata/weapondefs_post.lua
@@ -219,12 +219,18 @@ end
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 --
--- Set myGravity for Cannons because maps cannot be trusted. Standard is 120,
+-- Set myGravity for ballistic weapons because maps cannot be trusted. Standard is 120,
 -- gravity of 150 can cause high things (such as HLT) to be unhittable.
 
- for _, weaponDef in pairs(WeaponDefs) do
-	if weaponDef.weapontype == "Cannon" and not weaponDef.mygravity then
-		weaponDef.mygravity = 2/15 -- 120/(GAME_SPEED^2)
+local defaultMyGravity = 120 / (Game.gameSpeed^2)
+
+for _, weaponDef in pairs(WeaponDefs) do
+	local supportsMyGravity =
+		(weaponDef.weapontype == "Cannon") or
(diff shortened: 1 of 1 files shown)


gamedata/weapondefs_post.lua
```diff
@@ -219,12 +219,20 @@ end
...
+-- Set myGravity for ballistic weapons because maps cannot be trusted. Standard is 120,
 -- gravity of 150 can cause high things (such as HLT) to be unhittable.
 
- for _, weaponDef in pairs(WeaponDefs) do
-	if weaponDef.weapontype == "Cannon" and not weaponDef.mygravity then
-		weaponDef.mygravity = 2/15 -- 120/(GAME_SPEED^2)
+local defaultMyGravity = 120 / (Game.gameSpeed^2)
+
+for _, weaponDef in pairs(WeaponDefs) do
+	local isBallistic =
```
sprunk, line 228: `MissileLauncher` and `StarburstLauncher` are also ballistic after their "fuel" (flighttime) runs out.

CarabusX: > glancing at the engine code it looks like DGuns and expired torpedoes disregard this value and just use map gravity regardless?

The wiki said: Overrides the map gravity for **ballistic weapons**.

I guess I will have to dig into Spring code more to see where it is actually used.

CarabusX: Ok, looking at the Spring source only Cannon and BombDropper use myGravity. I will update the PR.

sprunk: BombDropper is actually Cannon too IIRC.

CarabusX: > BombDropper is actually Cannon too IIRC. 

It isn't. Spring source says:

> case hashString("AircraftBomb"): {

>  possibly adding a 0.0 comparison to the brutal checks instead (so that ZK base units never use mapside, but modders can).

Where to add that check then?

sprunk: Yeah disregard that part, AircraftBomb does spawn Cannon projectiles but is still a separate weapon type. I'll handle the brutal check part.

CarabusX: Updated PR with 0.0 check removed.

Issue #4454, opened 2021-07-14
Add a command that dynamically turns into the plate of the closest factory


GoogleFrog:
Rather than matching factory to plate, I would like an action that just turns my cursor into whatever plate is appropriate for the factory that is currently nearby. It should work like the current factory commands, switching based on context. The unplated version could have a little factory plop icon with a tiny arrow, like the terraform commands.

sprunk:
Maybe this could be tied to Caretaker since they are both "improve this fac's output" to an extent? Hold CTRL to make a plate instead of a Caretaker.

GoogleFrog:
I think a distinct command is best.

PR #4525, opened 2021-09-24
Added dynamic plate command to factory tab

Dave-tB:
Added a command to build a plate to the factory tab in spot BZ aka row, column 1. When command is issued, cursor shows a plate icon, and if cursor moves in range of a factory, the active command gets switched to building a plate of that factory.

It's implemented as a generic do-nothing command in LuaRules, and I edited cmd_factory_plate_placer.lua in LuaUI to add the behavior there in a way that does not disrupt the current plating behavior.  This allows it to also use the LuaUI glsl stuff in the current widget, which is familiar and looks nicer.

https://github.com/ZeroK-RTS/Zero-K/issues/4454


LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+
+function gadget:AllowCommand_GetWantedCommand()
+	return wantedCommands
+end
+
+function gadget:AllowCommand_GetWantedUnitDefID()
+	return true
+end
+
+-- local function dump(o)
```
GoogleFrog, line 57: This just looks like TableEcho. Should be removed.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+
+function gadget:Initialize()
+	gadgetHandler:RegisterCMDID(CMD_PLATE)
+end
+
+else
+function gadget:Initialize()
+	Spring.AssignMouseCursor("Plate", "cursorplate", true, true)
+end
+end
```
GoogleFrog, line 124: Newline.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+-- Speedup
+include("LuaRules/Configs/customcmds.h.lua")
+
+if gadgetHandler:IsSyncedCode() then
+
+local spInsertUnitCmdDesc   = Spring.InsertUnitCmdDesc
+local spGetUnitAllyTeam     = Spring.GetUnitAllyTeam
+local spGetUnitIsDead       = Spring.GetUnitIsDead
+
+local constructors          = 0
```
GoogleFrog, line 27: Why is the gadget keeping a list of units? It doesn't seem to need to do this.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+local constructor           = {}
+local constructorTable      = {}
+local constructors          = 0
+
+
+local exceptionArray = {
+	[UnitDefNames["athena"].id] = true,
```
GoogleFrog, line 34: Perhaps this could be generated by checking whether the unit has any factory construction options? (for mods and stuff)

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+local plateUnitDefIDs = {}
+for i = 1, #UnitDefs do
+	local ud = UnitDefs[i]
+	if ud and ud.isMobileBuilder and not ud.isFactory and not exceptionArray[i] then
+		plateUnitDefIDs[i] = true
+	end
+end
+
+function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOptions)
+	-- Don't allow non-constructors to queue terraform fallback.
```
GoogleFrog, line 79: Copypasta

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+
+local plateUnitDefIDs = {}
+for i = 1, #UnitDefs do
+	local ud = UnitDefs[i]
+	if ud and ud.isMobileBuilder and not ud.isFactory and not exceptionArray[i] then
+		plateUnitDefIDs[i] = true
+	end
+end
+
+function gadget:AllowCommand(unitID, unitDefID, teamID, cmdID, cmdParams, cmdOptions)
```
GoogleFrog, line 78: I don't see why any AllowCommand logic exists in this gadget beyond that required to block the command from ever being issued.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+	if plateUnitDefIDs[unitDefID] then
+		spInsertUnitCmdDesc(unitID, plateCmdDesc)
+		local aTeam = spGetUnitAllyTeam(unitID)
+		constructors = constructors + 1
+		constructorTable[constructors] = unitID
+		constructor[unitID]    = {allyTeam = aTeam, index = constructors}
+	end
+end
+
+function gadget:UnitDestroyed(unitID, unitDefID, teamID)
```
GoogleFrog, line 104: UnitDestroyed does not seem required.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+		constructors = constructors - 1
+		constructor[unitID]    = nil
+	end
+end
+
+function gadget:Initialize()
+	gadgetHandler:RegisterCMDID(CMD_PLATE)
+end
+
+else
```
GoogleFrog, line 120: Unsynced.

LuaUI/Widgets/cmd_factory_plate_placer.lua
```diff
@@ -112,22 +114,49 @@ local myAllyTeamID = Spring.GetMyAllyTeamID()
...
+local buildFactoryDefID
+local buildPlateDefID
 local closestFactoryData
 local activeCmdOverride
+local cmdFactoryDefID
+local cmdPlateDefID
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
+-- local function dump(o)
```
GoogleFrog, line 126: Remove.

LuaUI/Widgets/cmd_factory_plate_placer.lua
```diff
@@ -203,45 +232,84 @@ local function CheckTransformFactoryIntoPlate(factoryDefID)
...
-	if currentFactoryDefID then
-		currentFactoryDefID = nil
-		currentPlateDefID = nil
-		closestFactoryData = nil
+	if buildFactoryDefID or cmdFactoryDefID then
+		cmdFactoryDefID = nil
+		cmdPlateDefID = nil
+		buildFactoryDefID = nil
+		buildPlateDefID = nil
+		closestFactoryData = nil
```
GoogleFrog, line 312: ws

LuaUI/Widgets/cmd_factory_plate_placer.lua
```diff
@@ -203,45 +232,84 @@ local function CheckTransformFactoryIntoPlate(factoryDefID)
...
 		Spring.SetActiveCommand(buildAction[plateDefID])
+		return factoryDefID, plateDefID
+	else
+		Spring.SetActiveCommand("plate")
+		return
 	end
-	return true
 end
 
 function widget:Update()
```
GoogleFrog, line 267: This function has become a mess but I don't like this type of widget at the best of times (stateful UI), so idk.

GoogleFrog:
Summary:
 - The gadget should be pared down as it does a lot of useless stuff at the moment (from what I can tell).
 - I really don't like the widget, but I didn't like the widget before, and it is much much less worse than the lasso terraform UI widget. I just don't like stateful UI widgets at the best of times, so it's probably fine.
 - The BZ command icon needs to look better - more like a command. Commands have a particular border thickness and monocolour gradient. I don't recall anyone adding a command in the appropriate style since they were added, except when Aquanim added recall drones (which mostly matches), so perhaps he knows how to make a more fitting factory command.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+function gadget:GetInfo()
+	return {
+		name      = "Generic Plate Command",
+		desc      = "Implements CMD_PLATE as a do-nothing command, behaviour is defined in LuaUI",
```
sprunk, line 7: Is a gadget really needed at all? The command should always be converted to a factory build order and thus never reach gadgetside.

LuaUI/Configs/integral_menu_config.lua
```diff
@@ -421,12 +422,20 @@ for i = 1, 5 do
 	}
 end
 
+local factoryButtonLayoutOverride = {}
+	factoryButtonLayoutOverride[1] = {
+		[3] = {
+			buttonLayoutConfig = buttonLayoutConfig.command,
+			isStructure = false,
+		}
+	}
```
sprunk, line 431: Could initialize the [1] right away instead of a separate assignment:
```suggestion
local factoryButtonLayoutOverride = {
	{
		[3] = {
			buttonLayoutConfig = buttonLayoutConfig.command,
			isStructure = false,
		}
	}
}
```

LuaUI/Widgets/cmd_factory_plate_placer.lua
```diff
@@ -337,13 +405,13 @@ local function GetDrawDef(mx, mz, data)
 end
 
 local function DrawFactoryLine(x, y, z, drawDef)
-	local mx, mz = GetMousePos(not floatOnWater[currentFactoryDefID])
```
sprunk, line 340: Why remove the float check?

LuaUI/Widgets/cmd_factory_plate_placer.lua
```diff
@@ -318,7 +386,7 @@ function widget:PlayerChanged(playerID)
 	if myAllyTeamID == Spring.GetMyAllyTeamID() then
 		return
 	end
-	myAllyTeamID = Spring.GetMyAllyTeamID()
+	myAllyTeamID = Spring.GetMyAllyTeamIDs()
```
sprunk, line 389: Suspicious

LuaRules/Gadgets/cmd_generic_plate.lua
GoogleFrog: I wouldn't trust it to be ignored by everything.

sprunk: I would, because otherwise arbitrary nonexistent commands (ID 12345 or something) would also do this in which case we would have to make a general sweeping change that would cover CMD_PLATE anyway.

Dave-tB:
Re: Icon

How does this look?

image

GoogleFrog:
That looks decent as a shape. In terms of colour maybe do area mex grey, as both commands create structures. Perhaps the idle worker button would work too. Shape and colour are the only free variables here, as it should otherwise have a gradient and border style that looks like it fits with the other commands. I do not know how to create such a thing.

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,124 @@
...
+if gadgetHandler:IsSyncedCode() then
+
+local spInsertUnitCmdDesc   = Spring.InsertUnitCmdDesc
+local spGetUnitAllyTeam     = Spring.GetUnitAllyTeam
+local spGetUnitIsDead       = Spring.GetUnitIsDead
+
+local constructors          = 0
+local constructor           = {}
+local constructorTable      = {}
+local constructors          = 0
```
CarabusX, line 30: You have 2 same lines (L27 and L30):
`local constructors          = 0`

LuaRules/Gadgets/cmd_generic_plate.lua
CarabusX: Or should be moved to config file.

Dave-tB: I'm going to replace this functionality with a local CanBuildFactory function, although is striderhub considered a factory under unitdefs? or if someone mods in a new factory but not a plate that goes with it then you might still be left with a unit with a command that does nothing.

I think I need to add in parentOfPlate logic here as well to cover all edge cases.

CarabusX: You can most easily distinguish factories by having customParams.parentOfPlate tag. Striderhub doesn't have it.

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: done

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: done

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: removed

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: done

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: removed

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: Doesn't seem to be an issue when I remove them, so removed.

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: All of this was useless anyway, removed.

LuaUI/Widgets/cmd_factory_plate_placer.lua
Dave-tB: done

LuaUI/Widgets/cmd_factory_plate_placer.lua
Dave-tB: removed

LuaUI/Widgets/cmd_factory_plate_placer.lua
Dave-tB: removed.......

LuaUI/Widgets/cmd_factory_plate_placer.lua
Dave-tB: no idea, put it back in

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,72 @@
...
+local plateCmdDesc = {
+	id      = CMD_PLATE,
+	type    = CMDTYPE.ICON_MAP,
+	name    = 'plate',
+	cursor  = 'Plate',
+	action  = 'plate',
+	tooltip = 'Build a Plate of a nearby factory',
+}
+
+local function CanBuildFactoryPlate(unitDefID)
```
GoogleFrog, line 37: Cache this so it doesn't look up UnitDefs more than once per unitDefID. Invoking UnitDefs is slow.

LuaRules/Gadgets/cmd_generic_plate.lua
Dave-tB: removed.

LuaRules/Gadgets/cmd_generic_plate.lua
GoogleFrog: Ideally unitDefIDs would not be able to come into the function negative.

Dave-tB: Ok, tried implementing the cache as tables instead and it seems to work so:

setContains(defIDCache, unitDefID) -------> setContains(defIDCache, {unitDefID, true})
setContains(defIDCache, -unitDefID) -------> setContains(defIDCache, {unitDefID, false})

LuaRules/Gadgets/cmd_generic_plate.lua
```diff
@@ -0,0 +1,89 @@
...
+function addToSet(set, key)
+    set[key] = true
+end
+
+function setContains(set, key)
+    return set[key] ~= nil
+end
+
+local function CanBuildFactoryPlate(unitDefID)
+	if not (setContains(defIDCache, {unitDefID, true}) or setContains(defIDCache, {unitDefID, false})) then
```
GoogleFrog, line 46: This line spams tables. The setContains function seems redundant to me.

As a rule of thumb, if you have to check "if bla ~= nil then" rather than "if bla then", then the code is going to confuse the next person to look at it. I'd try to avoid it.

LuaRules/Gadgets/cmd_generic_plate.lua
sprunk: In that case I think the gadget would ideally still be a 1-liner that just does `if cmd == CMD_PLATE then return false end` with adding it to units done in a widget. But it's okay for now.

GoogleFrog: IF A COMMAND HITS THE ENGINE IT CANCELS THE COMMAND QUEUE.

LuaRules/Gadgets/cmd_generic_plate.lua
sprunk: Why not cache it once at init? Then you still look up UnitDefs once per unitDefID but also don't need to check for the existence of the table every time and don't need to convert from 0/1 numbers to bools (can just keep bools directly because there are no nils).

PR #4526, opened 2021-09-25
Add support for mod-side weapondefs_mod and explosions_mod files.

CarabusX:
Additionally standardize ExplosionDefs names and tags to lowercase like it is done with UnitDefs, because some ExplosionDefs (exactly 4 of them) have random case.

### DIFF
--- a/gamedata/explosions.lua
+++ b/gamedata/explosions.lua
@@ -5,13 +5,28 @@ local vfsDirList = VFS.DirList
 vfsInclude("LuaRules/Utilities/tablefunctions.lua")
 local suCopyTable = Spring.Utilities.CopyTable
 
+local explosionDefs = {}
+ExplosionDefs = explosionDefs
+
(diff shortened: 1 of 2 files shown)


gamedata/explosions.lua
```diff
@@ -2,17 +2,35 @@ local VFS = VFS
...
+local lowerkeys = system.lowerkeys
+
 vfsInclude("LuaRules/Utilities/tablefunctions.lua")
 local suCopyTable = Spring.Utilities.CopyTable
 
+local explosionDefs = {}
+ExplosionDefs = explosionDefs
+
+local shared = {} -- shared amongst the lua explosiondef enviroments
+Shared = shared
```
GoogleFrog, line 15: What is this for? These two lines appear to do nothing.

CarabusX: This is the same as done in UnitDefs/WeaponDefs, in case .lua files wanted to share data between each other.

PR #4530, opened 2021-10-02
Unhardcode Dirtbag mound and move to config file.

CarabusX:
See commits.


LuaRules/Gadgets/unit_terraform.lua
```diff
@@ -3619,64 +3620,34 @@ local function deregisterStructure(unitID)
...
-			local ux, uy, uz  = spGetUnitPosition(unitID)
+	if (terraformOnUnitDestroyed[unitDefID]) then
+		local  _,_,_,_,buildProgress = spGetUnitHealth(unitID)
+		if buildProgress == 1 then
+			local config = terraformOnUnitDestroyed[unitDefID]
+			local posX, posY, posZ = config.posX, config.posY, config.posZ
+
+			local ux, uy, uz = spGetUnitPosition(unitID)
 			ux = floor((ux+8)/16)*16
 			uz = floor((uz+8)/16)*16
```
CarabusX, line 3633: Btw @GoogleFrog why is center position rounded to nearest 16, while heightmap resolution is 8?
Perhaps this should be changed so the mound appears closer to exact Dirtbag position?
I can fix it if you agree.

sprunk: Aren't buildings snapped to 16?

LuaRules/Gadgets/unit_terraform.lua
```diff
@@ -3619,64 +3620,34 @@ local function deregisterStructure(unitID)
...
 function gadget:UnitDestroyed(unitID, unitDefID)
 
-	if (unitDefID == shieldscoutDefID) then
-		local  _,_,_,_,build = spGetUnitHealth(unitID)
-		if build == 1 then
-			local ux, uy, uz  = spGetUnitPosition(unitID)
+	if (terraformOnUnitDestroyed[unitDefID]) then
+		local  _,_,_,_,buildProgress = spGetUnitHealth(unitID)
+		if buildProgress == 1 then
+			local config = terraformOnUnitDestroyed[unitDefID]
```
sprunk, line 3628: Could avoid indexing the table twice:
```suggestion
	local config = terraformOnUnitDestroyed[unitDefID]
	if config then
		local  _,_,_,_,buildProgress = spGetUnitHealth(unitID)
		if buildProgress == 1 then
```

CarabusX: But then you would create an extra variable and perform assignment for all units, instead of extra indexing for Dirtbags only.
No idea what is better tbh.

LuaRules/Gadgets/unit_terraform.lua
CarabusX: > Isn't building placement snapped to 16? This means that for example blocking fac exits or building placement with Dirtbags is consistent.

It would still not be consistent between factories with even footprint/exit width (then you can block perfectly at center) vs odd footprint/exit width (then you can't block perfectly at center).

> Test 8 with some use cases: blocking fac exits, blocking buildings, bot pathing over mounds (incl. when stacked a bit), 

Ok, I will do some tests and come back with results.

> but also for example a line of Dirtbags - maybe if the spacing is 16-16-16-16 it doesn't leave a hole for vehicles but a 16-8-24-16 would.

Perhaps if you could get perfect 16-16-16 spacing between Dirtbags this would make sense, but IRL you will never have it especially that X and Z coordinates are rounded separately.
Lets assume close-to-ideal initial spacing of 17-15-17-15-17.
With rounding to closest 16 it can end with anything between 16-16-16-16-16 and 32-0-32-0-32 depending on how it aligned with the grid.
With rounding to closest 8 it can end with anything between 16-16-16-16-16 and 24-8-24-8-24, so it is less extreme as every point can only move at range [-4,-4] when rounded.

LuaRules/Gadgets/unit_terraform.lua
sprunk: I checked the "compiler" output (`$ luac -l -l -p indexing.lua`). This chunk:
```lua
local tab
local function single(index)
             local x = tab[index]
             if x then
             end
end

local function double(index)
             if tab[index] then
                          local x = tab[index]
             end
end
```
has this output:
```
function <indexing.lua:2,6> (5 instructions, 20 bytes at 0x19697c0)
1 param, 2 slots, 1 upvalue, 2 locals, 0 constants, 0 functions
	1	[3]	GETUPVAL 	1 0	; tab
	2	[3]	GETTABLE 	1 1 0
	3	[4]	TEST     	1 0 0
	4	[4]	JMP      	0	; to 5
	5	[6]	RETURN   	0 1
constants (0) for 0x19697c0:
locals (2) for 0x19697c0:
	0	index	1	5
	1	x	3	5
upvalues (1) for 0x19697c0:
	0	tab

function <indexing.lua:8,12> (7 instructions, 28 bytes at 0x1969710)
1 param, 2 slots, 1 upvalue, 2 locals, 0 constants, 0 functions
	1	[9]	GETUPVAL 	1 0	; tab
	2	[9]	GETTABLE 	1 1 0
	3	[9]	TEST     	1 0 0
	4	[9]	JMP      	2	; to 7
	5	[10]	GETUPVAL 	1 0	; tab
	6	[10]	GETTABLE 	1 1 0
	7	[12]	RETURN   	0 1
constants (0) for 0x1969710:
locals (2) for 0x1969710:
	0	index	1	7
	1	x	7	7
upvalues (1) for 0x1969710:
	0	tab
```
So my way is strictly superior. The temporary value from the first indexing has to be put on the stack anyway, so there is no extra assignment (it even gets put in the same stack slot, so no memory either). Indexing twice just results in an additional `GETTABLE` op (and `GETUPVAL` since the table is not local). It doesn't matter that the `if` is empty, putting things inside is symmetrical.

sprunk:
I think these are already a good backend improvement, the 8/16 thing is a gameplay change so could be a separate PR.

LuaRules/Gadgets/unit_terraform.lua
CarabusX: Thanks for doing the research!
Thats certainly some stuff worth to know. I will remember this for future code.
As for this case, I will fix it in next PM.

sprunk: Already taken care of 737a07c23bdcaa9282e17f8365b86fd124614d80

Issue #4535, opened 2021-10-06
Some gadgets contain hardcoded unitnames or weaponnames, which is an issue for moddability


CarabusX:
Here is the list of some gadgets that I found that contain hardcoded unitnames or weaponnames.
The list is not complete, as I mostly looked at unit_ and weapon_ gadgets, and could miss some.

1. **unit_air_plants.lua - L33-L45** - the list of air factories and plates, with their default Fly/Land setting. Could be turned to customParam.
2. **unit_bomber_command.lua - L55-L68** - list of units with airpads, with come config. Could be moved to config file.
3. **unit_factory_guard.lua - L50-L73** - list of factories and plates. Could probably be detected automatically.
4. **unit_refuel_pad_handler.lua - L60-L62** - list of mobile units with airpads. Could be detected automatically, or use config from 2.
5. **unit_show_shooter.lua - L35-L37** - list of units that don't decloak when firing. Could be turned to customParam.
6. **unit_teleporter.lua - L67** - teleporter beacon unit is hardcoded. Could be turned to customParam on the teleporter like the teleporting ability already is.
7. **unit_transport_load_unload.lua - L20-L23** - list of air transports.  Could be detected automatically.
8. **unit_transport_pickup_floating_amphib.lua - L71-L94** - list of units that can be dropped to sea. And some evil code that bases this ability on buildability by Amph Factory. Could be turned to customParam or even detected automatically if there are no exceptions.

CarabusX:
Point 6 fixed in PR #4542.

CarabusX:
Point 8 fixed in PR #4543.

GoogleFrog:
I think the ticks should be for pulled PRs rather than open ones.

CarabusX:
Points 10 and 12 fixed in PR #4546.

sprunk:
How did you perform the check? There are more, for example https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Gadgets/unit_starlight_handler.lua#L29-L30

CarabusX:
Yeah I didn't look for units with non-standard names.
I mostly searched for strings prefixed with amph..., hover... (for all mobile unit types including stuff like bomber..., drone..., sub...), and then for statics (static..., energy..., plate..., turret... etc.).
I skipped some I found are not worth fixing. (like the hardcoded windgen in some OD places)

GoogleFrog:
Or better, `UnitDefNames["`.

[...]

PR #4536, opened 2021-10-06
Unhardcode some unitnames hardcoded in gadgets

CarabusX:
See commits.

This fixes points 1, 2, 3, 4, 5, 7 from issue #4535.

Everything tested to check that resulting tables are exactly same as before.

### DIFF
--- a/LuaRules/Gadgets/unit_factory_guard.lua
+++ b/LuaRules/Gadgets/unit_factory_guard.lua
@@ -50,7 +50,7 @@ local GiveClampedOrderToUnit = Spring.Utilities.GiveClampedOrderToUnit
 local factoryDefs = {}
 
 for unitDefID, ud in pairs(UnitDefs) do
-	if (ud.isFactory and ud.buildOptions) then
+	if ud.isFactory and (not ud.customParams.notreallyafactory) and ud.buildOptions then
 		local buildOptions = ud.buildOptions
 
 		for i = 1, #buildOptions do


LuaRules/Configs/airpad_defs.lua
```diff
@@ -0,0 +1,20 @@
...
+	[UnitDefNames["shipcarrier"].id] = {
+		padPieceName = {"LandingFore","LandingAft"}
+	},
+}
+
+for unitDefID, config in pairs(airpadDefs) do
+    local ud = UnitDefs[unitDefID]
+
+    config.mobile = (not ud.isImmobile)
+    config.cap = config.cap or ud.customParams.pad_count
```
sprunk, line 17: I think I'd do just
```suggestion
    config.cap = ud.customParams.pad_count
```
because having two sources can be confusing.

gamedata/planetwars/pw_structuredefs.lua
```diff
@@ -123,6 +124,7 @@ local structureConfig = {
 		unitDef.buildoptions = { [[bomberprec]], [[bomberriot]], [[bomberdisarm]], [[bomberheavy]] }
 		unitDef.customparams.nongroundfac = [[1]]
 		unitDef.customparams.landflystate = [[0]]
+		unitDef.customparams.factory_land_state = 0
```
sprunk, line 127: Ditto
```suggestion
		unitDef.customparams.factory_land_state = [[0]]
```

LuaRules/Configs/airpad_defs.lua
CarabusX: I will make it default to #padPieceName instead (because they are closely related in the code)

LuaRules/Gadgets/unit_factory_guard.lua
```diff
@@ -47,30 +47,20 @@ VFS.Include("LuaRules/Utilities/ClampPosition.lua")
...
-	[UnitDefNames["platetank"].id] = 0,
-	[UnitDefNames["plateplane"].id] = 0,
-	[UnitDefNames["plategunship"].id] = 0,
-	[UnitDefNames["plateship"].id] = 0,
-}
+local factoryDefs = {}
+
+for unitDefID, ud in pairs(UnitDefs) do
+	if (ud.isFactory and ud.buildOptions) then
+		for _, buildOptionDefID in ipairs(ud.buildOptions) do
```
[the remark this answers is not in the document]
CarabusX: Ok

LuaRules/Configs/airpad_defs.lua
CarabusX: Done.

LuaRules/Gadgets/unit_transport_load_unload.lua
```diff
@@ -17,17 +17,19 @@ end
...
-		allowTransportCache[unitDefID] = {}
+
+for unitDefID, ud in pairs(UnitDefs) do
+	if (ud.isTransport and ud.canFly) then
+		airTransports[unitDefID] = true
+
+		if ud.customParams.islighttransport then
+			lightTransport[unitDefID] = true
+			allowTransportCache[unitDefID] = {}
+		end
```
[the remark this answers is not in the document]
GoogleFrog: Many assumptions will break in the face of non-air transports, but luckily I think the cost will mostly be on the non-air transport lacking ZK features, rather than breaking ZK.

LuaRules/Gadgets/unit_factory_guard.lua
```diff
@@ -47,30 +47,23 @@ VFS.Include("LuaRules/Utilities/ClampPosition.lua")
...
+for unitDefID, ud in pairs(UnitDefs) do
+	if (ud.isFactory and ud.buildOptions) then
+		local buildOptions = ud.buildOptions
+
+		for i = 1, #buildOptions do
+			local boDefID = buildOptions[i]
+			local bod = UnitDefs[boDefID]
+
+			if (bod and bod.isBuilder and bod.canAssist) then
+				factoryDefs[unitDefID] = true  -- only factories that can build builders are included
```
GoogleFrog, line 61: Maybe add the fake factory custom param?

CarabusX: Not necessary I think.
The 4 units that are isFactory=true but excluded here are:
- airpad (excluded by having no buildoptions)
- missilesilo, pw_dropfac, pw_bomberfac (excluded by having no builder)

The builder condition appears later in the gadget so it was quite natural to use it to filter the factories, and proved enough to produce the original table.

sprunk: Missile silo is probably a "true" factory but doesn't build builders. (inb4 construction missile)

CarabusX: Just tested missilesilo after giving it Crane buildoption and it can build these just fine :
screen00328
)
So in this case it is only dependent on buildoptions.
But Airpad can't build anything even when given buildoptions, so it could probably me marked by some customParam, even if not needed by this gadget.

CarabusX: Apparently there is "notreallyafactory" customParam already. I improved the check to use it.

PR #4537, opened 2021-10-07
Visual limit for rectangle terraform lasso

JennyDmnit:
Gives a visual indication if the rectangle select of the terraform lasso is too big and will be ignored.
It turns red when it's too big.

### DIFF
--- a/LuaUI/Widgets/gui_lasso_terraform.lua
+++ b/LuaUI/Widgets/gui_lasso_terraform.lua
@@ -1071,8 +1071,6 @@ function widget:MousePress(mx, my, button)
 		spSetActiveCommand(index)
 		currentlyActiveCommand = CMD_LEVEL
 		
-		local mx,my = spGetMouseState()
-		
 		setHeight = true
 		drawingRectangle = false
 		placingRectangle = false
--- a/LuaUI/Widgets/gui_lasso_terraform.lua
+++ b/LuaUI/Widgets/gui_lasso_terraform.lua
@@ -180,7 +180,9 @@ local posVolume   = {0, 1, 0, 0.1} -- posisive volume
 local groundGridColor  = {0.3, 0.2, 1, 0.8} -- grid representing new ground height
 
 -- colour of lasso during drawing
-local lassoColor = {0.2, 1.0, 0.2, 1.0}
+local lassoColorGood = {0.2, 1.0, 0.2, 1.0}
+local lassoColorBad  = {1.0, 0.2, 0.2, 1.0}
+local lassoColorCurrent = lassoColorGood
 
 -- colour of ramp
 local vehPathingColor = {0.2, 1.0, 0.2, 1.0}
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_lasso_terraform.lua
```diff
@@ -1069,7 +1077,7 @@ function widget:MousePress(mx, my, button)
 		spSetActiveCommand(index)
 		currentlyActiveCommand = CMD_LEVEL
 		
-		local mx,my = spGetMouseState()
+		-- local mx,my = spGetMouseState()
```
sprunk, line 1080: Don't leave unused lines lying around, delete.

LuaUI/Widgets/gui_lasso_terraform.lua
```diff
@@ -1053,6 +1058,9 @@ local function snapToHeight(heightArray, snapHeight, arrayCount)
 end
 
 function widget:MousePress(mx, my, button)
+	local mx,my = spGetMouseState()
+	local _, corner = spTraceScreenRay(mx, my, true, false, false, true)
+	startCorner = corner
```
sprunk, line 1063: The starting corner is already saved in some other variable, I believe it's somewhere in the first few elements of the `point` array (likely the first one but there is some math going on).

LuaUI/Widgets/gui_lasso_terraform.lua
```diff
@@ -1261,7 +1269,19 @@ function widget:MouseMove(mx, my, dx, dy, button)
 		return true
 		
 	elseif drawingRectangle then
-
+		local mx,my = spGetMouseState()
+		local _, corner = spTraceScreenRay(mx, my, true, false, false, true)
+		currentCorner = corner
```
sprunk, line 1274: `corner` and `pos` in line 1286 (or 1266 before this patch) are the same thing, reuse `pos`.

LuaUI/Widgets/gui_lasso_terraform.lua
```diff
@@ -1261,7 +1269,19 @@ function widget:MouseMove(mx, my, dx, dy, button)
...
+
+		if legalPos(startCorner) and legalPos(currentCorner) then
```
sprunk, line 1276: `startCorner` (or rather `point[x]` as mentioned above) shouldn't change so doesn't need repeated legality checks on mouse movement. `currentCorner` (or rather `pos`) legality is already checked below, so move the inside of this check there.

LuaUI/Widgets/gui_lasso_terraform.lua
```diff
@@ -2111,10 +2131,10 @@ function widget:DrawWorld()
 			
 			--glDepthTest(false)
 		elseif drawingLasso then
-			glColor(lassoColor)
+			glColor(currentColor)
```
sprunk, line 2134: Isn't `currentColor` only updated for rectangles (not lassos)? This looks as if interrupting a rectangle in its "bad" state will keep the bad colour for subsequent lassos. If that is the case, keep `lassoColor` here.

sprunk (changes requested):
Some implementation issues but a good idea!

PR #4540, opened 2021-10-13
make disruptor cannon trail a bit more sleek

Anarchid:
After:
- sleek shape
- round imposing head
- sparkles of ripping through enemy will 
image

Before:
- weird unicorn sparkles
- shape extends towards end 
- hard ribbing 
- no sparkles
image


Anarchid:
image

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
+-- vulcanfx
+
+return {
+	["disrupter_cannon_muzzle"] = {
```
sprunk, line 4: dominater, inquisiter, traiter, hoder

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+		alwaysvisible         = false,
```
sprunk, line 5: there are tons of spaces here but it's not even aligned with the line below

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+		usedefaultexplosions = false,
+		shaft = {
+			air                = true,
+			class              = [[CBitmapMuzzleFlame]],
+			count              = 1,
+			ground             = true,
+			underwater         = 1,
```
sprunk, line 12: should probably be `true`

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+				length             = 24,
+				sidetexture        = [[muzzleside]],
+				size               = 16,
+				sizegrowth         = -0.5,
+				ttl                = 16,
+			},
+		},
+	},
+	["disruptertrail"] = {
+	alwaysvisible         = false,
```
sprunk, line 28: the inside of the table could be indented relative to the outer brackets

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+	usedefaultexplosions = false,
+	head = {
+	  air                = true,
+	  class              = [[heatcloud]],
+	  count              = 2,
+	  ground             = true,
+	  water              = true,
+	  underwater         = 1,
+	  unit               = 1,
```
sprunk, line 37: `true` probably, see above

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+	  properties = {
+		  alwaysvisible      = false,
+		  heat               = [[100]],
```
sprunk, line 40: should probably be just 100 (i.e. number not a string)

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+		  heatfalloff        = 15,
+		  maxheat            = 100,
+		  pos                = 0,
```
sprunk, line 43: is `pos` not a vector but a number? sus scrofa

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+		  size               = [[10]],
```
sprunk, line 44: should prolly be a number, see above

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+		usedefaultexplosions = false,
+		shaft = {
+			air                = true,
+			class              = [[CBitmapMuzzleFlame]],
+			count              = 1,
+			ground             = true,
+			underwater         = 1,
+			water              = true,
+			properties = {
+				colormap           = [[0.9 0.7 0.9 0.01 0.5 0.1 0.8 0.01 0 0 0 0.01]],
```
sprunk, line 15: consider either some extra spaces between the phases or ideally something like this that separates them neatly:
```suggestion
				colormap           = [[0.9 0.7 0.9 0.01]]
				                 .. [[ 0.5 0.1 0.8 0.01]]
				                 .. [[ 0.0 0.0 0.0 0.01]],
```

effects/disrupter_cannon.lua
```diff
@@ -0,0 +1,100 @@
...
+	  water              = true,
+	  underwater         = true,
+	  properties = {
+		  airdrag            = 0.7,
+		  alwaysvisible      = false,
+		  colormap           = [[1 0.25 1 0.05   0.01 0.005 0.02 0.01]],
+		  directional        = true,
+		  emitrotspread      = 0.1,
+		  emitvector         = [[dir]],
+		  gravity			   = [[0,-0.01,0]];
```
sprunk, line 84: is this correct? all other numerical vectors seem to use spaces as delimiters instead of commas. also, `=` misaligned and `;` at the end

sprunk (changes requested):
Trail: looks good, but let's see what the local unexpectedly bearded person has to say about it.
Code: looks bad.

effects/disrupter_cannon.lua
Anarchid: it's a bool but using a bool instead of an int before has caused it to not work, so it's a cargo bool.

effects/disrupter_cannon.lua
Anarchid: consistent with `disrupter_beam`, `disrupter_missile`, etc

effects/disrupter_cannon.lua
Anarchid: Not a single example of the nice staggered thing exists in the codebase and i don't trust CEG to not explode with this

PR #4543, opened 2021-10-17
Unhardcode list of amphibious units in unit_transport_pickup_floating_amphib.lua gadget.

CarabusX:
Unhardcode list of amphibious units in unit_transport_pickup_floating_amphib.lua gadget.
(and remove evil dependency on factoryamph's buildOptions)

The resulting list includes all the same units as before and additionally - for completeness:
- Grebe
- all amphibious non-floating chickens (which is most of them)

This fixes point 8 from issue #4535.

### DIFF
--- a/LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
+++ b/LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
@@ -68,28 +68,25 @@ local sinkCommand = {
 	[CMD_WAIT_AT_BEACON] = true,
 }
 
-local dropableUnits = {
-	--all floatable unit will be dropped when regular unload fail (such as when unloading at sea), but some can't float but is amphibious,
-	--this list additional units that should be dropped.
-	[UnitDefNames["amphcon"].id] = true, --clam
-	[UnitDefNames["amphraid"].id] = true, --duck
-	[UnitDefNames["striderantiheavy"].id] = true, --ultimatum
-	[UnitDefNames["shieldshield"].id] = true, --aspis
-	[UnitDefNames["cloakjammer"].id] = true, --eraser
-	[UnitDefNames["striderdetriment"].id] = true, --striderdetriment
+local DROPPABLE_UNIT_MAXDEPTH = 100  -- Arbitrary value (there are no units with maxDepth between 22 and 5000, so any value from range 23-4999 will work here)
+
+local floatingMoveClasses = {
+	[Game.speedModClasses.Hover] = true,
+	[Game.speedModClasses.Ship] = true,
 }
 
-if UnitDefNames["factoryamph"] then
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
```diff
@@ -68,28 +68,25 @@ local sinkCommand = {
...
-	[UnitDefNames["striderantiheavy"].id] = true, --ultimatum
-	[UnitDefNames["shieldshield"].id] = true, --aspis
-	[UnitDefNames["cloakjammer"].id] = true, --eraser
-	[UnitDefNames["striderdetriment"].id] = true, --striderdetriment
+local DROPPABLE_UNIT_MAXDEPTH = 100  -- Arbitrary value (there are no units with maxDepth between 22 and 5000, so any value from range 23-4999 will work here)
+
+local floatingMoveClasses = {
+	[Game.speedModClasses.Hover] = true,
+	[Game.speedModClasses.Ship] = true,
 }
```
sprunk, line 76: The ultimate check is `not floatsOnWaterSurface` so I think it would be slightly better if this table listed things that sink in the water (i.e. Tank and KBot). But this is okay too.

LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
```diff
@@ -68,28 +68,25 @@ local sinkCommand = {
...
 
-for i=1, #UnitDefs do
-	if (UnitDefs[i].customParams.level) then
-		dropableUnits[i] = true
+--All floatable units will be dropped when regular unload fail (such as when unloading at sea), but some can't float but are amphibious,
+--This detects additional units that should be dropped.
+for unitDefID = 1, #UnitDefs do
+	local ud = UnitDefs[unitDefID]	
+
+	local floatsOnWaterSurface = floatingMoveClasses[ud.moveDef.smClass]
```
sprunk, line 85: Looks suspicious, do buildings have a proper `ud.moveDef` table with an `smClass` member?

LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
CarabusX: I just thought it is more characteristic for ships and hovers that they float, rather than for others that they not.
The check was all about readibility at the end.

LuaRules/Gadgets/unit_transport_pickup_floating_amphib.lua
CarabusX: Apparently they all have ud.moveDef table.
ud.moveDef.depth was nil for buildings and planes.
No idea about ud.moveDef.smClass, I will check.

CarabusX: Improved the check.

PR #4544, opened 2021-10-17
Adjust scalpel OKP timeout

adams164:
Current scalpel has OKP issues:
https://cdn.discordapp.com/attachments/278908415756206080/899346886127931462/ScalpelvFencer.mp4
https://cdn.discordapp.com/attachments/278908415756206080/899345857281945711/ScalpelOKPx2.mp4

Tweaking the time out gives better behavior:
https://cdn.discordapp.com/attachments/278908415756206080/899377365195558982/ScalpelOKPImproved.mp4

Concern was expressed re: behavior against raiders, with modified timeout:
https://cdn.discordapp.com/attachments/278908415756206080/899406021657755718/ScalpelGlaive.mp4

### DIFF
--- a/scripts/hoverskirm.lua
+++ b/scripts/hoverskirm.lua
@@ -148,7 +148,7 @@ function script.FireWeapon()
 end
 
 function script.BlockShot(num, targetID)
-	return GG.Script.OverkillPreventionCheck(unitID, targetID, 620.1, 440, 65, 0, true, 180, 0.5)
+	return GG.Script.OverkillPreventionCheck(unitID, targetID, 620.1, 440, 93, 0, true, 180, 0.5)
 end
 
 function script.Shot()


scripts/hoverskirm.lua
```diff
@@ -148,7 +148,7 @@ function script.FireWeapon()
 end
 
 function script.BlockShot(num, targetID)
-	return GG.Script.OverkillPreventionCheck(unitID, targetID, 620.1, 440, 65, 0, true, 180, 0.5)
+	return GG.Script.OverkillPreventionCheck(unitID, targetID, 620.1, 440, 93, 0, true, 180, 0.5)
```
CarabusX, line 151: Would it be best to determine the exact value by shooting Scalpel at Mex or Fencer at max range to see the exact time the missiles fly?

Issue #3922, opened 2020-05-08
Make Superweapons ramp up with energy grid


GoogleFrog:
I came across the idea of making superweapons better with more grid energy. Perhaps they could be disabled at no energy, be about about as powerful as they are now at 500 energy, and linearly extend their power from there.

DRP could rotate faster, causing it to fire faster. Starlight could aim faster and BZZT with higher frequency, Zenith could gather meteors faster.

Effects:
 - Adds a noob trap, because linking energy is an extra step.
 - There are more ways to disable superweapons.
 - Rushing a superweapon becomes a bit worse than making one with a large economy.

The rushing effect could actually be beneficial to the game. If it were possible to rush underpowered superweapons, ones that do not immediately end the game, then they could have more nuance.

Licho1:
I like that.. grid was meant to work for big defs but we nerfed big defs to near uselessness (annihilator)

alexpyattaev:
I think one way this can work is as follows:
- Starlight - more E => more range
- DRP, BB - more E => better rate of fire and more range
- Zenith - more E => more max asteroids controlled
- Nux - already has a nice dynamic
Another mechanic that I really liked in original TA that we could use is that superweapons actually USE energy. As in, for each BB shot you need to have e.g. 1000 energy or something. Same could work for others:

- BB reload time cut to 2 seconds, 1000 E per shot. This will make feeding more E to BB better than getting more of them, will also make counterplay with shields draining E much more fun - i.e. you have to choose if you bank energy for BB to keep shooting, or let shields charge.
- DRP reload time cut to 0.2 seconds, 1000 E per shot (effectively it is ratelimited by energy only). Idea being that if you run it non-stop it will drain just about any eco into oblivion. While tearing apart enemy eco, of course. But you have to be careful in terms of how you use it, or you'd drain your own eco to death.
- Zenith uses 100 energy per asteroid attracted, but has no ratelimit on asteroid attraction, so it can also drain all your E into nothing in seconds, but can also produce insane amount of DPS
- Starlight uses 100 energy per second while firing, more as range increases (give it unlimited range, but increase cost linearly with range, such that at BB max range it uses like 500 energy/second). This would make it into a nice counter-superweapon thing with unlimited (but expensive) range.
- Lucifer uses 300 energy per shot, reload time 5 seconds.
- Desolator and Cerberus use both 50 energy per shot, cut reload times to compensate for this nerf to keep them useful

[...]

PR #4545, opened 2021-10-17
⚖️ Zenith fire once now fires just 1/3 of meteors

sprunk:
Tries to alleviate the main issue of the unit (which is that meteors past
the first ~30 are mostly useless since everything except Det dies anyway)
without ruining the "firing spends way too much ammo" aspect of the unit.

This is a new mechanical ability for the unit and not just a bit of UI,
unlike Fire Once for other units. This is somewhat important not just
because of efficiency but also because overflow is bad so if you could
send a small, fixed amount of meteors into a corner it would dilute this
penalty.

### DIFF
--- a/scripts/zenith.lua
+++ b/scripts/zenith.lua
@@ -51,6 +51,7 @@ local spGetUnitIsStunned = Spring.GetUnitIsStunned
 local spGetUnitRulesParam = Spring.GetUnitRulesParam
 local spSetUnitRulesParam = Spring.SetUnitRulesParam
 local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
+local sputIsBitSet = Spring.Utilities.IsBitSet
 local CMD_ATTACK = CMD.ATTACK
 local gaiaTeam = Spring.GetGaiaTeamID()
 
@@ -58,6 +59,7 @@ local launchInProgress = false
 local isBlocked = true
 local currentlyEnabled = false
 
+local CMD_OPT_CTRL = CMD.OPT_CTRL
 local INLOS_ACCESS = {inlos = true}
 
 local ux, uy, uz
@@ -214,8 +216,11 @@ local function SpawnProjectileThread()
 	end
 end
 
-local function LaunchAll(x, z)
+local function LaunchN(x, z, n)
 	-- Sanitize input
+	if not n or n > projectileCount then
+		n = projectileCount
+	end
 	x, z = Spring.Utilities.ClampPosition(x, z)
 	local y = math.max(0, Spring.GetGroundHeight(x,z))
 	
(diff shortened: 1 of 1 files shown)


GoogleFrog:
So if I'm reading this correctly, a command with the fire-once switch makes it fires all the meteors? If it doesn't fire once by default then I doubt people will discover this ability, so it would take a long time to notice any balance impact. I also find the limit of 1/3 a bit weird.

How about up to 100 meteors at once, and a reload time of something like 5-10s. The reload time would give people a much better chance of noticing the batching. Then fire once would work as normal and the limit of 100 rather than 1/3 would prevent long tails of dregs.

sprunk:
> a command with the fire-once switch makes it fires all the meteors? If it doesn't fire once by default then I doubt people will discover this ability, so it would take a long time to notice any balance impact

Currently fire-once fires 1/3 and default fires all, since I was afraid of messing with the defaults and people's muscle memory (and presuming that nobody fires once at the moment). But since you agree I'll swap it.

> the limit of 100 rather than 1/3 would prevent long tails of dregs

The long tail is the point because it lets you send multiple bursts while making sure they're reasonably large. This maintains how Zenith is about burst damage but removing the "far too much burst damage than is useful" part that is keeping it down.

If you start from 300, then at flat-100 you can shoot 100, 100, 100.
At 1/3 you can shoot 100, 67, 45, 30, 20, 13, (25 remaining). Most of those are still brutal bursts, keeping with Zenith's theme, and enough to destroy most targets, but there are twice as many compared to flat.

If you start from 100 then at flat-100 you can just shoot 100.
At 1/3 you can shoot 34, 22, 15, (29 remaining). Four reasonable bursts instead of one but excessive.

The bursts can still get larger than you may prefer, but this happens slowly as you approach the cap, and you can trim the burst size somewhat gently (by using it). Compare to flat 100 which only ever lets you make one non-excessive burst (depending on how many meteors you have modulo 100).

---

Originally I wanted to make the burst a flat 25, but ultimately decided not to because I thought reaching the cap should still hurt a lot to keep incentivizing usage (and wasting 25 to prevent meteors falling on yourself hurts less than wasting 100) and to be Quant (since other superweapons don't have that kind of downside). Maybe the punishment is another thing keeping Zenith down though so perhaps the flat 25 isn't too bad after all, at least as long as the other supers don't have their own punishments? Idk. If you read this and still think that flat 100 is better (and that this is worth doing over the energy change suggested in #3922) then I'll just do that (in addition to the other changes).

GoogleFrog:
> If you start from 300, then with flat 100 you can shoot 100, 100, 100.
With 1/3 you can shoot 100, 67, 45, 30, 20, 13, (25 remaining). Most of those are still brutal bursts, keeping with Zenith's theme, and enough to destroy most targets, but there are twice as many compared to flat.

 > If you start from 100 then with flat 100 you can just shoot 100.
With 1/3 you can shoot 34, 22, 15, (29 remaining). Four reasonable bursts instead of one but excessive.

To me this implies a really kludgy model of the underlying (ie UI independent) abilities of the unit. The actions that a unit can take should have some level of in-world reasonableness. I see what you're doing UI-wise, but in terms of mechanics this is very complicated and awkwardly close to the much simpler rule: Zenith can fire any number of meteors it likes.

sprunk:
> in terms of mechanics this is very complicated and awkwardly close to the much simpler rule: Zenith can fire any number of meteors it likes.

Is this a suggestion (if yes, is 1 preferable to 100)? I considered making fire once fire just one meteor but decided that this would be un-Quant because this makes the "going over the cap punishes you via meteors fall on your head" that is unique to Zenith obsolete (just send 1 meteor into the corner right before spawning one over the cap) and also because Zenith is all about large bursts.

Maybe with fire 1,  Zenith is still all about bursts because sending 1 meteor is fairly pointless and maybe the punishment doesn't work in practice?

GoogleFrog:
> I considered making fire once fire just one meteor but decided that this would be un-Quant because this makes the "going over the cap punishes you via meteors fall on your head" that is unique to Zenith obsolete

I think any form of partial burst degrades this aspect to the point that the cap punishment may as well be removed. I think the big problem with partial burst is UI, as I don't recall hearing any suggestions that make such a Zenith have a good UI for common tasks.

I'm not sure that I'm at the point of suggestions yet. So far I feel more like I've been finding potential complications down various paths rather than picking a best path.

sprunk:
Keeping the branch, maybe reopen later when Zenith really needs looking at.

PR #4551, opened 2021-10-19
deduplicate halberd sounds

Anarchid:
Halberd weapon fire sound used to be identical to bandit fire sound. No more.
Halberd weapon impact sound used to be identical to buoy fire sound. No more.


GoogleFrog:
image

The existing sounds are shown as a comparison.

Technical issues:
- Clipping.
- Very asymmetric waveform may be a problem?
- Long periods of wasted silence or near-silence.
- No ramp up.

Otherwise, I don't get these sounds. It feels like we need a palette (or some equivalent concept for sounds) because I don't see how they fit ZK. Perhaps I can explain it by saying it sounds both more and less realistic at the same time, in that to me it is like a more realistic interpretation of an explosion but in an 8-bit game. The sound is fuzzier but also has less depth to it, like a lot of the frequencies were removed or equalised.

To draw an analogy, these sounds are snare drums while the rest of ZK is bongos. Neither are trying to sound like a drum set (real world sounds), but they stylised in different ways. I prefer the simpler but richer sounds of ZK to the more noisy and "flat" sounds in this PR.

sprunk:
Doesn't seem to be happening.

PR #4552, opened 2021-10-19
deduplicate disruptor bomb sound

Anarchid:
Proposed:

https://user-images.githubusercontent.com/3822768/137996509-35254b26-f882-48a4-ad1b-dc8f47ee1606.mp4

Current (aka gremlin/toad/raptor aa laser shoot sound):

https://user-images.githubusercontent.com/3822768/137996531-30852539-2b65-4853-bcab-78e585c7f0b9.mp4


GoogleFrog:
I have numerous issues.

image

Technical:
- Clipping.
- Asymetric waveform (I vaguely feel like this is bad but I don't know enough about sound encoding to justify the feeling.
- Very long wasted tail of near-silence.

Sound design:
- The first 10% sounds like a low-fi 8-bit explosion.
- The last 90% sounds like something is charging up.
- The combination of the above makes it sound like two sounds have occurred, ie, that two actions in the game world have occurred.
- I doesn't sound all that much like an explosion to me.
- Imagine it being used for DRP and slowbomb (for consistency), it sounds even less fitting there.

Perhaps you could send the two source files and it can be cleaned up, possibly combined with the current sound.

The primary advantages of the current sound are that it happens in sync with the visuals and it sounds more like an explosion.

a-nop:
Asymmetric blips in audio signal would only be a problem when: it prevents adding enough gain where you need it because headroom is reduced by the offset amount, but since it's not directly on the transient doesn't seem like a problem here.

It looks more like a very low frequency blip which could be part of the "thump".

A good explanation to put the mind at ease: https://www.soundonsound.com/sound-advice/q-why-do-waveforms-sometimes-look-lop-sided

sprunk:
Doesn't seem to be happening, reopen if there are new advances.

PR #4558, opened 2021-10-23
Unhardcode Puppy weapon-jump logic

CarabusX:
Unhardcode Puppy weapon-jump logic in unit_puppy_handler.lua gadget.

This fixes point 20 from issue #4535.


GoogleFrog:
When is teleporting units mod?

PR #4560, opened 2021-10-30
Swift can't fire while landed/landing

a-nop:
This change prevents Swift from firing while stationary or while only moving vertically.

### DIFF
--- a/scripts/planefighter.lua
+++ b/scripts/planefighter.lua
@@ -14,6 +14,7 @@ local flare = {
 	[0] = flare1,
 	[1] = flare2,
 }
+local isMoving = false
 
 local SPEEDUP_FACTOR = tonumber (UnitDef.customParams.boost_speed_mult)
 local BOOSTUP_FACTOR = tonumber (UnitDef.customParams.boost_accel_mult)
@@ -119,10 +120,12 @@ function script.Create()
 end
 
 function script.StartMoving()
+	isMoving = true
 	activate()
 end
 
 function script.StopMoving()
+	isMoving = false
 	deactivate()
 end
 
@@ -139,7 +142,11 @@ function script.AimFromWeapon(num)
 end
 
 function script.AimWeapon(num, heading, pitch)
-	return not (GetUnitValue(COB.CRASHING) == 1)
+	if not (GetUnitValue(COB.CRASHING) == 1) and isMoving then
+		x,y,z = Spring.GetUnitVelocity(unitID)
+		return (x ~=0 or z ~= 0)
+	end
+	return false
 end
 
 function script.FireWeapon(num)


a-nop:
The intention is to fix some game kludges produced by Swift grounding:
* Swift's blaster weapon should be tuned as a strafing weapon against ground targets, not as a potentially sustained fire weapon with 700 elmo range.
* AFAIK the campaign does not demonstrate that it's even possible to instruct Swifts to land.

It may still be possible to cheat this with command spam, because there is some horizontal movement during some parts of the takeoff and landing cycles.  However I think this is marginal because:
* The landing cycle rapidly decelerates to zero horizontal movement as soon as landing is initiated.
* The takeoff cycle sets nose up as soon as there is any horizontal movement, putting a ground target outside the firing arc.

It seems to me that any potential command spam hack would not result in more damage over time than a regular strafing attack order, which is the goal.

Further mitigation by looking at altitude/velocity is tricky because:
* Swifts may have low horizontal speed while in air combat, because they can turn quickly and "drift".
* Swifts may have low altitude (less than cruising) during strafing runs.
* Swifts may have non-zero vertical speed during strafing runs.

Reducing Swift's firing arc would potentially mitigate further.

Licho1:
Not sure if this is desirable.. wasn't it ok if people randomly landed them? It was rarely used tactics anyway

a-nop:
Landing Swifts is objectively better sustained DPS against ground targets compared to strafing runs.  Just because it's a Rarely Used tactic doesn't mean it's Not Good.  Just because it's mostly relevent earlygame and in teams doesn't mean it doesn't matter.  And on the other side of the coin, if it's rarely used, how does that justify it being in the game?  The stated goal is to remove any need to balance Swift as a potentially landed sustained damage platform.

Defensively, a parked Swift has better range and DPS than a Picket, meaning it has equivalent sustained DPS/cost to a static defense, except it can also boost away from danger and be used for many other purposes.

Offensively, parking Swifts near an expansion is the fastest and least exposure to AA coverage technique to burn it down by a big factor.

By removing this kludge further tweaks to Swift's unitdef are opened up.  We may find that the Swift gun/missile balance needs tweaking to compensate for less sustained damage potential against ground targets, which is a better problem to have.

Landing Swifts is one of those things where I'm annoyed that I **have to** use it to get the most out of the unit.

a-nop:
Landed Swifts were seen a lot in lobpot today (and in the past week).

Licho1:
Ok, let's merge it then.. but just regarding implementation performance, is it really necessary to check crashing and velocity from engine? Won't the isMoving be enough?

a-nop:
Checking crashing was existing behavior so I didn't change it.  This should prevent Swift firing while crashing, it doesn't share that gunship trait.

If you only check `isMoving` the Swift will fire during the takeoff and landing animations in which it switches between regular flight model and vertical movement.  During this time it hovers practically stationary, and if you alternate rapidly between a move order (right click) and cancelling the order (S) it will hang in a vertical movement cycle, giving you sustained fire.  Not wanting to wait for people to start widgeting this, I've added the velocity check to disallow firing during the vertical part of the takeoff/landing animation which should curb at least some of the potential abuse.

The logic is set up such that the more expensive velocity check is gated behind the less expensive checks.  Checking `isMoving` in the first place is sort of an edge case, it's only true when the plane is fully landed while the velocity check finds most of the aim-blocking states we want.  If we later want to check for horizontal motion under some threshold we would reach for some more expensive trig, but ponder it first.

Licho1 (approved):

a-nop:
I'm concerned that initializing `isMoving = false` might break savegames, will `StartMoving` be called?  Probably safer to initialize it to `true`.

a-nop:
Nevermind, looks like it works fine.

GoogleFrog:
Seems to work in testing.

PR #4569, opened 2021-11-15
🎵 Impaler noises: tank → vehicle

sprunk:
All veh, including the other ex-tanks, have veh sounds. But impalator is an order of magnitude heavier than other veh, so maybe it's okay. Dunno, just a suggestion.


GoogleFrog:
> But impalator is an order of magnitude heavier than other veh, so maybe it's okay.

This was the intention.

sprunk:
Reverted and added that as a comment bf6af027051ce97f0ff455b9bf651494c78649b3

PR #4575, opened 2021-12-05
Paladin: 20 combat regen

Aquanim:
Tested in nullAI skirmish and nothing broke


GoogleFrog (commented):
Should these lines exist? They are missing from Det.

sprunk:
Combat regen obsoletes them so effects-wise it doesn't matter if they exist, but for confusion reasons they shouldn't.

PR #4577, opened 2021-12-08
3rd-person COFC tracking disabled

dvr333:
*If this is a necessary feature, it should be its own camera script, considering both how different the camera should work in 3rd-person mode, and how internally we're changing engine camera mode anyway

### DIFF
--- a/LuaUI/Widgets/camera_cofc.lua
+++ b/LuaUI/Widgets/camera_cofc.lua
@@ -650,31 +650,31 @@ Complete Overhead/Free Camera has six actions:
 		path = cameraFollowPath,
 	},
 
-    thirdpersontrack = {
-		name = "Enter 3rd Person Trackmode",
-		desc = "3rd Person track the selected unit (mouse midclick to exit mode). Press arrow key to jump to nearby units, or move mouse to edge of screen to jump to current unit selection (will exit mode if no selection).",
-		type = 'button',
-		hotkey = {key='k', mod='alt+'},
-		path = cameraFollowPath,
-		OnChange = function(self)
-			local selUnits = Spring.GetSelectedUnits()
-			if selUnits and selUnits[1] and thirdperson_trackunit ~= selUnits[1] then --check if 3rd Person into same unit or if there's any unit at all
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/camera_cofc.lua
```diff
@@ -1738,7 +1739,7 @@ OverviewAction = function()
 				thirdperson_trackunit = selUnits[1]
 				local cs = GetTargetCameraState()
 				cs.px,cs.py,cs.pz=spGetUnitPosition(selUnits[1]) --move camera to unit position so no "out of LOD" case
-				cs.py= cs.py+25 --move up 25-elmo incase FPS camera stuck to unit's feet instead of tracking it (aesthetic)
+				cs.py= cs.py+spGetUnitHeight(selUnits[1])+50 --move up 25-elmo incase FPS camera stuck to unit's feet instead of tracking it (aesthetic)
```
sprunk, line 1742: Now all those comments ("move up 25-elmo...") are incorrect. Maybe this should be a helper function because it's duplicated so much.

LuaUI/Widgets/camera_cofc.lua
```diff
@@ -650,31 +650,31 @@ Complete Overhead/Free Camera has six actions:
...
-				-- Spring.SetCameraState(cs,0)
-				OverrideSetCameraStateInterpolate(cs,0)
-			else
-				Spring.SendCommands('trackoff')
-				Spring.SendCommands('viewfree')
-				thirdperson_trackunit = false
-			end
-        end,
-	},
+ --    thirdpersontrack = {
```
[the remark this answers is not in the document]
dvr333: Done

LuaUI/Widgets/camera_cofc.lua
dvr333: I was considering making it a helper function, but wasn't sure because it's a deprecated feature. I'll make it a helper function anyway, that might help someone if they want to extract it.

dvr333:
GoogleFrog has enabled more solid support for rotation while tracking units, so this PR is obsolete

PR #4584, opened 2022-01-02
Ronin fight AI skirmishes farther away

sprunk:


LuaRules/Configs/tactical_ai_defs.lua
```diff
@@ -1369,14 +1369,15 @@ local behaviourConfig = {
 		fightOnlyUnits = medRangeExplodables,
 		maxSwarmLeeway = 30,
 		minSwarmLeeway = 130,
-		jinkPeriod = 2,
+		jinkPeriod = 2.5,
```
sprunk, line 1372: As far as I can tell, only integers make sense here and this is equivalent to 3.
```suggestion
		jinkPeriod = 3,
```

palimpsestate:
Tidied that up.  Any further input?

sprunk:
From a technical PoV looks good, now GoogleFrog should say something from a design/gameplay PoV.

palimpsestate:
*crickets*

GoogleFrog:
I told a Ronin to attack move a Jack and it walked far away while reloading.
image

GoogleFrog:
Also Ronin very rarely fires at a Jack that is running from it.

PR #4587, opened 2022-01-05
Additional rmb command settings options

scrangos:
Adds settings for issuing target and attack as if it had alt pressed with rmb drag.  Also adds the ability to also issue line formation fight with the rmb drag.

### DIFF
--- a/LuaUI/Widgets/cmd_customformations2.lua
+++ b/LuaUI/Widgets/cmd_customformations2.lua
@@ -131,6 +131,8 @@ local requiresAlt = {
 local overrideCmds = {
 	[CMD.GUARD] = CMD_RAW_MOVE,
 	[CMD_WAIT_AT_BEACON] = CMD_RAW_MOVE,
+	[CMD.ATTACK] = 20,
+	
 }
 
 -- What commands are issued at a position or unit/feature ID (Only used by GetUnitPosition)
@@ -518,10 +520,10 @@ function widget:MousePress(mx, my, mButton)
 	
 	-- Get command that would've been issued
 	local _, activeCmdID = spGetActiveCommand()
+	
 	if activeCmdID then
 		if mButton==3 and options.RMBLineFormation.value then
 			usingCmd = activeCmdID
-			--If the option to use RMB for fight and attack as well is on we fake the rmb not being used. Is this too janky?
 			usingContextCommand = false
 		else
 			if mButton ~= 1 then
@@ -543,6 +545,7 @@ function widget:MousePress(mx, my, mButton)
 		end
 		
 		local overrideCmdID = overrideCmds[defaultCmdID]
+		
 		if overrideCmdID then
 			local targType, targID = CulledTraceScreenRay(mx, my, false, inMinimap)
 			if targType == 'unit' then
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I don't see why this is just for some commands. Shouldn't the setting simply make all RMB form line commands?

GoogleFrog:
I don't see why a list of commands to apply the change to is required. I'd just do all of them. This widget usually doesn't care about what a command actually does or how it is issued. There should be a single consistent UI unless there is a good reason to do otherwise. One example of doing otherwise is the difference in handling between commands with an area mode and commands without an area mode. But in this case, since nothing else happens on RMB with a command selected, I don't see why all the commands wouldn't do the same thing.

The only difference between set target and set target circle is that one draws a circle area while the other draws a box. Neither of them is "the one that works". Some poweruser bound the non-circle set target because they wanted rectangle select rather than a circle. But as far as custom formations is concerned, they're just commands that can be issued on a line.

scrangos:
Oh, I meant that in the option definition when I specify attack and set target I had tried the one that didnt say area first but when i tried to issue the command ingame it didnt work.  so the command bound to T ingame corresponds to the area circle im guessing since the other one didn't do anything.

Alright Ill change it to be a global implementation.

scrangos:
Okay I updated the implementation

LuaUI/Widgets/cmd_customformations2.lua
```diff
@@ -657,7 +671,10 @@ local function StopCommandAndRelinquishMouse()
 end
 
 function widget:MouseRelease(mx, my, mButton)
-	if (mButton == 1 or mButton == 3) and (not usingRMB) == (mButton == 3) then
+	--copied from press to check activecmd
+	local _, activeCmdID = spGetActiveCommand()
```
GoogleFrog, line 675: This is redundant now right?

scrangos: Is it, you're right. ill update the branch

GoogleFrog:
Right clicking no longer cancels the command. It issues a point command.

GoogleFrog:
It's weird that usingRMB is false when I use the right mouse button.

scrangos:
Well, I loaded the game again with your fixed code, I tried all the combinations I could think of and it seems to be fixed.  Anything else that I can do?

scrangos:
Hold up, theres something im not sure is wrong with plane context targetting behavior, need to test it without and with.

scrangos:
Kay fixed it, had to add it to the commands that get overriden if the press and release are on the same entity.  Has the behavior that if you move the mouse off the unit and put it back before release it overrides, but it also does that with guard so I'm assuming that is the intended behavior.

LuaUI/Widgets/cmd_customformations2.lua
```diff
@@ -131,6 +131,8 @@ local requiresAlt = {
 local overrideCmds = {
 	[CMD.GUARD] = CMD_RAW_MOVE,
 	[CMD_WAIT_AT_BEACON] = CMD_RAW_MOVE,
+	[CMD.ATTACK] = 20,
```
GoogleFrog, line 134: Replace 20 with the command it is meant to be.

LuaUI/Widgets/cmd_customformations2.lua
```diff
@@ -566,6 +569,10 @@ function widget:MousePress(mx, my, mButton)
 		
 		usingContextCommand = true
 		usingRMB = true
+		Spring.Echo("usingContextCommand")
```
GoogleFrog, line 572: Echos

scrangos: the heck, i thought i did find echo and i removed them all.
i tried using the attack line there but i got a not defined error, and CMD_RAW_MOVE isnt defined in that file.  wasnt sure what to do. still am not, should i just define it in the same file as a global?

PR #4592, opened 2022-01-15
Spread out units of same type in custom formation.

amnykon:
local function HasWaterWeapon(UnitDefID) is no longer used. and removed.


amnykon:
I have not seen any issues with maxHngTime and maxNoXTime; however, they may need to be lowered as now these timeouts are per unit type. I don't make large numbers of fleas, so I did not stress test it.

GoogleFrog:
I'm not sure how I feel about this situation:
image

amnykon:
If the cop has area cloak or shield, he needs to be centered.

PR #4593, opened 2022-01-16
make Scallop more assaulty

palimpsestate:
This is a first draft of a more assaulty Scallop.  It makes use of the shotgun weapon, but makes it very slow compared to other shotgun weapons.  Hopefully we can use this version of Scallop as a starting point to discuss how to further develop the Scallop.

### DIFF
--- a/units/amphriot.lua
+++ b/units/amphriot.lua
@@ -5,7 +5,7 @@ return { amphriot = {
   acceleration           = 0.54,
   activateWhenBuilt      = true,
   brakeRate              = 2.25,
-  buildCostMetal         = 260,
+  buildCostMetal         = 240,
   buildPic               = [[amphriot.png]],
   canGuard               = true,
   canMove                = true,
@@ -29,7 +29,7 @@ return { amphriot = {
   footprintZ             = 2,
   iconType               = [[amphtorpriot]],
   leaveTracks            = true,
-  maxDamage              = 2070,
+  maxDamage              = 1300,
   maxSlope               = 36,
   maxVelocity            = 1.6,
   movementClass          = [[AKBOT2]],
@@ -94,7 +94,7 @@ return { amphriot = {
       },
 
       damage                  = {
-        default = 22.01,
+        default = 48.01,
       },
 
       edgeEffectiveness       = 0.6,
@@ -126,8 +126,10 @@ return { amphriot = {
 
     FLECHETTE = {
       name                    = [[Flechette]],
-      areaOfEffect            = 16,
-      burst                   = 3,
+      areaOfEffect            = 32,
+      bounceRebound           = 1,
+      bounceSlip              = 1,
+      burst                   = 4,
       burstRate               = 0.033,
       coreThickness           = 0.5,
       craterBoost             = 0,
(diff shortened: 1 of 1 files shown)


Licho1:
This looks like a giant buff.. it has normal hp/cost for a riot..

palimpsestate:
I did play a test game with this unit, where I more or less monospammed the scallop with buoys treating it as a riot.  https://zero-k.info/Battles/Detail/1282375  .  I designed it to perform poorly as a riot on purpose and it does succeed at being bad at rioting.  The spray angle and reload time make it too high for it to function well as a riot and they were routinely overrun by glaives pretty easily.  The idea of this scallop is that it has to be lobbed in to point blank things for big dps.  Honestly I think it is probably underpowered right now and it probably needs to delete things harder at pb range

GoogleFrog:
I don't think designing Scallop as something that has to be lobbed is going to lead anywhere healthy. As an assault this has pretty terrible stats. I don't see it fighting at point blank range very often because enemy units can just move away. Even when lobbed, the impact location can be predicted and avoided.

Hermit is perhaps the closest comparison for raw stats. Hermit is:
- 62% of Scallop weight.
- 6% faster.
- 85% better HP/cost.
- 19% more range.

This Scallop has 266 empirical DPS, so it has 3.07x more DPS/cost than Hermit. But Scallop has to be pretty close to deal this damage. A single Scallop beats a Hermit if the distance between the units is less than around 250. Three Scallops shooting at max range die to a single Minotaur.

If we say that Scallop should be used at point-blank range then we can compare it to Jack. Jack is:
- 2.5x heavier.
- 12% faster.
- 85% better HP/cost.
- 42% of Scallop range.
- 45% of Scallop DPS.

High weight and jumpjets are vital for Jack as it lets it retreat and build up value. Even then, Jack is pretty niche. I could see this Scallop being dropped on factories and commanders to snipe them, but I'm doubtful of further use. Requiring a Lobster (or perhaps a cloaker) seems pretty damning.

I don't have any reliable ideas about what to do with Scallop. Maybe it could have an armoured movement mode (ala Halberd) so it can at least close with turrets. Mostly, I don't see a hole in the amph lineup around the role of Scallop.

On an implementation note, shotgun bouncing feels off to me.
- No other shotguns bounce.
- The bounce is very easy to miss, so it's a hidden mechanic.
- The bounce doesn't do anything in the sea.

palimpsestate:
I honestly agree with most of your critique. Needing to keep scallop hp fairly low is really a straitjacket on design, perhaps scallop needs to be conceived as a lower weight design in order to reach a useful place in the roster.  I like the idea of an armored movement mode, even if that were to come with a "stop to fire" penalty.  This would give amphs what I feel they desperately need - something that can be lobbed into riots without immediately dying.  Of course you felt that it wouldn't be healthy having a scallop that was designed to be lobbed.  I am willing to pursue a design based around armored movement if you think that could be a promising direction.

sprunk:
Swap Jack and Scallop between the facs?
 * melee makes sense underwater and Jack would make for a great lobbee (as an amph, it would lose the jump, but Lobster makes up for that and also now there's easy access to Limpet that prevents things from kiting Jack).
 * jumpy finally gets a comparatively low-jank riot unit, and even the kinda sucky Scallop would be a big help to its roster. Unsure if even needs to get jump.

GoogleFrog:
I'm not confident about an armoured mode. We'd also have to see what the model is capable of. I think we need to step back and find a hole in amph that would be reasonable to fill, as I don't know how a slow bursty Halberd is going to be a better option than Bulkhead encroachment.

Underwater melee has the fatal flaw of the sea height varying randomly per map.

palimpsestate:
I think we can identify two areas where amphs struggle.  One is that they do a very poor job of dealing with massed skirmishers shy of grizzly, and grizzly is a bit too expensive to deal with that problem.  The second is that they sacrifice all of their mobility in water on every unit other than ducks and scallops, but the scallop torpedo in water is quite little damage.

The first problem is one that I think could be lessened by a scallop with armor when moving.  This lets it survive being lobbed in to actually have a chance to attack the things that are doing a suspiciously good job against ducks.  I'm not sure the current degree of difficulty experienced by the amph player vs skirmishers is reasonable right now, with ducks being unable to close (and then having to fight riots, especially reaver) and neither buoy nor bulkhead really being able to land hits on ronins and reavers either.  Making bulkhead/buoy's projectile better so it could hit those units more easily might be an option (bulkhead projectile being better is probably better than buoy's projectile being better, but maybe I'm wrong).

If you don't like scallop as an armored lobster target, perhaps making the scallop be a sea skirmisher/riot that can actually move and not have to surface and stay still to be useful would be an acceptable niche for it.

sprunk:
Doesn't seem to be happening.

sprunk:
There's a merge conflict, though more importantly this idea doesn't seem to have gained traction.

PR #4594, opened 2022-01-16
added apm stats as third tab on endgame window

Dave-tB:
small issues with accuracy of stats, but otherwise should be good

### DIFF
--- a/LuaUI/Widgets/gui_chili_apm_stats.lua
+++ b/LuaUI/Widgets/gui_chili_apm_stats.lua
@@ -1,0 +1,174 @@
+----------------------------------------------------------------------------
+----------------------------------------------------------------------------
+
+function widget:GetInfo()
+	return {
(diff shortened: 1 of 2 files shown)


Dave-tB:
Issue #1: Only most recent session of stats are logged for player. (player dces, then reconnects, earlier stats are lost)  Spring.GetCurrentStatistics doesn't see stats prior to this.  For now, I've just compensated for that with tracking inactive time.

Issue #2: single unit line move can be abused to artificially raise apm.

Dave-tB:
a. stats are only tracked for a players most recent session atm, if they leave and then come back, the original stats are lost.
b. apm can be artificially raised to insane levels by single unit draw/drag line move, which spams move commands.

Dave-tB:
local player keeps track of own statistics but can't know active time...other players can know active time but not playerstatistics

PR #4598, opened 2022-01-18
fixes for apm stats crash

Dave-tB:
Most likely cause was that player was logging incorrect stats for themselves (logging raw stats like Mouse Pixels instead of final stats like MPS).  Also, adjusted endgamewindow so it would keep calling update after gameEnd and window setup, only for apmstats, in case their is a delay in sending.

### DIFF
--- a/LuaUI/Widgets/gui_chili_endgamewindow.lua
+++ b/LuaUI/Widgets/gui_chili_endgamewindow.lua
@@ -66,6 +66,7 @@ local endgame_fontcolor
 local gameEnded
 local showEndgameWindowTimer
 local myPlayerID = Spring.GetMyPlayerID()
+local updateFlag = true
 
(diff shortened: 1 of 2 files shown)


LuaUI/Widgets/gui_chili_endgamewindow.lua
```diff
@@ -626,32 +627,36 @@ function widget:GameOver(winners)
 end
 
 function widget:Update(dt)
-	showEndgameWindowTimer = showEndgameWindowTimer - dt
-	if showEndgameWindowTimer > 0 then
-		return
+	if updateFlag then
```
GoogleFrog, line 630: This should be `if not then stuff return end` rather than `if .. then giant block then end ...`

Dave-tB:
going to redo this

PR #4604, opened 2022-01-22
Add global build command selection override option and fix global build command toggle.

amnykon:


LuaUI/Widgets/gui_selection_hierarchy.lua
```diff
@@ -140,6 +154,30 @@ options = {
...
+		path = globalBuildPath,
+		OnChange = function (self)
+			globalBuildRankOverride = self.value
+		end
+	},
+	globalBuildRankOption = {
+		name = 'Global build selection override:',
+		desc = "Units controlled by global build command are treated as this selection rank, if override is enabled.",
+		type = 'number',
+		value = 0, -- This should be 0 because otherwise Ctrl selection keys work on the unit.
```
amnykon, line 172: Not sure if the default should be 0, I am starting to think 2 would be better.

LuaUI/Widgets/gui_selection_hierarchy.lua
```diff
@@ -140,6 +154,30 @@ options = {
 			retreatingRank = self.value
 		end
 	},
+	globalBuildOverrideOption = {
+		name = "Global build overrides selection rank",
```
amnykon, line 158: Since the rank effectively uses min(), the boolean override options do not add any value. Primarily added due to Retreat overrides having one, which also effectively uses min().

LuaUI/Widgets/gui_selection_hierarchy.lua
```diff
@@ -70,7 +72,19 @@ end
 
 options_path = 'Settings/Interface/Selection'
 local retreatPath = 'Settings/Interface/Retreat Zones'
```
amnykon, line 74: I feel the path for retreatPath and globalBuildPath should be in 'Settings/Interface/Selection', as that makes more sense to me. These adjust the behavior of selection not retreat and global build command. If one has selection issues, they should look in the selection settings. Used globalBuildPath, mainly to match existing code. Can we change these without resetting everyone's settings?

GoogleFrog:
I dislike how the changes are spread across the widgets. The PR seems to add options that won't make any sense to anyone who is using selection rank and not using global build command. I expect this describes most players.

I suspect what I want is for https://github.com/ZeroK-RTS/Zero-K/pull/4604/files#diff-1f9e46326efcc900e8d72fda47d4979d217ad1b2b7b68ec33296aca04676ec5cR306 to be the only line changed in selection rank. The options can be in global build command to prevent anyone who doesn't use that widget from seeing them.

amnykon:
Moved the options over to global-build-command Lua file. I did this on a Mac, so I have not tested it yet. I probably will not test it before Monday.

Also should the retreat options be moved to its respected Lua file? I feel we should use the same pattern for both cases.

GoogleFrog:
> Also should the retreat options be moved to its respected Lua file? I feel we should use the same pattern for both cases.

In theory yes, but that would be a separate PR. I don't particularly care because retreat is enabled by default so players will see zero difference.

amnykon:
Tested. As expected, untested code crashes. Fixed.

GoogleFrog:
Note that I did not test the GBC side of things.

PR #4618, opened 2022-01-29
Fixed most of the issues, SHADOW/no-SHADOW shaders are left as an excercise to the reader

lhog:
Plus ==> https://github.com/beyond-all-reason/spring/commit/85fc43f9b7fad62ebf1883f5f6cc79e7c8175f44

Should do the trick.


lhog:
https://github.com/beyond-all-reason/spring/actions/runs/1764337627 (when ready) to test

PR #4644, opened 2022-02-24
Add formation Ranks.

amnykon:
Adds formation ranks to units other then planes, as it doesn't make much since for planes. Default formation rank is 2 for all units, which makes the default behavior the same as the current behavior.


LuaUI/Widgets/unit_start_state.lua
```diff
@@ -600,6 +607,18 @@ options = {
 		path = "Settings/Unit Behaviour/Default States/Misc",
 		tooltipFunction = tooltipFunc.selectionrank,
 	},
+
+	commander_formation_rank = {
+		name = "  Formation Rank",
+		desc = "Selection Rank: when selecting multiple units only those of highest rank are selected. Hold shift to ignore rank.",
```
sprunk, line 613:
```suggestion
		desc = "Formation Rank: units of lower rank line up in front of units of higher rank when given line movement orders.",
```

GoogleFrog:
I haven't had a chance to give this a review. But I like the idea. Here are some comments going into this which could speed things up.
- I think this should have the same level of discoverability as selection rank. Ie, can be set in Unit Start State and the toggle can be enabled in Interface/Commands.
- I think the defaults should be (almost?) all at some central value, such as formation rank 2. Perhaps AA and artillery could be 3. I would have to test to see how well it works, but I feel like the default should not do too much for the player.
- The behaviour of the widget when everything has the same formation rank should be identical to the existing behaviour.

amnykon:
GoogleFrog, I agree and already accounted for. I think having aa and artillery default to 3 would look like a bug to a new player. It also makes a formation going to the left different than going to the right, which takes awhile to get used to. Dragging the wrong way puts artillery in front, which is usually bad. Adding dots on the formation line helps a lot and is basically needed for this. It was difficult to use this until I updated that piece.

amnykon:
If I was not clear, defaulting everything to the same value makes leftwards mouse drags identical to rightward drags, except for minor unit ordering from spacing out like units.

PR #4649, opened 2022-03-07
more better randomness for wind turbine/tidal gen

SteelBlueZK:
math.random() will return a real number, but math.random() with
arguments returns only int eg. args (-0.5,0.5) the result will
always be 0.

Wind turbine randomness in stable a little less random than probably intended due to above. Also, in tidal form cradle no longer double turned by random value and is given randomness.

### DIFF
--- a/scripts/energywind.lua
+++ b/scripts/energywind.lua
@@ -11,21 +11,27 @@ local UPDATE_PERIOD = 1000
 local BUILD_PERIOD = 500
 
 local turnSpeed = math.rad(20)
+local waterSpin = math.rad(30)
+local waterMaxSpin = math.rad(2)
 
 local isWind, baseWind, rangeWind
 
 local function BobTidal()
-	baseDirection = baseDirection + math.random(0,math.rad(2))
+	local spinSpeed	= 0 -- radian speed
 	while true do
-		Turn(cradle, y_axis, baseDirection, math.rad(1))
+		
+		spinSpeed = math.max(-waterMaxSpin),math.min(waterMaxSpin, 0.99*spinSpeed + (math.random() - 0.5) / 100))
+		Spin(cradle, y_axis, spinSpeed)
+		Spin(fan, z_axis, waterSpin + spinSpeed)
 		
 		Move(cradle, x_axis, math.random(-2,2), 0.2)
-		Move(cradle, y_axis, math.random(-0.5,0.5) - 51, 0.05)
+		Move(cradle, y_axis, math.random(-1,1) * 0.5 - 51, 0.05)
 		Move(cradle, z_axis, math.random(-2,2), 0.2)
 		Sleep(1000)
 		
 		if GG.Wind_SpinDisabled then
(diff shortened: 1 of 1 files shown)


scripts/energywind.lua
```diff
@@ -15,12 +15,13 @@ local turnSpeed = math.rad(20)
 local isWind, baseWind, rangeWind
 
 local function BobTidal()
-	baseDirection = baseDirection + math.random(0,math.rad(2))
+	local tideDirection = 0
 	while true do
-		Turn(cradle, y_axis, baseDirection, math.rad(1))
+		tideDirection = tideDirection + math.random() - 0.5 * math.rad(2)
```
sprunk, line 20: Right now this line increments the direction value by between 0.0 and 1.0 radians (i.e. up to 57ish degrees with an average of 28ish) and then decrements by 1 degree (i.e. half of two degrees) for an average change of 27ish. This is much faster than the turn speed of 1 degree/s can keep up (the direction target runs circles around the turbine).

I believe this was supposed to be
```suggestion
		tideDirection = tideDirection + (math.random() - 0.5) * math.rad(2)
```
On the other hand not keeping up looks marginally better in that it keeps swaying while the "fixed" version will turn (for example) 0.3 degrees in 300ms, then stop for the remaining 700ms of a cycle. Perhaps save the random roll in a variable and scale the speed accordingly as well?

sprunk (changes requested):
See comment.

scripts/energywind.lua
SteelBlueZK: Thanks for the find.

scripts/energywind.lua
```diff
@@ -11,21 +11,27 @@ local UPDATE_PERIOD = 1000
...
 
 local isWind, baseWind, rangeWind
 
 local function BobTidal()
-	baseDirection = baseDirection + math.random(0,math.rad(2))
+	local spinSpeed	= 0 -- radian speed
 	while true do
-		Turn(cradle, y_axis, baseDirection, math.rad(1))
+		
+		spinSpeed = math.max(-waterMaxSpin),math.min(waterMaxSpin, 0.99*spinSpeed + (math.random() - 0.5) / 100))
```
sprunk, line 23: Looks incorrect:
```suggestion
		spinSpeed = math.max(-waterMaxSpin, math.min(waterMaxSpin, 0.99*spinSpeed + (math.random() - 0.5) / 100)))
```

But also, `0.99*spinSpeed + (math.random() - 0.5) / 100)` will never put `spinSpeed` beyond 1. Maybe it could be tweaked so that the min/max calls aren't needed (keeping in mind 1 radian is ~57 degrees).

SteelBlueZK:
TODO: better maths on this. I'll also look into how expensive math.min/math.max are in comparison to lua number math.

It might be an option to make it sway back and forth in a +- 60 degree range from the randomly defined baseDirection heading, which would be Turn().

PR #4656, opened 2022-03-24
Small speedup to avoid length() calcs on each sample, disable unused …

Beherith:
…paths of geo shader


Beherith:
THIS IS A BLIND EDIT

PR #4671, opened 2022-04-10
Optionally show WHR ratings in Tab Window

jklw:
Hi, my username in the game is "dunno".

Previous forum post: https://zero-k.info/Forum/Thread/34990

It can't display the rating of players who joined after the start; I think that would require adding the rating to the `MidGameJoinUser` message:

https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/9da8b2369c9d63e87ff4ab779f6f54cc45090c44/Shared/LobbyClient/DedicatedServer.cs#L75


LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -54,8 +55,10 @@ local images = {
 	giftenergy = 'LuaUI/Images/energy.png',
 }
 local defaultamount = 100
+local zeroIfNumWrhDisabled = 0
```
sprunk, line 58: Weird name. Also consider a proper type and apply logic later:
```suggestion
local whrEnabled = false
```
and then later:
```diff
- stuff + zeroIfNumWrhDisabled * whrWidth
+ stuff + (whrEnabled and whrWidth or 0)
```

LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -1354,6 +1398,7 @@ local function SetWantRebuild()
 		wantRebuild = true
 	end
 end
+SetWantRebuild_ = SetWantRebuild
```
sprunk, line 1401: Just move the definition of `SetWantRebuild` above the first use. Alternatively if there's ugliness and/or circular dependency then forward-declare the actual function.

sprunk (changes requested):

LuaUI/Widgets/gui_chili_share.lua
jklw: Was trying to keep the usage of Lua's somewhat dangerous replacement for the ternary operator in 1 place (actually 2), but ok :)

LuaUI/Widgets/gui_chili_share.lua
jklw: Lua lob here; was following the precedent in that file (`UpdateListFunction = Buildme`, but without the change of name ;)).

Moving would require moving lots of existing code to above the options.

I tried forward-declaring `local SetWantRebuild` on top and removing the highlighted line (and obviously, changing the call in `OnChange` to `SetWantRebuild()`); that strangely doesn't work (see error below. If `SetWantRebuild` is nil at that point, my version shouldn't work either, but somehow it does?).


```
[t=00:01:31.372525][f=0008188] Chili-Error in `Show current WHR (Elo) rating of players` (couldn't detect the owner widget): [string "LuaUI/Widgets/gui_chili_share.lua"]:124: attempt to call upvalue 'SetWantRebuild' (a nil value)
[t=00:01:31.372600][f=0008188] stacktrace:
	[string "LuaUI/Widgets/gui_chili_share.lua"]:124: in origOnChange
	[string "LuaUI/Widgets/gui_epicmenu.lua"]:1342: in origOnChange
	[string "LuaUI/Widgets/gui_epicmenu.lua"]:1178: in eventListener
	[string "LuaUI/Widgets/chili_old/controls/object.lua"]:631: in CallListeners
	...
	(tail call): in [?]
	[string "LuaUI/Widgets/chili_old/controls/screen.lua"]:233
```

LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -1,6 +1,6 @@
 function widget:GetInfo()
 	return {
-		name    = "Chili Share menu v1.24",
+		name    = "Chili Share menu v1.25",
```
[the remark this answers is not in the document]
jklw: Done

LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -1041,11 +1068,12 @@ local function InitName(subject, playerPanel)
 			noFont = true,
 		}
 	end
-	local country, icon, badges, clan, avatar, faction, admin
+	local country, icon, rating, badges, clan, avatar, faction, admin
 	if (subject.player) then
 		local pdata = select(10, Spring.GetPlayerInfo(subject.player))
 		country = select(8, Spring.GetPlayerInfo(subject.player, false))
 		icon = pdata.icon
+		rating = pdata.elo
```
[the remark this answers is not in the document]
jklw: Done

LuaUI/Widgets/gui_chili_share.lua
```diff
@@ -69,6 +72,18 @@ local pingCpuColors = {
...
 
+-- RGBA. Used as font color of players' WHR.
+local rankColors = {
+	["7"] = {1  , 0   , 1  ,  1},
+	["6"] = {0  , 0.6 , 1  ,  1},
+	["5"] = {0.7, 0.8 , 1  ,  1},
+	["4"] = {1  , 1   , 0  ,  1},
+	["3"] = {1  , 0.65, 0  ,  1},
+	["2"] = {0.8, 0.4 , 0.1,  1},
+	["1"] = {1  ,   0 ,   0,  1},
```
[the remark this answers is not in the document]
jklw: Done

LuaUI/Widgets/gui_chili_share.lua
jklw: To be clear, I've left this unchanged (using `SetWantRebuild_`) in the commit below, which works. So this PR should be functionally mergeable.

GoogleFrog:
I think options at the bottom of widgets can make a lot of sense. See integral menu.

LuaUI/Widgets/gui_chili_share.lua
GoogleFrog: @jklw for the future, don't learn precedents from this widget. Rebuilding as much as it does is bad, things like `UpdateListFunction = Buildme` are bad. Most of the other chili widgets are written in a better style.

PR #4675, opened 2022-04-16
Widget Performance Improvements

Porkch0p:


Porkch0p:
Changed: Reclaim Highlight, Pause Screen, Newton Fire Zone, Vertical Line on Radar Dots v2, Highlight Geos, Decloak Range, and Chili Minimap

LuaUI/Widgets/gui_pauseScreen.lua
```diff
@@ -183,9 +185,9 @@ function widget:DrawScreen()
 		if now - disablePauseSlideTimestamp > slideTime then
 			drawPause(paused, now)
 		end
+	ResetGl()
 	end
 	
-	ResetGl()
```
Porkch0p, line 188: This was being called even outside of the loop where it would draw to the screen

LuaUI/Widgets/cmd_newton_firezone.lua
```diff
@@ -41,9 +41,19 @@ local spGiveOrderToUnitArray = Spring.GiveOrderToUnitArray
...
 --local ech = Spring.Echo
 
 local floor = math.floor
+local min = math.min
+local max = math.max
+local abs = math.abs
+local sin = math.sin
+local cos = math.cos
+local atan2 = math.atan2
+local sqrt = math.sqrt
```
Porkch0p, line 56: The full set of changes are localizations. I couldn't reliably replicate high allocation rates, but the widget peaked in some late-game replays (https://zero-k.info/Battles/Detail/1342845 was one)

LuaUI/Widgets/gui_vertLineAid.lua
```diff
@@ -14,9 +14,22 @@ local last_frame = 0
 local disabled = false
 local enemyDots = {}
 local allyDots = {}
+local dot = { x, y, z, losState, r, g, b}
```
Porkch0p, line 17: All local allocations of x,y,z,losState,r,g,b instead happen to 'dot'

LuaUI/Widgets/gui_vertLineAid.lua
```diff
@@ -14,9 +14,22 @@ local last_frame = 0
...
+local removals = {}
```
Porkch0p, line 18: No longer being created for each DrawWorld

LuaUI/Widgets/gui_highlight_geos.lua
```diff
@@ -26,14 +26,18 @@ local glCallList = gl.CallList
...
 
+local abs = math.abs
 
 local geoDefID = UnitDefNames["energygeo"].id
 
 local mapX = Game.mapSizeX
 local mapZ = Game.mapSizeZ
 local mapXinv = 1/mapX
 local mapZinv = 1/mapZ
+local lastX, lastY, lastGeo
```
Porkch0p, line 40: lastGeo is cached value of WG.mouseAboveGeo, only recalculated on mouse move

LuaUI/Widgets/unit_decloak_range.lua
```diff
@@ -103,6 +105,33 @@ end
...
+				selectionCanCloak = true
+				return
+			end
+		end
+	end
+	selectionCanCloak = false
+end
+
+function widget:UnitCloaked(curID, curUnitDefID, teamID)
+	if (not currentSelection) or ((teamID ~= Spring.GetMyTeamID()) and not Spring.GetSpectatingState()) then
```
Porkch0p, line 125: Is there another exception beyond Spectator that might need to be handled here? Team Share?

LuaUI/Widgets/gui_chili_minimap.lua
```diff
@@ -1093,7 +1104,6 @@ local function DrawMiniMap()
 end
 
 function widget:DrawScreen()
-	local cs = Spring.GetCameraState()
```
Porkch0p, line 1096: Safe to have GetCameraState and GetViewSizes only called once per second and on resize?

LuaUI/Widgets/gui_reclaim_field_highlight.lua
```diff
@@ -664,7 +664,7 @@ function widget:GameFrame(frame)
 	end
 	UpdateFeatures(frame)
 	--spEcho("featuresUpdated", featuresUpdated)
-	if featuresUpdated then
+	if featuresUpdated and drawEnabled then
```
Porkch0p, line 667: Noticed in some intense late-game chicken the Wave indicator hitching every 1 second exactly, followed it back to this performing multi-megabyte allocations

LuaUI/Widgets/gui_vertLineAid.lua
GoogleFrog: I don't get this. Aren't all the variables in this list nil?

LuaUI/Widgets/unit_decloak_range.lua
GoogleFrog: Check for fullview.

LuaUI/Widgets/gui_vertLineAid.lua
Porkch0p: Right, this was more of a self-comment as I was making the change and doesn't need to be there. See update.

LuaUI/Widgets/unit_decloak_range.lua
Porkch0p: Okay, check for update to confirm

LuaUI/Widgets/gui_vertLineAid.lua
sprunk: Why? `local x, y, z = Foo()` is faster than `a.x, a.y, a.z = Foo()`

Porkch0p: This appears to have been a bad assumption on my part about the garbage collector. I re-profiled this with a hectic save state to get a better sense of what was happening re: performance. For better or worse it appeared to make very little difference compared to the drawing part of the widget. Scratch this one, will return to original.

PR #4681, opened 2022-04-21
Added a gadget to unstick caretakers

### LINKED ISSUE #4640: Caretakers attempt to reclaim wreck out of range
1. Have idle caretakers
2. Have a wreck fly above them (from an aircraft dying or a singularity explosion flinging them)
3. Caretakers will attempt to reclaim it while it's in range, however as it leaves range they all keep staring at it and won't attempt to assist with construction

Arch-Shaman:
Caretakers occasionally will get stuck reclaiming (typically when they catch onto a flying wreck) something out of range. This gadget introduces a 2hz check for if a feature (or unit) is range of a caretaker.

Fixes #4640.

### DIFF
--- a/LuaRules/Gadgets/unit_unstick_staticcon.lua
+++ b/LuaRules/Gadgets/unit_unstick_staticcon.lua
@@ -100,7 +100,7 @@ OR (after finding a feature)
 [5] = radius
 ]]
 
-local function HandleUnit(unitID, f)
+local function HandleUnit(unitID)
 	if not spValidUnitID(unitID) then
 		return
 	end
--- a/LuaRules/Gadgets/unit_unstick_staticcon.lua
+++ b/LuaRules/Gadgets/unit_unstick_staticcon.lua
@@ -45,18 +45,8 @@ local spGetUnitDefID = Spring.GetUnitDefID
 local spGetFeaturePosition = Spring.GetFeaturePosition
 --local EMPTY = {}
 
-local function DebugEcho(str, arg1, arg2, arg3, arg4)
-	if arg4 then
-		spEcho("[unit_unstick_staticcon]: " .. str, arg1, arg2, arg3, arg4)
-	elseif arg3 then
-		spEcho("[unit_unstick_staticcon]: " .. str, arg1, arg2, arg3)
-	elseif arg2 then
-		spEcho("[unit_unstick_staticcon]: " .. str, arg1, arg2)
-	elseif arg1 then
-		spEcho("[unit_unstick_staticcon]: " .. str, arg1)
-	else
-		spEcho("[unit_unstick_staticcon]: " .. str)
-	end
+local function DebugEcho(...)
+	spEcho("[unit_staticcon_unsticker]", unpack(arg))
 end
 
 local function GetDistance(unitID, targetID, isReclaim)
@@ -75,7 +65,7 @@ local function GetDistance(unitID, targetID, isReclaim)
 		if debug then DebugEcho("Target doesn't exist?") end
 		return 9999
 	end
-	return ((x2 - x) * (x2 - x)) + ((z2 - z) * (z2 - z))
+	return (x2 - x)^2 + (z2 - z)^2
 end
 
 local function IsObjectCloseEnough(unitID, targetID, cmd) -- Note: build ranges are 2D usually.
@@ -120,7 +110,7 @@ local function HandleUnit(unitID, f)
 	if cmdParam1 and spValidFeatureID(cmdParam1 - Game.maxUnits) then
 		cmdParam1 = cmdParam1 - Game.maxUnits
 	end
-	if debug then DebugEcho(unitID .. ": " .. tostring(cmdID) .. ", " .. tostring(cmdParam1) .. ", " .. tostring(cmdParam2)) end
+	if debug then DebugEcho(unitID .. ": ", cmdID, cmdParam1, cmdParam2) end
 	if cmdID and cmdParam1 and (spValidFeatureID(cmdParam1) or spValidUnitID(cmdParam1)) then
 		if debug then DebugEcho(unitID .. ": Valid command, checking distance.") end
 		if not IsObjectCloseEnough(unitID, cmdParam1, cmdID) then
--- a/LuaRules/Gadgets/unit_unstick_staticcon.lua
+++ b/LuaRules/Gadgets/unit_unstick_staticcon.lua
@@ -15,12 +15,11 @@ function gadget:GetInfo()
 end
 
 local IterableMap = VFS.Include("LuaRules/Gadgets/Include/IterableMap.lua")
-local handled = IterableMap.New()
-
+local handledUnits = IterableMap.New()
 
 -- Config --
-local delay = 15
-local debug = false
+local UPDATE_RATE = 45
+local debugMode = false
 
 local buildRangesSq = {} -- [unitDefID] = buildDistance
 
@@ -37,7 +36,6 @@ local CMD_REPAIR = CMD.REPAIR
 local CMD_RECLAIM = CMD.RECLAIM
 local CMD_RESURRECT = CMD.RESURRECT
 local CMD_REMOVE = CMD.REMOVE
-local spEcho = Spring.Echo
 local spGiveOrderToUnit = Spring.GiveOrderToUnit
 local spValidUnitID = Spring.ValidUnitID
 local spGetGameFrame = Spring.GetGameFrame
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,142 @@
...
+					spGiveOrderToUnit(unitID, CMD_REMOVE, cmdTag, 0)
+					if cmdParam5 and cmdParam5 ~= buildRanges[spGetUnitDefID(unitID)] then -- this was an area command added by the user.
+						if debug then DebugEcho("Readding area command: " .. tostring(cmdParam2) .. ", " .. tostring(cmdParam3) .. ", " .. tostring(cmdParam4) .. ", " .. tostring(cmdParam5)) end
+						spGiveOrderToUnit(unitID, CMD_INSERT, {0, cmdID, CMD.OPT_SHIFT, cmdParam2, cmdParam3, cmdParam4, cmdParam5}, CMD.OPT_ALT) -- reissue the command (just in case)
+					end
+				end
+			elseif debug then
+				DebugEcho("Invalid command handled for " .. unitID)
+			end
+			handled[f + delay] = handled[f + delay] or {} -- check back in 15 frames.
```
sprunk, line 117: Why build the table from scratch? `handled[f + delay] = handled[f]` and remove anything that died. Better yet, nanos are rare enough that it would be fine not to have a per-frame table and just a single iteratable map.

sprunk (changes requested):

LuaRules/Gadgets/unit_unstick_staticcon.lua
Arch-Shaman: Attempted optimization for FW as players are expected to build both rejuv (static repairer, probably less rare) and caretaker. Figured it would be less of a perf hit this way, though admittedly this is poorly understood. I just submitted the first working draft for FW. I'll update it to use iterable map. 

I'd like to know if I should do the same for my copy in FW since there's expected to be more static cons.

sprunk: Iterable map should be faster, regardless of what unit types there are (ie. still fastest in your mod when rejuv exists) or how many of each type get built (ie. still fastest if people build 9001 rejuvs and caretakers). A caveat is that it may disallow removal of dead units during iteration but `UnitFinished` and `UnitReverseBuilt` should be used for that regardless.

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,142 @@
+function gadget:GetInfo()
+	return {
+		name      = "Static Con Unsticker",
+		desc      = "Implements Caretaker / rejuvenator unstick.",
+		author    = "Shaman",
+		date      = "4-20-2022",
```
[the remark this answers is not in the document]
Arch-Shaman: Any particular reason for this format, out of curiosity?

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,142 @@
...
+		return dist <= buildRanges[spGetUnitDefID(unitID)]
+	elseif (cmd == CMD_RECLAIM or cmd == CMD_RESURRECT) and spValidFeatureID(targetID) then
+		local dist = GetDistance(unitID, targetID, true)
+		if debug then
+			DebugEcho(unitID .. ": Distance: " .. dist .. " (Build Range: " .. buildRanges[spGetUnitDefID(unitID)] .. ")")
+		end
+		return dist <= buildRanges[spGetUnitDefID(unitID)]
+	else
+		return false
+	end
```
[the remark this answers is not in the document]
Arch-Shaman: Will Echo(x, y) would work the DebugEcho func here?

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,142 @@
...
+local spGetUnitIsStunned = Spring.GetUnitIsStunned
+local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
+local spValidFeatureID = Spring.ValidFeatureID
+local spGetUnitPosition = Spring.GetUnitPosition
+local spGetUnitDefID = Spring.GetUnitDefID
+local spGetFeaturePosition = Spring.GetFeaturePosition
+--local EMPTY = {}
+
+local function DebugEcho(str)
+	spEcho("[unit_staticcon_unsticker]: " .. str)
```
sprunk, line 55:
```suggestion
local function DebugEcho(...)
	spEcho("[unit_staticcon_unsticker]", unpack(arg))
```

LuaRules/Gadgets/unit_unstick_staticcon.lua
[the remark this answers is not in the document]
sprunk: Right, `DebugEcho` needs to support it. See the comment above at that func

LuaRules/Gadgets/unit_unstick_staticcon.lua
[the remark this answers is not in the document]
sprunk: That is the international standard when writing dates using numbers-with-dashes ( https://en.wikipedia.org/wiki/ISO_8601 ). Feel free to use something like "April 20th, 2022" if you want the month in front though.

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,142 @@
...
+			x2, _, z2 = spGetUnitPosition(targetID)
+		end
+	else
+		x2, _, z2 = spGetUnitPosition(targetID)
+	end
+	if not x2 then
+		if debug then DebugEcho("Target doesn't exist?") end
+		return 9999
+	end
+	return math.sqrt(((x2 - x) * (x2 - x)) + ((z2 - z) * (z2 - z)))
```
sprunk, line 74:
```suggestion
	return (x2 - x)^2 + (z2 - z)^2
```

LuaRules/Gadgets/unit_unstick_staticcon.lua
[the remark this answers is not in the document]
Arch-Shaman: No need, just wanted to know the why behind it. Updated.

LuaRules/Gadgets/unit_unstick_staticcon.lua
Arch-Shaman: That's fine with me. I'll update both versions to use it.

Arch-Shaman: Added iterablemap.

Arch-Shaman:
Untested, but setting to ready for review for now. The FW version is relatively stable, but i'd need to do testing in zk.

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,152 @@
...
+local spValidFeatureID = Spring.ValidFeatureID
+local spGetUnitPosition = Spring.GetUnitPosition
+local spGetUnitDefID = Spring.GetUnitDefID
+local spGetFeaturePosition = Spring.GetFeaturePosition
+
+local function DebugEcho(...)
+	spEcho("[unit_staticcon_unsticker]", unpack(arg))
+end
+
+local function GetDistance(unitID, targetID, isReclaim)
```
GoogleFrog, line 55: GetDistanceSq

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,152 @@
...
+			end
+		end
+	elseif debug then
+		DebugEcho("No action needed for " .. unitID)
+	end
+end
+
+function gadget:GameFrame(f)
+	if f%delay == 0 then
+		for id, _ in pairs(IterableMap.Iterator(handled)) do
```
GoogleFrog, line 129: Iterator is slow. Use Apply. Also, doesn't Iterator work without putting it in a pairs?

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,152 @@
...
+
+local function HandleUnit(unitID)
+	if not spValidUnitID(unitID) then
+		return
+	end
+	local cmdID, _, cmdTag, cmdParam1, cmdParam2, cmdParam3, cmdParam4, cmdParam5 = spGetUnitCurrentCommand(unitID)
+	if cmdParam1 and spValidFeatureID(cmdParam1 - Game.maxUnits) then
+		cmdParam1 = cmdParam1 - Game.maxUnits
+	end
+	if debug then DebugEcho(unitID .. ": ", cmdID, cmdParam1, cmdParam2) end
```
GoogleFrog, line 111: Whitespace, missing newline after then.

LuaRules/Gadgets/unit_unstick_staticcon.lua
GoogleFrog: ApplyFraction would be even better.

GoogleFrog:
`for id, _ in pairs(IterableMap.Iterator(handled)) do` simply crashes. How does this work in FW?

The gadget does not work through luarules reload.

LuaRules/Gadgets/unit_unstick_staticcon.lua
```diff
@@ -0,0 +1,152 @@
...
+local spGetGameFrame = Spring.GetGameFrame
+local spGetUnitIsStunned = Spring.GetUnitIsStunned
+local spGetUnitCurrentCommand = Spring.GetUnitCurrentCommand
+local spValidFeatureID = Spring.ValidFeatureID
+local spGetUnitPosition = Spring.GetUnitPosition
+local spGetUnitDefID = Spring.GetUnitDefID
+local spGetFeaturePosition = Spring.GetFeaturePosition
+
+local function DebugEcho(...)
+	spEcho("[unit_staticcon_unsticker]", unpack(arg))
```
GoogleFrog, line 52: This simply crashes.

PR #4692, opened 2022-05-14
Fix custom formations error caused by selecting and giving orders to units in fog

Arch-Shaman:
This is more developer QoL than anything and is a pretty simple fix anyways, so thought I'd pr it.


Arch-Shaman:
[t=00:02:50.344104][f=0001163] [cawidgets.lua] Error: Error in CommandsChanged(): [string "LuaUI/Widgets/cmd_customformations2.lua"]:993: attempt to index local 'ud' (a nil value)
[t=00:02:50.344180][f=0001163] [cawidgets.lua] Error: Removed widget: CustomFormations2

This is the error this avoids, to be exact. It has been annoying me for a while when doing FW development tests.

PR #4693, opened 2022-05-14
Add Highlight Unit (under cursor) widget from BAR

completesomething:
I noticed that while Zero-K has the Highlight Selected Units widget, it does not have the probably more useful Highlight Unit widget (highlights units under cursor) that BAR does. This pull request adds https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_highlight_unit_gl4.lua to Zero-K, except disabled by default.


GoogleFrog:
Don't change the names of existing widgets. This PR will cause people to have unit highlight enabled due to naming conflicts.

completesomething:
Alright, I renamed it back. Just an FYI not necessarily related to the PR, but the "Highlight Selected Units" widget in BAR seems to be less overbearing in that it basically only brightens up the units instead of giving the whole animated look.

GoogleFrog:
You can do anything you like to the highlight unit widget. I just added it as a test during adding reclaim highlight.

Issue #3338, opened 2018-11-15
Cyclops will not fire its' main weapon against elevated targets


Godde:
https://imgur.com/j7560Fl.png
Cyclops will move into range and fire its' slow beam but not far enough to hit it with its' projectile weapon.

PR #4698, opened 2022-05-25
Fixing Cyclops futily slow beaming a stinger by instead making it inch forwards

danfireman:
Fixes https://github.com/ZeroK-RTS/Zero-K/issues/3338

### DIFF
--- a/LuaRules/Gadgets/unit_cyclops_range_fix.lua
+++ b/LuaRules/Gadgets/unit_cyclops_range_fix.lua
@@ -1,0 +1,68 @@
+function gadget:GetInfo()
+    return {
+        name    = "Cyclops attack move fixing",
+        desc    = "Makes Cyclops move closer when it can't hit with canon",
+        author  = "dyth68",
+        date    = "2022-05-12",
+        license = "Public Domain",
+        layer   = 83,
+        enabled = true
+    }
+end
+
+local cyclopsi = {}
+local cyclopsDefID = UnitDefNames["tankheavyassault"].id
+local spGetUnitWeaponTestRange = Spring.GetUnitWeaponTestRange
+local spGetUnitWeaponTarget    = Spring.GetUnitWeaponTarget
+local spGiveOrderToUnit        = Spring.GiveOrderToUnit
+local spGetUnitPosition        = Spring.GetUnitPosition
+local spGetTeamUnitsByDefs     = Spring.GetTeamUnitsByDefs
+local Echo                     = Spring.Echo
+local spGetUnitCommands        = Spring.GetUnitCommands
+
+
+function gadget:UnitCreated(unitID, unitDefID, teamId)
+	if unitDefID == cyclopsDefID then
+		cyclopsi[unitID] = true
+	end
+end
+
+
+function gadget:UnitDestroyed(unitID, unitDefID, teamId)
+	if unitDefID == cyclopsDefID then
+		cyclopsi[unitID] = nil
+	end
+end
+
+function gadget:UnitGiven(unitID, unitDefID, newTeamID, teamID)
+	if unitDefID == cyclopsDefID then
+		cyclopsi[unitID] = true
+	end
+end
+
+function gadget:Initialize()
+	for _,teamID in ipairs(Spring.GetTeamList()) do
+		for _, unitID in ipairs(spGetTeamUnitsByDefs(teamID, cyclopsDefID)) do
+			cyclopsi[unitID] = true
+		end
+	end
+end
+
+function gadget:GameFrame(frame)
+	for cyclopsId, _ in pairs(cyclopsi) do
+		local _, _, canontargetID = spGetUnitWeaponTarget (cyclopsId, 1)
+		local targetType, _, slowBeamTargetID = spGetUnitWeaponTarget (cyclopsId, 2)
+		if (targetType == 1) and (canontargetID == nil) then
+			--Echo("Retargetting Cyclops")
+			local x, y, z = spGetUnitPosition(cyclopsId)
+			local tx, ty, tz = spGetUnitPosition(slowBeamTargetID)
+			local nx, nz = x * 0.95 + tx * 0.05, z * 0.95 + tz * 0.05
+			local ny = math.max(0, Spring.GetGroundHeight(nx, nz))
+			local cmd1 = spGetUnitCommands(cyclopsId, 1)[1]
+			if cmd1 and cmd1.options.internal then
+				spGiveOrderToUnit(cyclopsId, CMD.REMOVE, {1, cmd1.tag}, {"ctrl"})
+			end
+			spGiveOrderToUnit(cyclopsId, CMD.INSERT, {0, CMD.MOVE, CMD.OPT_INTERNAL, nx, ny, nz}, CMD.OPT_ALT)
+		end
+	end
+end


GoogleFrog:
I feel like the engine is meant to have something that sets which weapon is for aiming. Failing that, how much range does slowbeam need to lose for the main gun to fire at reasonable elevation differences? If the fix is as simple as 20 less slowbeam range, then perhaps that's the one to go for.

danfireman:
After a bit of looking I couldn't find any way to use the engine for this.

The slow beam needs to lose about 60 range before a modest slope or a hill 1.5 lotus high prevents it from just sitting there.
However, this is not a viable solution as losing any range at all means a Cyclops on attack move won't move into slow beam range against stingers or other static things like a afk paladin.

Shall I go ahead with my TODOs above on the assumption this will be merged?

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+        author  = "dyth68",
+        date    = "2022-05-12",
+        license = "Public Domain",
+        layer   = 83,
+        enabled = true
+    }
+end
+
+local cyclopsi = {}
+local cyclopsDefID = UnitDefNames["tankheavyassault"].id
```
GoogleFrog, line 14: Ideally this would be a customParam so that people could make similar units use the same behaviour, or so that modders could easily remove this behaviour if they're using cyclops for a different purpose.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+local spGetUnitWeaponTestRange = Spring.GetUnitWeaponTestRange
+local spGetUnitWeaponTarget    = Spring.GetUnitWeaponTarget
+local spGiveOrderToUnit        = Spring.GiveOrderToUnit
+local spGetUnitPosition        = Spring.GetUnitPosition
+local spGetTeamUnitsByDefs     = Spring.GetTeamUnitsByDefs
+local Echo                     = Spring.Echo
```
GoogleFrog, line 20: This should at least be spEcho, although in practice I never localise Spring.Echo.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+end
+
+
+function gadget:UnitDestroyed(unitID, unitDefID, teamId)
+	if unitDefID == cyclopsDefID then
+		cyclopsi[unitID] = nil
+	end
+end
+
+function gadget:UnitGiven(unitID, unitDefID, newTeamID, teamID)
```
GoogleFrog, line 37: Unit given is redundant if the behaviour is completely team agnostic.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+	if unitDefID == cyclopsDefID then
+		cyclopsi[unitID] = true
+	end
+end
+
+function gadget:Initialize()
```
GoogleFrog, line 43: I like to put Init at the bottom of the gadget, with create/destroy above, with more frequent logic and callins above that.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+	for _,teamID in ipairs(Spring.GetTeamList()) do
+		for _, unitID in ipairs(spGetTeamUnitsByDefs(teamID, cyclopsDefID)) do
+			cyclopsi[unitID] = true
+		end
+	end
+end
+
+function gadget:GameFrame(frame)
+	for cyclopsId, _ in pairs(cyclopsi) do
```
GoogleFrog, line 52: Ideally pairs would not be used. IterableMap with ApplySelf fits. It would even cleanly implement my next suggestion, which is that cyclops probably doesn't need to be checked every frame, and that we may as well stagger the checks over frames.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+			local x, y, z = spGetUnitPosition(cyclopsId)
+			local tx, ty, tz = spGetUnitPosition(slowBeamTargetID)
+			local nx, nz = x * 0.95 + tx * 0.05, z * 0.95 + tz * 0.05
+			local ny = math.max(0, Spring.GetGroundHeight(nx, nz))
+			local cmd1 = spGetUnitCommands(cyclopsId, 1)[1]
+			if cmd1 and cmd1.options.internal then
+				spGiveOrderToUnit(cyclopsId, CMD.REMOVE, {1, cmd1.tag}, {"ctrl"})
+			end
+			spGiveOrderToUnit(cyclopsId, CMD.INSERT, {0, CMD.MOVE, CMD.OPT_INTERNAL, nx, ny, nz}, CMD.OPT_ALT)
+		end
```
GoogleFrog, line 66: Why are commands being removed and move commands inserted? Firstly, this should probably be CMD_RAW_MOVE. But secondly, does this need to be a command at all? It _might_ be neater to implement this behaviour using a short move goal.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: Switched

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: Done

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+local spGetUnitWeaponTarget    = Spring.GetUnitWeaponTarget
+local spGiveOrderToUnit        = Spring.GiveOrderToUnit
+local spGetUnitPosition        = Spring.GetUnitPosition
+local spGetTeamUnitsByDefs     = Spring.GetTeamUnitsByDefs
+local Echo                     = Spring.Echo
+local spGetUnitCommands        = Spring.GetUnitCommands
+
+
+function gadget:UnitCreated(unitID, unitDefID, teamId)
+	if unitDefID == cyclopsDefID then
```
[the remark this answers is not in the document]
danfireman: Done

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: Removed

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: Moved

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: "IterableMap with ApplySelf " <- no idea what this is

I'd say you want to check the Cyclops often enough to make the inching forwards fairly smooth, so probably not less than once every 3 frames. Will wait to know more about this "ApplySelf" business before making a change.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
```diff
@@ -0,0 +1,68 @@
...
+	for cyclopsId, _ in pairs(cyclopsi) do
+		local _, _, canontargetID = spGetUnitWeaponTarget (cyclopsId, 1)
+		local targetType, _, slowBeamTargetID = spGetUnitWeaponTarget (cyclopsId, 2)
+		if (targetType == 1) and (canontargetID == nil) then
+			--Echo("Retargetting Cyclops")
+			local x, y, z = spGetUnitPosition(cyclopsId)
+			local tx, ty, tz = spGetUnitPosition(slowBeamTargetID)
+			local nx, nz = x * 0.95 + tx * 0.05, z * 0.95 + tz * 0.05
+			local ny = math.max(0, Spring.GetGroundHeight(nx, nz))
+			local cmd1 = spGetUnitCommands(cyclopsId, 1)[1]
```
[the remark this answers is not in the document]
danfireman: Added check for not having commands.

Not idea what you're referring to by the second sentence. Can't see anything relevant except GetUnitCommands at https://springrts.com/wiki/Lua_SyncedRead#CommandQueues

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: The intent is to remove any existing inch-forwards command (otherwise you get a whole bunch of them and they mess things up).

Using CMD_RAW_MOVE seems to result in the Cyclops not inching.

No idea what a short move goal is.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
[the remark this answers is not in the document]
danfireman: Done

LuaRules/Gadgets/unit_cyclops_range_fix.lua
GoogleFrog: Grep iterableMap.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
GoogleFrog: Move goals are direct orders for units to move somewhere. They sit below the command system. They can be neater for unit AI because they don't mess with commands at all.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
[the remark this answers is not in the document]
sprunk: `Spring.GetUnitCurrentCommand`, search the zk repo for examples (it's not on wiki)

LuaRules/Gadgets/unit_cyclops_range_fix.lua
danfireman: That's not very helpful to me actually implementing it.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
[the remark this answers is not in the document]
danfireman: Errors when I try and use this.

LuaRules/Gadgets/unit_cyclops_range_fix.lua
sprunk: `Spring.SetUnitMoveGoal(unitID, x, y, z, radius)` makes the unit move towards given point (and be satisfied once in given radius)

GoogleFrog:
image
I tried to move a Cyclops through Crab range and it gave itself a small move order and became stuck. The units in this picture are immobile.

image
This time I spawned a Cyclops in range and tried to move away. It got to the point where it could no longer fire, then got stuck on a small move order towards the Crab.

image
Moving a Cyclops away from a Nimbus is impossible.

GoogleFrog:
I don't see any part of the code that would deal with these cases.

> "IterableMap with ApplySelf " <- no idea what this is
Grep for examples. Eg
image

image
Grep for examples again. Other gadgets use it. Experiment a bit.

PR #4710, opened 2022-06-18
Add widow to unit marker

Mach565:
What the title says.
As an aside, would setting the cloaked striders like ultimatum and scorp to "MARK_EACH" be a worthwhile change too? Only marking them once could be annoying if you've already marked them once before, but they showed up again somewhere you weren't paying attention to.


sprunk:
> would setting the cloaked striders like ultimatum and scorp to "MARK_EACH" be a worthwhile change too?

sounds good bc5ed215d27abfe64090dfa9918dd88e0b917f59

PR #4715, opened 2022-06-23
Add "Hold Ctrl during placement to morph" widget

jklw:
Adds the ability to hold Ctrl while placing a Cornea, Aegis, Radar Tower or Geothermal Reactor to issue a morph order to the nanoframe when it's created (which, in turn, will make it start morphing once construction has finished).

Works with 'repeat' state on con, allowing convenient mass production of aspis without a shield fac.

[Forum thread](https://zero-k.info/Forum/Thread/35147)

I've been using this for a month now and found it quite useful.

### DIFF
--- a/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
+++ b/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
@@ -1,8 +1,6 @@
-local widgetName = "Hold Ctrl during placement to morph";
-
 function widget:GetInfo()
 	return {
-		name      = widgetName,
+		name      = "Hold Ctrl during placement to morph",
 		version   = '1.0',
 		desc      = "Hold the Ctrl key while placing a Cornea, Aegis, Radar Tower or Geothermal Reactor to issue a morph order to the nanoframe when it's created (which will make it start morphing once finished).",
 		author    = "dunno",
--- a/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
+++ b/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
@@ -112,10 +112,7 @@ function widget:UnitCreated(unitID, unitDefID, unitTeam, builderID)
 end
 
 function widget:UnitDestroyed(unitID, unitDefID, unitTeam)
-	if buildingsToMorphByBuilder[unitID] then
-		-- Spring.Echo('Clearing buildingsToMorph for builder '..unitID..' because it was destroyed.')
-		buildingsToMorphByBuilder[unitID] = nil
-	end
+	buildingsToMorphByBuilder[unitID] = nil
 end
 
 function widget:UnitIdle(unitID, unitDefID, unitTeam)
--- a/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
+++ b/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
@@ -40,6 +40,7 @@ local buildingsToMorphByBuilder = {}
 function widget:Initialize()
 	myTeamID = Spring.GetMyTeamID()
 end
+widget.PlayerChanged = widget.Initialize
 
 function widget:UnitCommand(unitID, unitDefID, unitTeam, cmdId, cmdParams, cmdOpts, cmdTag)
 	if cmdId > 0 
--- a/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
+++ b/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
@@ -19,7 +19,8 @@ end
 ---@return table<UnitDefId, boolean>
 local function CreateUnitDefIdSet(names)
 	local result = {}
-	for _, name in ipairs(names) do
+	for i = 1, #names do
+		local name = names[i]
 		result[UnitDefNames[name].id] = true
 	end
 	return result
--- a/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
+++ b/LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
@@ -1,0 +1,126 @@
+local widgetName = "Hold Ctrl during placement to morph";
+
+function widget:GetInfo()
+	return {
+		name      = widgetName,
+		version   = '1.0',
+		desc      = "Hold the Ctrl key while placing a Cornea, Aegis, Radar Tower or Geothermal Reactor to issue a morph order to the nanoframe when it's created (which will make it start morphing once finished).",
+		author    = "dunno",
+		date      = "2022-05-18",
+		license   = "MIT",
+		layer     = 0,
+		enabled   = true
+	}
+end
+
+---@alias UnitInternalName string
+---@alias UnitDefId integer
+---@alias UnitId integer
+
+---@param names UnitInternalName[]
+---@return table<UnitDefId, boolean>
+local function CreateUnitDefIdSet(names)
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
+local widgetName = "Hold Ctrl during placement to morph";
+
+function widget:GetInfo()
+	return {
+		name      = widgetName,
```
sprunk, line 5: No need for an extra var.
```suggestion
function widget:GetInfo()
	return {
		name      = "Hold Ctrl during placement to morph",
```

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+	end
+
+	-- Spring.Echo('This building is not to be morphed!')
+end
+
+function widget:UnitDestroyed(unitID, unitDefID, unitTeam)
+	if buildingsToMorphByBuilder[unitID] then
+		-- Spring.Echo('Clearing buildingsToMorph for builder '..unitID..' because it was destroyed.')
+		buildingsToMorphByBuilder[unitID] = nil
+	end
```
sprunk, line 118: Avoiding the needless assignment is as slow as the assignment itself, so offers no benefit and doubles the work in case the value wasn't already `nil`.
```suggestion
	buildingsToMorphByBuilder[unitID] = nil
```

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+end
+
+function widget:UnitIdle(unitID, unitDefID, unitTeam)
+	if buildingsToMorphByBuilder[unitID] then
+		-- Spring.Echo('Clearing buildingsToMorph for builder '..unitID..' because it is idle.')
+		buildingsToMorphByBuilder[unitID] = nil
+	end
```
sprunk, line 125: Ditto
```suggestion
	buildingsToMorphByBuilder[unitID] = nil
```

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+		version   = '1.0',
+		desc      = "Hold the Ctrl key while placing a Cornea, Aegis, Radar Tower or Geothermal Reactor to issue a morph order to the nanoframe when it's created (which will make it start morphing once finished).",
+		author    = "dunno",
+		date      = "2022-05-18",
+		license   = "MIT",
+		layer     = 0,
+		enabled   = true
+	}
+end
+
```
sprunk, line 15: Ideally this would be an epicmenu option.

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+---@alias UnitInternalName string
+---@alias UnitDefId integer
+---@alias UnitId integer
+
+---@param names UnitInternalName[]
+---@return table<UnitDefId, boolean>
+local function CreateUnitDefIdSet(names)
+	local result = {}
+	for _, name in ipairs(names) do
```
sprunk, line 24: Avoid `ipairs`, it's slow compared to regular iteration (in this case it doesn't really matter since it's just 4 entries in the table but maybe someone else would copy-paste code in the future):
```suggestion
	for i = 1, #names do
		local name = names[i]
```

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+
+local morphableUnitDefIds = CreateUnitDefIdSet({'staticjammer', 'staticshield', 'staticradar', 'energygeo'})
+
+---@type table<UnitId,{x:number,z:number}[]>
+local buildingsToMorphByBuilder = {}
+
+function widget:Initialize()
+	myTeamID = Spring.GetMyTeamID()
+end
+
```
sprunk, line 44: Handle changing team (debugging with `/team`, comm sharing, etc):
```suggestion
widget.PlayerChanged = widget.Initialize

```

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+	-- Spring.Echo('UnitCreated(unitDef = '..UnitDefs[unitDefID].name..', builderID = '..(builderID or 'nil')..', x = '..ux..', z = '..uz..')')
+
+	for i, point2 in pairs(buildingsToMorph) do
+		if (point2.x) then
+
+			-- Note: unit_building_starter.lua, which does something similar, uses a location tolerance of 16
+			-- here. But this appears to be unnecessary for the morphable buildings considered here(?)
+			-- (in fact, it workers with exact equality, but being defensive)
+
+			if abs(point2.x - ux) < 1e-3 and abs(point2.z - uz) < 1e-3 then
```
sprunk, line 102: Would this work well with morphable mobiles built by Athena if they were handled by the widget?

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
```diff
@@ -0,0 +1,126 @@
...
+			buildingsToMorph = {}
+			buildingsToMorphByBuilder[unitID] = buildingsToMorph
+		end
+		table.insert(buildingsToMorph, point)
+
+		-- Spring.Echo('Building at '..show(point).. ' marked as to be morphed when placed by builder '..unitID)
+
+	elseif buildingsToMorph then
+		for i, point2 in pairs(buildingsToMorph) do
+			if (point2.x) then
```
sprunk, line 68: Consider checking whether `cmdParams[1]` and `cmdParams[3]` are not nil at the beginning, then this check can be dropped.

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
jklw: This was intended as a prefix for debug echos, but not actually used.

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
jklw: I don't understand? The purpose of the `elseif` branch is a scenario like this:

1. Place a Cornea while holding ctrl
2. ...
3. Before the Cornea has been created or the builder has become idle, place a Cornea or another morphable building on the same spot as in 1, but without holding ctrl

In this case, no morphing should happen.

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
jklw: I left it default-enabled undeliberately. It's a single-function widget, but I suppose an epic option would be more discoverable if we want the functionality of the widget to be disabled by default.

LuaUI/Widgets/cmd_hold_ctrl_to_place_and_morph.lua
sprunk: I mean the `if (point2.x) then`. Putting points where `x` value is `nil` into the `buildingsToMorph` table can be avoided since these are probably bogus orders anyway.

Issue #4704, opened 2022-05-30
luaerror in unit_proximity_targetting


Arch-Shaman:
[t=00:32:35.715537][f=0063510] [Game::ClientReadNet][LOGMSG] sender="hoaqhieu08" string="[Internal Lua error: Call failure] [string "LuaRules/Gadgets/unit_proximity_targetting...."]:55: attempt to perform arithmetic on local 'x2' (a nil value)
stack traceback:
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:55: in function 'DistanceSq'
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:79: in function 'GetTargetToClosest'
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:122: in function 'funcToApply'
	[string "LuaRules/Gadgets/Include/IterableMap.lua"]:131: in function 'Apply'
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:169: in function 'GameFrame'
	[string "LuaRules/gadgets.lua"]:967: in function <[string "LuaRules/gadgets.lua"]:965>
	(tail call): ?"

See: https://zero-k.info/Battles/Detail/1376633 (I believe this is related to the recent starlight targeting pr)

danfireman:
So there's a unit which is in line of sight but has a nil position coordinate? 🤯

johnsonjsyuen:
![image](https://user-images.githubusercontent.com/44350915/171065821-51ab379a-b519-412f-a785-118174221f39.png)

mushroomraider:
here is the error from my log:


[t=02:28:47.772173][f=0046596] [Game::ClientReadNet][LOGMSG] sender="Xamoguitar" string="[Internal Lua error: Call failure] [string "LuaRules/Gadgets/unit_proximity_targetting...."]:55: attempt to perform arithmetic on local 'x2' (a nil value)
stack traceback:
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:55: in function 'DistanceSq'
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:79: in function 'GetTargetToClosest'
	[string "LuaRules/Gadgets/unit_proximity_targetting...."]:122: in function 'funcToApply'
	[string "LuaRules/Gadgets/Include/IterableMap.lua"]:131: in function 'Apply'
[...]

PR #4716, opened 2022-06-23
Hacky fix for starlight targetting

danfireman:
Fixes https://github.com/ZeroK-RTS/Zero-K/issues/4704

Couldn't figure out WHY it was happening, so here's a workaround.


sprunk:
>Couldn't figure out WHY

Try this setup:
image

PR #4718, opened 2022-06-28
Rosetta metal map fix

Arch-Shaman:
Normalize mex values to 2.0 and set a reasonable metal value for the supermexes based on Dave[tB]'s suggestions.


SteelBlueZK:
noooo, it was really fun having 40m/s in a 14v14!

sprunk:
Maybe Moose intended +0.7 mexes to exist so that people would have a tempo choice to make, and the +10 mexes to add some volatility to what is otherwise a slogfest. Or maybe he was just balancing for BAR where these are sensible balues. I can't read minds though so go ask him.

PR #4733, opened 2022-07-26
Optimize unit_morph gadget

Porkch0p:
Add caching to morphing units list, remove an extra synced data access per draw frame


Porkch0p:
In replays, killing a morph-in-progress commander appeared to repeatably improve frame rates... Looking into it, there does appear to be some low-hanging fruit in the gadget. This change reduces the allocation rate to about a third of the original (at 60fps).

sprunk:
Looks good. Ideally changes in that list would just become Add/Remove events since there's ever maybe like 3 things morphing at once.

PR #4734, opened 2022-07-26
Support scancodes and engine passed actions

badosu:


badosu:
Closed for lack of review

GoogleFrog:
what?

Licho1:
Please reopen the PR @badosu it looks useful if engine really sends it.
We just don't have much time to review and test some changes. 
It will get looked into.

badosu:
@Licho1 Scancodes have been supported for a while already, the change here adds handling for engine passed actions on 13xx+. Should be retrocompatible with older versions. The advantage of using engine passed actions is fixing keychain bindings for lua actions (e.g. `bind m,m say hi` says hi on double tapping m) and avoiding reconstructing keychains from provided keycodes (a very finicky and hacky method due to unplanned lua action implementation by original authors)

badosu:
There should be a performance improvement but that's not significant at this runtime section

badosu:
If your deployed engine version is 9xx+ scancodes should work, I removed the retrocompatibility for older versions so keep this in mind before merging.

badosu:
A minor consideration engine passed actions allow is handling action chain halt with whitelists or blacklists by widgets receiving actions on keypress/keyrelease/actionreceiver.

Licho1 (approved):

PR #4741, opened 2022-08-14
Persistent Build Spacing - Big update, various improvements and fixes, slight rewrite

Helwor:
-Adapted my code to be compatible with the Zero-K version

-Rewrote slightly the code to be more concise and efficient

-Fixed: The Zero-K version is broken since an edit made Sep 2020 (!) when user start to change spacing, the rectangles would never disappear because newspacing variable was never falsified at the end of the  Update round.

-Fixed: the param for spTraceScreenRay was the opposite of what it should be

-Improvement: The widget now adapt to the wrong engine grid which is off and also sometimes misleading when it comes to place units on water

-Added: On this particular case, widget gives the possibility for the user to either follow the grid coherence, or show the reality of where the unit will actually be placed. In that case, an extra rect is drawn to see where the first placement will land.  By usage, it seems the latter is more interesting so I put it on by default.
(More details in the comments that could help fix the problem with the engine)

-Improvement: the option panel now stop resetting scroll anytime an option is changed, this fix could easily be implemented universally by changing WG.crude.OpenPath, all it have to do is remembering 'scrollPosY' of the last panel and apply it to the new.

-Added: an option  for the user to choose wether rects should spread straightforward (if user want to only have a basic glance of the separation with no simulation of placing on the ground) or follow the ground, as it was before. I'm undecided which is the best as default for new users.

-Added: an option to have different colors of rects for bad placements, this on default if follow ground is on default, it seems to be adding a plus mostly in this case.

-Changed Description that's been edited, it does not just memorize spacing. I understand that it need to be concise, but a description is meant also to say what it does. User won't be able to tell if that widget make the previsualization or add mouse wheel to spacing by reading its description.

-Changed back title of an option. The purpose of how they are written is to build a sentence obviously, to make it more straightforward and user friendly. This change broke that concept. When user, especially new user, dive into options they can be lost as I was, to understand what option do what, if the option is dependent of another, etc... that's the purpose of my system to make all that as clear as I can.

-Changed back some default for new user, I think having the shift+mousewheel is a must, it's faster and more convenient, especially when discovering the game. I would have been glad to have that when I started, (that's, by the way, why I rewrote this widget in the first place).
Also changed back previsualization, with only 2 rects and only when spacing change, and for 0.7 sec
It is, in my opinion, also much more convenient for a new user to be able to see in a quick glance how your placement will be separated without the need of start posing queue and THEN realize the spacing is not good. If non-new user get tired of it, they will be able to find the options to undo it.
Indeed, new user will not be anymore really a new user when he discovers previsualization, if it's not on by default.

-Few minor things:
  -preGame is now set by GameFrame and then GameFrame removes itself
  -MouseWheel gets added/removed whenever shift+wheel option is changed
  -I don't see any reason to change the name spaced_rects in spaced_rects_2. As far as I know, one can only want that for resetting his own option to default for testing  purpose. Reverted it since it didn't make sense.
  -Upped the max time before infinite for show value and show rects to 10 sec
  -Didn't reverted it but imho, changing 'p' to 'placement' made the code uselessly more cumbersome than it already is.
  -Changed 'if not placementCache[unitDefID] then'  to 'if not placeTable[unitDefID] then' as the code was intended to mean
  note about this: since the vast majority of building have same sizex and sizez, maybe it could be improved by checking if it's relevant to consider the off-facing in 
  before.
 -Estomped the double dots before option titles when they are active, to make them less present.
 -update placement on more relevant changement of facing
 -Made the options defaults in concordance with the widget defaults
 -And maybe some other stuff I forgot

Lots of change, so please, let me know what you think.

### DIFF
--- a/LuaUI/Widgets/gui_persistent_build_spacing.lua
+++ b/LuaUI/Widgets/gui_persistent_build_spacing.lua
@@ -1,9 +1,7 @@
 function widget:GetInfo()
 	return {
 		name      = "Persistent Build Spacing",
-		desc      = "-Maintains build spacing between matches"..
-			    "\n-Add Mousewheel+Shift to control it"..
-			    "\n-Add previsualization of future placements",
+		desc      = "-Maintains build spacing between matches\n-Add Mousewheel+Shift to control it\n-Add previsualization of future placements",
 		author    = "Helwor",
 		date      = "August 2020",
 		license   = "GNU GPL, v2 or later",
@@ -27,7 +25,6 @@ local spGetGameSeconds      = Spring.GetGameSeconds
 local spGetModKeyState      = Spring.GetModKeyState
 local spuGetMoveType        = Spring.Utilities.getMovetype
 local spTestBuildOrder      = Spring.TestBuildOrder
-local Echo                  = Spring.Echo
 
 local floor, round, huge, max = math.floor, math.round, math.huge, math.max
 
@@ -65,7 +62,6 @@ local colors = {  -- red, yellow, green
 }
 -- related to options
 local grdots = "\255\155\155\155" .. " .." .. "\255\255\255\255"
-local requestUpdate
 local wheelSpacing, wheelValue = false, 1
 local showSpacingRects, only2Rects, followGround, stickToWrongGrid, showRectsOnChange
 	=      true,          true,        true,          false,             true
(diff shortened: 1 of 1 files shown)


GoogleFrog:
Was this split into commits? It looks like a lot of the changelog is just refactoring old silly stuff.

GoogleFrog:
There is a lot going on here. I got through some of it and removed the epicmenu hax. It could be implemented as an option for epicmenu itself, such as options that greys out bool options which are disabled and supply sliderbars with two strings, which epicmenu should use on either side of an updating number.

GoogleFrog:
I'll copy it locally and see what happens after some games.

GoogleFrog:
No issues after a week.

PR #4742, opened 2022-08-15
⚙️ Ronin now steps away from LLT when reloading

### LINKED ISSUE #3038: Unit AI for poking LLT
Ronin etc... could have unit AI that has them only enter LLT range when they are reloaded. Support for this exists, but not per enemy, and I think this would be best to only apply to a few targets (slow or static ones).

sprunk:
Fixes #3038 and is a major buff.


Licho1:
Make ronin OP again :)

Licho1 (approved):

PR #4754, opened 2022-08-26
Fix Pyro's flamethrower

Arch-Shaman:
Pyro's flamethrower animation doesn't fully reveal the flamethrower on the model. This PR fixes this by adjusting the position of the lower head and upper head to reveal that sweet flamethrower underneath.

image


sprunk:
Now better reveals the inner mouth flamer

PR #4759, opened 2022-09-08
Improvements to blast radius widget

### LINKED ISSUE #3524: Consider showing the big explosion radii around geothermal vents
Please deliberate making the vent display a distinct ring after the player has selected a building to warn of the explosion radius of the advanced geothermal building. The ring colors could range from yellow to red for a fiery impression. I believe an apt warning text should be clearly visible near the ring that might link to the circle through a line for example. If you are choosing the location of a structure, its color could even keep shifting between yellow and red while it's inside the ring.

Arch-Shaman:
* Added Geothermal
* Units with morphs that get added to the always visible table now also show their morph's death explosion size and damage.

Preview:
image

image

Fixes #3524.

### DIFF
--- a/LuaUI/Widgets/gui_blastRadius.lua
+++ b/LuaUI/Widgets/gui_blastRadius.lua
@@ -1,5 +1,5 @@
 include("keysym.lua")
-local versionNumber = "1.2"
+local versionNumber = "1.1"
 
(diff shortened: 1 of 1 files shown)


Arch-Shaman:
acaff9d preview:

image

PR #4761, opened 2022-09-17
Fix caretaker guard orders

Arch-Shaman:
The caretaker unsticker gadget assumes all commands are reclaim or resurrect.

I haven't tested this in the wild via FW yet, but lab testing shows caretakers no longer drop guard commands after https://github.com/Arch-Shaman/ZK-Futurewars-Mod/commit/6f94610f25c60d3fe09808e4b666bc7937cd12af


sprunk:
Wasn't half the point to have repair and guard be dropped as well though?

Arch-Shaman:
Guard is dropped regardless of range in the current code. This can lead to
poor ux.

la 17. syysk. 2022 klo 21.09 sprunk ***@***.***> kirjoitti:

PR #4763, opened 2022-09-23
fix paletted images

Beherith:


sprunk:
image
There seems to be a subtle difference, the old one is RGB 0, 31, 3 while the new one is 0, 25, 0. Did the transformation somehow break?

PR #4780, opened 2022-11-22
✨ Krow lasers now look like Redback

sprunk:
How about this? After the recent buffs it now has laser closer (in power, role, range, damage) to somewhere between Mace and Commander, but it still uses the red pew visuals, which is used for weapons with roles all over the place (Bandit's low intensity pew, Razor's AA pew, Krow's powerful pew). It's also somewhat hard to see where the shots are going given how fast they move.

Sad effects:
 * loses unique pew sound. On the other hand hearing Mace is also (already) a "shit's melting, yo" audio cue so maybe it's fine?
 * minor balance change, probably mostly vs Swifts or Fleas. Probably doesn't matter at this weight.
 * balance change was already stabled so it's a bit late.

image

### DIFF
--- a/scripts/gunshipkrow.lua
+++ b/scripts/gunshipkrow.lua
@@ -111,7 +111,7 @@ local function DeathAnim()
 				EmitSfx(jets[i], 258)
 			end
 			if count % 18 == 0 then
-				EmitSfx(jets[index], 1025)
+				EmitSfx(jets[index], 1024)
 				index = index + 1
 				if index > #jets then
 					index = 1
@@ -318,13 +318,6 @@ function script.AimWeapon(num, heading, pitch)
 	StartThread(RestoreAfterDelay)
 	return true
 end
-
-function script.FireWeapon(num)
-	if num ~= 3 then
-		EmitSfx(gunpoints[num].fire, 1024)
-	end
-end
-
 
 function script.Killed(recentDamage, maxHealth)
 	local severity = recentDamage/maxHealth
--- a/units/gunshipkrow.lua
+++ b/units/gunshipkrow.lua
@@ -46,7 +46,6 @@ return { gunshipkrow = {
   sfxtypes               = {
 
     explosiongenerators = {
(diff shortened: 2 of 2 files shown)


Anarchid:
I think it is possible to keep the pew sound if rather than being a constant beam it's something like a rapid-fire Stinger, with downtime. I think a nonzero beamtime weapon will also look more epic on a strider-level gunship, but hopefully not too janky if it's short (duration about as long as pew sound, reload as long as previous reload).

Anarchid:
https://user-images.githubusercontent.com/3822768/203450566-55039c94-2381-4636-bf61-3fe83eeee6b7.mp4

sprunk:
Perhaps Redback? Stats-wise it's even closer than Mace/Commander and it retains the redness and the discreteness. On the other hand, it goes in the other direction compared to spray/stream (reload 0.2 → 0.3 as opposed to 0.2 → 0.1) and Redback sound doesn't carry as much danger as Mace.
image

GoogleFrog:
> 0001-0551.mp4

This looks like a good approach. It sounds scary, the flicker dodges the smooth aiming problem, and the beam time can be used to tune its effective DPS against fast aircraft. The main balance concern of a Redback laser is it suddenly becomes 100% accurate against planes. I don't know how accurate current Krow is against planes, but until it is tested it's a concern.

Someone should run that test. Where is green HLT Krow?

Anarchid:
> Someone should run that test. Where is green HLT Krow?

Looks like i have a local branch for it, but it contains no commits. The local tree has long been overwritten. It is lost and will have to be redone.

sprunk:
> The main balance concern of a Redback laser is it suddenly becomes 100% accurate against planes. I don't know how accurate current Krow is against planes, but until it is tested it's a concern.

I tested and was unable to get current Krow below 100% hit rate against aircraft (normal Swift, Likho, Thunderbird, Gnat, Blastwing). I managed to get Krow to miss a shot by using Swift boost.

sprunk:
Included in #4975

PR #4785, opened 2022-11-28
Add support for MiniMapCanFlip

badosu:
Remarks:

- `Gadgets/unit_cloak_shield.lua` callin for `DrawInMiniMap` seems to be noop. Disabling the gadget does not change anything, the range is still drawn correctly, flipped or not.

Untested widgets:

- `Widgets/minimap_events.lua`

These should be safe when minimap cant flip or is not flipped, otherwise they follow same structure for drawing as other tested ones.

All changes, unless stated otherwise above, were tested.

### DIFF
--- a/LuaUI/Headers/minimap_utilities.lua
+++ b/LuaUI/Headers/minimap_utilities.lua
@@ -1,0 +1,12 @@
+local function getMiniMapFlipped()
+	if ((not Spring.GetMiniMapRotation) or
+	    (Spring.GetConfigInt("MiniMapCanFlip", 0) == 0)) then
+		return false
+	end
+
+	local rot = Spring.GetMiniMapRotation()
+
+	return rot > math.pi/2 and rot <= 3 * math.pi/2;
+end
+
+return { getMiniMapFlipped = getMiniMapFlipped }
--- a/LuaUI/Widgets/api_set_springsettings.lua
+++ b/LuaUI/Widgets/api_set_springsettings.lua
@@ -12,6 +12,7 @@ function widget:GetInfo()
 end
 
 function widget:Initialize()
+	Spring.SetConfigInt("MiniMapCanFlip", 1)
 	Spring.SetConfigInt("RotateLogFiles", 1)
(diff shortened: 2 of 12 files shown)


LuaUI/Headers/minimap_utilities.lua
```diff
@@ -0,0 +1,12 @@
+local function getMiniMapFlipped()
+	if ((not Spring.GetMiniMapRotation) or
+	    (Spring.GetConfigInt("MiniMapCanFlip", 0) == 0)) then
```
sprunk, line 3: Is the config check needed? I think if this config value is set to false then `Spring.GetMiniMapRotation` should always return 0.

badosu: Not necessary probably, I just like to decouple from specific implementation cases

badosu: If you send the suggestion I can incorporate it

sprunk:
`minimap_startbox_new.lua` seems not to work. I think this is because of `gl.CallList(boxMinimapList)` which is above the flip check and transforms.
screen00055
screen00056

sprunk:
Here's the startscript I use if you need to reproduce (drag'n'drop onto spring.exe, should work with any other map as well) [script.txt](https://github.com/ZeroK-RTS/Zero-K/files/10200375/script_pentos.txt)

badosu:
Thank you! Let me reproduce

sprunk:
@badosu have you done anything in this area? I'd like to look into it myself but wouldn't want to duplicate work.

badosu:
I'm trying to run Zero-K from skylobby for the startbox start but it seems to not work (it can't find "Zero-K $VERSION").

If you know how to test this scenario running directly from the engine binary I can address it.

badosu:
@sprunk This should be addressed now

badosu:
I'm stupid, didn't check again that you actually shared the startscript previously.

sprunk:
Thanks!

PR #4786, opened 2022-11-29
Gunships no longer teleport units after crashing

### LINKED ISSUE #4784: Apparently transports teleport units when they die
https://zero-k.info/Battles/Detail/1501517

Arch-Shaman:
Fixes #4784

Gunships that are in the "crashing" state do not call their "killed" callin until they smack the ground. This causes the transportee to teleport.

Tested via [Future Wars 242eaa2](https://github.com/Arch-Shaman/ZK-Futurewars-Mod/commit/242eaa29c65967f717f39af87fe83115e5895e1f) using a special unitdef to cause the gunship to crash reliably by dealing low damage. (You can test this by creating a razor with a laser weapon and deals around 1 damage every 1/30 second.)

### DIFF
--- a/LuaRules/Gadgets/unit_aircraft_crash.lua
+++ b/LuaRules/Gadgets/unit_aircraft_crash.lua
@@ -25,6 +25,7 @@ local spGetUnitIsStunned	= Spring.GetUnitIsStunned
 local spGetUnitHealth		= Spring.GetUnitHealth
 
 local aircraftDefIDs = {}
+local emptyTable = {}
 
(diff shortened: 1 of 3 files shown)


sprunk:
Polling sounds bad. Make the gadget that triggers crashing drop the unit at the time instead.

GoogleFrog:
Is there a gadget that triggers crashing? I like how these are entirely in LUS rather than an associated gadget. Why not poll 5 times a second rather than 30? If gadgets control crashing then go ahead with an event though.

PR #4787, opened 2022-12-04
Add Contrast Adaptive Sharpen widget

badosu:
Review is welcome so code adheres to Zero-K standards.

## Before

2022-12-03T23:51:46,648767007-03:00

## After

2022-12-03T23:51:28,032739402-03:00

### DIFF
--- a/LuaUI/Widgets/Shaders/cas.frag.glsl
+++ b/LuaUI/Widgets/Shaders/cas.frag.glsl
@@ -1,0 +1,67 @@
+#version 330
+#line 20058
+
+uniform sampler2D screenCopyTex;
+uniform float sharpness;
+
+#if 0 // in case AMD drivers refuse to compile the shader, though according to GLSL spec they shouldn't
+	#define TEXEL_FETCH_OFFSET(t, c, l, o) texelFetch(t, c + o, l)
+#else
+	#define TEXEL_FETCH_OFFSET texelFetchOffset
+#endif
+
+in vec2 viewPos;
+out vec4 fragColor;
+
+vec3 CASPass(ivec2 tc) {
+	// fetch a 3x3 neighborhood around the pixel 'e',
+	//  a b c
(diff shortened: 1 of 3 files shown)


LuaUI/Widgets/Shaders/cas.frag.glsl
```diff
@@ -0,0 +1,67 @@
+#version 330
+#line 20058
+
+uniform sampler2D screenCopyTex;
+uniform float sharpness;
+
+#if 0
```
sprunk, line 7:
```suggestion
#if 0 // the trade-off is X so use this in circumstances Y
```

badosu: No idea, would have to ask @lhog, otherwise I'll remove the branch

lhog: Use that branch in case AMD drivers refuse to compile the shader, though according to GLSL spec they shouldn't

badosu: Thank you!

sprunk:
Note, this is fairly low priority so is likely to sit around a bit but I think it would be fine to have it in eventually.

sprunk:
@badosu I rebased and made some changes on top of your commit, would you mind reviewing them?

badosu:
1. The layer can be anything you want, I assume the widget author wanted to wait for everything to be drawn before applying cas.

2. That's a good question, I'm unaware cc @lhog

GoogleFrog:
Looks good. @sprunk since the default is 0 at the moment perhaps it would be worth not doing all the stuff in Initialize until it is required.

lhog:
I don't think `sharpness` has easy definition. Just consider it "degree or sharpening" from 0 to 1 float.

sprunk:
Thanks everybody involved!

PR #4791, opened 2022-12-10
Making bombers display their actual reload times (time on the pad)

danfireman:
Before:
image

After:
image

### DIFF
--- a/units/bomberdisarm.lua
+++ b/units/bomberdisarm.lua
@@ -82,6 +82,7 @@ return { bomberdisarm = {
         disarmDamageMult = 1,
         disarmDamageOnly = 1,
         disarmTimer      = 16, -- seconds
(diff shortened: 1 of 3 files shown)


GoogleFrog:
This is too fragile, in the sense that I doubt whoever changes reammoseconds next is going to think to change script_reload in the weapondef. Context menu could pick up reammoseconds and replace the reload times of weapons. This wouldn't work for bombers with a tailgun though.

Ideally weapons with ammo would have custom param that flags the reload requirement, and the context menu would then print "needs reammo" under the weapon, and not bother displaying reload time or DPS (since they are nonsense). Then there would be a unit ability section that states how long it takes for the unit to rearm. This is a lot more descriptive than just writing a reload time in the weapondef.

danfireman:
@sprunk : Did that commit really fix things?
image

PR #4800, opened 2022-12-22
Fix Campaign chassis having two HDP modules

Arch-Shaman:
Thanks @LeojEspino for the report.


sprunk:
Thanks, though this removes the wrong module.

d1947a01626eec93f8fcafacb97683f4c5ad9a14

PR #4820, opened 2023-02-12
Healthbars: Fixed a bug where the "shield" text was not rendered

Arch-Shaman:
50be39ee introduced the bug, apparently by changing "shield" to "shield_bar", but never updating the reference in the widget.


sprunk:
good catch, thanks!

PR #4825, opened 2023-02-26
Use PrintWorld and a single font:Begin/End block.

moreginger:
Attempt to make gl text code when rendering several reclaim areas faster. I still need to double check this is actually faster, but in previous testing a single Begin/End block was improving time from ~30ms -> 5ms

However there's a technical problem. The text is rendered in the right place, but it's rotated at a seemingly arbitrary angle. Any idea what is going on here? I can't find any other uses of :PrintWorld so maybe it's not usable?

image

Ah, oh, it seems to be oriented to the camera at the time that the elements are created. Any ideas how to negate this?

Note: I wasn't able to use :Print because it didn't seem to be possible to set a different height for each text element.

### DIFF
--- a/LuaUI/Widgets/gui_reclaim_field_highlight.lua
+++ b/LuaUI/Widgets/gui_reclaim_field_highlight.lua
@@ -188,7 +188,7 @@ local function UpdateDrawEnabled()
 			or (options.showhighlight.value == 'conandecon' and (conSelected and WG.showeco)) then
 		return true
 	end
-	
+
 	local currentCmd = spGetActiveCommand()
 	if currentCmd then
 		local activeCmdDesc = spGetActiveCmdDesc(currentCmd)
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_reclaim_field_highlight.lua
```diff
@@ -589,22 +589,17 @@ local function DrawFeatureConvexHullEdge()
 end
 
 local function DrawFeatureClusterText()
+	glPushMatrix()
```
moreginger, line 592: Don't need this glPush/PopMatrix if we aren't transforming? But we'd probably need to apply one to fix the camera rotation anyway.

moreginger:
Note: @lhog investigated the slow performance of gl.Text and found it to be an interaction between displaylists (which IIUC are deprecated) and AMD. It seems that rewriting it without displaylists could solve the original performance issue in a more useful way. Also, he mentioned that the widget has been ported to BAR, so it's possible that we could use that as a better foundation for things.

GoogleFrog:
Thanks for looking into it. Porting the drawing from the BAR versions sounds like the way to go.

PR #4831, opened 2023-03-04
⚖️ Fix #2774, transportable heavy ships

### LINKED ISSUE #2774: Lobster-shogun horrible inconsistency
It is now possible to lobster transport shogun.
However hercules cannot transport a shogun.

This sould be mitigated by making all ships transportable again.

sprunk:
Reef and Shogun are now transportable, they could already be moved via Lobster anyway.

Reef might be somewhat problematic since its drones can still act, but maybe that's based and protosspilled.

### DIFF
--- a/units/chickenbroodqueen.lua
+++ b/units/chickenbroodqueen.lua
@@ -40,7 +40,6 @@ return { chickenbroodqueen = {
   canMove             = true,
   canPatrol           = true,
   canSubmerge         = true,
-  cantBeTransported   = true,
   category            = [[LAND]],
 
   customParams        = {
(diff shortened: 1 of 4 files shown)


GoogleFrog:
I thought there was a reason this didn't happen. Have you tested this, maybe the transport can't make it close enough to the units to actually pick them up? If it works though, I guess it sounds fine.

sprunk:
The transport can pick them up fine, Shogun looks slightly silly because has a tall superstructure:
image
 
There are some potential problems with Reef though.
1) It becomes a Protoss Carrier and can attack, because the drones aren't stunned. Maybe this is fine though because it can't rebuild them.
2) Planes can land on it fine, but are stuck because a stunned airpad doesn't actually refuel them.
image
3) gunships can't land on it. They arrive at the pad, shake a bit and fall down.
4) the default right-click command for a Hercules is to land on Reef since it's an airpad.

So perhaps Reef stays untransportable?

sprunk:
brutal

PR #4832, opened 2023-03-04
Add startboxes for Supreme_Crossing_V2 1

completesomething:
This pull request adds startboxes for https://zero-k.info/Maps/Detail/63186 , which takes a more minimalistic and desaturated approach to visuals, while also adding some new elements such as small mountains rising from the two seas.


sprunk:
thanks!

PR #4835, opened 2023-03-05
Add map info widget

completesomething:
This PR ports the Map Info widget from BAR (with some minor edits: Author localization call removed, ZK font used instead, description changed). The widget displays the map info on the bottom left corner of the map when zoomed in.

For reference: https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_mapinfo.lua

mapinfo

### DIFF
--- a/LuaUI/Widgets/gui_mapinfo.lua
+++ b/LuaUI/Widgets/gui_mapinfo.lua
@@ -1,0 +1,227 @@
+function widget:GetInfo()
+  return {
+    name      = "Map Info",
+    desc      = "Draws map info on the bottom left of the map.",
+    author    = "Floris",
+    date      = "20 May 2015",
+    license   = "GNU GPL, v2 or later",
+    layer     = -10,
+    enabled   = true,  --  loaded by default?
+  }
+end
+
+local scale					= 1
+local offset				= 5
+local thickness				= 6
+local fadeStartHeight		= 800
+local fadeEndHeight			= 4800
+local dlistAmount			= 20		-- amount of dlists created, one for each opacity value
+
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+local spGetCameraPosition	= Spring.GetCameraPosition
+local spGetGroundHeight		= Spring.GetGroundHeight
+local spIsAABBInView		= Spring.IsAABBInView
+
+local glColor           = gl.Color
+local glScale           = gl.Scale
+local glPushMatrix      = gl.PushMatrix
+local glPopMatrix       = gl.PopMatrix
+local glTranslate       = gl.Translate
+local glBeginEnd        = gl.BeginEnd
+local glVertex          = gl.Vertex
+local glBlending		= gl.Blending
+local glCallList		= gl.CallList
+local glDeleteList		= gl.DeleteList
+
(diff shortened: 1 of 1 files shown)


sprunk:
What for? The name is probably alright. The description and the author are generally useless to know and the size can be seen in world.

GoogleFrog:
I don't really see a reason for it and it will hit map edge extension, so it shouldn't be enabled by default. But I guess it is fine to chuck it in the widget list.

LuaUI/Widgets/gui_mapinfo.lua
```diff
@@ -0,0 +1,227 @@
...
+	for opacity, list in pairs(mapinfoList) do
+		glDeleteList(mapinfoList[opacity])
+		mapinfoList[opacity] = nil
+	end
+end
+
+function widget:RecvLuaMsg(msg, playerID)
+	if msg:sub(1,18) == 'LobbyOverlayActive' then
+		chobbyInterface = (msg:sub(1,19) == 'LobbyOverlayActive1')
+	end
```
sprunk, line 204: Why? It's an in-world thing so shouldn't care about the overlay.

LuaUI/Widgets/gui_mapinfo.lua
```diff
@@ -0,0 +1,227 @@
...
+	-- find the lowest map height
+	for i=math.floor(offset*scale), math.floor((mapInfoWidth+offset)*scale) do
+		if spGetGroundHeight(i, Game.mapSizeZ) < mapInfoBoxHeight then
+			mapInfoBoxHeight = spGetGroundHeight(i, Game.mapSizeZ)
+		end
+	end
+end
+
+function widget:GameFrame(gf)
+	if gf == 1 then
```
sprunk, line 177: Does frame == 1 mean it renders incorrectly if the map is shifted pregame (waterlevel modoption, random map generators)? What if the terrain shifts later (terraform)?

LuaUI/Widgets/gui_mapinfo.lua
```diff
@@ -0,0 +1,227 @@
...
+local vsx,vsy = Spring.GetViewGeometry()
+
+local mapInfoWidth = 400	-- minimum width
+local mapinfoList = {}
+
+local font, mapInfoBoxHeight, chobbyInterface
+
+local success, mapinfo = pcall(VFS.Include,"mapinfo.lua") -- load mapinfo.lua confs
+
+function widget:ViewResize()
```
sprunk, line 50: I don't see why an in-world thing would care about viewport size.

LuaUI/Widgets/gui_mapinfo.lua
```diff
@@ -0,0 +1,227 @@
...
+    enabled   = true,  --  loaded by default?
+  }
+end
+
+local scale					= 1
+local offset				= 5
+local thickness				= 6
+local fadeStartHeight		= 800
+local fadeEndHeight			= 4800
+local dlistAmount			= 20		-- amount of dlists created, one for each opacity value
```
sprunk, line 18: I am not competent at GL enough to know how bad 20 dlists are, it might be a nothingburger but also why bother fading.

PR #4851, opened 2023-03-22
WIP: Add SSAO widget

completesomething:
This PR adds the SSAO widget (screen-space ambient occlusion) from https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gfx_ssao.lua, simulating the impacts of bounce light by adding soft shadows to cornering/compact areas.

The effect is more prominent on objects closer to the camera, and doesn't seem to apply when the camera is zoomed out a fair bit.

All the relevant shaders are already included in the ZK repo, so it should probably be fine, though the widget warns of inactive "modelDiffTex" and "miscTex" on activation.

TODO:
* It would probably be wise to add GUI configuration. At least on my PC, it seems to cost ~4-10 FPS depending on zoom level.

### DIFF
--- a/LuaUI/Widgets/gfx_ssao.lua
+++ b/LuaUI/Widgets/gfx_ssao.lua
@@ -1,0 +1,660 @@
+
+local isPotatoGpu = false
+local gpuMem = (Platform.gpuMemorySize and Platform.gpuMemorySize or 1000) / 1000
+if Platform ~= nil and Platform.gpuVendor == 'Intel' then
+	isPotatoGpu = true
+end
+if gpuMem and gpuMem > 0 and gpuMem < 1800 then
+	isPotatoGpu = true
+end
+
+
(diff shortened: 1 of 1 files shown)


sprunk:
> the widget warns of inactive "modelDiffTex" and "miscTex" on activation.

I haven't seen a warning. Perhaps ZK moved to a more modern version of engine?

> It would probably be wise to add GUI configuration.

Yes. For starters it could be as simple as a checkbox though.

GoogleFrog:
I this supposed to have progressed? I have been ignoring it since it says WIP.

sprunk:
No progress but the only thing missing is to wire it up to epicmenu so I may end up doing that myself.

sprunk:
8bc3bc4cd696a98c9f9f0cc3a779a6a4d2f21e85

PR #4855, opened 2023-03-24
Normalized Carrot Mountain Mexes

Arch-Shaman:
Values would produce some unfairness before in proposed startboxes (Some would have 2.38 mexes and some would have 2.37's)


sprunk:
> Some would have 2.38 mexes and some would have 2.37's

Why the 1.9 then?

Arch-Shaman:
> > Some would have 2.38 mexes and some would have 2.37's
> 
> Why the 1.9 then?

image

The map has smaller mexes scattered throughout the map.

PR #4856, opened 2023-03-25
WIP: Fire is hot

Anarchid:
missileslow is you 

- Modifies area damage gadget to be able to spawn customParam-defined CEGs as long as the effect is active
- Adds Niarteloc's gas simulation sprite
- Uses the gas simulation sprite to create an animated fireball
- Configures Inferno to use this fireball as the continual spawn particle for its field of effect
- Does a similar thing with Zeno
- Damage-at-point value of the field is passed to the CEG, so it can adjust dynamically (set density, or size etc) to reflect effect strength.
- There is now much less need to use write-only cegexps to spawn a simple spherical volume.

https://user-images.githubusercontent.com/3822768/227731370-53740512-f28f-4cc2-87bb-d15f5e8e624c.mp4

Future re~~search~~cent advances :
- Replace the groundflash component of Inferno with a dynamic light. Do the same for Zeno.
- Feather the edges of the spritesheet with black, since Springcoil ignores alpha channel here and just uses sum(rgb) as transparency.
- Use LUPS jitter to make the fire more lively and turbulent.
- Use the miasma spritesheet (`in:#zkdev from:NRLC has:file`) for Zeno


sprunk:
Flame seems a bit too pastel (monocolor blobs) and the purple looks a bit too opaque (obscures stuff inside)

Anarchid:
> The Zeno gas has some horizontal lines 'artifacts' on the bottom. Should be ez fix.

Actually is not technically fixable [in this paradigm] at all. This is sprites clipping into the ground, the only way to avoid it is either spawn them far away from the ground to never intersect, or to make them smaller so that they're too small to intersect.

A proper solution would be to use a raymarched fog volume instead, but that's expensive.

M5PR:
Might be exacerbated by the camera angle; still good enough, don't get me wrong. Perhaps rotate two over eachother and drop transparancy, dumb trickery. Clipping as such isn't bad, but only when an unatural long line is visible. I'm noob at this so ignore me if I'm talking nonsense.

GoogleFrog:
image

GoogleFrog:
image

GoogleFrog:
I am not pleased.
- The second thing I notice (after testing what was changed) is that the other fireball effects were replaced with flickering squares. Do some basic testing, why should I even look further while the PR is in such a broken state?
- I would rather you attempt to use the existing system than have SpawnCeg called 150 times per second.  
- The whitespace is terrible.

See https://github.com/ZeroK-RTS/Zero-K/pull/4915 for minimal example.

Anarchid:
> I would rather you attempt to use the existing system than have SpawnCeg called 150 times per second.

I tried that and found everything i could do with it to be garbage.

The same animation playing on loop over a static hovering particle is garbage. 

Spawning them through CEG has no option to spawn particles continuously on its own, except for pre-spawning them all at once and giving them a different start timer - that's also quite crap, as we've seen in Area17.

> The second thing I notice (after testing what was changed) is that the other fireball effects were replaced with flickering squares. Do some basic testing, why should I even look further while the PR is in such a broken state?

Huh. Other fire should have been untouched, i guess i'll fix that.

Anarchid:
Restored everything except Inferno and Zeno.

Anarchid:
@GoogleFrog how much of a blocker is the SpawnCEG spam? I feel like continuous particle creation is a critical requirement, so i'm wary of going further with this if that's a blocker.

sprunk:
Perhaps making the particles a bit bigger and longer-lived (loop the animation a few times) would let you reduce the spawnage to normal levels.

sprunk:
I think a non-zero number would be fine. As always with artwork, it can only be judged after spending the mana though.

GoogleFrog:
Spamming spawn ceg isn't completely out of the question but it does raise the difficulty. For one thing I thought I was going to look at some ceg specifications but a lot of it ended up being highly looped code. It also gets ugly when a lot of cegs use the old system and some use the new, and it is harder to apply to non-area damage applications.

Spawning a big ceg in the gameplay loop, to handle a second or do of ceg activity, rather than spamming the spawns in the inner loop seems like a good place to aim for.

GoogleFrog:
The blocker for me is that I perpetually want to make a stable within a month, and I think polishing this up to releasable would take me a significant part of that month. The main issue is that we shouldn't have two wildly-different representations of AoE fire. So I would want to replace all the fire.

This is what I was feeling when I said.
> I thought I was going to look at some ceg specifications but a lot of it ended up being highly looped code.

A PR that just changed how a ceg is defined is simple. It doesn't touch any _code_ code so is very unlikely to break. And pure cegs are nicely abstracted, in the sense that Pheonix and DRP and Pyro can all just have the ceg they spawn changed to update how they look. I'm not saying that the PR can't include lua code, just that once it did, it crossed a complexity threshold.

With lua-augmented cegs I become less confident about the answer to the following questions, because cegs are more constrained.
- What happens if this hits a shield, a cliff, or water? Does it work?
- Is this code something that could be scaled down for Kodachi or Firewalker, or will there be unseen issues?
- Does it look any good when scaled down?
- What does lua spawning like this do to particle count culling?

More generally, gating animated cegs behind using this particular area damage gadget seems bad. It isn't generalisable. If a modder or wants to use it for something else they'll have to invent it from scratch. Which is why I would much prefer this simply be a ceg. But it could at least be an animated ceg framework, rather than something added to the area damage gadget.

I also have not been looking forward to saying that the Zeno effect seems like a step backwards in actual gameplay terms.

Anarchid:
> With lua-augmented cegs I become less confident about the answer to the following questions, because cegs are more constrained.

There are three things here.

1) An animated CEG particle that goes poof up. It is not gated behind the gadget.
2) A visual that i am trying to create: a hot, burning fire.
3) A gadget modification for an existing area-damage gadget, that spawns the particle in 1 over the effect area.

That makes it easy to reason about this:

>   What happens if this hits a shield, a cliff, or water? Does it work?

The gadget that spawns the particles is the same gadget that sets units on fire in the AOE of an inferno. If it currently works when hitting a shield, then it will work when hitting a shield. 

This seems actually less confounded than with CEG, which my electively not spawn if a correct flag is not set.

>    Does it look any good when scaled down?

The particle spawn parameters are configured in customparams fields of the area weapon def. They can be tuned to look good if needed.

>    What does lua spawning like this do to particle count culling?

Each flamelet counts as one particle, equivalent to a single orange ball in the current Inferno field. There are approximately as many short-lived flamelets at any point in time as there are long-lived orange balls.


---

 The visual i want in 2) cannot be expressed in pure CEG.

---

@sprunk i wonder how difficult it would be to do something like a repeater version of CExpGenSpawner?

Currently that thing can spawn, say, N instances of other cegs. WIth `i` it's also possible to delay each of those by an increasing amount of time, and thus separate them in time to create 1000 repetitions - but all the "sleeping agents" are already pre-created and drain the particle count limit (and do bad things to performance, as Area17 showed).

What if it could spawn on a timer without the i+delay hack?

sprunk:
> If it currently works when hitting a shield, then it will work when hitting a shield.

Current napalm is a vapor cloud though, so it looks good if centered mid-air. Meanwhile a realistic flame looks great when the ground is burning, but less so if it's an arbitrary point and it's just _space itself_ that's burning.

> separate them in time to create 1000 repetitions - but all the "sleeping agents" are already pre-created and drain the particle count limit (and do bad things to performance, as Area17 showed)
> What if it could spawn on a timer without the i+delay hack?

Didn't Area 17 spawn every individual particle ahead of time? What if you define a sub-explosion that spawns 200 repetitions and then have just 5 sleeper agents. I believe they still count as 1 each while sleeping so you should get a reasonably cheap effect as far as perf goes. I think spawning particles for long-lived effects should generally be done continuously via gadget though. This is for two reasons:
 * being a 'virtual' particle that doesn't count against the particle limit ruins part of the point of having a limit. They may well be cheaper (no rendering, physics etc) but they're still entities that need to be stored and handled. Is there a meta-limit on particles that don't obey the regular limit? The flip-side of this is that if the particles _do_ obey the limit (i.e. the current i+delay hack), they can themselves get dropped when the effect spawns and there's no control over the distribution of droppings. So perhaps it is sleeping itself that is a problem.
 * particles are unsynced* so get dropped on save/load but the underlying area damage is still there, so becomes invisible. This doesn't mean spawning all particles individually, but maybe have the gadget apply like 2-3s worth of them at a time (via a sub-explosion that lasts this much).

<sub>\* CEGs are technically synced, but there is no reason for them to be so hopefully engine advancements would make them unsynced in the foreseeable future. I wouldn't be surprised if save/load already skips them</sub>

Anarchid:
> Didn't Area 17 spawn every individual particle ahead of time, by the thousands? What if you define a sub-explosion that spawns 200 repetitions and then have just 5 sleeper agents. I believe they still count as 1 each while sleeping so you should get a reasonably cheap effect as far as perf goes. 

This would work, but assuming 1000 flamelets and 5 second-tier sleeper-agents, it would count as 204 sleeper agents (200 delayed flamelets of the first spymaster, and 4 inheritor spymasters).

The spymaster could also get killed by the particle limit, which would kill all of its sleepers.

Anarchid:
> Is this code something that could be scaled down for Kodachi or Firewalker, or will there be unseen issues?

Should a mergeable version of this PR include matching-style fire configs for Koda, Firewalker, etc?

GoogleFrog:
> The gadget that spawns the particles is the same gadget that sets units on fire in the AOE of an inferno. If it currently works when hitting a shield, then it will work when hitting a shield

I mean, does it look good?

> This seems actually less confounded than with pure CEG, which may kinda not spawn if a correct flag is not set.

Just set the correct tags. It's like saying a lua solution might not spawn in the right place if you assume Z is up. Just don't do that.

> The particle spawn parameters are configured in customparams fields of the area weapon def. They can be tuned to look good if needed.

Looking good is needed, and it isn't the case that every framework is capable of being tuned to look good.

>> What does lua spawning like this do to particle count culling?
> Each flamelet counts as one particle, equivalent to a single orange ball in the current Inferno field. There are approximately as many short-lived flamelets at any point in time as there are long-lived orange balls.

You've said what it does to particle count, but not what it does it culling. If there are enough particles to cull 20% of them, will the gadget spawn 100% or 80% of particles? It seems bad for lua to force-spawn all particles if they meant to be lodded essentially the same as engine cegs. Perhaps Spring.SpawnCeg should have a canCull parameter that applies the same probabilistic (I hope) culling of usual effects. Or perhaps this already exists.

I wouldn't block this by a lack of culling, but it would be nice to know how it works at the moment, and for there to be some way to get it in the future.

> I think spawning particles for long-lived effects should generally be done continuously via gadget though. This is for two reasons:

This seems fine, I would just prefer the gadget be its own thing so lots of mechanics can spawn these effects. If I were to implement Firewalker and Kodachi I would first refactor the lua. 

> Should a mergeable version of this PR include matching-style fire configs for Koda, Firewalker, etc ground-dot-fire sources?

Yes.

Anarchid:
>This seems fine, I would just prefer the gadget be its own thing so lots of mechanics can spawn these effects. If I were to implement Firewalker and Kodachi I would first refactor the lua.

Does it make sense to break out the particle-spawner thing into a separate gadget and have it coupled to weapon_area_damage only by config?

Anarchid:
> It seems bad for lua to force-spawn all particles if they meant to be lodded essentially the same as engine cegs

I would think that the force-spawning doesn't spawn them, since they're still CEG and still subject to engine culling. But that needs investigation i guess.

GoogleFrog:
> Does it make sense to break out the particle-spawner thing into a separate gadget and have it coupled to weapon_area_damage only by config?

I think it makes the most sense for the separate gadget to be callable, and for the area damage gadget to call it based on its config. Having two gadgets happen to directly use the same config seems bad if it makes sense to split them. The new gadget could essentially define effects which can be called by other gadgets.

> I would think that the force-spawning when above the limit does nothing, since they're still CEG and still subject to engine culling. But that needs investigation i guess.

Yes, it needs investigation. I have a suspicion that culling has a probabilistic range and that it is ignored by lua.

Anarchid:
Nananaf energy

PR #4864, opened 2023-04-02
Update zh_CN translation

Anarchid:
Some lines seem to be deleted, i assume this is because of the changes in the source material. Some are changed by translators on TX.

Anyone with chinese skillz could review, and or fix then on transifex.

### DIFF
--- a/LuaUI/Configs/lang/units.zh.json
+++ b/LuaUI/Configs/lang/units.zh.json
@@ -6,7 +6,7 @@
 	},
 	"gunshipcon" : {
 		"name" : "黄蜂",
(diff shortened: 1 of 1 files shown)


gounson:
Thanks a lot! 
Hoping zh_TW unit update soon.

sprunk:
I just noticed that in here most of the ones that reverted to English are because the spelling of "defense" changed to "defence" and Transifex thought it's a new string. So I can fix most of these.

sprunk:
Some strings were reverted to English in what looks like a regression, but there were changes to the source string that would leave the old translation incorrect in a subtle way that's hard to notice.

PR #4895, opened 2023-04-25
TTS: Replays ignore messages for commander select

Porkch0p:
Optimization: Prevent backlog of selection messages from being spoken at start of replays

### DIFF
--- a/LuaUI/Widgets/snd_text_to_speech.lua
+++ b/LuaUI/Widgets/snd_text_to_speech.lua
@@ -94,6 +94,10 @@ function widget:AddConsoleMessage(msg)
 	if not (msg and msg.msgtype == "player_to_allies" and msg.playername ~= myPlayerName) then
 		return
 	end
+	local comselect = "I choose:"
+	if Spring.IsReplay() and msg and msg.argument:sub(1, #comselect) == comselect then
+		return
+	end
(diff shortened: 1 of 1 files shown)


Porkch0p:
There is some processing of TTS input [on the infra side](https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/f228355e9e828921a66df4d6a20a1f3a28c2ce70/ChobbyLauncher/TextToSpeechBase.cs#L29) but it's mainly performing a sanity check on the inputs as far back as the commit history goes.

sprunk:
Looks like it was a ZKL thing? https://github.com/ZeroK-RTS/Zero-K-Infrastructure/blob/master/ZeroKLobby/ChatToSpeech.cs#L72

In that case it sounds fine to block gameside, perhaps it would also be good to replicate the other things the code above did like not reading "added point" or muted people.

Porkch0p:
> In that case it sounds fine to block gameside, perhaps it would also be good to replicate the other things the code above did like ignoring "I choose" regardless of replay, or not reading "added point" or muted people.

It appears "added point" is already left unspoken due to TTS only voicing allied chat at the moment. But as a downside "added label: HELP" also isn't voiced with current implementation

PR #4903, opened 2023-05-13
Add optional huge nuke warning

jklw:
This is for people who tend to tunnel-vision on a different part of the screen and miss the nuke warning completely :)

Haven't found a way to put this below the color blindness options instead of above.


sprunk:
Looks good, though my instinct is that if the regular message is not visible enough then perhaps that is something to work on. Maybe GF doesn't mind though.

screen00181

Works with translations:
screen00182

GoogleFrog:
I agree with Sprung. This is a niche solution by virtue of being behind a toggle in the menu. That's fine if its a niche problem. But if it isn't a niche problem then we're still waiting for a solution.

PR #4909, opened 2023-05-26
Make aegis spin proportional to shield health.

Mach565:
Doesn't spin at all at 0 shield strength, spins at the current speed when shields are full. I could also make it so that it doesn't move up and down at 0 shield strength too, if that's wanted.


sprunk:
Cool, thanks!

PR #4928, opened 2023-06-12
⚙️ Change some defs keys to match UnitDefs

sprunk:
In wupgets, you use for example `UnitDefs[x].metalCost`. But in unit def files it's `buildCostMetal`. This is needlessly confusing. New engine added support for matching keys recently, so apply some of these:

 * buildCostMetal → metalCost
 * buildCostEnergy → energyCost
 * energyUse → energyUpkeep
 * metalUse → metalUpkeep (vanilla ZK doesn't use this)

Mods who use the old spelling are supported for now.


GoogleFrog:
I am a bit conflicted about the unitdefs. On one hand, they're supposed to be alphabetical. On the other hand, I am used to looking near the top for metal cost. There is also the fact that units with energy cost specified would now have their costs spread around the file. There is precedent for related parameters to be grouped though (collision volume and selection volume).

GoogleFrog:
So that there is a decision, I'll go with making metal cost alphabetical, and putting energy cost and worker time below metal cost in the cases where they exist.

sprunk:
> and worker time

I'll assume this is buildTime and not workerTime which is buildpower.

GoogleFrog:
Yes.

PR #4953, opened 2023-06-30
Add username teamcoloring in chat message bodies

Birdulon:
Adds an option (defaults to on) which will find any usernames (within a game) in chat messages and will color them with their current teamcolor.
Note that changing teamcolors will not update past messages, though reloading the chat widget by toggling it off and on will.

Also bumps up the maximum height of the chat panel/window.

### DIFF
--- a/LuaUI/Widgets/gui_chili_proconsole_test.lua
+++ b/LuaUI/Widgets/gui_chili_proconsole_test.lua
@@ -121,6 +121,7 @@ local HIGHLIGHT_SURROUND_SEQUENCE_2 = ' <<<'
 local DEDUPE_SUFFIX = 'x '
 
 local MIN_HEIGHT = 50
+local MAX_HEIGHT = 2160
 local MIN_WIDTH = 300
 local MAX_STORED_MESSAGES = 300
 
@@ -201,6 +202,7 @@ options_order = {
 	
 	'color_chat_background','color_console_background',
 	'color_chat', 'color_ally', 'color_other', 'color_spec',
+	'color_usernames',
 	
 	'hideSpec', 'hideAlly', 'hidePoint', 'hideLabel', 'hideLog',
 	'error_opengl_source',
@@ -525,6 +527,14 @@ options = {
 		type = 'colors',
 		value = { 0.8, 0.8, 0.8, 1 },
 		OnChange = onOptionsChanged,
+		path = color_path,
+	},
+	color_usernames = {
+		name = "Color usernames in messages",
+		type = 'bool',
+		value = true,
+		OnChange = onOptionsChanged,
+		advanced = true,
 		path = color_path,
 	},
 	color_dup = {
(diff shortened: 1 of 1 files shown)


Birdulon:
https://cdn.discordapp.com/attachments/278908415756206080/1124206081091047474/image.png

LuaUI/Widgets/gui_chili_proconsole_test.lua
```diff
@@ -121,6 +121,7 @@ local HIGHLIGHT_SURROUND_SEQUENCE_2 = ' <<<'
 local DEDUPE_SUFFIX = 'x '
 
 local MIN_HEIGHT = 50
+local MAX_HEIGHT = 2160
```
sprunk, line 124: Won't this be a problem for smaller screens? If there are enough messages to go beyond the top of the screen, will it still be possible to scroll up to see the first message?

LuaUI/Widgets/gui_chili_proconsole_test.lua
```diff
@@ -905,7 +915,29 @@ end
...
+	-- Find and colour any usernames in the message body
+	if options.color_usernames.value then
+		-- This could be slightly faster by caching the value for each format rule,
+		-- but that's more effort elsewhere, risks cache bugs,
+		-- this operation is much faster than the N regex replacements,
+		-- and this code path isn't that hot to begin with (not many chat messages per second)
+		local last_colour_code = format:match('.*(#%w)') or '#o'
+		local message_colour = incolors[last_colour_code]
+
+		-- Lua lacks \b, so we match with spaces instead and strip the added ones afterwards
```
sprunk, line 928: Is `\b` about matching brackets? IIRC there's something like `%b()` where you specify the two characters.

Birdulon:
On my display it doesn't automatically expand under any circumstances, and
manually resizing it hits a rather small cap.

Birdulon:
regex `\b` is word boundary, the typical "find whole words only" search
option behaves as if you did `\bsearch term\b`, which in other regex terms
is similar to starting with a non-capturing lookbehind for non-word
characters or start of line or start of input, and ending with a
non-capturing lookahead for non-word characters or end of line or end of
input.
Lua's `%b` seems to be entirely different, I saw some suggestions for Lua's
frontier pattern match when searching around for alternatives, but the
example queries didn't give the behaviour I was after.
Basically the ideal is matching the word (username), and only the word, and
only when it is not attached to other legal characters (`[a-zA-Z0-9_]` for
regex `\w`, `[a-zA-Z0-9]` for lua `%w`, will need to do `[%w_]` to prevent
`Blue` from matching the tail of `Steel_Blue`)
Might be able to replace the gsub substitution function call with
```lua
local pattern = '([^%w_])('..name..')([^%w_])'
local sub = '%1'..colour..'%2'..message_colour..'%3'
formatted_arg, _ = formatted_arg:gsub(pattern, sub)
```
but I'll need to test it later.

The padding is for messages starting with or ending with a username.

GoogleFrog:
What? The white border at least needs fixing.
image

Birdulon:
Fixed in #4976

Issue #4955, opened 2023-06-30
Consider `wupget:Initialize(*allUnits*)`


sprunk:
Many wupgets iterate over all existing units on `/lua reload`. Perhaps the wupget handler could do a single `Spring.GetAllUnits` call and pass this table to all initializing wupgets so they don't have to do this on their own?

sprunk:
Here's some obvious benefits and downsides.
```diff
++ teaches newbie modders writing wupgets (via the presence of the arg hinting that it is there for a reason) that it may be a good idea to handle existing units
+ marginal (idk, measure!) speed-up for rare use cases such as loading a save game (LuaUI only), disabling LuaUI when catching up, and possibly reloading during development.
-- it requires talking to other games since otherwise ZK diverges.
- is a new standard that people will need to adhere to if it is going to be useful, which is cognitive clutter.
```
It would require easy but extensive changes, ideally in most spring games, and it's hard to tell how much of a benefit it would really be. I think it may be too mana-inefficient at this point to try and convince other games (and engine basecontent) to inflict a change like this to their wupget handlers, but idk.

GoogleFrog:
The downsides only occur if you think widgets are portable between games. In practise they aren't due to the different includes and ways we handle settings.

sprunk:
I don't think portability is binary and I think it's good not to move towards the world where it's closer to the "not portable at all" world as opposed to the "essentially portable but you need to change a handful of lines".

sprunk:
Engine development is now in a state where this needn't be done by ZK.

GoogleFrog:
Also another downside, what if a unit is created in Initialize?

PR #4958, opened 2023-06-30
Draft: ``wupget:Initialize(*allUnits*)`` implementation

Sharkgamedev:
This PR aims to explore the concept of #4955   
I thought this would be pretty straightforward but as it turns out...  
Anyway this is a pretty big changeset, so to make it simpler I have split things up via commits, I'll explain the levels here:
  
- 1. 5b5a486fe945dd3c2e28a8fc483b37b60bdefa76 Simply updates the widget/gadget handlers to pass through all units on initialisation
- 2. This 68436b90cd672e1559711f89ba32166533dae390 is a bulk search-and-replace operation to bring the new variable to widgets which don't have it. ALL THIS DOES is add 'allUnits' as a parameter to every widgets initialise function, it leaves out any widgets which might actually use it. I have manually reviewed this entire changeset to ensure that there are no mistakes so you should be good to ignore this one for now.
- 3. This fc02d68743d9127c7ab4e685413e6f1d7762ee14 actually changes the behaviour of those widgets which can be easily updated to take advantage of the new allUnits parameter  

## Those are the last of the easy changes, the following are special cases broken up by category
- 4. This 1a4f0078918adef0c42572e7ef18e3adb1828dd3 and this cfd336898f94d37d16ddecb18c91ecafddb11320 deal with widgets where initialisation happens in a separate function which is called multiple times from functions that don't have access to allunits. This comes in two flavours:
Flavour 1:
```lua
function widget:GameFrame()
   someOtherInitFuncThatWantsUnits([ nothing is passed through ]) -- I call this function, but don't want to run spring.getall every frame especially when it has its own logic for determining whether to or not

function someOtherInitFuncThatWantsUnits([should this take allUnits?])
    if someStateThatEnsuresWeAreNotConstantlyCallingGetAllUnits:
         -- do stuff with allUnits/Spring.GetAllUnits(), this is where it would be helpful to have allUnits when it is called in Initialize

function widget:Initialize(allUnits)
   someOtherInitFuncThatWantsUnits([do we pass through allUnits?])
```
We could ignore these widgets, or a possible solution to the above situation (as is implemented in some of the widgets as an example) is to include a 
```lua
(allUnits or Spring.GetAllUnits())
```
Flavour 2:
```lua
function widget:GameFrameOrSomeOtherFuncCallledOften()
   someOtherInitFuncThatWantsUnits([ we could do Spring.GetAllUnits() ]) -- since the function would have called getallunits anyway we are ok to pass this through

function someOtherInitFuncThatWantsUnits([should this take allUnits?])
     -- do stuff with allUnits/Spring.GetAllUnits(), this is where it would be helpful to have allUnits when it is called in Initialize

function widget:Initialize(allUnits)
   someOtherInitFuncThatWantsUnits([do we pass through allUnits?])
```
The possible solution to this is included above.
- 5. This db43d966f73b8fc6658d88e998e4b31521ac125a deals with widgets where other code calls widget:Initialize() again, often in places where we would not want to call Spring.GetAllUnits() again, yet we don't want to call widget:Initialize() without any parameters because that means Initialize() cant use allUnits without the possibility of it being null.
Solutions to this are:
- Ignore these widgets
- Separate initialisation code into a separate function that the other code can call again (as in most cases these widgets don't actually use allUnits)
- Use a conditional in these again `(allUnits or Spring.GetAllUnits)`

Another possible solution to all of these problems is to not retro-actively update widgets which do not make use of Spring.GetAllUnits(), that would bring the changeset down to about 20 files and remove a lot of these icky situations.   
  
Thus far I think this should be a helpful overview of all the commits and changes (most files have comments in them explaining the various issues faced)  
  
Right now this PR only covers Widgets, as I figured I would wait until we had decisions on these design choices before I attempt to work on Gadgets where I am sure the same conundrums will crop up again.


sprunk:
Most of the work involved is to sit down a bit and figure out whether it's even worth doing anything, since it goes beyond the scope of ZK. I'll write something on the ticket.

Sharkgamedev:
Yup! That totally makes sense. I just hoped to give some insight into the pitfalls and what would be required to make this happen with this PR.

sprunk:
5b5a486fe945dd3c2e28a8fc483b37b60bdefa76 doesn't achieve the optimisation part, it even makes it worse because now all widget initialisations involve a GetAllUnits call. IDK if it's worth fixing at the moment though because I still am unsure if it's worth doing the change in the first place.

Sharkgamedev:
> 5b5a486fe945dd3c2e28a8fc483b37b60bdefa76 doesn't achieve the optimisation part, it even makes it worse because now all widget initialisations involve a GetAllUnits call. IDK if it's worth fixing at the moment though because I still am unsure if it's worth doing the change in the first place.

Ohh wow I wasn't paying attention 😅 that shouldn't be too difficult to change but yeah Ill leave it for now

sprunk:
There's a bunch of merge conflicts (in addition to the remarks above) and now that engine development is approachable ZK would probably want to take something from upstream. Thanks for the attempt though!

PR #4975, opened 2023-07-10
Big weapon update nobody asked for

sprunk:
Krow has Redback pew. See #4780 for reasoning.
image

Zephyr has WW2 guns. See 7064d2dd0bfdd74a4d36f253b3c2a2ff377d2a9a for reasoning (tl;dr slightly helps #3500 but mostly for the vibe, ww2 is the shipyard fantasy).
image

Shotguns no longer look like lasers. Slight mechanics implications but burnblow should keep them tame.
image
image

Lightning looks nicer. See #4465 and #3900. Possibly large mechanics and balance implications.
image


GoogleFrog:
image
Alternate Krow proposal. See the commit for comments.

GoogleFrog:
Shotgun looks good to me.

GoogleFrog:
Lightning difference needs resolving
image

GoogleFrog:
image
The lightning needs more work. Wiggle width seems to scale based on beam length and wiggle segments appears fixed. The Knight in the above screenshot looks fine, but the Paladin just looks like a scaled-up version of the Knight. When put side-by-side it ruins scale verisimilitude, and if people mod longer lightning beams it will get even worse. We probably need more ways to configure this visual in the engine.

image

GoogleFrog:
I am not sure what the Zephyr change is doing, besides giving it burst. Some thoughts
- It still seems to miss quite a bit against turning bombers, which was one of the main issues.
- This makes Razor unique and impinges on the EMG sound space, which already has a lot of different types.
- Technically Toad uses this weapon type, but it is hard to hear over the teal laser.
- Why couldn't laser Zephyr be given burst?

GoogleFrog:
image
Zephyr

PR #4976, opened 2023-07-10
Default fonts to autoOutlineColor = false

### LINKED ISSUE #4954: Remove all white outlines from player names
image

Some widgets show white outlines around dark-ish player names, others do not. The white outlines make the text hard to read and the UI generally feel messy. If we have player colours that are too dark then they should be lightened, not given an outline. That said, I don't consider the blue in the screenshot above too dark. So the solution is not to simply tweak the default colours, but to edit the outline settings in offending widgets.

Birdulon:
Removes the behaviour of "dark" font colours having white outlines.
Should fix #4954.
Additional small refactor on chat widget following up from comments on #4953.


Birdulon:
Silent_AI and Maxwus have colours that would have white text outlines/shadows in these views without this change.

Issue #4968, opened 2023-07-08
Merge improved music widget


GoogleFrog:
https://git.ufeff.net/birdulon/Zero-K_Custom_Widgets/src/branch/master/snd_music.lua

It seems possible to configure it to do basically the same thing as the existing widget, but with a bunch of options to do extra stuff.

PR #4982, opened 2023-07-20
[WIP] Port of additional music moods (#4968)

Birdulon:
Rework war calculations, add `war2`, `winning`, `losing` music moods related to higher combat levels and large attrition ratios, respectively. Rework dynamic transition logic to accomodate more than just peace<->war.
This is a considerable refactor of my additions of my personal widget linked in #4968, which forked in June 2020. It is nowhere near as used and tested as that, but should be a little bit nicer than it by the end.

### DIFF
--- a/LuaUI/Widgets/snd_music2.lua
+++ b/LuaUI/Widgets/snd_music2.lua
@@ -15,8 +15,8 @@ function widget:GetInfo()
 	return {
 		name    = "Music Player v2",
 		desc    = "Plays music based on situation",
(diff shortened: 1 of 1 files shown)


sprunk:
Splitting off as many small standalone changes as possible into commits would be helpful, a monolith is a bit harder to review.

GoogleFrog:
That might not be necessary. We could just replace one black box with another - that sounds easier. It's relatively safe to do with a widget as I could  just run it for a month and see if anything comes up.

PR #5063, opened 2023-11-16
Propagate attacker info to widgets for UnitDamaged callin.

moreginger:
This should be safe to do since https://github.com/beyond-all-reason/spring/issues/391 was fixed in the engine.

As player I get the callin info when I have line of sight, but not when I don't - I think this is OK?

image
image

### DIFF
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -2348,9 +2348,9 @@ end
 
 function widgetHandler:UnitDamaged(unitID, unitDefID, unitTeam, damage, paralyzer, weaponDefID, projectileID, attackerID, attackerDefID, attackerTeam)
 	for _, w in r_ipairs(self.UnitDamagedList) do
-		-- The engine only provides attacker information when visible. However weapon information is always passed - arguably
-		-- the weapon is always visible - but we don't want widgets to be able to infer the attacker when it is not visible.
-		w:UnitDamaged(unitID, unitDefID, unitTeam, damage, paralyzer, nil, nil, attackerID, attackerDefID, attackerTeam)
+		-- The engine only provides attackerID etc if the attacker is visible.
+		-- OTOH weaponDefID and projectileID are always provided - elide projectileID so that widgets can't aquire projectile vector and locate the attacker.
+		w:UnitDamaged(unitID, unitDefID, unitTeam, damage, paralyzer, weaponDefID, nil, attackerID, attackerDefID, attackerTeam)
 	end
 end
 
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -2346,9 +2346,9 @@ function widgetHandler:UnitCmdDone(unitID, unitDefID, unitTeam, cmdID, cmdParams
 end
 
 
-function widgetHandler:UnitDamaged(unitID, unitDefID, unitTeam, damage, paralyzer)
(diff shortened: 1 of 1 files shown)


sprunk:
Looks reasonable, widgets that use the callin need to be checked whether they make any assumptions around this.

moreginger:
Good point, I'll have a look later.

moreginger:
Look safe to me in ZK widgets. Only unit_attack_warning.lua and unit_global_build_command.lua are parameterized with these args, and neither use them.

GoogleFrog:
Doesn't projectileID need poisoning?

moreginger:
> Doesn't projectileID need poisoning?

In @lhog s MR he wrote "these two do not count as information leaks", but it does seem out of keeping. Although the missile callins seem to have WeaponDefID. Do you think this needs a similar engine change, or can we just pass nil, or something else?

sprunk:
Just pass `nil`.

moreginger:
Just so I can comment right, is this because we don't want widgets to be able to infer out of sight units (even though a human could)?

GoogleFrog:
I was just thinking you could use projectile velocity to pinpoint things like Phantoms, Tremors, Bertha and Missile Silos. If the projectile exists enough for its velocity to be read.

moreginger:
> I was just thinking you could use projectile velocity to pinpoint things like Phantoms, Tremors, Bertha and Missile Silos. If the projectile exists enough for its velocity to be read.

I'm assuming you can't read the projectile information as a widget even if you have the id, otherwise you could try to scan with random ids if nothing else :shrug: It's probably best if you clean this MR up as you'd like it, if you're OK with the changes to attacker info. If we can propagate all the information then great... it might come in useful :)

moreginger:
Any thoughts here @GoogleFrog? Are the values OK to pass after all, or some of them?

GoogleFrog:
> I'm assuming you can't read the projectile information as a widget even if you have the id, otherwise you could try to scan with random ids if nothing else

I don't recall ZK being secure against projectile hax, we've just blocked the non-obscure things. I would expect things can be read from any existing projectileID. In any case, the projectile IDs from this callin need to be tested by someone.

moreginger:
OK I've given it another shot and revised my comment. Does this cover your concerns? I've checked the engine and the weaponDefID/projectileID are always passed, so we absolutely do need to nil either of them that you think could be abused.

GoogleFrog:
In general, the engine doesn't decide what counts as an information leak.

image

WeaponDefID is the only way to tell these two projectiles apart. The main thing is I don't know what this PR is for, so I don't know what can be tweaked easily or what removes the point.

moreginger:
When a unit is destroyed the callin is told which unit did it if they are in LoS. This allows replaycam to display the identity of the assailant, which is useful for commentary. I wanted the same for unit damaged callin, see the screenshots. Before the indicated engine change this would have leaked information on units outside the LoS, but now it does not.
As well as this, this makes the damaged callin consistent with unit destroyed. If this information were considered secret it should be removed there too.

Then there are the weapon infos. I am not using these but I have considered it. For example I could use them to determine if a unit is being "attacked" by impulse weapon? These don't appear to be available to many callins, only some missile ones IIRC. They are not restricted to LOS. You could argue, that the weapon type (Def) is visible information already and not a leak. As you said, with the projectile id you might be able to retrieve a vector or other information about the attacker. I can see a good argument for not passing the projectile id, and I don't have any intention to use it.

PR #5070, opened 2023-11-22
restore Odin liftoff sound in fog-safe manner

Anarchid:


scripts/bomberassault.lua
```diff
@@ -85,17 +85,19 @@ local function IsManualFireTargetValid()
...
-
---function script.Activate()
---	Spin(turbineTop, z_axis, 8,2);
---	Spin(turbineLeft, z_axis, 8,2);
---	Spin(turbineRight, z_axis, -8,2);
---end
+function script.Deactivate()
+	StopSpin(turbineTop, y_axis, 0.5);
+	StopSpin(turbineLeft, y_axis, 0.5);
+	StopSpin(turbineRight, y_axis, 0.5);
```
sprunk, line 91: doesn't look sound related and there is no corresponding Spin

GoogleFrog:
Has this been tested? It just sounds like a brief click to me. Odin already makes a noise when it receives an order, so this is redundant for the order-giver, and it just seems like noise to anyone else. Sounds are used to tell people that something happened, and I see no value in telling people that Odin took off. Perhaps something like this is good for ambience, but I think we at least would need someone designing the overall audio landscape and designating bits of it for ambience for that to work. When I heard this in the wild I thought my speakers were broken or a sound was clipping. It has a piercing pitch.

Anarchid:
The sound was originally tested when Eclipse was added. It's supposed to sound like a gas burner being turned on, which does include a bit of clicking.

PR #5072, opened 2023-11-24
ReplayCam

moreginger:
This widget takes over the camera when in spec mode, and attempts to show highlights of the battle with commentary.

See this thread: https://zero-k.info/Forum/Thread/36302 for background and discussion

It has some minimal options that are useful for debugging / streaming:
- Show tracking reticle (default: false)
- Disable tracking (default: false)
- Pause tracking on user input (default: true)

This isn't ready to be merged yet, at a minimum the following are required, but no doubt there will be more:

- Update the UI to support ZK standards, placement, resizing etc
- How should the functionality be turned on/off (currently you have to enable/disable the widget)?

### DIFF
--- a/LuaUI/Widgets/replaycam.lua
+++ b/LuaUI/Widgets/replaycam.lua
@@ -1,0 +1,1667 @@
+function widget:GetInfo()
+	return {
+		name    = "ReplayCam",
+		desc    = "Pan to and comment on interesting events",
+		author  = "moreginger",
+		date    = "2023-11-24",
+		license = "GNU GPL v2",
+		--layer        = 0,
+		enabled = false
+	}
+end
+
+local atan2 = math.atan2
+local cos = math.cos
+local deg = math.deg
+local exp = math.exp
+local floor = math.floor
+local huge = math.huge
+local max = math.max
+local min = math.min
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I am not sure about merging this at level 4 as it seems like a waste. See https://zero-k.info/Forum/Post/226780#226780

I am also not sure about putting it under camera, since this feels pretty distinct from different camera modes and settings for manual cameras. Perhaps start the submenu Settings/Interface/Spectating and put this there, so spectating options aren't spread all over the place.

moreginger:
> I am not sure about merging this at level 4 as it seems like a waste. See https://zero-k.info/Forum/Post/226780#226780
> 
> I am also not sure about putting it under camera, since this feels pretty distinct from different camera modes and settings for manual cameras. Perhaps start the submenu Settings/Interface/Spectating and put this there, so spectating options aren't spread all over the place.

Yeah, these sound like good ideas. I didn't know about the levels TBH.

I was wondering if UI-wise a button could be integrated into the replay controls to start/stop the widget... does that sound like it might be reasonable? Of course it doesn't really help you if you are speccing a live game.

PR #5074, opened 2023-11-25
Adding new widget for drawing lobster jump effects

danfireman:
image

(screenshot shots slightly thicker lines than PR)

Activates when lobsters are selected and the dgun is chosen.

Shows:
* Range of selected lobsters.
* Range from lobster for units to be eligible to be thrown (in yellow)
* Which units will be launched (in green)
* Which units you might think would be launched but aren't (e.g. a unit just at the edge of range or the launching lobster in the middle of your ball) (in red)
* Which units are in range of a lobster, but the cursor is positioned too far from a launching lobster so they won't be lobbed (in orange)

### DIFF
--- a/LuaUI/Widgets/gui_attack_aoe.lua
+++ b/LuaUI/Widgets/gui_attack_aoe.lua
@@ -29,8 +29,6 @@ local depthLineWidth       = 1
 local circleDivs           = 64
 local minSpread            = 8 --weapons with this spread or less are ignored
 local numAoECircles        = 9
-local pointSizeMult        = 2048
-
 --------------------------------------------------------------------------------
 --vars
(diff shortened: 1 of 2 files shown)


GoogleFrog:
You can see in the screenshot how one of the circles is a bit bolder because its also being drawn by attack AoE. I was hoping that the attack AoE widget would be updated to draw all the circles, and for them to match the weight of normal attack circles a bit better.

danfireman:
Done

GoogleFrog:
I think the main lobster effect and the stacking AoEs should be level 3 (https://zero-k.info/Forum/Post/226780#226780).

The stacking AoE adds more information to the screen and

PR #5082, opened 2023-11-30
unit team names widget

moreginger:
Rework comm nametags widget as a general unit team names widget.

This picks one unit to act as an avatar for each team using a simple algorithm:

- Nearest to centre of screen
- Mobile always beats static
- Keeps selection unless unit leaves the centre of the screen

Text is log scaled to make it readable over a wider zoom range

Performance seems alright - I'm seeing 0.5% of time with a few labels. When I checked it was mostly the label rendering - this is known to be slow in Spring.

See spec demo here:

https://clipchamp.com/watch/gnQWGEiHksM


moreginger:
> There should be settings that subsume the commander name tags widget if it is to be deleted. It doesn't have to be deleted though.

I considered that, but I thought on balance it's OK to delete things to provide a more streamlined experience. Superfans can retrieve the original comm widget if they need it.

What is the feature of the original widget that you think is valuable? For me the value of both was "know your friend/enemy", not "commander tagging".

moreginger:
I'm struggling with finding the centre of the viewport. It might not be possible with the current API when running windowed. see discussion here: https://discord.com/channels/278805140708786177/911790552687730689/1180535092846800906

moreginger:
I'm just wondering if it's better to make this centre on the cursor instead of the screen.

GoogleFrog:
Screen seems better as the cursor would be noisy. Also, for the replay camera it wouldn't want to care about cursor.

PR #5089, opened 2023-12-07
Commanders no longer spawn on top of each other

Arch-Shaman:
This is an enhancement added in FW v0.38.22 to help players understand that they have multiple comms. Comms that spawn on top of each other will also (very) slowly float away due to impulse sliding, which is just silly.

### DIFF
--- a/LuaRules/Gadgets/start_unit_setup.lua
+++ b/LuaRules/Gadgets/start_unit_setup.lua
@@ -302,7 +302,7 @@ local function CanUnitDropHere(unitDefID, x, y, z, facing, checkFeature)
 	end
 end
 
-local function GetAdjustedDropPosition(unitDefID, facing, x, z)
+local function GetClosestValidSpawnSpot(unitDefID, facing, x, z)
 	local radius = 16 -- FIXME: take the actual unit footprint size, perhaps also facing for nota style comms
 	local y = Spring.GetGroundHeight(x, z)
 	local canDropHere = CanUnitDropHere(unitDefID, x, y, z, facing, false)
@@ -367,7 +367,7 @@ local function SpawnStartUnit(teamID, playerID, isAI, bonusSpawn, notAtTheStartO
 		
 		-- get facing direction
 		local facing = GetFacingDirection(x, z, teamID)
-		x, y, z = GetAdjustedDropPosition(startUnit, facing, x, z) -- adjust for new location.
+		x, y, z = GetClosestValidSpawnSpot(startUnit, facing, x, z) -- adjust for new location.
 
 		if CAMPAIGN_SPAWN_DEBUG then
 			local _, aiName = Spring.GetAIInfo(teamID)
--- a/LuaRules/Gadgets/start_unit_setup.lua
+++ b/LuaRules/Gadgets/start_unit_setup.lua
@@ -305,6 +305,7 @@ local function GetClosestValidSpawnSpot(teamID, unitDefID, facing, x, z)
 	local dir = 1 -- 1: right, 2: up, 3: left, 4 down
 	local nx, ny, nz
 	local offsetX, offsetZ = 0, 0
+	local aborted = false
 	repeat -- 1 right, 1 up, 2 left, 2 down, 3 right, 3 up
 		nx = x + offsetX
 		nz = z + offsetZ
@@ -319,7 +320,7 @@ local function GetClosestValidSpawnSpot(teamID, unitDefID, facing, x, z)
 				movesLeft = mag
 				dir = dir%4 + 1
 			elseif mag == 8 and movesLeft == 0 and dir == 4 then -- abort
-				canDropHere = true 
+				aborted = true 
 			else -- move to the next offset
 				if dir == 1 then
 					offsetX = offsetX + radius
@@ -333,8 +334,12 @@ local function GetClosestValidSpawnSpot(teamID, unitDefID, facing, x, z)
 				movesLeft = movesLeft - 1
 			end
 		end
-	until canDropHere
-	return nx, ny, nz
+	until canDropHere or aborted
+	if canDropHere then
+		return nx, ny, nz
+	else -- aborted: give original position back.
+		return x, Spring.GetGroundHeight(x, z), z
+	end
 end
 	
 local function SpawnStartUnit(teamID, playerID, isAI, bonusSpawn, notAtTheStartOfTheGame)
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+local function GetAdjustedDropPosition(unitDefID, facing, x, z)
+	local radius = 16
+	local y = Spring.GetGroundHeight(x, z)
+	local canDropHere = CanUnitDropHere(unitDefID, x, y, z, facing, false)
+	if canDropHere then return x, y, z end
+	local mag = 1
+	local index = 1
+	local nx, ny, nz
+	repeat
+		nx = x + (offsetGrid[index][1] * radius * mag)
```
sprunk, line 314: This has the checked positions in a star pattern, perhaps it could be a spiral?

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+	else
+		return blocking > 1
+	end
+end
+
+local function GetAdjustedDropPosition(unitDefID, facing, x, z)
+	local radius = 16
+	local y = Spring.GetGroundHeight(x, z)
+	local canDropHere = CanUnitDropHere(unitDefID, x, y, z, facing, false)
+	if canDropHere then return x, y, z end
```
sprunk, line 309: Could be part of the loop below if the starting conditions involved `mag = 0` (incl. if it was a spiral)

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+	local blocking, feature = Spring.TestBuildOrder(unitDefID, x, y, z, facing)
+	if checkFeature then
+		return blocking == 3 -- Recoil engine now has 3 for "free", 2 for "blocked by feature"
+	else
+		return blocking > 1
+	end
+end
+
+local function GetAdjustedDropPosition(unitDefID, facing, x, z)
+	local radius = 16
```
sprunk, line 306:
```suggestion
	local radius = 16 -- FIXME: take the actual unit footprint size, perhaps also facing for nota style comms
```

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+		ny = Spring.GetGroundHeight(nx, nz)
+		canDropHere = CanUnitDropHere(unitDefID, nx, ny, nz, facing, false)
+		if not canDropHere then 
+			index = index + 1
+			if index == 9 then
+				index = 1
+				mag = mag + 1
+			end
+		end
+	until canDropHere
```
sprunk, line 325: Could use a loop limiter in case the map is super degenerate.

LuaRules/Gadgets/start_unit_setup.lua
sprunk: Also it should probably check map bounds in case somebody starts on the edge.

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -323,6 +367,7 @@ local function SpawnStartUnit(teamID, playerID, isAI, bonusSpawn, notAtTheStartO
 		
 		-- get facing direction
 		local facing = GetFacingDirection(x, z, teamID)
+		x, y, z = GetAdjustedDropPosition(startUnit, facing, x, z) -- adjust for new location.
```
sprunk, line 370: This seems to ignore startboxes, which is bad but not necessarily worrisome (since I don't know any maps off the top of my head that have areas where it would lead you to spawn outside the box). It also looks like it handles cliffs, which is nice (but would ideally happen at the placing stage).

Arch-Shaman: image

Indeed does handle cliffs nicely. (Adding in case someone just reads the PR and not the chat around it)

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+local function CanUnitDropHere(unitDefID, x, y, z, facing, checkFeature)
+	local blocking, feature = Spring.TestBuildOrder(unitDefID, x, y, z, facing)
+	if checkFeature then
+		return blocking == 3 -- Recoil engine now has 3 for "free", 2 for "blocked by feature"
+	else
+		return blocking > 1
+	end
+end
+
+local function GetAdjustedDropPosition(unitDefID, facing, x, z)
```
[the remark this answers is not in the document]
Arch-Shaman: Marking as resolved by a4f01d1

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+		nz = z + (offsetGrid[index][2] * radius * mag)
+		ny = Spring.GetGroundHeight(nx, nz)
+		canDropHere = CanUnitDropHere(unitDefID, nx, ny, nz, facing, false)
+		if not canDropHere then 
+			index = index + 1
+			if index == 9 then
+				index = 1
+				mag = mag + 1
+			end
+		end
```
sprunk, line 324:
```suggestion
		if canDropHere then
			return nx, ny, nz
		end
		
		index = index + 1
		if index == 9 then
			index = 1
			mag = mag + 1
		end
```

LuaRules/Gadgets/start_unit_setup.lua
```diff
@@ -282,6 +282,50 @@ local function GetStartPos(teamID, teamInfo, isAI)
...
+	until canDropHere
+	return nx, ny, nz
```
sprunk, line 326: If nothing valid was found I'd return the original/center spot instead of whatever random spot happened to be the last one checked.
```suggestion
	return x, y, z
```

Arch-Shaman: resolved by c8f8d67

GoogleFrog:
What is WIP here? The code itself looks pretty self contained, but I would want to test it on people so don't just want to merge an hour before stable.

Arch-Shaman:
WIP is "I haven't tested this so I'm worried about merging" (aka confidence
issue). The FW version used original and I've tested up to the second to
last commit. I can inflict it on fw:test in 15-20 minutes.

Arch-Shaman:
While this code works, see #5099. I have no idea if `shuffleboxes` works properly though. This may need @sprunk's insight on that.

PR #5106, opened 2023-12-11
fix: Make the first ally color a bit whiter overall, as it doesn't contrast well with the dark outline

moreginger:
Plus some cleanup of unused code.

As discussed in Discord the first ally blue isn't very readable next to the other colors. Tried to tweak it a little without losing too much of the blueness.

Before:

image

After:

image


LuaUI/Configs/TeamColorConfig.lua
```diff
@@ -132,13 +130,6 @@ local simpleColors = {
 	enemyByTeamColors = colors.enemyColors,
 }
 
-local simpleColorsTeams = {
```
moreginger, line 135: Unused.

PR #5107, opened 2023-12-12
make thermite a bit angrier when it bites

Anarchid:
Only in case of damage:
- Sparks fly longer
- Sparks have gravity, so they arc down
- More sparks
- Bigger sparks
- Sparks fade to orange and red as they dim

image


Anarchid:
Superceded by direct commit

PR #5123, opened 2023-12-20
fix: Add guard against invalid events.

moreginger:
This is currently triggered being triggered by fakeunit_los.


sprunk:
What about fakeunit_los causes it to fail? It's supposed to behave like terraunit just with LoS, and presumably terraunit doesn't fail.

moreginger:
> What about fakeunit_los causes it to fail? It's supposed to behave like terraunit just with LoS, and presumably terraunit doesn't fail.

Specifically I wasn't able to get a location for it when it was created, or destroyed. Maybe it's a timing thing that terraform tends to exist for long enough? I have a guard on UnitDestroyed against terra but that's just because we aren't interested in terra units being "destroyed".

PR #5127, opened 2023-12-22
BAR selection widget 

moreginger:
Credit: this is largely a copy of https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_selectedunits_gl4.lua and https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_allySelectedUnits.lua

This screenshot shows how it looks on default settings.

image

Ally/other (in replays) selections are shown in team color. Local selections are always in green and superimposed. Local selections are also animated with a breathe, see video later.

image

I included a few options but tried to keep things basic.

image

This is intended to replace unit_shapes.lua. Can it do this, being GL4, or do we need both? The principle advantage is that it is considerably faster (~1% vs ~10% in some cases):

image

It differs from the BAR widgets in the following ways:

- The two widgets have been combined into one
- Highlights are more ZK-style. Notably:
   - Local selections are green
   - Buildings are squares, not cornerrects
   - Selection highlighting doesn't float above ground

The main visual difference from unit_shapes is that the selection lines are solid, not gradients. As discussed in Discord, it is plausible to write a fragment shader to replicate this style, but I'm not sure it's worth the effort - I actually like the look of this.

Video in action here: https://clipchamp.com/watch/PjEF4vvYxxJ

### DIFF
--- a/LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.frag.glsl
+++ b/LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.frag.glsl
@@ -1,0 +1,33 @@
+#version 420
+#extension GL_ARB_uniform_buffer_object : require
+#extension GL_ARB_shading_language_420pack: require
+
+// This shader is (c) Beherith (mysterme@gmail.com)
+
+//__ENGINEUNIFORMBUFFERDEFS__
+//__DEFINES__
+
+#line 30000
+uniform float addRadius = 0.0;
+uniform float iconDistance = 20000.0;
+in DataGS {
+	vec4 g_color;
+	vec4 g_uv;
+};
+
+uniform sampler2D DrawPrimitiveAtUnitTexture;
+out vec4 fragColor;
+
+void main(void)
+{
+	vec4 texcolor = vec4(1.0);
+	#if (USETEXTURE == 1)
+		texcolor = texture(DrawPrimitiveAtUnitTexture, g_uv.xy);
+	#endif
+	fragColor.rgba = vec4(g_color.rgb * texcolor.rgb + addRadius, texcolor.a * TRANSPARENCY + addRadius);
+	POST_SHADING
+	//fragColor.rgba = vec4(1.0);
+	#if (DISCARD == 1)
+		if (fragColor.a < 0.01) discard;
+	#endif
+}
--- a/LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.vert.glsl
+++ b/LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.vert.glsl
@@ -1,0 +1,105 @@
+#version 420
+#extension GL_ARB_uniform_buffer_object : require
+#extension GL_ARB_shader_storage_buffer_object : require
+#extension GL_ARB_shading_language_420pack: require
+// This shader is (c) Beherith (mysterme@gmail.com)
+
+#line 5000
+
+layout (location = 0) in vec4 lengthwidthcornerheight;
+layout (location = 1) in uint teamID;
+layout (location = 2) in uint numvertices;
+layout (location = 3) in vec4 parameters; // lifestart, ismine
+layout (location = 4) in vec4 uvoffsets; // this is optional, for using an Atlas
+layout (location = 5) in uvec4 instData;
+
+//__ENGINEUNIFORMBUFFERDEFS__
+//__DEFINES__
+
+struct SUniformsBuffer {
+    uint composite; //     u8 drawFlag; u8 unused1; u16 id;
+    
+    uint unused2;
+    uint unused3;
+    uint unused4;
+
+    float maxHealth;
+    float health;
+    float unused5;
+    float unused6;
+    
+    vec4 drawPos;
+    vec4 speed;
+    vec4[4] userDefined; //can't use float[16] because float in arrays occupies 4 * float space
+};
+
+layout(std140, binding=1) readonly buffer UniformsBuffer {
+    SUniformsBuffer uni[];
+}; 
+
+#define UNITID (uni[instData.y].composite >> 16)
+
+#line 10000
+
+uniform float addRadius = 0.0;
+uniform float iconDistance = 20000.0;
+
+out DataVS {
+	uint v_numvertices;
+	float v_rotationY;
+	vec4 v_color;
+	vec4 v_lengthwidthcornerheight;
+	vec4 v_centerpos;
+	vec4 v_uvoffsets;
+	vec4 v_parameters;
+	#if (FULL_ROTATION == 1)
+		mat3 v_fullrotation;
+	#endif
+};
+
+layout(std140, binding=0) readonly buffer MatrixBuffer {
+	mat4 UnitPieces[];
+};
+
+
+bool vertexClipped(vec4 clipspace, float tolerance) {
+  return any(lessThan(clipspace.xyz, -clipspace.www * tolerance)) ||
+         any(greaterThan(clipspace.xyz, clipspace.www * tolerance));
+}
+
+void main()
+{
+	uint baseIndex = instData.x; // this tells us which unit matrix to find
+	mat4 modelMatrix = UnitPieces[baseIndex]; // This gives us the models  world pos and rot matrix
+
+	gl_Position = cameraViewProj * vec4(modelMatrix[3].xyz, 1.0); // We transform this vertex into the center of the model
+	v_rotationY = atan(modelMatrix[0][2], modelMatrix[0][0]); // we can get the euler Y rot of the model from the model matrix
+	v_uvoffsets = uvoffsets;
+	v_parameters = parameters;
+	v_color = teamColor[teamID];  // We can lookup the teamcolor right here
+	v_centerpos = vec4( modelMatrix[3].xyz, 1.0); // We are going to pass the centerpoint to the GS
+	v_lengthwidthcornerheight = lengthwidthcornerheight;
+	#if (ANIMATION == 1)
+		float animation = clamp(((timeInfo.x + timeInfo.w) - parameters.x)/GROWTHRATE + INITIALSIZE, INITIALSIZE, 1.0) + sin((timeInfo.x)/BREATHERATE)*BREATHESIZE;
+		v_lengthwidthcornerheight.xy *= animation; // modulate it with animation factor
+	#endif
+	POST_ANIM
+	v_numvertices = numvertices;
+	if (vertexClipped(gl_Position, CLIPTOLERANCE)) v_numvertices = 0; // Make no primitives on stuff outside of screen
+	// TODO: take into account size of primitive before clipping
+
+	// this sets the num prims to 0 for units further from cam than iconDistance
+	float cameraDistance = length((cameraViewInv[3]).xyz - v_centerpos.xyz);
+	if (cameraDistance > iconDistance) v_numvertices = 0;
+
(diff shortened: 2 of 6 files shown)


GoogleFrog:
I plan to run this locally live on the weekend to check it out.

GoogleFrog:
The new shapes are drawn behind the wiggly water effect, even for units above the water. This old shapes did this for underwater units and it possibly wasn't ideal there either. The new shapes also seem to suffer more from being drawn behind water, since the wide line makes the wiggle more obvious (or something).
image
image

GoogleFrog:
The pulse seems like too much. It makes the squares push beyond unit footprints so they no longer line up nicely.
image
image

GoogleFrog:
The colours are wasted on team colour selection. The colour for hovering an ally unit should at least be distinct from having a unit selected, which should be distinct from hovering an enemy unit too. Currently these are all green. Selections from allies don't need to be by teamcolour.

GoogleFrog:
Selection boxes feel "laggy" due to how long it takes the selection preview to appear. Drag a box around some units in a fraction of a second and the preview doesn't even feel like it appears.
spring_ilxU4zUg3Z
spring_cCSaTWxw39

GoogleFrog:
The aircraft triangles appear to be equilateral.
image
image

GoogleFrog:
Offscreen culling of selection box drawing is premature.
image
image
image

GoogleFrog:
Here is the culling problem at a realistic zoom. This will cause disconcerting flickering.
image
image

GoogleFrog:
The new selection boxes draw below units that should be behind them.
image

moreginger:
Thanks for the thorough review. I'll see what I can do about the issues.

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -0,0 +1,380 @@
+function widget:GetInfo()
+	return {
+		name = "Selected Units GL4",
```
moreginger, line 3: Let's try and work through the issues here. Note that a lot of these are moving us further from the BAR selection widget.

- [ ] Don't like misalignment caused breathe animation
  - Let's remove it, or we could tone it down if you'd prefer, but see below
- [ ] Want to use colors for selections like the current widget (instead of above)
  - OK this is straightforward for the selection. I guess we need to do preselection too. This will reduce performance since we now have 3 selection levels (preselection, selection and ally selection) needing separate shader passes, but it shouldn't be too bad?
- [ ] Initial selection growth animation feels unresponsive
  - unitshapes scales 95 -> 100% in 1/20s. This PR scales 50 -> 100% in 1/2s, so we can easily speed that up to match what we did before.
- [ ] Want elongated / isosceles triangles for planes
  - Should be easy enough with some fiddling, current code supports elongated quads after all.
- [ ] Don't like wiggly selections affected by water
  - Needs investigation
- [ ]  Offscreen culling
  - Needs investigation
- [ ] Need to draw boxes above other units when they are "above" them.
  - Needs investigation

LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.frag.glsl
```diff
@@ -0,0 +1,33 @@
...
+uniform sampler2D DrawPrimitiveAtUnitTexture;
+out vec4 fragColor;
+
+void main(void)
+{
+	vec4 texcolor = vec4(1.0);
+	#if (USETEXTURE == 1)
+		texcolor = texture(DrawPrimitiveAtUnitTexture, g_uv.xy);
+	#endif
+	fragColor.rgba = vec4(g_color.rgb * texcolor.rgb + addRadius, texcolor.a * TRANSPARENCY + addRadius);
```
moreginger, line 27: Review this. I don't understand what it's achieving in practise.

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -0,0 +1,400 @@
...
+	local InitDrawPrimitiveAtUnit = DPatUnit.InitDrawPrimitiveAtUnit
+	local InitDrawPrimitiveAtUnitVBO = DPatUnit.InitDrawPrimitiveAtUnitVBO
+	local shaderConfig = DPatUnit.shaderConfig -- MAKE SURE YOU READ THE SHADERCONFIG TABLE!
+	shaderConfig.BILLBOARD = 0
+	shaderConfig.ANIMATION = 1
+	shaderConfig.INITIALSIZE = 0.95
+	shaderConfig.GROWTHRATE = 4.0
+	shaderConfig.HEIGHTOFFSET = 0
+	shaderConfig.USETEXTURE = 0
+	shaderConfig.POST_GEOMETRY = "gl_Position.z = (gl_Position.z) - 64.0 / gl_Position.w;" -- Reduce pull-forward to 1/4 default otherwise the selection plane tends to clip through Krow.
```
moreginger, line 186: Risky?

GoogleFrog:
Terrain clipping.
image

It looks somewhat clean so it might not be the worst thing.

GoogleFrog:
image
This welder draws no selection circle.

GoogleFrog:
customParams.selection_scale is ignored.

GoogleFrog:
Everyone's teamcolour selection shape is rapidly flashing.

GoogleFrog:
UpdateCmdColorsConfig should not be present as-is. It will end up turning on engine boxes when people really just want to toggle between box widgets. I forget exactly how the toggle options for selection shapes deals with this.

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -0,0 +1,417 @@
...
+	shaderConfig.POST_SHADING = "fragColor.rgba = vec4(g_color.rgb, texcolor.a * " .. platterOpacity .. " + texcolor.a * sign(addRadius) * " .. (outlineOpacity - platterOpacity) .. ");"
+	selectionShader = InitDrawPrimitiveAtUnitShader(shaderConfig, "selectedUnits")
+	hoverSelectionVBO = InitDrawPrimitiveAtUnitVBO("selectedUnits_hover")
+	localSelectionVBO = InitDrawPrimitiveAtUnitVBO("selectedUnits_local")
+	otherSelectionVBO = InitDrawPrimitiveAtUnitVBO("selectedUnits_other")
+
+	return selectionShader ~= nil
+end
+
+options_path = 'Settings/Interface/Selection/Selected Units'
```
GoogleFrog, line 208: Options should go at the top of the widget so people can find them. I would do local init above options in this case, then define init below.

GoogleFrog:
image
Selections still disable on the edge of the screen too early.

GoogleFrog:
Does fractional linewidth cause issues?

GoogleFrog:
I had a look into it and don't even know how the widget decides whether a unit is visible. Unit Shapes uses Spring.GetVisibleUnits(-1, nil, false) (which I believe is cached via a ZK api) and fails a little, but less than Selections GL4.

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -0,0 +1,421 @@
...
+			Init()
+		end,
+	},
+	linewidth = {
+		name = 'Outline Width',
+		desc = '',
+		type = 'radioButton',
+		items = {
+			{ name = 'Thin',     key = '1' },
+			{ name = 'Standard', key = '2' },
```
moreginger, line 84: @GoogleFrog I don't think that fractional values will cause an issue. I don't have a particularly refined taste so if you want to tweak default to 1.7 and/or allow configuration over a continuum have at it!

LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.vert.glsl
```diff
@@ -0,0 +1,112 @@
...
+	#if (ANIMATION == 1)
+		// No animation when parameters.y is 0
+		float initial = clamp(((timeInfo.x + timeInfo.w) - parameters.x)/GROWTHRATE + INITIALSIZE, max(INITIALSIZE, float(1 - parameters.y)), 1.0);
+		float breathe = (BREATHESIZE + sin((timeInfo.x)/BREATHERATE)*BREATHESIZE) * parameters.y;
+		v_lengthwidthcornerheight.xy *= initial;
+		v_lengthwidthcornerheight.xy += breathe;
+	#endif
+	POST_ANIM
+	v_numvertices = numvertices;
+	if (vertexClipped(gl_Position, CLIPTOLERANCE)) v_numvertices = 0; // Make no primitives on stuff outside of screen
```
moreginger, line 95: @GoogleFrog This is where offscreen culling happens.
I only see it when somewhat zoomed in. If increase CLIPTOLERANCE I can still make the box disappear by zooming in even more.
I don't understand what gl_Position.www is... is there a better function that we can use here to clip - possibly as suggested in comment below.

moreginger:
Copying original summary here: Credit: this is largely a copy of https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_selectedunits_gl4.lua and https://github.com/beyond-all-reason/Beyond-All-Reason/blob/master/luaui/Widgets/gui_allySelectedUnits.lua

This screenshot shows how it looks on default settings.

image

Ally/other (in replays) selections are shown in team color. Local selections are always in green and superimposed. Local selections are also animated with a breathe, see video later.

image

I included a few options but tried to keep things basic.

image

This is intended to replace unit_shapes.lua. Can it do this, being GL4, or do we need both? The principle advantage is that it is considerably faster (~1% vs ~10% in some cases):

image

It differs from the BAR widgets in the following ways:

- The two widgets have been combined into one
- Highlights are more ZK-style. Notably:
   - Local selections are green
   - Buildings are squares, not cornerrects
   - Selection highlighting doesn't float above ground

The main visual difference from unit_shapes is that the selection lines are solid, not gradients. As discussed in Discord, it is plausible to write a fragment shader to replicate this style, but I'm not sure it's worth the effort - I actually like the look of this.

Video in action here: https://clipchamp.com/watch/PjEF4vvYxxJ

LuaUI/Widgets/gui_selectedunits_gl4.lua
moreginger: TODO:

- [ ] Default width and selectable range of outline
- [ ] Float selection boxes slightly to help reduce clipping e.g. on factory plate?
- [ ] Configurable value for above?
- [ ] Edge culling issue, needs input
- [ ] UpdateCmdColorsConfig not correct?

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -13,7 +13,8 @@ function widget:GetInfo()
 end
 
 -- Configurable Parts:
-local lineWidth, showOtherSelections, drawDepthCheck, platterOpacity, outlineOpacity
+local lineWidthDefault, platterOpactiyDefault, outlineOpacityDefault, selectionHeightDefault = 2, 0.15, 0.75, 2.0
```
GoogleFrog, line 18: This seems redundant. Just read the values from options.bla.value?

moreginger: I use them as defaults in Init if tonumber returns nil. I picked that up from somewhere and assumed it was needed for safety.

LuaUI/Widgets/Shaders/DrawPrimitiveAtUnit.vert.glsl
GoogleFrog: Where is CLIPTOLERANCE defined?

moreginger: It's in the shader config. The default is 1.1.
So we can override it with `shaderConfig.CLIPTOLERANCE= 1.5` for example.

PR #5128, opened 2023-12-23
Integrate replaycam into replay controls

moreginger:
Also did some cleanup and i18n on replay controls.

image


LuaUI/Widgets/camera_replaycam.lua
```diff
@@ -133,7 +132,7 @@ options = {
 		name = 'Pause tracking on user input',
 		desc = 'Pause camera tracking when the user moves the mouse etc',
 		type = 'bool',
-		value = true,
+		value = false,
```
moreginger, line 135: It's confusing to have this default to true as it gives the impression the widget doesn't do anything

LuaUI/Widgets/gui_replay_controls.lua
```diff
@@ -22,83 +24,51 @@ local Chili
...
 
 -- Defined upon learning the appropriate colors
 local BUTTON_COLOR
 local BUTTON_FOCUS_COLOR
 local BUTTON_BORDER_COLOR
 
----------------------------------
--- Epic Menu
----------------------------------
-options_path = 'Settings/HUD Panels/Replay Controls'
```
moreginger, line 64: This option isn't worthwhile.

sprunk: It's for spoilers

moreginger: Sorry can you explain what that is? Why would you ever want to hide the progress bar?

sprunk: Say you're casting a game for a youtube video (or just watching somebody else's replay). You can gauge the total and remanining length of the game from the progress bar which spoils what is gonna happen (eg. nuke about to finish -> if the bar is near 100% it means it will have succeeded). So you can disable the bar to avoid spoilers. But you still want the rest of the widget to be around for the x1 x10 speed buttons.

moreginger: Damn OK, I would not have thought of that. Will restore.

moreginger: Restored

moreginger:
Put in draft while I work out how to best handle toggling off in spec mode.

moreginger:
Closing this for now.

PR #5151, opened 2024-01-04
cmd_send_one_exlude_pad.lua fixed wrong option code and wrong WG var

Helwor:
Also the filename miss the 'c' in 'exlude'.

### DIFF
--- a/LuaUI/Widgets/cmd_send_one_exlude_pad.lua
+++ b/LuaUI/Widgets/cmd_send_one_exlude_pad.lua
@@ -10,6 +10,9 @@ function widget:GetInfo()
 		handler   = true,
 	}
 end
+local Echo 				= Spring.Echo
+local spGiveOrderToUnit 		= Spring.GiveOrderToUnit
+local spGetSelectedUnitsSorted 		= Spring.GetSelectedUnitsSorted
 
 local airUnitDefID = {}
 for defID, def in pairs(UnitDefs) do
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/cmd_send_one_exlude_pad.lua
```diff
@@ -23,15 +26,29 @@ do
...
 	if cmd ~= CMD_EXCLUDE_PAD then
 		return false
 	end
-	local selTypes = WG.selectionMap or Spring.GetSelectedUnitsSorted()
-	for defID, t in pairs(selTypes) do
+	local selTypes = WG.selectionDefID or spGetSelectedUnitsSorted()
+	for defID, units in pairs(selTypes) do
 		if airUnitDefID[defID] then
-			Spring.GiveOrderToUnit(t[1],cmd,params,opts)
+			spGiveOrderToUnit(units[1], cmd, params, opts.coded or GetOptionsCode(opts))
```
sprunk, line 51: you should be able to just pass `opts` and it is supposed to work (engine accepts both number and table)

LuaUI/Widgets/cmd_send_one_exlude_pad.lua
```diff
@@ -10,6 +10,9 @@ function widget:GetInfo()
 		handler   = true,
 	}
 end
+local Echo 				= Spring.Echo
```
GoogleFrog, line 13: Don't localise Echo

LuaUI/Widgets/cmd_send_one_exlude_pad.lua
GoogleFrog: Doesn't this comment make the PR a noop then? Is there a way to reproduce a bug here?

Helwor: Indeed my bad for the options, I thought engine worked with indexed table only (if table).
Remaining only the WG var change.

sprunk:
Where is `WG.selectionDefID` defined? It doesn't seem to actually exist.

Helwor:
It doesnt exist in here, only exists in my SelectionAPI widget to centralize selection infos.
Since GoogleFrog added my widget 'as is' and let this var, I just correct it. It's up to you if you want to remove it or let it just in case it would be something in the future.

PR #5154, opened 2024-01-13
Fix strider hub not inflicting initial states onto units

Arch-Shaman:


LuaUI/Widgets/unit_start_state.lua
```diff
@@ -146,6 +146,10 @@ for i = 1, #UnitDefs do
 	end
 end
 
+local notAFactory = {
+	[UnitDefNames["athena"]] = true, -- use stupid hax to avoid not treating striderhubs like a factory.
+}
```
sprunk, line 151: Use customparam: `notreallyafactory`

Arch-Shaman: c176132

Issue #4952, opened 2023-06-30
Check out BAR "glowing hot barrel after shot" shader


sprunk:
IIRC it's present on their Core Bertha. If it's reasonably easy to add to existing units then perhaps we could use it on units like Sniper, Emissary or Bertha (or in general anything powerful with a reasonably long reload, maybe even Stardust if the high RoF isn't a problem).

sprunk:
![image](https://github.com/ZeroK-RTS/Zero-K/assets/2573076/007e682c-b202-4cb7-b621-4f0a636dd78f)

Licho1:
looks pretty cool

6AKU66:
Steal from BAR can actually become a thing?

GoogleFrog:
I thought it was something that depends on the atlas. Still, try it out.

PR #5164, opened 2024-01-20
✨ Likho remodel.

sprunk:
image
image

Folds wings when landed (like existing Likho apparently). Both modes shown here + comparison with other planes.
image

On buildpad.
image

Colvol XZ size of existing Likho seemed alright so I didn't touch it. Only made it less thicc in the Y dimension (and removed a random sideways shift).
image
image

Minor issue: doesn't enter the "landed" state on airpad -> doesn't fold wings and can clip the superstructure on airfac pad.
image


GoogleFrog:
Probably not worth touching the colvol on a unit that has been so tricky to balance. A visual difference of 4 elmos seems ok.

Anarchid:
> Minor issue: dethklon is not mangled, just a retexture.

image

sprunk:
> A visual difference of 4 elmos seems ok.

vertical?

sprunk:
Apparently this part is bugged and not supposed to look like that, but I hear it looks correct in Blender
image

Anarchid:
Deleting the armature object in the model fixes this, at the cost of breaking the script, which seems to be using the armature for some reason.

image

sprunk:
Not necessarily, maybe it is a badly UV'd copy of the whole geometry, perfectly overlapping so that neither is "on top" of the other, and Blender happens to draw the real geometry on top while Spring/Recoil happens to draw Armature on top?

sprunk:
Also yours looks matte bright-gray compared to my shiny dark-gray, though maybe it's just map lighting (I tested on `duck`).

Anarchid:
> Not necessarily, maybe it is a badly UV'd copy of the whole geometry, perfectly overlapping so that neither is "on top" of the other, and Blender happens to draw the real geometry on top while Spring/Recoil happens to draw Armature on top?

Seems implausible. There is only one mesh object in the file altogether - `likho2023`. Additionally, the armature is essentially this:
image

Essentially it's a hierarchy of transform matrices for each bone. It has no polygons, so it can't be drawn on top.

GoogleFrog:
The darker, slightly shinier version looks better to me. In terms of colour, not in terms of the texture being messed up around the edges. In general a bit of variation of colour could be good too.

Anarchid:
> The darker, slightly shinier version looks better to me. In terms of colour, not in terms of the texture being messed up around the edges. In general a bit of variation of colour could be good too.

There is no darker/brighter version. The one in my screenshot has different lighting (it's on Folsom, not on Duck) and is highlighted by the cursor hover shader, but it's the same texture. There were no changes to texture, or even to UVs on the model vertices, all i did in that test was to delete the armature to see if it was related to the garbled texture problem. Which it was.

sprunk:
> There is no darker/brighter version. The one in my screenshot has different lighting (it's on Folsom, not on Duck) and is highlighted by the cursor hover shader, but it's the same texture. 

I think GF means that the texture should be edited so that it would look on Folsom (and other normal maps) like it currently looks on Duck (and on Duck it would then probably end up being black and mirror-level shiny?).

Anarchid:
> I think GF means that the texture should be edited (made darker and shinier) so that it would look on Folsom (and other normal maps) like it currently looks on Duck (and on Duck it would then probably end up being black and mirror-level shiny?).

It seems as dark as other "core" planes on Folsom atm:

image

That said, maybe it could be significantly blacker and glossier. It's made out of denser stuff.

Anarchid:
> That said, maybe it could be significantly blacker and glossier. It's made out of denser stuff. The texture could also be a bit less noise grain and more metal-material-ish.

I made a quick attempt to evaluate/iterate 
image

sprunk:
Looks like the early levels of Tomb Raider 3

GoogleFrog:
I see three different versions above.
- Sprung's screenshots with a decent black plane, which are perhaps bit too shiny and a bit close to grey, but which have broken texture alignment.
- Anarchid's concrete grey version that fixes the texture alignment, but is really grey and matt.
- Anarchid's terrible cloudy version.

Why are there so many differences, which one works? Is it a map lighting thing? And all three versions have a stretched texture issue in the air intakes on the top.

sprunk:
Cloudy is a retexture attempt. Difference between mine and gray Anarchid is map lighting but unfortunately mine is on an map with uncommon/broken light.

GoogleFrog:
I guess I could crosspost this feedback here.

image

> actually the ravager picture is a good example of things I think new Likho is missing
Ravager has a bit of variation. It has black and while, different texture of black on the wheels, and a little red window
Likho is just shiny black with teamcolour trying to do a lot of the jobs that little highlighted details would try to do. But it all ends up being one colour with teamcolour on top.
and usually I think teamcolour exists in reasonably thick blocks, not things like thin edges that fade into the rest of the texture. It should stand out and not be everywhere
the blacks of Likho turn into grey, but it all depends on the angle of the unit and the sun settings of the map. Which is bad for consistency

But idk if anything should be changed based on it. I've been saying stuff like that for ages already. I'm not sure we have any options beyond fixing the technical issues with the existing PR and just calling it good enough.

GoogleFrog:
image
It's pretty grey and indistinct.

image
Being grey and _this_ shiny seems like a bad approach since sometimes the whole plane flashes white.

image
Overall it seems to be too small, or built at a small scale. The details disappear at this distance. You can easily tell what the other bombers are. It doesn't fell like something that just scaling up the model will fix.

image

Here are some shots of it flying around. The old Likho is much easier to pick out imo.

GoogleFrog:
image
How many of each type of Likho is in this picture? (Added airjets to old Likho for fairness)

GoogleFrog:
image

sprunk:
> Added airjets to old Likho for fairness

Commit that? It was supposed to have them AFAIK (for the new one I just changed piece name but kept the rest of the config the same)

bruno-cb:
@Anarchid @sprunk @GoogleFrog 
Sorry to ping you, but 3 days of silence implies that you haven't been notified.

sprunk:
PR updated with latest model. @GoogleFrog

GoogleFrog:
I like the shape overall and the ball is looking good in the holder. My main concern is the model for the ball when it is fired, and the texture being very reflective and saturated. The gold bits look a bit too much like teamcolour for my liking due to their general brightness. Also the texture might be a bit too fine grained compared to other units, making it look like a shrunk version of a larger unit. The teamcolour is quite decent though.
image

I might be worth trying increasing the size of the model by 10%. The hitvolume will need a bit of adjusting anyway. I am also not sure about the wings folding all the way up while landing. Retracting halfway into the sides seems better, just to help people recognise the unit when it is on a pad.

It is extremely bright/reflective
image
image

The whole model seems to be offset
image

The in-flight ball is jarringly large.
image

image
image

GoogleFrog:
A gold version could be used for Likhos that attain a high level.

sprunk:
@GoogleFrog new version.

GoogleFrog:
I feel like it is too gold, and that is going to look good with some teamcolours and weird with others. It looks a bit weird with player-green, but I also suspect this would be the type of thing that grows on me.
image
Does it look fine with gold as dull as Odin? I think we're at the point where the texture can be left as-is, and I'd get used to it and it would be fine. But when I run my "what would a real art director do?" algorithm, it says "There are large areas of non-teamcolour with much more saturation than the standard range for Zero-K units". 

Actually, I can give an approximate answer to my own question:
image
That is dropping saturation by 18 "points" in Paint.net, where 100 points reduces everything to grey.

It is still noticeably gold, but the teamcolour sticks out clearly as more saturate, and it fits better with the rest of the units and environment. It is no longer weirdly bright.
image
I probably subtlety ruined the DDS somehow when exporting it, so I don't want to mess with the texture myself, but I think this slight toning down of the gold makes it fit a lot better so something like this would be good.

Here is my messed with file:
[likho2024-1.zip](https://github.com/ZeroK-RTS/Zero-K/files/15403614/likho2024-1.zip)

sprunk:
The white squares in the no-ammo state are supposed to be a glowing-hot radiator. I think the idea was spawned at the time the glowing-hot barrels shader (#4952) was discovered, idk how realistic it would be to implement (given it is a shader). So far it's just flat white and produces (or is supposed to produce, I forget) some LUPS heat dissipation effects (distortion + glow).
image

bruno-cb:
I can't see the diference in saturation in your changes, perhaps it can only be noticed when actively looking for it?
When comparing with 34 player colors, I noticed that desatured yellows and pinks tend to blend in with the gold:
Screenshot 2024-05-23 143150
Screenshot 2024-05-23 143446

So I made 2 new textures:
One that halves saturation, as @GoogleFrog proposed:
Screenshot 2024-05-23 144231
Screenshot 2024-05-23 144355

And another that halves value, trying to match Odin:
Screenshot 2024-05-23 145141
Screenshot 2024-05-23 145234

The radiator fins aren't parallel because at some angles they wouldn't be visible.
They are white because they only appear when in a "overheat" state. On a previous version there was some texture to emphasize the glow, but it needed the glowing-hot shader, which I couldn't use, so I ditched it.

[50sat-50val.zip](https://github.com/ZeroK-RTS/Zero-K/files/15421241/50sat-50val.zip)

GoogleFrog:
The saturation thing is subtle. It is about making the gold pop less than the teamcolour, so it doesn't look like a teamcolour, and so it fits in with the rest of the game world.

The fins not being parallel, is ok, or at least it is a minor detail. My point about the fins is that, in actual gameplay, they look like one of two things.
- A pure white square on an otherwise textured model.
- A red dot.

Which thing it looks like depends on the angle of the camera and the Likho. This makes the white square flash red momentarily when the plane banks. The idea of a radiator is cool, but currently it looks like a white square that flashes red when the plane turns. I don't really know what to do about it.

How much more work are you looking to put in here? These comments are about the model reaching its full potential, but I could just pull, then:
- Pick a slightly lower saturation texture.
- Disable the cooling fins.
- Possibly spruce up the CEG.

bruno-cb:
I completely desaturated the texture so it can be interpolated to an ideal value:
Screenshot 2024-05-25 155124

I noticed that the skybox adds some color to the neutral Likho, so reflection channel might make it more yellowish depending on the map:
Screenshot 2024-05-25 155452

I also added a bit of player color so it's not fully colorless:
Screenshot 2024-05-25 163813

Golden + player color:
Screenshot 2024-05-25 163314

[sat00+color.zip](https://github.com/ZeroK-RTS/Zero-K/files/15444946/sat00%2Bcolor.zip)

I will try to do something with the radiators so they don't disappear when looking from above, and add some red/orange gradient.

> How much more work are you looking to put in here?

Zero.

It has been over a year that I started working on this and most of this time might have been waiting for feedback. I know it isn't in the high priority list but it doesn't need to be this slow.

I will make the final alterations this week.

_"Better done than perfect."_

GoogleFrog:
You can do zero work right now if you want. I don't think fixing the radiators is going to be that easy.

Also, I think teamcolour should generally be in blocks, rather than being subtly mixed in with other colours. Reflectiveness will control how much the skybox colours a unit, or perhaps just the sun colour. Different maps have different sun colours too.

bruno-cb:
Fixed an inverted texture that caused bad reflections.
Made player color a bit more saturated by tuning emission down.
Screenshot 2024-05-27 214859
Changed radiator and some particles.
Screenshot 2024-05-28 173253
Screenshot 2024-05-28 173334

> I think teamcolour should generally be in blocks, rather than being subtly mixed in with other colours

I agree with the blocks of color for primary identification, but I think there is no harm in adding some hints as a subtle differentiation. 
Screenshot 2024-05-27 215046

[likho2024-RC3.zip](https://github.com/ZeroK-RTS/Zero-K/files/15475938/likho2024-RC3.zip)

GoogleFrog:
Oh, I pulled but those aren't in the PR.

GoogleFrog:
Ok, I don't understand the technical structure of this PR. [likho2024-RC3.zip](https://github.com/ZeroK-RTS/Zero-K/files/15475938/likho2024-RC3.zip) contains those textures, but it also contains a bunch of other specification etc files that are not the same as the PR, so what is going on?

sprunk:
For earlier PRs I cleaned up the Lua files each time.

GoogleFrog:
I guess they will need cleaning again, because there were some changes.

GoogleFrog:
The model doesn't seem to have a piece hierarchy. From what I can tell from the dae, all the pieces are roots.

`Move(base, y_axis, 30, 60)`
image

GoogleFrog:
I am happy at this point. That leaves technical issues.
- [ ] As I found above, the model doesn't seem to have a piece hierarchy. All the pieces just sit somewhere in space, which is going to make any further modification or use by modding difficult.
- [ ] The "base" piece has geometry. Models should have an empty root because that is convenient for moving things around in the script.
- [ ] The main body piece could be improved by splitting it in three. This would let the main left and right engines fly off as gibs independently.
- [ ] The midpos is on the ground, when it should come from the middle of the unit, like with Odin.
- [ ] The collision volume is offset for some reason. I can mess with it after other things are fixed.

Body cuts:
image

Midpos:
image
image

bruno-cb:
The base files I modified are very outdated, from January or earlier.
Ignore changes unrelated to likho.
Some effects weren't updated after the 10% model increase.

> The model doesn't seem to have a piece hierarchy

Indeed, after the problems with the armature I looked at other models and they worked like this, so I just went with it.
I remember trying some hierarchy but some pieces would disappear.

> The collision volume is offset for some reason

Is it correct in the old model?
I didn't touch that and both models are aligned in blender.

GoogleFrog:
> The base files I modified are very outdated, from January or earlier.
Ignore changes unrelated to likho.

I don't know what to ignore or not, I'm just looking at the PR. Ideally the PR would be synced with master. I suppose @sprunk will figure it out?

> Indeed, after the problems with the armature I looked at other models and they worked like this, so I just went with it.
I remember trying some hierarchy but some pieces would disappear.

Which models? That should be fixed.

> Is it correct in the old model?
I didn't touch that and both models are aligned in blender.

image

Yes. But the old model isn't a dae. I wouldn't vouch for importing it to blender being a faithful recreation of how the model will actually work.

Anarchid:
>  The "base" piece has geometry. Models should have an empty root because that is convenient for moving things around in the script.

Dae importer always creates a root `Scene` piece that does this by default.

> The midpos is on the ground, when it should come from the middle of the unit, like with Odin.

This can be set in the metafile

bruno-cb:
> Which models?

Old likho, spherebots, spiders, detriment, scorpion and paladin.

> I wouldn't vouch for importing it to blender being a faithful recreation of how the model will actually work

How should one know how a model works if the related guides are outdated?

When I started working on this I found a BAR tool that converted all .s3o into .obj, which I don't think supports hierarchy.
Since my original plan was to animate with armature, everything could be one mesh.
When I moved into animation by script, hierarchy wouldn't matter since each piece had to be individually animated and, in this case, was independent from other pieces.

Can't hierarchy be created by script?

GoogleFrog:
Piece hierarchy is set by the model, and idk how it is done in dae. I wouldn't be surprised if a s3o converter doesn't support hierarchy. I don't know how hierarchy is set in dae, but Odin has one, so @Anarchid seems to know.

GoogleFrog:
image
WTF is going on? Why did I have to merge several conflict when I pulled just now?

GoogleFrog:
What is likho2024.dae for?

GoogleFrog:
It's confusing that the blinky lights aren't on the wing tips.
image

GoogleFrog:
How do the wing ribbons work without a piece hierarchy?
image
Oh

GoogleFrog:
Luckily it was fairly easy to put the pieces in the right hierarchy.

GoogleFrog:
image
I think the projectile is inside out.

PR #5170, opened 2024-01-25
🐛 Fix #3227, Raven OKP letting extra bombs through

### LINKED ISSUE #3227: Bomber OKP doesn't give enough time for projectiles to land
When bombers are attacking statics from high altitude with OKP enabled, they will start firing again before their first salvo has reached the ground, wasting shots.

sprunk:


GoogleFrog:
I was going to write some thoughts but instead found those same thoughts on the linked ticket.

> I only added Raven OKP to prevent the situations in which many bombers stack on one location and overkill a structure with a big blob of bombs. It is especially important that Ravens never fly over their target without dropping a bomb in cases where a bomb is required. This is why the leeway is so high in this case.

Is the entire change the increase to 50 frames, and what was the test scenario? Raven already has AI that removes attack commands from doomed targets if it has an attack command queued after it. If you area attack two mexes, then Raven should be more wary about overkill than if it only has one mex as a target. Raven not dropping a bomb against its only target seems pretty bad if there is any reasonable doubt of killing it. But also, OKP is ignored for units on hold fire in general now, so idk. I'll probably have to check this.

sprunk:
> Is the entire change the increase to 50 frames

I am unaware of there being any second order effects so hopefully yes.

> what was the test scenario?

I spawned a bunch of enemy mexes in a line, on flat land. Then I spawned a somewhat elongated rectangle of bombers (such that there would be a constant flow of bombers flying over each mex while the bombs are falling) and gave them a queue of attack orders on the mexes from the closest.
image
Before the change, a second bomb would be dropped by one of the later bombers right before the first connects. I increased the timeout until that stopped happening.

> Raven already has AI that removes attack commands from doomed targets if it has an attack command queued after it.

Yes, see above. The test scenario involved a queue but all mexes (also those who had further ones queued afterwards, i.e. not just the last) were double-bombed.

> Raven not dropping a bomb against its only target seems pretty bad if there is any reasonable doubt of killing it.

Makes sense. Note though that there are two checks, one for unqueuing the current order and one for the actual shot. Perhaps the second one (for the shot) could stay more trigger-happy.

GoogleFrog:
The case to avoid is as follows:
- Raven A fires at Lance
- Lance has 800 expected damage and 1000 health.
- Raven A hits Lance.
- Lance has 800 expected damage and 200 health.
- Raven B considers shooting at Lance, decides not to because expected damage > health.
- 50 seconds elapse, removing the 800 expected damage.
- Lance survives on 200 health.

PR #5179, opened 2024-01-30
⚙️ Make some awards less trivial to get

sprunk:
Difficulty factors:

* Spoils of War (reclaim) 0.06 → 0.30 Right now it gets awarded basically every game regardless of how little reclaim you actually do.

* Slow King (slow) 1 → 2 I recently got it from just the Cyclops sidearm in a lobpot and I didn't even spam Cyclops.

* Peacemaker (disarm) 1 → 4 Absurdly high damage which you can keep reapplying since it expires. Would still get awarded often if the multiplier was like 10.

### DIFF
--- a/LuaRules/Gadgets/awards.lua
+++ b/LuaRules/Gadgets/awards.lua
@@ -50,11 +50,6 @@ local cappedComs = {}
 
 local awardData = {}
 
-local basicEasyFactor = 0.5
-local veryEasyFactor = 0.3
-
-local empFactor     = veryEasyFactor*4
-local reclaimFactor = veryEasyFactor*0.2 -- wrecks aren't guaranteed to leave more than 0.2 of value
 
 local minReclaimRatio = 0.15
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
What does the reclaim factor mean? I think giving it out fairly often is good because it is a pretty important aspect of the economy, and can be otherwise easy to miss. What is the 30% of? Total metal produced?

sprunk:
Of average metal damage dealt by players, except yourself ([exact mechanics wikipage](https://zero-k.info/mediawiki/index.php?title=Award)). So if the average player has dealt 10000 metal's worth of damage you are currently eligible for the reclaim award after reclaiming a mere 600m, which will become 3000m after the change. In lobpots there's many players both driving down the average damage and feeding metal so it is trivial.

In 1v1 "other players except yourself" essentially means your own damage taken. So if you suffered 1000m of damage then you're eligible after reclaiming 60m (will be 300m). Keep in mind an ~equal battle where you lost 1000m produces 800m worth of reclaim (400m from you and 400 from the enemy) and you can increase it via being efficient (400m from you and more from the enemy) so 300m sounds reasonable. The change is mostly aimed at lobpot though so if you don't want to affect 1v1 it's fine and I can keep reclaim as-is.

GoogleFrog:
20% sounds better for reclaim. I think it should be a pretty common award, but not given out if everyone almost completely neglects it.

PR #5204, opened 2024-03-02
WIP: modern scars

Anarchid:
https://github.com/ZeroK-RTS/Zero-K/assets/3822768/06aa3971-ae36-43fa-8b65-d4f5858fbcf6


GoogleFrog:
My first impression is that the lines are too pronounced. It ends up looking too regular. The lines give a clear guide to distinguish the explosion "on top" from the ones underneath, which we want to avoid. Also at some point the lines lie too much about the shape of the underlying terrain, eg Comet Catcher Remake with its crater textures on completely flat ground.

This version dealt with the issue above better:
image

Also, the PR should avoid breaking things for pre-2399, so there needs to be conditional loading in resources.lua.

Anarchid:
I might have some time to do another attempt within March, so it's good to have critique. This is intended mainly as a just-in-case better-than-nothing thing at the moment. My supply of wind for this is very sporadic and largely driven by exploring the new tech itself.

The scars on the Folsom screenshot are just the first two scars without the other two. It is a distinct visual, but it is not a distinct set of assets or a configuration thereof. With some low chance (specifically, (2/4)^10) one could even have this output with the full set of 4 decals.

Lying about terrain seems to happen when they grow large enough. As far as i understand, the finalized version of this feature in the engine will allow size binning, so larger scars can be flatter progressively, while keeping relief on mid-sized ones; and small ones (for glaive-to-reaver size) can then be made nicely circular and possibly even more extreme in relief.

TL:DR

- if when i have next gust of wind, the engine provides size api: larger scars will be flatter, smaller ones rounder
- otherwise, all scars should be made flatter

sprunk:
> Doing the compat thing seems boring, so i won't think about it.

If we're still on old engine by the time this is close to completion I'll handle it

GoogleFrog:
There are two distinct things going on there, and it is unclear which is being referred to. Here are the answers to both:
- ZK has to work on pre-2399 engines regardless of which engine we are on. Whether we are still on the old engine doesn't come into it.
- I would like PR that makes us able to use post-2399 engines, as soon as possible, so the bugs don't stack up in engine development for months. But that does not seem to be this PR.

GoogleFrog:
image
They look fine when small, and the glow is cool.

image
image
image
image
As a medium scar they look too much like a place that has been converted to desert, rather than explosion. Or perhaps they could be read as an airburst explosion that cracked some brittle ground. This looks weird on inappropriate ground.

There should be more ejecta that indicates an explosion that happened at the ground level and that deformed the ground. See the old scars. It should be easy to read the explosion location from the implied motion of the scar. The lines are sort of doing this in the new one, but don't fit.
scar1

image
image
image
image
As a large scar they veer into lying about the terrain territory. At this scale they become the kind of terrain that would at least slow down vehicles. Their transparency is also a too high, making them look like the painted-on overlays that they are, and they don't deal well with stretching on cliffs. At this scale I think we want something quite simple that will mesh well with terrain.

GoogleFrog:
image
image
image

I bumped up the alpha of scar1, mostly at the top end. I think it works best. Mostly because it has the most movement to it, rather than being cracked ground.

image
Lots of lying about terrain from the normals here though.

image
The increased alpha probably warrants reduced normal strength, but idk how to do that operation in paint.net.

GoogleFrog:
image
The main problem with scar 1 is that it is offset. I read the explosion as coming from the red X, but it actually came from the map marker.

GoogleFrog:
Scar 1 on small to medium explosions is looking quite cool though.

sprunk:
Tasks:
- [ ] add this snippet to weapondef posts (modified as appropriate - GF says "for small aoe scars, use the same large images the medium aoe scars use")
```lua
local SMALL_SCARS = {1, 3, 5} -- indices in resources.lua, `scars` subtable
local MID_SCARS = {1, 2, 3}
-- etc

for id, wd in pairs(WeaponDefs) do
  if (wd.areaofeffect or 0) < 123 then
    wd.scarIndices = SMALL_SCARS
  else
    -- etc
  end
end
```

- [ ] check out engine default scars because these seem to be AI generated or otherwise deficient (flat normal isn't #808080, weird non-flatness in the corners)

- [ ] needs something of a higher resolution for the really big explosions (singu, nux etc)

GoogleFrog:
image
image
image
Here are some examples of the current scars working. They are fairly circular for simplicity, and don't try to take over the place with detail.

sprunk:
These scars have issues (flat areas aren't 0x8080FF or whatever was the neutral upwards vector; probably due to AI generation which is a minor problem in itself due to Steam policy). There seems to be nothing in the PR that we can take.

PR #5211, opened 2024-03-09
Added shieldball healthbars widget

danfireman:
image

I haven't done enough performance checking to be sure this is okay to be enabled by default, but a couple of hundred shields didn't cause an FPS drop I could detect.

### DIFF
--- a/LuaUI/Widgets/gui_shieldball_healthbars.lua
+++ b/LuaUI/Widgets/gui_shieldball_healthbars.lua
@@ -1,0 +1,283 @@
+function widget:GetInfo()
+	return {
+		name         = "Shieldball healthbars UI",
+		desc         = "Give a healthbar to shieldballs. Version 1.0",
+		author       = "dyth68",
+		date         = "2024",
+		license      = "PD", -- should be compatible with Spring
+		layer        = 11,
+		enabled      = true
+	}
+end
+
+local UPDATE_FRAME=7
+local glCallList = gl.CallList
+local glPushMatrix = gl.PushMatrix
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_shieldball_healthbars.lua
```diff
@@ -0,0 +1,283 @@
+function widget:GetInfo()
+	return {
+		name         = "Shieldball healthbars UI",
+		desc         = "Give a healthbar to shieldballs. Version 1.0",
+		author       = "dyth68",
+		date         = "2024",
+		license      = "PD", -- should be compatible with Spring
+		layer        = 11,
```
sprunk, line 8: why 11?

danfireman: Because copypaste, change to whatever you wish.

GoogleFrog:
I think I will have to run this locally for a bit.

danfireman:
The opacity, size and position of the healthbars are probably things to tweak. Also how much they should change size by zoom level.

GoogleFrog:
I was license baited for a while, but will now run it.

GoogleFrog:
It looks like the visual is hidden until I do /luaui reload, so I have barely seen it.

GoogleFrog:
image
This picture doesn't make any sense.

https://zero-k.info/Battles/Detail/1931178

danfireman:
I had apparently made a very important update I forgot to push which actually refreshes the list of shielded units. This explains both your observations.

Now pushed. 🤦

PR #5220, opened 2024-03-29
fix: Split outline drawing between over/underground phases like unit_shapes

moreginger:
fix: Split outline drawing between over/underground phases rather than double drawing it. This allows us to avoid world shaders (including water) for any part that is over ground and also to render with full opacity in both passes.

### DIFF
--- a/LuaUI/Widgets/gui_selectedunits_gl4.lua
+++ b/LuaUI/Widgets/gui_selectedunits_gl4.lua
@@ -257,11 +257,7 @@ function Init()
 	drawDepthCheck = options.drawdepthcheck.value
 	platterOpacity = options.platteropacity.value
 	outlineOpacity = options.outlineopacity.value
-	
-	if drawDepthCheck then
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -298,11 +294,20 @@ local function DrawSelectionType(vbo, preUnit)
...
 	glDepthTest(false)
-	glColorMask(preUnit) -- Only draw in preUnit, stencil later
+	glColorMask(preUnit) -- Only color in preUnit, stencil later
 	vbo.VAO:DrawArrays(GL_POINTS, vbo.usedElements)
 
 	-- Draw outlines
-	glDepthTest(not preUnit) -- No depth check for world
+	-- Draw the below ground half of the outline in the preUnit pass, and the above ground half in the postUnit pass.
+	-- This allows us to cheat the depth check for the below ground pass. 
+	if not drawDepthCheck then
```
moreginger, line 303: Results:

image

PR #5237, opened 2024-04-04
Bulk i18n of epicmenu widget options

moreginger:


moreginger:
This doesn't cover everything, but it's a good start.

GoogleFrog:
I wish translations didn't make the rest of development so inconvenient.

moreginger:
They are a pain, I've gotten used to them being annoying. One option is to have the English translation with the code, and not in a resources file. But you'd have to extract and store it for switching back to English.

LuaUI/Widgets/gfx_deferred_rendering.lua
```diff
@@ -80,18 +80,37 @@ local spGetSmoothMeshHeight  = Spring.GetSmoothMeshHeight
...
+i18nPrefix = 'deferredrendering_'
 options_path = 'Settings/Graphics/HDR (experimental)'
 options_order = {'enableHDR', 'enableBloom', 'illumThreshold', 'maxBrightness'}
 
 options = {
-	enableHDR      = {type = 'bool',   name = 'Use High Dynamic Range Color',  value = true,},
-	enableBloom    = {type = 'bool',   name = 'Apply Bloom Effect (HDR Only)', value = true,},
+	enableHDR      = {
+		type = 'bool',
+		alue = true,
```
sprunk, line 90: `alue`. Extra scary because there could be more errors like this lurking around the patch.

moreginger: Thanks for spotting this.

Maybe the answer is: run script on another copy of ZK, review differences?
For ~100 files there should be no change.

Or maybe review only added/changed lines and assume deleted are OK, if there's a way to do that.

LuaUI/Configs/lang/epicmenu.en.json
```diff
@@ -1228,6 +1260,7 @@
 	"unitmarkerzerok_disableall_desc" : "Mark nothing.",
 	"unitmarkerzerok_enableall" : "Enable All",
 	"unitmarkerzerok_enableall_desc" : "Marks all listed units.",
+	"unitmarkerzerok_unitslabel" : "unitslabel",
```
sprunk: Label name is probably technical and shouldn't be put into i18n.

sprunk:
Monoliths are hard to review and there's a bunch of issues as above. I think ideally there would be a bunch of commits with, say, 10 widgets per commit and then they would be easy to review and mergeable individually.

moreginger:
You can safely replace all the % with %% in that file.

> Monoliths are hard to review

Hmm I thought you'd prefer to get it all done in one.

I'm not sure that I have the staying power to create a series of 14 PRs. Especially after having done a conversion twice already. I appreciate it's a massive pain to review the large change.

I think maybe if I make a couple of changes to the script it might be best if you use it in batches you are happy with. Perhaps you can feel happier that it is acting in a sensible way? I can let you know the 2 widgets that it breaks. That will leave 10 or 20 that either partially or don't i18n, but we can chase those later.

LuaUI/Configs/lang/epicmenu.en.json
moreginger: I'm interested in what the labels are. I'll have a look later.

moreginger:
You might also want to review the technique. I wonder if it's better not to move the English translation to the resources file, @GoogleFrog didn't seem happy with it. I can imagine how that might work - providing the current value as a default when requesting translations, and store it as English if you don't have an English translation stored, but I think you'd want to consider the API carefully together. If I start trying to guess what you want then we'll be here a long time!

I guess that makes it harder to translate them though 🤷

LuaUI/Configs/lang/epicmenu.en.json
moreginger: Yeah this one is weird, and good spot. It looks like the "value" is the text that we'd want here. I searched for "unitslabel" and found no references, it could be constructed of course. But as a label I find it hard to imagine the string has a technical purpose.

```
	unitslabel = {name="unitslabel", type = 'label', value = "Individual Toggles", path = options_path},
```

Generally, labels are UI elements as you no doubt know, and the i18n of them is working in other places.

LuaUI/Widgets/api_grabinput.lua
```diff
@@ -35,7 +36,6 @@ options_order = {
 -- toggle grabinput easily.
 options = {
 	grabinput = {
-		name = "Lock Cursor to Window",
 		tooltip = "Prevents the cursor from leaving the Window/Screen",
```
sprunk, line 39: tooltip seems unhandled?

moreginger: Oooh, good spot. I'm not sure if options work at all, certainly my script isn't smart enough to do them properly.
I guess we really need to cut this into smaller pieces.

GoogleFrog:
@sprunk I hope you're not waiting on me for anything.

sprunk:
No, I'm waiting for chore mana from myself.

PR #5259, opened 2024-04-21
fix: Check the correct index for current avatar when removing a unit.

moreginger:


moreginger:
This is much more obvious when capturing a unit, since it can remain assigned to both teams indefinitely.

image

LuaUI/Widgets/gui_name_tags.lua
```diff
@@ -142,7 +142,7 @@ end
 
 local function _removeUnit(unitTeam, unitID)
 	unitsByTeam[unitTeam][unitID] = nil
-	if teamAvatars[unitTeam] and teamAvatars[unitTeam][1] == unitID then
+	if teamAvatars[unitTeam] and teamAvatars[unitTeam][2] == unitID then
```
sprunk, line 145: Tangential to the fix, but: `teamAvatars[x][1]` is always `x`, right? Why keep it in the table?

sprunk:
I knew the picture looked familiar.
image

PR #5265, opened 2024-04-29
fix: Change render order of various decorations: icons > names > bars > selection

moreginger:


moreginger:
Before:

image

After:

image

LuaUI/Widgets/gui_selectedunits_gl4.lua
```diff
@@ -7,7 +7,7 @@ function widget:GetInfo()
 		license = "GNU GPL, v2 or later",
 		-- Somewhere between layer -40 and -30 GetUnitUnderCursor starts
 		-- returning nil before GetUnitsInSelectionBox includes that unit.
-		layer = -30,
+		layer = -8, -- Draw below health bars and selection.
```
sprunk, line 10: selection is this one
```suggestion
		layer = -8, -- Draw below health bars, icons, and nametags
```

LuaUI/Widgets/gui_name_tags.lua
```diff
@@ -5,7 +5,7 @@ function widget:GetInfo()
 		author  = "fiendicus_prime",
 		date    = "2023-11-29",
 		license = "GNU GPL, v2 or later",
-		layer   = -9,
+		layer   = -11,
```
sprunk, line 8: -- above healthbars and selections, but below icons

sprunk, line 8:
```suggestion
		layer   = -11, -- above selections and healthbars, but below icons
```

LuaUI/Widgets/gui_selectedunits_gl4.lua
moreginger: Oops. Is it a good idea to enumerate all the options (which isn't a complete list here anyway). Or I suppose we could import constants?

sprunk (approved):

Issue #5197, opened 2024-02-21
Extract spirals from start unit spawner


sprunk:
#5089 added a spiral generator into the start unit spawner gadget. It sounds like generally useful math and doesn't really belong there. Extract it to a `Spring.Utilities` function, perhaps one that returns a closure that successively yields the spiral when called repeatedly.

sprunk:
Test cases.
```lua
local spiralWithExtraParameters = Spring.Utilities.GetSpiralGenerator(12, 34, { step = 3, startingDirection = "n",  clockwise = false })
local x1, z1 = spiralWithExtraParameters.get()
local x2, z2 = spiralWithExtraParameters.get()
local x3, z3 = spiralWithExtraParameters.get()

if  x1 == 12 and z1 == 34
and x2 == 12 and z2 == 31
and x3 ==  9 and z2 == 31 then
  Spring.Echo("OK")
else
  error("fail")
end
```
```lua
local defaultSpiral = Spring.Utilities.GetSpiralGenerator()
local spiralWithDefaultsExplicitlySpecified = Spring.Utilities.GetSpiralGenerator(0, 0, { step = 1, startingDirection = "n",  clockwise = false })

for i = 1, 20 do
  local xD, zD = defaultSpiral.get()
  local xE, zE = spiralWithDefaultsExplicitlySpecified.get()
  if xD ~= xE or zD ~= zE then
    error("fail")
  end
end
Spring.Echo("OK 2")
```

PR #5270, opened 2024-05-13
trying to make GetSpiralGenerator utility for the issue #5197

IcarusRTS:
I need some assistance with this one as right now, start_unit_setup that has this utility integrated causes Zero-K to freeze then crash.


sprunk:
Perhaps leave `start_unit_setup.lua` alone for now and just make sure the utility passes the provided tests (via a widget).

Freeze and a subsequent crash most likely means you have an infinite loop.

IcarusRTS:
I did actually make two test widgets prior to the pull request and the debug log gave zero errors ingame (unless there is something else I need to do for testing, if so, tell me)

sprunk:
> the debug log gave zero errors ingame

Check `infolog.txt`, Ctrl+F "test1_sg.lua" or whatever was the name of the file.

IcarusRTS:
> > the debug log gave zero errors ingame
> 
> Check `infolog.txt`, Ctrl+F "test1_sg.lua" or whatever was the name of the file.

Good news, both test widgets load in properly now.

Bad news is that there are new issues to tackle.

image

PR #5278, opened 2024-05-20
add guard selection rank override.

amnykon:
Adds an selection rank option for "Units issued a guard command are reduced to this selection rank."

### DIFF
--- a/LuaUI/Widgets/gui_selection_hierarchy.lua
+++ b/LuaUI/Widgets/gui_selection_hierarchy.lua
@@ -20,6 +20,13 @@ end
 local SUC = Spring.Utilities.CMD
 local CMD_SELECTION_RANK = SUC.SELECTION_RANK
 
+local isGuardCommand = {
+	[CMD.GUARD     ] = true,
+	[SUC.ORBIT     ] = true,
+	[SUC.ORBIT_DRAW] = true,
+	[SUC.AREA_GUARD] = true,
+}
+
 local spDiffTimers = Spring.DiffTimers
 local spGetTimer = Spring.GetTimer
 
@@ -72,6 +79,7 @@ options_order = {
 	'altBlocksHighRankSelection',
 	'doubleClickFlattenRankOption',
 	'controlGroupFlattenRank',
+	'guardRankOverrideOption',
 	'retreatOverrideOption',
 	'retreatingRankOption',
 	'retreatDeselects'
@@ -156,6 +164,15 @@ options = {
 		OnChange = function (self)
 			controlGroupFlattenRank = self.value
 		end
+	},
+	guardRankOverrideOption = {
+		name = 'Guard rank reduction:',
+		desc = "Units currently executing the guard command are reduced to this selection rank, if higher.",
+		type = 'number',
+		value = 3,
+		min = 0, max = 3, step = 1,
+		tooltip_format = "%.0f",
+		noHotkey = true,
 	},
 	retreatOverrideOption = {
 		name = "Retreat overrides selection rank",
@@ -308,6 +325,14 @@ local function RawGetFilteredSelection(units, subselection, subselectionCheckDon
 
 		if WG.GlobalBuildCommand and WG.GlobalBuildCommand.IsSelectionOverrideSet and WG.GlobalBuildCommand.IsControllingUnit(unitID) and (rank > WG.GlobalBuildCommand.SelectionOverrideRank) then
 			rank = WG.GlobalBuildCommand.SelectionOverrideRank
+		end
+
+		local guardRankOverrideRank = options.guardRankOverrideOption.value
+		if rank > guardRankOverrideRank then
+			local cmdID = Spring.GetUnitCurrentCommand(unitID)
+			if isGuardCommand[cmdID] then
+				rank = guardRankOverrideRank
+			end
 		end
 
 		if rank then


LuaUI/Widgets/cmd_orbit.lua
```diff
@@ -83,3 +120,20 @@ function widget:CommandNotify(cmdID, params, options)
...
+	WG.Orbit = { -- add compatibility functions to a table in widget globlals
+                IsUnitGuarding = function(id)
+			local cmdQueue = Spring.GetUnitCommands(id, 1);
+
+                        if (#cmdQueue == 0) then
+				return false
+			end
+
+			local cmd = cmdQueue[1].id
+			return cmd == CMD_ORBIT or cmd == CMD_ORBIT_DRAW or cmd == CMD_GUARD or cmd == CMD_AREA_GUARD
```
sprunk, line 134: If you only look at the 1st command, prefer `Spring.GetUnitCurrentCommand` since that doesn't allocate two tables. Also, conventionally unit and command ID vars are called `unitID` and `cmdID`:
```suggestion
                IsUnitGuarding = function(unitID)
			local cmdID = Spring.GetUnitCurrentCommand(unitID)
			return cmdID == CMD_ORBIT or cmdID == CMD_ORBIT_DRAW or cmdID == CMD_GUARD or cmdID == CMD_AREA_GUARD
```

LuaUI/Widgets/cmd_orbit.lua
```diff
@@ -21,12 +22,48 @@ local spTraceScreenRay = Spring.TraceScreenRay
...
+
+options = {
+	GuardOverrideOption = {
+		name = "Global overrides selection rank",
+		desc = "Units issued a guard command will be treated as a different selection rank.",
+		type = "bool",
+		value = true,
+		noHotkey = true,
+		OnChange = function (self)
+	        WG.Orbit.IsSelectionOverrideSet = self.value
```
sprunk, line 40: some bad whitespace here
```suggestion
			WG.Orbit.IsSelectionOverrideSet = self.value
```

LuaUI/Widgets/cmd_orbit.lua
```diff
@@ -21,12 +22,48 @@ local spTraceScreenRay = Spring.TraceScreenRay
...
+options_path = 'Settings/Interface/Circle Guard'
+
+options_order = {
+	'GuardOverrideOption',
+	'GuardRankOption'
+}
+
+options = {
+	GuardOverrideOption = {
+		name = "Global overrides selection rank",
```
sprunk, line 34:
```suggestion
		name = "Guard overrides selection rank",
```

LuaUI/Widgets/cmd_orbit.lua
```diff
@@ -6,7 +6,8 @@ function widget:GetInfo()
 		date      = "11 August 2015",
 		license   = "GNU GPL, v2 or later",
 		layer     = 0,
-		enabled   = true  --  loaded by default?
+		enabled   = true,  --  loaded by default?
+		handler   = true,
```
sprunk, line 10: I don't think `handler` should be needed?

sprunk (changes requested):

LuaUI/Widgets/cmd_orbit.lua
amnykon: I thought WG required it. Testing showed it didn't, so removing it.

amnykon:
Update PR to address Sprunk's comments.

GoogleFrog:
The PR lacks a description so I went looking for one.

> desc = "Units issued a guard command will be treated as a different selection rank.",
desc = "Units issued a guard command are treated as this selection rank, if override is enabled.",

I thought "this seems marginally more useful as an upper bound, rather than a strict override".

But then
```

		if WG.Orbit and WG.Orbit.IsSelectionOverrideSet and WG.Orbit.IsUnitGuarding(unitID) and (rank > WG.Orbit.SelectionOverrideRank) then
			rank = WG.Orbit.SelectionOverrideRank
		end
		```
So just make the description accurate?

**Comment by GoogleFrog** [comment]
Hold up.

Why was the orbit widget modified at all? This should be a setting under selection rank, and it should work on guard regardless of whether the orbit widget is enabled. The WG-ness of the check for a guard command is redundant, as it doesn't ask the orbit widget anything. This entire PR should be in the selection rank widget, because the new option does nothing with selection rank disabled, and should do basically everything with orbit disabled.

**Comment by amnykon** [comment]
updated to address GoogleFrog's comments, and removed the bool option since guardRankoverrideOption set at 3 is the same as having it off.

**Comment by GoogleFrog** [comment]
Removing the tickbox option is nice too. But now, shouldn't the default be 3?

**Comment by amnykon** [comment]
Updated default value to 3. As I just returned my new laptop due to too many crashes and boot issues, I am not able to currently test this change; however, it's simple enough that it should work, as long as the comma is still there.

**Review by sprunk — APPROVED** [review]
Already looks fine tho

File: LuaUI/Widgets/gui_selection_hierarchy.lua [hunk 5]
```diff
@@ -310,6 +334,12 @@ local function RawGetFilteredSelection(units, subselection, subselectionCheckDon
 			rank = WG.GlobalBuildCommand.SelectionOverrideRank
 		end
 
+		local cmdID = Spring.GetUnitCurrentCommand(unitID)
+		local isGuarding = cmdID == CMD_ORBIT or cmdID == CMD_ORBIT_DRAW or cmdID == CMD_GUARD or cmdID == CMD_AREA_GUARD
+		if isGuarding and (rank > guardSelectionOverrideRank) then
+			rank = guardSelectionOverrideRank
+		end
```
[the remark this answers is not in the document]
GoogleFrog: I also like how the refactor encapsulates the behaviour within a check for the setting that controls the behaviour. It makes it more readable.

GoogleFrog:
I have yet had a chance to test, but I suspect the code has a bug. OnChanged doesn't call when the value is read as the default from settings, and the default is 3. However, the local value is 0, so guard selection rank will be applied until the player toggles it from 3 then back. I would remove `guardSelectionOverrideRank` to prevent this and future bugs like it, just use `options.<setting>.value`.

amnykon:
Updated, like I said, I returned my laptop so please test this before merging it. As the funds for my new laptop is allocated elsewhere, I cannot test until ~ December...

LuaUI/Widgets/gui_selection_hierarchy.lua
```diff
@@ -310,6 +329,14 @@ local function RawGetFilteredSelection(units, subselection, subselectionCheckDon
 			rank = WG.GlobalBuildCommand.SelectionOverrideRank
 		end
 
+    local guardRankOverrideRank = settings.guardRankOverrideOption.value
```
sprunk, line 332:
```suggestion
    local guardRankOverrideRank = options.guardRankOverrideOption.value
```

sprunk (approved):

sprunk:
Tested, seems to work.

PR #5285, opened 2024-06-01
Fix useless and potential harmful loading redundancies

Helwor:
Remove duplicate path in widgetFiles:
When user want to replace zip widget with a local one, the widget will be loaded twice because listed twice in widgetFiles using different slash. This cause a second load of the same file for nothing (and will be discarded due to duplicate name) but also may be harmful if the widget rely on the discarded body instead of the first one (ie using global function set up at loading stage)

### DIFF
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -486,6 +486,20 @@ local function InitPlayerData(playerID)
 	return {team = teamID, spectator = spectator}
 end
 
+local function RemoveDuplicateFilenames(files)
+	local commons = {}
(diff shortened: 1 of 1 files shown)


sprunk:
If `VFS.DirList` lists the same file twice with different slashes it sounds like an engine bug. Here is an old engine ticket https://github.com/beyond-all-reason/spring/issues/139 Would be useful to say what OS you are on and which file (zip, raw) has which slash on that ticket.

The PR looks good as a workaround though, I'll try to find some time tomorrow to test.

PR #5294, opened 2024-06-08
fix side detection

Helwor:
myAllyTeamID can be 1

### DIFF
--- a/LuaUI/Headers/allyteam_selection_utilities.lua
+++ b/LuaUI/Headers/allyteam_selection_utilities.lua
@@ -4,9 +4,9 @@ local function GetLeftRightAllyTeamIDs()
 		return {}
 	end
 
-	local myAllyTeamID = 0 -- FIXME Spring.GetLocalAllyTeamID()
-	local enemyAllyTeamID = 1 -- FIXME assumes teams are 0 and 1
-	local myTeamID = Spring.GetTeamList(myAllyTeamID)[1] -- FIXME Spring.GetLocalTeamID()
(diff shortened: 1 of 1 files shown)


GoogleFrog:
What is this for? This utility is used in spectator panels and attrition counter, where it would seem good to have a consistent left/right arrangement. Local ally team is a bit meaningless for spectators.

Helwor:
the left/side arrangement in this function is determined in part by myLocalTeamID which was hardcoded as 0 in it.
myLocalTeamID was used to get the myTeamID, myTeamID is used to get the spectated player start box.
if myLocalTeamID is wrong and user doesn't have fullread mode (doesnt see other player), the starting box will not be retrieved, the function will return {0,1} as in AllyTeam 0 at left which can be wrong.
The current left/right arrangement is not currently consistent because of that, you can witness it if you spec some 1v1 with the spectator panel widget while spectating only one player (Settings/Spectating/PlayerView => Chosen player Only)
. That's why I made that fix

sprunk:
Looks slightly worrisome. The code was added in 4569851fb51c262c5d9ec2594bb7903d5a714460 and there was already:
```
local myAllyTeamID = 0 -- Spring.GetLocalAllyTeamID()
```
So it was known it should use that function but was left hardcoded anyway. Maybe there was a good reason for that? But maybe it's fine and we will discover it organically if something breaks.

Otherwise looks good.

Helwor:
using the fix atm, and speccing sometime not in full read and all good, no more wrong side

GoogleFrog:
Ok I get it now.

PR #5297, opened 2024-06-11
Enforce FFA Noelo through a hack

Arch-Shaman:
> FFA games are supposed to be noelo but there isn't anything in the server which forces them to be, at the moment. They will retroactively become no-elo at some point in the future. An unfortunate hack solution, but it is what it is.

This solves this issue by forcing Springiee to tell the server the game is noelo at game end.


GoogleFrog:
Huh, I thought it was read from the modoptions rather than controlled from the game. This should have an "automatic" mode rather than being enforced. Surely?

GoogleFrog:
Well I guess so. I just hope someone remembers this when it is time for the next round of hax.

PR #5311, opened 2024-06-18
added GetSpiralGenerator.lua utility

IcarusRTS:
Me and Niarteloc managed to make a working spiral generator. I have decided to make it so that the radius is based on how small or big the user wants spiral to be.


sprunk:
1) doesn't work, it's idempotent
2) if it worked it would implement a "proper" circular spiral, which is still useful, but the request was for a "square" spiral
image

PR #5317, opened 2024-07-05
Fixed idempotence for option random album spam

Helwor:
My bad I didn't realize that by implementing tracks consistent randomness it would break the sequence for rolling the albums. This is now fixed.

### DIFF
--- a/LuaUI/Widgets/snd_music2.lua
+++ b/LuaUI/Widgets/snd_music2.lua
@@ -254,7 +254,6 @@ end
 
 function widget:Update(dt)
 	if not initialized then
-		math.randomseed(os.clock()* 100)
 		initialized = true
 		-- these are here to give epicmenu time to set the values properly
 		-- (else it's always default at startup)
@@ -270,14 +269,15 @@ function widget:Update(dt)
 				defeatTracks    = VFS.DirList(dir .. 'defeat/', '*.ogg', vfsMode),
 			}
 		end
-		
+		if gameID then
+			-- update the tracklistName case: reload, random chosen at start
+			widget:GameID(gameID)
+		else
+			math.randomseed(os.clock()* 100)
+		end
 		trackList = includedAlbums[trackListName].tracks
 	elseif randomAlbumUseSeed == nil then
-		-- wait the second round for widget:GameID() to trigger (for the case: not in replay and not after a reload)
-		if gameID and not seed then
-			-- case: reload
-			widget:GameID(gameID)
-		end
+		-- case replay: widget:gameID() hasn't been triggered yet
 		randomAlbumUseSeed = false
 		continueAlbum = false
 	end
@@ -369,6 +369,7 @@ function widget:GameID(id)
 		end
 	end
 	randomAlbumUseSeed = false
+	continueAlbum = false
 end
 function widget:GameStart()
 	if not gameStarted then
--- a/LuaUI/Widgets/snd_music2.lua
+++ b/LuaUI/Widgets/snd_music2.lua
@@ -27,15 +27,25 @@ end
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 
--- see `widget:GameID` below. Used just for rolling tracks,
--- so that you get the same set of tracks in a replay.
-local seed = false
-local function SetRandomSeed()
-	if seed then
-		math.randomseed(seed)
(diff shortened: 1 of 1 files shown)


sprunk:
How about something like
```diff
+local wasAlreadyRandom = false -- default is denny
 
 OnChange = function(self)
 	local value = self.value
 	if value == 'random' then
+		if wasAlreadyRandom then
+			-- don't reroll
+			return
+		end
+		wasAlreadyRandom = true
 
 		SetRandomSeed()
 		local r = math.random(#self.items - 1)
```

Helwor:
So now you want to prevent from reclicking on random doing anything? I don't understand. Isn't it what you wanted to have the same rolling sequence happening with the option.OnChange ?

sprunk:
I think we understood "idempotent" differently. Here's a bigger description in plain words to avoid misunderstandings.

I want OnChange not to do anything if "changing" option back to the same option, for example by clicking the button again when it is already selected. E.g. from 'random' to 'random', or from 'denny' to 'denny':
* no rerolling current track
* no rerolling current album (for the 'random' option)
* further roll sequence (after current track) doesn't change
Basically, equivalent to "if same option then return".

If "really" changing option (e.g. from 'random' to 'denny', or from 'denny' to 'random', or from 'denny' to 'superintendent') then:
* album gets rolled if changing into random. It can roll anything and doesn't have to follow any specific sequence
* if you get the same album (for example changing from 'random'-rolled-as-denny into 'denny', or vice versa) then I don't mind either way whether the current track rerolls or not. Your call

If the player never went into settings to change anything, then it's desirable if:
* after `/luaui reload`, if using 'random' since the start of the game, the same album is chosen as was before the luaui reload. Track order can change
* if watching a replay, the same album is chosen as in the real game (if using random), and the same track sequence is rolled
* if the player changes settings during a game then both of these are invalidated and anything is fine to roll.

sprunk:
All of those are just high-quality polishing, it's all low-priority and the current state of the widget is already very good. So don't feel like this is a request and don't feel compelled to do anything.

Helwor:
Ok, this makes more sense to me as an improvement,.
The only difficulty is to get the gameID after a /luaui reload, since widget:GameID will not trigger again.
The best solutions I find for that is
1. make the engine set the key 'id' in Game table but I have trouble to test it myself, I didn't find yet how to compile the engine using Windows 8.1. But that would be useful to have it for other purpose aswell
2. Make widgetHandler set key id in 'Game' table
3. Make the widget set a temporary 'RandomMusicSeed' key in the config with Spring.SendCommands('TSet RandomMusicSeed num')'
(not setting gameID because value set with TSet still persist until lobby shutdown and might be wrong if the widget got deactivated and a new game is started then after)

sprunk:
gameID becomes available via `Spring.GetGameRulesParam("GameID")` after the event
https://github.com/ZeroK-RTS/Zero-K/blob/7f3afed7a3c784c4748c578ec9e3819bc41bc86f/LuaRules/Gadgets/game_id.lua#L17-L18

Helwor:
Ok, I think here it should work as you wanted, except about:
"after /luaui reload, if using 'random' since the start of the game, the same album is chosen as was before the luaui reload. Track order can change".
I think the user would just want to stay on the current album playing in any case in case of a reload so I Just write down the current album playing along with the gameID. What do you think?

sprunk:
> I think the user would just want to stay on the current album playing in any case in case of a reload so I Just write down the current album playing along with the gameID. What do you think?

Sounds good. I'll test soon (probably on the weekend)

sprunk:
Seems like rolling tracks within the same album and peace/war condition is now broken?
* album is set to 'denny' and I didn't touch it during a game.
* I keep getting the same one peace track over, and then when combat starts I keep getting the same war track (and if it reverts back to peace it's the same track as previous again). Not just a coincidence because I got the same one like 4-5 times in a row.
* the same thing happens when setting music volume to 0% (via the slider at the top) and then back above. Previously this re-rolled track but now it keeps giving me the same one right back.

It would be sad to waste work but I think I would be happy with just an "if option already set then return" check on top of the current code.

Helwor:
damn, ok I will look into it

Helwor:
That was just that, a simple oversight

sprunk:
Looks like the widget is now feature complete. Thanks!

PR #5321, opened 2024-07-11
Add Team controller to Cheat Sheet

Helwor:
And allow to update the global command bar button on widget reload, instead of making a new one.


sprunk:
When spectating an AI vs AI game, activating `/cheat` to make the menu show up automatically assigns me to some team.

Here's a startscript, drag-n-drop onto `spring.exe`.
[script_6v6.txt](https://github.com/user-attachments/files/16184656/script_6v6.txt)

Helwor:
I don't understand, this is not for spectating, this is for switching team during a game.  For example, taking control of your opponent AI or chicken.... this can be great for testing and making game with game master etc...

sprunk:
Yes, it's not for spectating but somehow breaks spectating anyway.

LuaUI/Widgets/gui_chili_cheats.lua
```diff
@@ -748,22 +797,33 @@ function languageChanged()
...
+			controlTeamPicker.items[i] = teamPicker.items[i]
+		end
+		if teamID == myTeamID then
+			controlTeamPicker.selected = i
 		end
 	end
 	teamPicker.tooltip = WG.Translate("interface", "cheatsheet_teampick_desc")
+	controlTeamPicker.tooltip = 'Take control of that team.'
 	teamPicker:Select(teamPicker.selected)
+	controlTeamPicker:Select(controlTeamPicker.selected)
```
sprunk, line 824: Maybe it's this line? It makes it so that when the menu is created (when you type `/cheat`) it automatically selects some team, which sends a `/team x` command and moves you from spectator to that team.

Helwor:
ah my bad then I'm gonna look into it

Helwor:
Fixed and improved

sprunk:
Looks good, nice functionality and no more bugs. A remaining task would be to wire it up to translation (i18n) like the rest of the widget since currently the new stuff is hardcoded to English.

image

Helwor:
Yes I thought I would need to know the translation to do so, therefore I didnt look into it

Helwor:
Also bring the Cheat Sheet window to front when toggled on.
By the way, the ukrainian translation is broken, because the filename is wrong: intreface.uk_UA.json.

sprunk:
Thanks! Split away the French translation fixes into a separate commit since they're unrelated

Helwor:
alright

PR #5322, opened 2024-07-13
Fix: Find better unit Team to switch to

Helwor:
Even though the user select some visible units to switch to, it might miss because of invisible selected among them, now fixed.

### DIFF
--- a/LuaUI/Widgets/gui_spectate_selected_team.lua
+++ b/LuaUI/Widgets/gui_spectate_selected_team.lua
@@ -74,11 +74,16 @@ end
 
 function widget:SelectionChanged(selection)
   if selection and #selection > 0 then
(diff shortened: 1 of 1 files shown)


sprunk:
Doesn't this guarantee you will switch away any time you select a mixed box? Sounds bad. @GoogleFrog thoughts?

Helwor:
Nop, the loop break at first visible unit, it can be the same team and then nothing happens.
It works exactly as before, but only skip the invisible units which is probably not what you want to select

sprunk:
So the situation that gets fixed is this, correct? Where currently it can randomly fail if enemy units are in the fog of war, but with the patch it will always correctly switch to the green team.
image

Helwor:
Yes exactly

PR #5325, opened 2024-07-14
Fix: wrong icons showing when selecting invisible units (Spectator/Ch…

Helwor:
…eat)

While it is not possible to retrieve the defID through spGetUnitDefID, spGetSelectedUnitsSorted can give them. By the way some optimization, saving multiple call of spGetUnitDefID aswell as the whole process widget:SelectionChanged()

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -1906,7 +1906,6 @@ end
 
 local function GetSingleUnitInfoPanel(parentControl, isTooltipVersion)
 	local selectedUnitID
-	local selectedUnitDefID
 	
 	local leftPanel = Chili.Control:New{
 		name = "leftPanel",
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -37,6 +37,7 @@ local GetHumanName = Spring.Utilities.GetHumanName
 local GetUnitCost = Spring.Utilities.GetUnitCost
 local GetDescription = Spring.Utilities.GetDescription
 local GetHelptext = Spring.Utilities.GetHelptext
+local spGetSelectedUnitsSorted = Spring.GetSelectedUnitsSorted
 
 local strFormat = string.format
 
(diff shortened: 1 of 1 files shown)


sprunk:
> While it is not possible to retrieve the defID through spGetUnitDefID, spGetSelectedUnitsSorted can give them

Sounds like engine bug if these functions are so inconsistent. And generally, the whole "selecting invisible units" behaviour is probably engine bug.

But if this patch also works as an optimisation for the normal case then I think we can take it. I'll test on the weekend.

Helwor:
I don't know if I would consider a bug being able to select and control units in god mode just because not having global vision.  But I can tell that's pretty useful as dev to have this ability.

LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
```diff
@@ -1026,7 +1027,6 @@ local function SelectionsIconClick(button, unitID, unitList, unitDefID)
 			newSelectedUnits = selectedUnitsList
 		end
 		spSelectUnitArray(newSelectedUnits)
-		widget:SelectionChanged(newSelectedUnits)
```
sprunk, line 1029: This was always redundant and doesn't depend on the rest of the patch, correct?

LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
```diff
@@ -1904,6 +1906,7 @@ end
 
 local function GetSingleUnitInfoPanel(parentControl, isTooltipVersion)
 	local selectedUnitID
+	local selectedUnitDefID
```
sprunk, line 1909: Unused?

LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
```diff
@@ -2806,7 +2809,18 @@ function widget:KeyRelease(key, modifier, isRepeat)
 	end
 end
 
-function widget:SelectionChanged(newSelection)
+function widget:CommandsChanged()
```
sprunk, line 2812: Why change from `SelectionChanged` to `CommandsChanged`?

Helwor:
hi, so SelectionChanged is used for modifying the selection on the fly by returning an array, and if it happens, SelectionChanged will be immediately called again. That's why I was saying in the description "aswell as the whole process widget:SelectionChanged()". In the opposite, CommandsChanged is called only once when the selection is finally made. You can check that by right clicking some units of your selection IIRC, SelectionChanged will be called twice. So using CommandsChanged prevent redundancies of the whole process.

sprunk:
Cool, thanks!

PR #5327, opened 2024-07-17
Fix: Complete team vision detection for spectator in fullread

Helwor:
UnitEntered/Left Radar, UnitEntered/Left Los deliver more information when spectating in fullread. So it become possible to know which unit got discovered by which ally team. Without those informations,  it is not possible to make sense of those calls when speccing FFA. It also save some useless checkings that would have to be done without.

### DIFF
--- a/LuaUI/cawidgets.lua
+++ b/LuaUI/cawidgets.lua
@@ -2387,30 +2387,30 @@ function widgetHandler:UnitStunned(unitID, unitDefID, unitTeam, stunned)
 end
 
 
-function widgetHandler:UnitEnteredRadar(unitID, unitTeam)
+function widgetHandler:UnitEnteredRadar(unitID, unitTeam, forAllyTeamID, unitDefID)
 	for _, w in r_ipairs(self.UnitEnteredRadarList) do
-		w:UnitEnteredRadar(unitID, unitTeam)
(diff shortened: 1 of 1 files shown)


LuaUI/cawidgets.lua
```diff
@@ -2387,30 +2387,30 @@ function widgetHandler:UnitStunned(unitID, unitDefID, unitTeam, stunned)
...
+function widgetHandler:UnitLeftRadar(unitID, unitTeam, forAllyTeamID, unitDefID)
 	for _, w in r_ipairs(self.UnitLeftRadarList) do
-		w:UnitLeftRadar(unitID, unitTeam)
+		w:UnitLeftRadar(unitID, unitTeam, forAllyTeamID, unitDefID)
 	end
 end
 
 
-function widgetHandler:UnitLeftLos(unitID, unitDefID, unitTeam)
+function widgetHandler:UnitLeftLos(unitID, unitDefID, unitTeam, forAllyTeamID, unitDefID)
```
sprunk, line 2411: Is this one correct? There are two `unitDefID`s. Looks like it was always broken and nobody ever noticed.

Helwor: Indeed no it isn't. I didn't even notice it, thinking it was correct and copy pasted

sprunk (approved):
Looks good, I'll test in the evening

PR #5329, opened 2024-07-18
Small mistake form the previous translation fix

Helwor:


sprunk:
Thanks, but translation fixes should generally be done through Transifex https://app.transifex.com/zero-k/zero-k/language/fr

PR #5336, opened 2024-08-04
Fix broken hovering when using scaled UI

Helwor:


GoogleFrog:
I think the fix should be in `function widget:IsAbove()` in api_chili, since that is where the rest of the uiScale shifts are.

PR #5338, opened 2024-08-14
Add link to development guide for new contributors

### LINKED ISSUE #5337: What's a way for new members to contribute?
Looking for something like a contributing.md file or similar, that lists how to set up/debug/test/folder structure/etc,

cktang88:
Fixes #5337 

New contributors might not be aware that wiki doc exists, so point them to the right direction (I ran into this myself haha).

I can also copy the wiki text directly and put in a `contributing.md` file but a single source of truth might be preferable so just linking for now.

Side note: The README has HTML tags and centered styling which seems to be very uncommon among Github projects, personally would love to use just vanilla Markdown and no styling but fine either way :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,12 @@ To play the game, you need to **[Download]** the lobby.
 
 [*» Learn how to import content directly into the engine.*][Importing]
 
+Development guide: https://zero-k.info/mediawiki/Zero-K:Developing
+
 </div>
 
 <br>
+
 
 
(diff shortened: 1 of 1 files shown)


sprunk:
IIRC the centered style seems to be because the previous contributor who touched the readme thought it would help the layout look better when viewed on a phone (and I wasn't bothered by it enough to raise fuss). If you (or anybody) expands the readme then feel free to make it normal.

PR #5364, opened 2024-09-27
Tech-K

GoogleFrog:
Switched attributes to use the generic system. It makes a few more tables but is easy for mods to hook into. Also added some new attributes that can be modified.

Then there is the Tech-K mod itself. I don't really claim that it is balanced, or that it will work when people go to tech level 8+. Will probably pull tomorrow for testing if there are people around.


GoogleFrog:
Spring.Utilities.GetUnitCost now includes cost multipliers, so many instances of GetUnitCost * Multiplier were removed.

sprunk:
All the supporting tech is good but the mod itself should probably be an actual mod

PR #5369, opened 2024-10-12
Fix crude playerlist PING_TIMEOUT

Arch-Shaman:
Previously it would only tag users who are 33 minutes and 33 seconds behind.


sprunk:
All other ping calculations seem to use seconds, perhaps change `PING_TIMEOUT` to seconds too instead?

sprunk:
Thanks!

PR #5370, opened 2024-10-15
correct flipped minimap mex placement

christianweinz:
when using the flipped map, the minimap mex placement is incorrect for non-quadratic maps. this pull request corrects it.


sprunk:
Nice, thanks.

PR #5372, opened 2024-10-21
Add pregame timer support

sprunk:
Requires infra, doesn't do anything without (singleplayer, or multi before the infra patch).


sprunk:
With infra support the ingame timer will count down to 0. See https://github.com/ZeroK-RTS/Zero-K-Infrastructure/pull/2995
image

Licho1 (approved):

Issue #5330, opened 2024-07-18
Custom music: which album?


sprunk:
Historically all the music was dumped into a music folder and that was it. Nowadays there are two albums, Denny and Superintendent, hardcoded into the music widget and you have to pick one. Where does one put custom music and how does one pick the "custom" album?

Helwor:
How it works currently:
The Denny's album is in sounds/music/ while the SuperIntendent's album is in sounds/music/ost23_uf (not well organized). 
Subfolders "war"', "peace", "briefing", "victory" and "defeat" are checked and used to run random track in whichever chosen album (subfolder 'loading' is used elsewhere, not exactly sure where, but the one of  SuperIntendent is the music at Zero-K Interface loading).

If the user want to put own music, those subfolders has to be filled locally.
The option 'Use Included Album' will add the tracks filenames of the game of the corresponding albums.
In that case, if some files have same name there will be duplicates, upping the random chances of that file to be rolled and the local version will be played no matter what.
Disabling that option will allow the user to have his own exclusive set of music, but will be constrained in the two hardcoded album storage place mentioned above.

sprunk:
Thanks for the write-up. With #5384 merged, now the remaining task is to somehow document to players how to create a custom album, ideally in a way they could discover themselves. Perhaps Steam or chobbyla could create an example album with the correct folder hierarchy in their install folder?

Licho1:
Yes it could be added to both. Perhaps adding it to steam builder is cleaner?

sprunk:
Sounds good. The correct folder hierarchy looks like this
```
(steam install folder)
└──── sounds
       └──── music
              └─── example_custom_album
                    ├─── peace/
                    ├─── war/
                    ├─── victory/
                    ├─── defeat/
                    ├─── briefing/
                    └─── metadata.lua
```
And contents of `metadata.lua`:
```lua
-- Copy the "example_custom_album" folder and/or fill in the subfolders with music tracks.
-- There has to be at least one "peace" track, and the valid formats are .ogg and .mp3
return {
  humanName = "My Custom Album",

  -- if you're going to share to others it's probably good to fill those in
  author = "somebody",
  licence = "CC?",
}
```

PR #5382, opened 2024-11-27
Fix non default, non random album initial choice

Helwor:
trackListName has to be changed prior .tracks has been filled in, because options.OnChange will be triggered before initialization in Update. Ideally a bigger rewrite should be made to change the way tracks are registered, suggestion: -remove the 'use included option'
-keep the vanilla albums in the pool of  album, add whatever user extra album has locally


sprunk:
Thanks. Yea albums probably need a rethink at some point, here's a ticket #5330

PR #5383, opened 2024-11-28
increase max fade height

amnykon:
increases the max fade height. This allows a player (me) to see health bars when fully zoomed out.

Since unit_healthbars.lua uses GetVisibleUnits with the radar dot param as false, `/disticon` needs to be set to 10000 to see the bars fully zoomed out, as well as this setting.

If the param is set to true, the bars are under the radar dots anyway.

For the time being, I will be using gui_icontransition.

### DIFF
--- a/LuaUI/Widgets/unit_healthbars.lua
+++ b/LuaUI/Widgets/unit_healthbars.lua
@@ -187,7 +187,7 @@ options = {
 		name = 'Unit Bar Fade Height',
 		desc = 'If the camera is above this height, health bars will not be drawn.',
 		type = 'number',
-		min = 0, max = 8000, step = 50,
+		min = 0, max = 10000, step = 50,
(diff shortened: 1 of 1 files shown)


amnykon:
10000 is enough of to always be in n when zoomed out. Unless there is a way to zoom out further. (I can try a higher map) It also matches gui_icontransition's max values.

amnykon:
I don't think a value option can have an infinite max.

amnykon:
Tried it on pinch point (32x20) and 10000 is not high enough; however, the bars would be too small enough to see at this level of zoom. Maybe If I had a larger monitor... currently I am good with 10000. I feel 8000 is a little too small. I may revisit after I add circle style health bars.

GoogleFrog:
I suspect this setting is barely used anyway.

PR #5384, opened 2024-12-01
Update: User can create own albums

Helwor:
- User can now add brand new albums of his own in sounds/music. Name of album is the name of the folder created by user.

- Fix duplicating filenames

- No more hardcoding, the widget will adapt to the game albums in the future aswell.

- useIncludedAlbums option is no longer needed, it was only useful to deactivate it to create own user albums from the place holders of the 2 (hardcoded) existing original albums paths, which was kind of a wonky way of doing so.

--
Reminder:
- creating album can be done by making a new directory in sounds/music/ that must contains at least a "peace" subfolder with at least one track .ogg or .mp3. All possible subs: "briefing", "peace", "war", "victory", "defeat".

- if the user reuse the existing original paths (sounds/music and sounds/music/ost23_uf) it will update or complete the corresponding albums (Denny and Superintendent)


LuaUI/Widgets/snd_music2.lua
```diff
@@ -40,35 +40,72 @@ local function SetRandomSeed()
...
+		albums[name] = {
+			tracks = tracks,
+			dir = dir,
+			humanName = humanName,
+		}
+	end
+	return albums
+end
+
+local includedAlbums = FindAlbums('sounds/music/')
```
Licho1, line 100: This now executes on different time than first widgetUpdate. It probably does not matter if you are not using stuff from other widgets

LuaUI/Widgets/snd_music2.lua
```diff
@@ -40,35 +40,72 @@ local function SetRandomSeed()
...
+
+local trackListName = includedAlbums.denny and 'denny' or next(includedAlbums)
+
+local trackList = includedAlbums[trackListName]
```
Licho1, line 104: You removed ability for users to pick tracklist subset. 
It could be kept and instead "custom" could be added to it?

LuaUI/Widgets/snd_music2.lua
Helwor: Yes it was filled up at widget:Update due to the fact that it needed to update the option useIncludedTracks to get the vfsMode the user want, which is no longer needed, and which complicate things because .
I don't see how another widget would change anything, unless it delete or copy musics to the folder meanwhile.
The albums filling can be done at widget Initialize if you want but better not at widget Update because the option that select album verify if the tracks exists.

LuaUI/Widgets/snd_music2.lua
Helwor: I didn't remove any ability for the user to pick album (I even fixed it in a previous PR). I don't understand where you get that from?

Licho1: I mean this switch: 

image

It seemed to work ok and according to code diff it is missing after this PR, or am I reading the PR wrong?

Licho1 (approved):

LuaUI/Widgets/snd_music2.lua
```diff
@@ -40,35 +40,72 @@ local function SetRandomSeed()
...
+		for _, musicType in ipairs(musicTypes) do
+			local files = VFS.DirList(path .. musicType .. '/' , supportedFileTypes, vfsMode)
+			Deduplicate(files)
+			tracks[musicType .. 'Tracks'] = files
+		end
+
+		local dir = path:gsub('sounds[\\/]music[\\/]', '')
+		local name = dir:gsub('[\\/]$', '')
+		local humanName
+		if name == '' then
```
GoogleFrog, line 82: The information in this if/else should be stored in config table near the top of the file (like the one that was deleted). Otherwise people looking to add tracks to the game will have to dig through the code.

PR #5385, opened 2024-12-01
to fix nimbus animation wrong gun smoke

XNTEABDSC:
When nimbus shot slow, problem revealed. (see silly wafr fix:)
fn Shot seems called before fn QueryWeapon so in shot you need to change gun before emit effect


sprunk:
Thanks, good catch!

PR #5387, opened 2024-12-03
fix initialization of icontransition.

amnykon:
Variables need to be set when widget initializes and when toggling back to dynamic mode.  The Initialization is important after the game begins (`/luaUI reload`).

### DIFF
--- a/LuaUI/Widgets/gui_icontransition.lua
+++ b/LuaUI/Widgets/gui_icontransition.lua
@@ -39,7 +39,7 @@ local renderAtPos = {}
 
 -- Forward function declarations
 local GotHotkeypress = function() end
-local UpdateDynamic = function() end
+local UpdateDynamic = function(shouldInit) end
 
 -- Localized Spring functions
 local echo = Spring.Echo
@@ -178,7 +178,7 @@ function GotHotkeypress()
 		target_mode = nil
 		kp_timer = nil
 		current_mode = "Dynamic"
-		UpdateDynamic()
+		UpdateDynamic(true)
 	else
 		waiting_on_double = true
 		kp_timer = spGetTimer()
@@ -194,7 +194,7 @@ function GotHotkeypress()
 	end
 end
 
-function UpdateDynamic()
+function UpdateDynamic(shouldInit)
 	local cs = spGetCameraState()
 	local gy = spGetGroundHeight(cs.px, cs.pz)
 	testHeight = cs.py - gy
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_icontransition.lua
```diff
@@ -208,12 +208,12 @@ function UpdateDynamic()
...
 	end
-	
-	if showing_icons and testHeight < options.icontransitiontop.value - tolerance then
+
+	if (shouldInit or showing_icons) and testHeight < options.icontransitiontop.value - tolerance then
 		spSendCommands("disticon " .. 100000)
 		showing_icons = false
 		drawIcons = true
-	elseif not showing_icons and testHeight > options.icontransitiontop.value + tolerance then
+	elseif (shouldInit or not showing_icons) and testHeight >= options.icontransitiontop.value + tolerance then
```
sprunk, line 216: Note that here it's `+tolerance` and in the one above it's `-tolerance` so this will still fail to apply anything if `testHeight` happens to be within the tolerance bound

LuaUI/Widgets/gui_icontransition.lua
```diff
@@ -369,7 +369,7 @@ function widget:Initialize()
 	target_mode = nil
 	kp_timer = nil
 	current_mode = "Dynamic"
-	UpdateDynamic()
+	UpdateDynamic(true)
```
sprunk, line 372: IMO instead of adding the bool param, do a manual check here without the tolerance bound.

LuaUI/Widgets/gui_icontransition.lua
amnykon: The problem I have is sowing_icons is false (after switching from "Off" mode to "Dynamic") or nil (on initial widget load after `/luaUI reload`) and I have options.icontransitiontop maxed out. Since this code is written to toggle based on zoom height, the code never triggers for me once one of those two conditions is meet.

I have not had issues with the tolerances, thought I can see that could be a problem. I see 2 fixes for it.
1) Remove tolerance altogether, as I don't think a dead zone of 50 really does much.
2) Remove the tolerance for initial load only. If I do this, I think adding a setModeToDynamic function may be better, and just copy the logic over without the sowing_icons and tolerance.

which approach do you prefer?

sprunk: `setModeToDynamic` function sounds good.

PR #5389, opened 2024-12-06
to_fix_bulkhead_stuck

XNTEABDSC:
bulkhead can't be undeployed when slowed.
GG.StartStopMovingControl used thresholdSpeed=0.3, which forgot the situation of slowed by 50%.
48*0.25*0.5/30=0.2.  When bulkhead try to StartMoving, StartStopMovingControl will call StopMoving.


sprunk:
Nice, thanks!

PR #5398, opened 2024-12-13
Invert health bars

amnykon:
sorry in advanced for the large PR.
may be easier to look at each commit separately
1) I removed the water tank, as archer has not used it in a long time.
2) Chagned barDrawer.AddBar to barDrawer.AddPercentBar and barDrawer.AddDurationBar
The new AddXbar functions primarily have 2 simple arguments. Had to add a 3rd to percent due to the color of Health
To reduce the number of parameters, messages and colors names needed to match, and to make this clearer, reordered them to match as well.
DE and TR translations were missing speed and heat translation.
DE's was added (proved by NRHD) TR's was left blank.
3) added the invert options using the same naming as messages and colors.


GoogleFrog:
- I'd like invert option in a subpath, as they spam the menu a bit.
-

sprunk:
Perhaps just a single 3-way radiobutton for inversion?
```
(o) default
( ) some healthbars are inverted so they are all 100% = good, 0% = bad (like health)
( ) some healthbars are inverted so they are all 100% = bad, 0% = good (like emp)
```

amnykon:
I prefer the sub menu. It's more configurable. A future feature may be menu actions that sets the inversion options.

LuaUI/Widgets/unit_healthbars.lua
```diff
@@ -873,43 +1071,42 @@ do
...
 		
 		--// SLOW
 		local slowState = GetUnitRulesParam(unitID, "slowState")
 		if (slowState and (slowState > 0)) then
 			if slowState > 0.5 then
-				barDrawer.AddBar(addTitle and messages.slow, 1, "slow", (addPercent and floor((slowState - 0.5)*25) .. 's'))
+				barDrawer.AddDurationBar("slow", (slowState - 0.5)*25)
 			else
-				barDrawer.AddBar(addTitle and messages.slow, slowState*2, "slow", (addPercent and floor(slowState*100) .. '%'))
+				barDrawer.AddPercentBar("slow", slowState)
```
GoogleFrog, line 1093: This looks like a bug to me. And not a simple 1-line change bug, because the bar should say 45% when 90% full.

sprunk: b5bc14fbf44a04dd3132901aac7d53eb89a1ed68

Issue #5394, opened 2024-12-12
Replay controls widget (gui_replay_controls.lua) misbehaves on ultrawide monitors


ThingsWillCalmUp:
RE: This discussion on discord: https://discord.com/channels/278805140708786177/432874986064052235/1316560761623023638

TL;DR:
The replay controls set the window Y offset based on a value calculated by window width, which causes them to hide beneath the minimap or other elements when ZK is running at ultrawide resolutions.

Someone should probably fix this properly and make the widget able to save its location after being dragged around. For now, I propose the following changes:

gui_replay_controls.lua::L119-120:
```
	local screenWidth, screenHeight = Spring.GetViewGeometry()
	local windowY = math.floor(screenHeight/8)
``` 

This will probably place the widget over the user's playerlist, but it should be draggable from there. Worst case as a dirty hack is to change the window's "right" value to something higher as well, slap it in the middle of the display, and rely on the user to relocate it every time, because it'd at least be somewhere grabbable at that point.

GoogleFrog:
Sounds like something to fix. Saving the location is a bit worrying since it might end up saved somewhere inaccessible. We'd probably want a type of window that doesn't save if it is too far behind another window.

ThingsWillCalmUp:
Okay, I put in a PR to do this. I THINK I did it right? Please let me know if I need to fix/change anything.

https://github.com/ZeroK-RTS/Zero-K/pull/5400

PR #5400, opened 2024-12-17
gui_replay_controls.lua: Calculate window vertical position based on …

ThingsWillCalmUp:
…window height, not window width.

Ref: https://github.com/ZeroK-RTS/Zero-K/issues/5394

Small two-line patch. This could be done better, but at least this will keep the replay controls from being offscreen or below the minimap with default UI layouts when the game window is an ultrawide aspect ratio. This will PROBABLY place the controls over the default playerlist, but it's draggable from that position.

### DIFF
--- a/LuaUI/Widgets/gui_replay_controls.lua
+++ b/LuaUI/Widgets/gui_replay_controls.lua
@@ -116,8 +116,8 @@ end
 
 function CreateTheUI()
 	--create main Chili elements
-	local screenWidth, _ = Spring.GetViewGeometry()
-	local windowY = math.floor(screenWidth*2/11 + 32)
+	local screenWidth, screenHeight = Spring.GetViewGeometry()
+	local windowY = math.floor(screenHeight/8)
 
 	local currSpeed = 2 --default button setting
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_replay_controls.lua
```diff
@@ -116,8 +116,8 @@ end
 
 function CreateTheUI()
 	--create main Chili elements
-	local screenWidth, _ = Spring.GetViewGeometry()
-	local windowY = math.floor(screenWidth*2/11 + 32)
+	local screenWidth, screenHeight = Spring.GetViewGeometry()
+	local windowY = math.floor(screenHeight/8)
```
sprunk, line 120: Doesn't Y start from the bottom for screen coordinates? I.e. shouldn't this be 7/8?

ThingsWillCalmUp: I think Y is top-down. When testing this change, the window starts as displayed now:
replaywidget

sprunk (approved):

PR #5401, opened 2024-12-19
Proposed balance changes:

Aquanim:
* Cutter cost 65 -> 50
 * Flail reload time 5.2 -> 4
 * Locust movespeed 212 -> 250
 * Phoenix cost 460 -> 540, Rearm time 13 -> 15
 * Cyclops turn rate decreased by 10 percent
 * Magpie rearm time 13s -> 12s
 * Bulkhead cost 230 -> 250
 * Felon speed 45 -> 39
 * Aspis cost 600 -> 700, Aegis cost 525 -> 600


Aquanim:
The thing about Flail is that presently it fairly efficiently three-hits Raven and four-hits Phoenix and yet it still feels like it sucks. It would need to get quite a bit more damage per shot to change those numbers.

PR #5404, opened 2024-12-20
Bring back old engine share behaviour

### LINKED ISSUE #5253: Check gifting a sneaky nano
There's a claim (trivial repro https://zero-k.info/Battles/Detail/1875393 ) that you can gift somebody a nano in a way that makes it build a thing. Engine recently stopped clearing commands when giving IIRC so perhaps it's just a matter of sharing while it has a build/assist command.

Perhaps also worth checking other weird venues when that is fixed (roam as a start state? the "give unit" command used in the dirtbag campaign mission?).

### LINKED ISSUE #5393: Gifting caretakers does not reset their action queue
1. Start a nanoframe (say a superweapon).
2. Make a caretaker.
3. Gift the caretaker to an ally.
4. The caretaker continues to spend metal on the nanoframe.

I understand the expected behavior is that the caretaker should at least pause. It currently continues to spend and the ally may not be aware they are funding another player's project.

sprunk:
Fixes #5393
Fixes #5253

### DIFF
--- a/LuaUI/cache.lua
+++ b/LuaUI/cache.lua
@@ -17,6 +17,7 @@ local GetTeamColor = Spring.GetTeamColor
 local SetTeamColor = Spring.SetTeamColor
 local GetVisibleUnits = Spring.GetVisibleUnits
 local MarkerAddPoint = Spring.MarkerAddPoint
+local ShareResources = Spring.ShareResources
 
 -- Block line drawing widgets
 --local MarkerAddLine = Spring.MarkerAddLine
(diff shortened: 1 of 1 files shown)


GoogleFrog:
The old behaviour wasn't the greatest. I'm mostly thinking of cases where someone goes AFK, then comes back, and all their queues are cancelled. I thought sneaky nano was dealt with by setting them to wait. Perhaps that was only factories though, and we could apply the wait thing to nanos as well. Constructors could be thrown in too.

sprunk:
I don't think AFK (or gadgeted transfers in general) was affected by the recent engine change so they should still have Wait applied and units should still not lose their queues.

The engine change only applied to widgets, essentially the crude/deluxe/tab playerlists method of sharing. Applying whatever lagmonitor does on transfer sounds good.

GoogleFrog:
Ah sorry, that makes a lot of sense. STOP on manual share is good.

PR #5423, opened 2024-12-29
Fix merlin middle shot

XNTEABDSC:
Makes merlin and grizzly no longer shot at odd middle when they have too much to shot.

### DIFF
--- a/scripts/amphassault.lua
+++ b/scripts/amphassault.lua
@@ -283,11 +283,7 @@ end
 
 function script.QueryWeapon(num)
 	if num == 1 then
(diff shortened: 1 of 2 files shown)


GoogleFrog:
What is the issue being resolved?

Do these units still deal correct damage and shoot in the right direction after the change?

XNTEABDSC:
e.g. high tech merlin, after merlin shot 40 missiles, new missiles are spamed at the middle of merlin, because gun_1>20.
now gun_1 will be set to 1 and loop

PR #5425, opened 2024-12-30
fix unit_windmill_control GG.MoveWindmill

XNTEABDSC:
Wtf I spend 2 hours on this.
This makes statictele cry when teleport structures.


XNTEABDSC:
or mb it should be `not (windDefs[unitDefID] and IterableMap.Get(windmills, unitID))` ?

Issue #4943, opened 2023-06-22
Dual weapon commander latency spikes


Porkch0p:
Upgrading Strike and Guardian Commanders to have a second main weapon can cause repeatable sim performance drops in combat. Reload rate does not appear to be important, as it happens even with slow weapons like a lightning gun or shotgun.

This is mostly a continuation of the strider performance issues in https://github.com/ZeroK-RTS/Zero-K/issues/4842 and shows the same behaviour. 

Problem is easy to replicate on a Steam Deck running AMD hardware, but doesn't have as much of an effect on an Intel+Nvidia desktop.

GoogleFrog:
@sprunk vindicator for script.Create() StartThread(MotionControl)?

Porkch0p:
Some other notes about this issue:
- It is hard to measure on a fast machine. The same amount of FPS slowdown at 2Ghz that happens with four basic strike commanders required at least 21 commanders when the same machine had its clock speed doubled.
- The anti-jitter changes in https://github.com/ZeroK-RTS/Zero-K/commit/5933447a68caf38fe6aa92d358ea0f0800de0d36 that improve re-aim speeds appear at least partly related to slowdowns, as dialing commander weapon re-aim times up reduces the effect.

PR #5443, opened 2025-01-23
Improve performance of strike/guardian weapons

Porkch0p:
Reworked AimWeapon threading to remove performance loss when Strike Commander fires (#4943) 
Smaller improvements for Guardian.

### DIFF
--- a/scripts/dynassault.lua
+++ b/scripts/dynassault.lua
@@ -195,7 +195,7 @@ local function RestoreTorsoAfterDelay(sleepTime)
 		end
 		if resetRestoreTorso then
 			resetRestoreTorso = false
-			counter = RESTORE_DELAY
+			counter = (sleepTime or RESTORE_DELAY)
 		end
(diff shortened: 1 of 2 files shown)


scripts/dynassault.lua
```diff
@@ -197,14 +184,28 @@ end
...
 
+local function RestoreTorsoAfterDelay(sleepTime)
+	local counter = (sleepTime or RESTORE_DELAY)
+	while true do
+		if counter > 0  and not Spring.GetUnitIsStunned(unitID) then
+			counter = counter - 100
+		end
+		if resetRestoreTorso then
+			resetRestoreTorso = false
+			counter = RESTORE_DELAY
```
sprunk, line 198: Shouldn't it use the supplied sleep time?
```suggestion
			counter = (sleepTime or RESTORE_DELAY)
```
Although the time is never actually supplied so it's somewhat moot.

PR #5445, opened 2025-01-23
fix selections and cursortip crash.

amnykon:


sprunk:
How did this ever work? I don't think unit defs even have shieldPower (weapons do) so this should've been nil always and crashed for all units.

GoogleFrog:
Maybe the engine copies shield power to the unitdef, for laziness.

PR #5453, opened 2025-02-07
fix Chili.Object.AddChild

XNTEABDSC:
When use index, AddChild forgot to
```
children[hobj] = index
children[objDirect] = index
```


GoogleFrog:
Does anything use index currently?

XNTEABDSC:
layoutpanel use index to sort controls.
RemoveChild do nothing if not self.children[childDirect]. then add it again will cause "There is already a control with the name"

GoogleFrog:
Sorry, I mean does anything use the index parameter in `AddChild(obj, dontUpdate, index)`?

XNTEABDSC:
In zk, none.

But people may use it.

PR #5466, opened 2025-03-02
Japanese localization

syuumu200:


LuaUI/Configs/lang/common.ja.json
```diff
@@ -0,0 +1,81 @@
...
+	"select_all" : "全て選択",
+	"select_and_go_to" : "Select and go to",
+	"go_to" : "Go to",
+	"x%_done" : "%{x}%% done",
+	"append_to_current_selection" : "Append to current selection",
+	"unit_under_attack" : "is under attack",
+	"roster_factorycloak" : "ステルスボット",
+	"roster_factoryshield" : "シールドボット",
+	"roster_factoryveh" : "ロバーボット",
+	"roster_factorytank" : "タンクボット",
```
sprunk, line 70: Should vehicle factories also have `ボット` at the end? E.g. why not just `タンク` instead of `タンクボット`? Why not something more Japanese, for example `戦車`?

syuumu200: In an SF game, it is better to actively use katakana rather than kanji that feel more traditionally Japanese.

As you may know, Japan has three writing systems: kanji, hiragana, and katakana. Japanese people adjust the ratio of these scripts depending on the type of content.

Literature, government documents, and academic texts tend to have a high proportion of kanji, while content related to foreign cultures tends to use more katakana. When a softer, more feminine tone is desired, hiragana is often preferred.

Since SF is not originally a Japanese genre, using katakana is more appropriate. If kanji were used excessively in this translation, it would create an atmosphere similar to Hollywood movies that have a slightly mistaken perception of Japan, which could make it harder for Japanese players to immerse themselves in the game.

Therefore, my translation policy is to use katakana as much as possible instead of kanji.

That being said, I am still trying to understand where each translation appears in the game. There may be parts that I haven't fully translated yet or some mistakes. I'd appreciate it if I could test how the text is displayed in practice.

sprunk:
Thanks! This stable I'm adding Japanese to the ingame language picker so you'll be able to test ingame.

Also it would be good to use Transifex to provide further translations (ping me if you make an account there) https://app.transifex.com/zero-k/zero-k/

PR #5468, opened 2025-03-05
Fix unpredictable jump damage from excessive velocity on unit collision

cktang88:
Took a stab at fixing https://zero-k.info/Forum/Thread/38027, this approach just bounds max collision damage to the specified landing weapon. If there's no specified landing weapon, bounds the max val to 20% of health (arbitrary, happy to change this to whatever val is best).

If that's not ideal, I can also do it by clamping velocity as Googlefrog suggested in that thread.

NOTE: most of diff is from file formatting in line w the Lua style guide (first commit).

The commit to actually do the change is all in [this commit](https://github.com/ZeroK-RTS/Zero-K/commit/da0d13a64b3bf9069087b4de1ccb4efdadeace86)

Highly recommend split view in GH review UI for viewing formatting changes.

### DIFF
--- a/LuaRules/Gadgets/unit_fall_damage.lua
+++ b/LuaRules/Gadgets/unit_fall_damage.lua
@@ -45,6 +45,32 @@ local attributes                      = {}
 local unitWantedVelocity
 local unitAlreadyProcessed
 
+-- Table to store maximum landing damage for each unit type
+local maxLandingDamage                = {}
+
+-- Find landing weapons and their damage values for all units
+for unitDefID = 1, #UnitDefs do
+	local ud = UnitDefs[unitDefID]
+	if ud.weapons then
+		for i, weapon in pairs(ud.weapons) do
+			local weaponDefID = weapon.weaponDef
+			local weaponDef = WeaponDefs[weaponDefID]
+			if weaponDef and weaponDef.name and (weaponDef.name:find("Landing") or weaponDef.name:find("landing")) then
+				-- Found a landing weapon, store its damage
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_fall_damage.lua
```diff
@@ -291,26 +336,37 @@ function gadget:UnitPreDamaged(unitID, unitDefID, unitTeam, damage, paralyzer, w
...
 		GG.AddGadgetImpulseRaw(unitID, vx, vy, vz, true, true)
-		
+
 		if env and env.script.StopMoving then
 			env.script.StopMoving()
 		end
-		
-		local damgeSpeed = math.sqrt((nx + tx*TANGENT_DAMAGE)^2 + (ny + ty*TANGENT_DAMAGE)^2 + (nz + tz*TANGENT_DAMAGE)^2)
+
+		local damgeSpeed = math.sqrt((nx + tx * TANGENT_DAMAGE) ^ 2 + (ny + ty * TANGENT_DAMAGE) ^ 2 +
```
GoogleFrog, line 356: This newline (and the one in UnitPreDamaged) seems needless.

GoogleFrog:
My concern is that this PR seems to bound all fall damage.

I like most of the whitespace changes, think that all the space before `= {}` is a bit much, but am neutral on it. I don't like the two extra newlines for line wrap.

GoogleFrog:
No response, and the PR has the basic issue of capping fall damage. Most units should die when launched all the way across the map, regardless of whether they hit the ground or an enemy unit.

PR #5470, opened 2025-03-08
Add trails to plasma cannon and artillery projectiles

Thorneel:
Adds glow particle and trails to plasma cannon projectiles
Those with large AoE have additional thin trails
Bertha trail is replaced by one with a longer straight part that transitions to a fiery trail
DRP is given long trails

https://github.com/user-attachments/assets/580a4164-3846-4add-a9d7-69720596293e

https://github.com/user-attachments/assets/e07e0a5d-bdf7-454a-9299-45f7843b3c02

https://github.com/user-attachments/assets/885a5d19-eed5-42a8-a476-d44cb490ad3c

https://github.com/user-attachments/assets/eed41b15-f507-4478-9d04-067ffa572c19

https://github.com/user-attachments/assets/eeacc018-8f3d-42ac-b606-77cb7351698c

https://github.com/user-attachments/assets/e00d2f9f-9e89-47f5-b476-b38915e211ef

https://github.com/user-attachments/assets/acbb555c-31d7-4bd2-84d3-3a856cb6ef31


cktang88:
love this!

units/tankheavyarty.lua
```diff
@@ -107,6 +108,7 @@ return { tankheavyarty = {
       noSelfDamage            = false,
       range                   = 1160,
       reloadtime              = 0.333,
+      rgbColor                = "0.5, 1, 0", -- TODO why doesn't this work, it seems to work in modularcomms/weapons/heavymachinegun_disrupt.lua
```
sprunk, line 111: commas?

Thorneel: goddammit
yes that was it, thanks
I will remove that one for the moment, getting the right look will take more time

GoogleFrog:
I mostly prefer ThornEel-SFX-more-tweaks. The trails there are a bit on the thin side though, especially for Crab and Desolator. I'd have to fiddle around with them over an afternoon. The main differences are:

1. More tweaks has glow around the main projectile. I think this works quite well.
2. The base PR has long trails for some long ranged artillery with slow projectiles (see below). The trail fades out just slowly enough to look more like a static piece of UI than something related to movement.

spring_KzV47iBEHk

I think it is also worth considering Nimbus, Ripper, and DRP. Nimbus would need a cheap version of the ceg.

GoogleFrog:
Here is my once-over ingame.
- Bulkhead - The head looks like it is detaching from the tail.
- Sling - Trail might be a bit short.
- Thug - Looks fine, maybe a tad short too.
- Envoy - Good. Doesn't have detachment.
- Shogun - Good length and falloff smoothness.
- Hermit - Trail is a bit short and detaches.
- Crabe - Looks good. Maybe a bit wider.
- Cerberus - Good, could maybe be slightly wider than Shogun.
- Emissary - Perhaps a bit short.
- Minotaur - Good. For some reason this doesn't detach, where other trails of this length tend to.
- Tremor - Good.
- Cyclops - Maybe a bit too wide. There also seems to be a gap (but not jittery detachment) between the trail and the head. Other trails look like they fall off from the projectile, whereas this one has a second peak of intensity about 1/3 of the way down the tail from the head.
- Desolator - Good.
- Ravager - A bit too short?

Detachment is when there is a gap between the tail that seems to jitter back and forth. I remember having a lot of trouble with this for Buoy.

The Bertha effect looks good in a vacuum, but I am not sure about removing the big visible smoke. The Bertha exists beyond the current battlefield, so having some distinctive smoke to look out for helps determine the cause of a mysterious explosion. It could have smoke and the trail. DRP could also have the new Bertha trail (no smoke).

GoogleFrog:
Bulkhead and Minotaur with alpha 0.81 instead of 0.01.
image
image
I see what is going on. The glow of the projectile can be used to fill in the tail gap:
image

GoogleFrog:
Full Old vs New.
image
image

GoogleFrog:
The shafts stack up and create a weird effect at the top of DRP trajectory. They're also jittering a bit in general.
image

GoogleFrog:
The remaining things are minor, with the biggest being a tweak of Cyclops and Crab. In doing DRP I saw that there are quite a few extra shafts being used to fix problems with longer trails.

The other thing to consider is whether to put a cheap version of the effect on Nimbus, and perhaps Reaver and Stardust. I think it is worth experimenting with. Also Ripper might look more consistent with other plasma if it had a trail, although it would be barely visible due to the speed of the shot.

Thorneel:
image
New(2) Bertha trail

Thorneel:
image

GoogleFrog:
spring_4AgiPqBBo1

DRP jitters too much.

GoogleFrog:
- [ ] Hermit
- [ ] Bulkhead
- [ ] Sling
- [ ] Thug
- [ ] Ravager
- [ ] Ripper
- [ ] Emissary
- [ ] Minotaur
- [ ] Cyclops
- [ ] Tremor
- [ ] Crab
- [ ] Envoy
- [ ] Shogun
- [ ] Cerberus
- [ ] Desolator
- [ ] Bertha
- [ ] DRP
- [ ] Glaive
- [ ] Reaver
- [ ] Nimbus
- [ ] Commander HMG
- [ ] Stardust

GoogleFrog:
Thanks @Thorneel, it all looks really good. I tweaked a few things and pulled.
- The smaller trails were looking a bit stumpy, so I reduced their size reduction and sometimes gave them an extra frame of life. This makes them taper more smoothly, rather than abruptly end.
- I felt like the medium and heavy trails deserved a bit more weight. So Envoy, Emissary, Minotaur, and especially Crab and Cyclops.
- I fixed the jitter just behind the head of the trail for Shogun, Cerb, and DRP. This uses a trail that extends in front of the projectile, but that starts off invisible until it falls back to the right spot to plug the gap.
- I culled about half of the particles from Tremor without changing all that much visually, and added an even more culled version of the trail to Nimbus and other heavy machine gun projectiles.
- I added a simple trail to Ripper, which looked fine with no adjustment.
- I went back to my rework of the original Bertha style, because the new one looked a bit thin and small.

PR #5481, opened 2025-04-12
Revert "Fix merlin middle shot"

GoogleFrog:
Reverts ZeroK-RTS/Zero-K#5423


GoogleFrog:
image
The weird hax is in place to make the beams hit either side of the attack command. Some other fix will be required for Grizzly.

Issue #5498, opened 2025-04-30
When smooth icon zoom is enabled, switching between Team color options ingame does not update unit colors


Azim-Palmer:
**Steps to reproduce**

1. Be ingame
2. Ensure `Enable Toggle Smooth Icon Zoom` is selected (Settings/Graphics/Unit Visibility/Icon Zoom Transition)
3. Open the `Team Color Mode` menu (Settings/Interface/Team Colors)

4. Zoom out, observe the color of the team icons

![Image](https://github.com/user-attachments/assets/0f12b470-ebd8-48f3-b501-62b65dedba63)

5. Zoom in, observe the matching color of the team icons

![Image](https://github.com/user-attachments/assets/fbe8fbe8-2c71-4fa3-96ce-83ddf438576b)

6. Zoom back out
7. Change the Team Color Mode to `ColorBlind`
8. Observe the icon color changes while zoomed out

![Image](https://github.com/user-attachments/assets/00d70bd3-ea73-440e-8995-6e033f1d9f6d)

10. Zoom in, observe the icon color is the same as it was previously

![Image](https://github.com/user-attachments/assets/8f911ad1-f7f9-4e85-8c7e-9cc8517669c9)

**Expected**
Icon colors should match the selected Team color mode irrespective of zoom level

**Actual**
Icon colors are set on loading and not updated when the setting is updated

**Use case**
I'm colorblind so when I just need to quickly figure out if I should be killing something I have the Team Color modes set to hotkeys and switch to a simple one, then switch back if I need to co-ordinate with allies.

**Dev Notes**
I think the issue is that the team colors are set in `gui_local_colors.lua` using `Spring.SetTeamColor`, the smooth transition stuff is in `gui_icontransition.lua`, the color is cached on unit creation in `addUnitIcon`

**Potential solutions**
1. I see there's 'event' functions like `widget:UnitCreated' but don't know if there's one for TeamColor setting etc that could be hooked into?
2. Is there a way to set a callback/register an event in gui_local_colors to which the gui_IconTransition widget could connect?

sprunk:
> there's 'event' functions like `widget:UnitCreated' but don't know if there's one for TeamColor setting etc that could be hooked into?

Custom events can be defined, e.g. here's UnitResurrected https://github.com/search?q=repo%3AZeroK-RTS%2FZero-K%20unitresurrected&type=code

sprunk:
[...]

PR #5499, opened 2025-04-30
fix: #5498 When smooth icon zoom is enabled, switching between Team color options ingame does not update unit colors

Azim-Palmer:
Fix: #5498

Icons now correctly update color when Team Color mode is toggled ingame

- Added new widget handler `TeamColorsChanged`
- Minor refactoring to reuse code in `gui_icontransition.lua`

### DIFF
--- a/LuaUI/Widgets/gui_icontransition.lua
+++ b/LuaUI/Widgets/gui_icontransition.lua
@@ -359,6 +359,11 @@ local function DrawWorldFunc()
 	glDepthMask(false)
 end
 
+local function clearUnitIcons()
+	renderOrders = { {}, {}, {} }
+	unitDefsToRender = {}
+end
+
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 -- Event Call-ins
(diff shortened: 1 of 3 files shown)


GoogleFrog:
It looks like a good callin to add. On the widget itself, I'll pull whatever looks reasonable (on the understanding that someone tested it) because it isn't enabled by default.

LuaUI/Widgets/gui_icontransition.lua
```diff
@@ -370,7 +375,10 @@ function widget:Initialize()
 	kp_timer = nil
 	current_mode = "Dynamic"
 	UpdateDynamic()
+	widget:addUnitIconsForAllUnits()
+end
 
+function widget:addUnitIconsForAllUnits()
```
sprunk, line 381: Why not a local function? Putting stuff in the `widget:bla` scope is, in theory, for exposing interfaces for widgetHandler.

sprunk (approved):
Haven't tested but looks good (with one minor nitpick comment).

PR #5511, opened 2025-06-14
WIP : A more newbie friendly GUI selection mode

Azim-Palmer:
**Intent**
- Make it easier for people less intimately familiar with the game (like me) to find the right units in the selection panel. 

image

**Notes**
- code is very WIP, the intent was to do just enough work to enable posting several variations of the icons and ask devs & community to help land on a design, then implement
- has an option in the Selection GUI
- default icon size has been changed, ( `uniticon_size.value`) - this will need to be reverted
- construction borders on builders isn't present

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -1,4 +1,7 @@
 
+
+
+
 function widget:GetInfo()
 	return {
 		name      = "Chili Selections & CursorTip v2",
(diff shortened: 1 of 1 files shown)


sprunk:
The screenshot doesn't seem to show the icons in the selection panel but in the main worldview instead?

sprunk:
Does this mean this line is incorrect and shouldn't say "and icon"?
https://github.com/ZeroK-RTS/Zero-K/blob/0e4327208f0d3fb2ba3cba5123a3d9833a5b34f6/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua#L440

FrequentPilgrim:
WIP icon unit selection panel overlay

This was the prototype screenshot I thought showed the most promise in terms of unifying the Selection Panel, Factory Build Panel, and the Icons used when zoomed out.

PR #5516, opened 2025-06-23
New .wav files & edits to Unit's .lua files

FrequentPilgrim:
clean up left over files from the workflow of creating the new audio mix.


GoogleFrog:
I went through everything and made a list of issues. A lot of them are probably volume adjustments. Everything else is great.

- [ ] Wind reply too quiet
- [ ] Factory construction reply, perhaps a bit short and too quiet
- [ ] Rogue fire lacks oomph, much quieter than the missile hit. Could just be a volume thing.
- [ ] Glaive firing pitch sounds like it varies too much. Could be an interaction with the hit sound, as they are hard to tease apart.
- [ ] Phantom firing sound could be louder to match its impact.
- [ ] Gremlin actually sounds ok.
- [ ] Scorcher firing feels flat, like it is happening down a well.
- [ ] Ripper firing sound is barely audible. So the weapon overall is missing the 'thomp' part of the thomp-pow.
- [ ] Dominatrix sounds weird and messy. I recall it being more of a coherent wom-wom sound before.
- [ ] Badger impact may even be too noticable, for how spammy it is.
- [ ] Raptor sounds really weird. For some reason Gremlin works with this sort of sound, but it is too understated for Raptor.
- [ ] Magpie fire is very weird half the time. Very attenuated perhaps? This might be pitch mod causing beating on two simulatenous sounds? If we want pitch mod here, then magpie will have to get soundTrigger.
- [ ] Pheonix pitch mod is very noticable. Not exactly bad, but it stands out and could be reduced.
- [ ] Thunderbird works great. Maybe it has no pitch mod, would have to check how it works exactly.
- [ ] Odin sounds are great. The launch and impact sounds of its shields are the same as Badger, but this is a one-shot ability so warrant the more noticable sound.
- [ ] Blastwing may feel a bit like uniform noise when a lot of them explode at once.
- [ ] Gnat sounds off-pitch when it fires, and a lot of them firing can turn into noise too.
- [ ] Nimbus is too quiet.
- [ ] Hercules sounds appropriately weak, like Gremlin.
- [ ] Dagger is very hard to hear and has a pop at the start that sounds like clipping.
- [ ] Scalpel firing sounds overly attenuated. Flat.
- [ ] Halberd is perhaps a bit quiet, especially the firing sound.
- [ ] Lance is ok, but if it was changed, it could stand to gain a bit of its wider range back. Since it is an important event.
- [ ] Duck torpedo impact seems to be inaudible.
- [ ] Archer impact may have too much echo after the initial impact.
- [ ] Buoy is too quiet, especially the impact.
- [ ] Scallop fire sounds weird. Too brief and like the detail has been lost. Impact sound could be a bit too lound. Torpedo impacts are too quiet.
- [ ] Grizzly sounds really weird. I suspect it will be a gain mod thing.
- [ ] Placeholder impact lacks punch. Like the body of the sound has been removed.
- [ ] Firewalker firing sounds really weird. Probably pitch mod. Impact is quite weird too.
- [ ] Jugglenaut gravity gun sounds muted.
- [ ] Toad is probably fine.
- [ ] Ogre impact feels a bit quiet and lacking.
- [ ] Cyclops firing is too quiet, impact is probably even a bit too quiet too. The slowbeam overpowers the other things that are heard from Cyclops.
- [ ] Emissary firing is quiet and feels like it lost some of its length. Perhaps it is just hard to make out the kathum. Impact may be a it quiet too.
- [ ] Tremor firing overpowers the impact sound. The impact sounds too short and at some pitches it doesn't really feel like an impact. It feels like the firing sound might take even more attention than before? There is a lot going on behind it
- [ ] Hunter firing and impact are barely audible.
- [ ] Seawolf firing is a bit quiet for its role. Impact barely audible.
- [ ] Corsair impact is too loud, overpowers firing sound, which may be a bit soft.
- [ ] Ultimatum impact may need less pitch mod.
- [ ] Dante missiles sound overly attenuated. Heatray is a bit weird. The burst sounds a bit weird too.
- [ ] Merlin probably has a pitch mod problem. The burst generates a background hissy-swishy noise.
- [ ] Paladin lightning isn't loud enough for its importance. Sounds sort of muted. Scorpion and Knight seemed to work though. Laser could be stronger, similar to Lance comment.
- [ ] Detriment has the quick clipping sound on gauss impact, same as Dagger. The weapons don't sound like they do much, very muted. Manualfire and jump sounds good.
- [ ] Reef missile launch and impact are barely audible. Drone capture beams are very loud, and have a similar issue to Dominatrix.
- [ ] Newton gravity gun sounds muted too.
- [ ] Stinger sounds fine. Maybe a bit low pitch.
- [ ] Gauss has the gauss problem.
- [ ] Lucifer see Lance. For some reason Lucifer sounds a bit better though, but aren't they identical?
- [ ] Desolator is too muted. The heatrays sound ok when doubled. Overall lacks impact.
- [ ] Urchin launch is good (much better than Hunter), impact is ok, maybe too soft.
- [ ] Artemis impact could be louder.
- [ ] Cerberus shots sound very weird sometimes when they all hit at once. Probably a gain mod thing. Probably applies to Shogun too.
- [ ] Quake impact is too quiet.
- [ ] Shockley launch and impacts are too quiet.
- [ ] Zeno launch too quiet as well.
- [ ] Starlight burst shot is too quiet.
- [ ] DRP yellow too loud. Green slightly too quiet. Blue too quiet. Purple too loud.

GoogleFrog:
Many of the sounds in the PR were truncated and their volume was normalised along their duration. I think this is a backwards step, since weapons should have an initial peak volume and then a tail. Many sounds ended up feeling truncated. I put in longer placeholder attempts at Lance firing, Crab impact, and Desolator firing. The selection and OK sounds of Bandit felt a bit too attenuated, so I had a go from the base files.

Some remaining issues:
- [ ] I wasn't able to fix the Gnat problem, where 20 Gnats firing sound like a real mess, compared to stable ZK.
- [ ] I also wasn't able to fix the weird noisy swish at the end of Archer firing.

PR #5532, opened 2025-07-10
Fix "mobile factories" (IE athena) not having "factory waypoints"

Arch-Shaman:
This adds a janky state that switches athena and strider hub (?) between "queue" mode and command mode.
TODO: Test


Arch-Shaman:
May be missing files? 
- TODO: test

PR #5569, opened 2025-09-05
Sound Balance #3

FrequentPilgrim:
Iterations on sound balance.


FrequentPilgrim:
People have a harder time telling that there was a change, but the general feedback is still that some sounds are quite flat or insufficiently present in the world. The majority of the sounds seem like improvements this time. Eg people like the new Venom.
    • Teal AA is understated

- [ ] Redid slower teal laser to feel closer to base game and slightly tweaked the faster 10sps teal laser to have more presence

    • weapon/missile/rapid_rocket_fire is much too flat.

- [ ] Restored to original

a lot of weapons sound like under water weapons now, if i had to put a finger on it
swift attack rocket muffled guess that where the main ear effect comes from

- [ ] weapon/missile/missile_fire7 restored fidelity

reaver sound muffled

- [ ] Boosted fidelity, returned more bass

ripper sound very different but kind of good, however sounds less aoe like

- [ ] Weird results from fidelity. Googlefrog implemented a fix

cant hear razor during phoenix drop

- [ ] Boosted Razor Volume

gremlin doesnt sound like laser anymore

- [ ] New slower teal laser for gremlin and hercules. Made more sense to differentiate the slower firing laser

ravager impact sound seems too present sounds like mino lvl

- [ ] Adjusted the volume down to be in line with other sounds. Previous volume was an error.

need to get used to claw landing sound from badger

- [ ] This is the same sound as Bulkhead, Magpie, Sling, Puppy, Thug, Hermit, Claw. I would like to differentiate more of these impacts but I don’t have a great library to do that with. Claw is probably at the front of the priority here since it’s more spammy than the others with the largest range and the sound of a mine landing should probably be different than a plasma ball impacting. As it is I don’t think it’s terrible.

stinger seems a bit too quite

- [ ] Slightly boosted volume of stinger

Some more from me:
    • Sniper impact is hard to hear, and the firing sound is not noticeable enough either.

- [ ]     • Boosted sniper volume. Slightly adjusted detriment volume to keep it the same.

    • Claw firing is too flat

- [ ]     • Boosted fidelity of claw fire

    • Raven drop a little too flat

- [ ]     • Used a totally new raven dropping sound. Created a higher fidelity version of the old one as bomb_drop2.wav for A/B testing but I think the new one is higher quality and crisper. Hard to tell without testing.

    • I also agree that Ripper impact sounds really weird now

- [ ]     • Addressed

The sounds in this game got tiring because it all ended up quite muffled, and I couldn't tell so well when a Sniper was firing:

the main thing that stood out to me was Nimbus being hard to hear
if one starts eating away at your base, you want to hear it

- [ ] Boosted nimbus ajnd gave it some extra maxconcurrent

plus the dante/rev rocket launch sounding a bit flat,

- [ ] weapon/missile/rapid_rocket_fire Restored to original

the Ogre impact having quality issues still

- [ ] Redid ogre hit with more preserved bass frequencies

also Dante sounded weird, I'll try the most recent heatray

- [ ] had already been changed

and it was hard to hear what Raptor was doing

- [ ] Split up gremlin from raptor sounds even more. raptor is more present.

- [ ] Returned recluse impact to base game as It sounds better untouched.

FrequentPilgrim:
I made some tweaks to sounds.lua. A shakeup on maxconuccrent allowing raptor and nimbus to get a bit louder.

I also found a typo that was preventing rolloff from functioning. I've applied a new maxconcurrent accross the board of 12 instead of the previous default of 16.  I suspect this change will be mostly unnoticeable except for avoiding some of the worst layering situations. May need some exceptions for lasers but I'd like to see how this works.

The other notable change is applying the new working default rolloff as 1.2 instead of 1.0. This should also be mostly unnoticeable to players but give a little more separation from the surrounding soundscape when the player zooms in. May need some tweaking but 1.2 feels more natural to me so far than 1.0.

PR #5577, opened 2025-09-15
Update jumpbomb.lua

Toykom1:
ignore this mb

### DIFF
--- a/units/jumpbomb.lua
+++ b/units/jumpbomb.lua
@@ -21,9 +21,9 @@ return { jumpbomb = {
 
   customParams           = {
     canjump          = 1,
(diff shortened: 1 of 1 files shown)


units/jumpbomb.lua
```diff
@@ -21,9 +21,9 @@ return { jumpbomb = {
 
   customParams           = {
     canjump          = 1,
-    jump_range       = 400,
+    jump_range       = 10000,
```
sprunk, line 24: a tad too big
```suggestion
    jump_range       = 9800,
```

GoogleFrog, line 24: Bold.

PR #5584, opened 2025-09-27
Add max zoom option to COFC Camera

ykc1n:
Sets a maximum zoom-in level for the COFC Camera. 
image

image

VERY useful when playing with high zoom speeds. Can help prevent over-zooming.


GoogleFrog:
It should be called "Minimum zoom distance".

PR #5585, opened 2025-09-27
unit transport load range is 30 elmo

ykc1n:
Transports should now load most moving allied units when they are 30 elmo from them, instead of just following them around and constantly stopping.  At this value, charons should be able to consistently pick up moving glaives. Anything smaller and faster than that will still have trouble being picked up though (fleas, darts, daggers). Not really sure if that matters much though.

Also changed the code so its a bit more clear.

### DIFF
--- a/LuaRules/Gadgets/unit_transport_load_unload.lua
+++ b/LuaRules/Gadgets/unit_transport_load_unload.lua
@@ -21,6 +21,8 @@ end
 local airTransports = {}
 local lightTransport = {}
 local allowTransportCache = {}
+local transportLoadRangeElmo = 30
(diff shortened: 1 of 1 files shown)


LuaRules/Gadgets/unit_transport_load_unload.lua
```diff
@@ -21,6 +21,8 @@ end
 local airTransports = {}
 local lightTransport = {}
 local allowTransportCache = {}
+local transportLoadRangeElmo = 30
+local transportLoadRangeDistanceSquared = (transportLoadRangeElmo*2)^2
```
sprunk, line 25: why *2?

GoogleFrog: Descriptions rot.

PR #5608, opened 2025-11-26
my new unit_attributes_generic

XNTEABDSC:
a unit_attributes_generic system which is more flexible and more readable.

It split attribute handling into files at LuaRules/Configs/UnitAttributeHandlers

also separate unit death explosion and unit reload pause into 2 gadgets.

Added ability to handle
    burst(mult),
    burstRate(mult),
    sprayAngle(add),

Tested behaviors:
    Techk level up health, projectileMult, rangeMult, minSpray, death explosion, econ
    Jump,

### DIFF
--- a/LuaRules/Configs/unit_attributes_generic_handlers.lua
+++ b/LuaRules/Configs/unit_attributes_generic_handlers.lua
@@ -3,20 +3,31 @@ if not Attributes then
     Attributes={}
     GG.Attributes=Attributes
 end
-
+--[=[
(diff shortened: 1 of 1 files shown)


XNTEABDSC:
Most UnitAttributeHandlers files are generated by Grok and I'm sure that AI is more careful than me, and I think the commit can just be cut into 1 file per commit, or you want me to show what are got taken out from unit_attributes_generic.lua to handlers. I will do it if you really want me to do.

PR #5610, opened 2025-11-26
Better comm system

XNTEABDSC:
merged related datas about commanders to single files. (just find, move and Include, so they keeps origin structure, just in another files)

Tested with Chicken comm, start select, morph, weapon, shield, damage and range mod.


gamedata/modularcomms/chassises/assult.lua
```diff
@@ -0,0 +1,245 @@
+return {
```
GoogleFrog: Typo in file name.

GoogleFrog:
Also it should be indented with tabs.

XNTEABDSC:
some unnecessary changes are removed

Issue #4503, opened 2021-09-03
Award tooltips


Hawkybird:
Far as I can tell there's not really an in-game way to know what each award is actually awarded for. I've had to look up what the "Land Sweeper" award is via the code for example.

The end game award screen gives a hint with the stats, but doesn't completely outline it. The main menu profile tab lets you hover for the names, but doesn't tell you either.

I believe that showing what the criteria is would be a good feature, even just added to the main menu version's current tooltip.

PR #5616, opened 2025-12-03
Show award descriptions using tooltips on the end game award screen

strategineer:
ref: https://github.com/ZeroK-RTS/Zero-K/issues/4503

Award descriptions mostly taken from this wiki page https://zero-k.info/mediawiki/Award

The "Rage Inducer" and "Experimental Engineer" awards aren't mentioned anywhere and don't seem to be used but just in case I've given them a description as well.

screen_2025-12-04_16-33-29-040


LuaRules/Configs/award_descs.lua
```diff
@@ -0,0 +1,28 @@
+return {
+	pwn     = 'Always awarded to the player who dealt the most damage.',
```
Mach565:
```suggestion
	pwn     = 'Award for dealing the most damage.',
```

Mach565 (commented):
Just a suggestion to make it consistent with the other awards. Players don't need to know that those awards are always given out, just what they're for.

Issue #5503, opened 2025-05-10
"Quick selection bar", the GUI element, has no option to enable clicking through


Sgathec:
![Image](https://github.com/user-attachments/assets/5a96c506-4e62-48ab-9ab3-0fbf4ac641d0)
There's a GUI element called "Quick selection bar" that is on by default as far as I know. It includes commanders and constructors. It consists of buttons the player can click on to select them. However, there's usually some empty space beside the buttons and the game doesn't let one click through that space. I suggest making that possible by default and enlarging the buttons if necessary so the player is less likely to miss while trying to select things.

PR #5618, opened 2025-12-05
Implement options for allowing clicking through the Quick Selection Bar and the Command Panel

strategineer:
NB: The Selected Units Panel already has an option like this.

I put these new options right after the `background_opacity` option in each panel because the option in the Selected Units Panel is ordered similarly (after the `selection_opacity` option for that panel).

These options will be unset (can't click through the panels) by default preserving the existing behavior.

Implements https://github.com/ZeroK-RTS/Zero-K/issues/5503

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -389,7 +389,7 @@ options = {
 	},
 	allowclickthrough = {
 		name = 'Allow clicking through', type='bool', value=false,
-		desc = 'Mouse clicks through empty parts of the panel to act on whatever is underneath.',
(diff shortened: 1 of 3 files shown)


LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
```diff
@@ -389,7 +389,7 @@ options = {
 	},
 	allowclickthrough = {
 		name = 'Allow clicking through', type='bool', value=false,
-		desc = 'Mouse clicks through empty parts of the panel act on whatever is underneath.',
+		desc = 'Mouse clicks through empty parts of the panel to act on whatever is underneath.',
```
sprunk: "clicks" are a noun here so the sentence is already correct (ditto the description for the new options)

PR #5627, opened 2025-12-07
Fix unit images in selected units panel and tooltip stretching beyond what's expected when the unit has abnormally large extra energy production/use numbers

### LINKED ISSUE #4483: Icons stretch with large energy production/use numbers
Having a large number for the metal/energy production/usage will stretch the icon of the unit/building in question. This affects both the selection UI and the hover tooltip.

This seems to start to happen with 4 positive digits, such as a multiplier on a singularity reactor, or 5 negative digits.

Example image [here](https://imgur.com/a/H2kvz5R).

strategineer:
Fixes https://github.com/ZeroK-RTS/Zero-K/issues/4483

demo video:
https://github.com/user-attachments/assets/a5b36b24-629a-4868-abc7-90172df9c626

### DIFF
--- a/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
+++ b/LuaUI/Widgets/gui_chili_selections_and_cursortip.lua
@@ -1279,7 +1279,39 @@ local function GetImage(parentControl, name, initY, imageFile, iconSize, xOffset
 	return Update
 end
 
-local function GetImageWithText(parentControl, name, initY, imageFile, caption, fontSize, iconSize, textOffset, xOffset)
+local function UpdatePanelWidthsIfNeeded(caption, parentPanel, otherPanel)
+	if parentPanel.name ~= "leftPanel" or caption == nil then
+		return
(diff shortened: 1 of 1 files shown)


strategineer:
It needs a refactoring but I took another stab at this that I'm much happier with compared to what I had previously. I added a demo video to the PR description.

strategineer:
Ok, I think this is ready for review.

units/energysolar.lua
```diff
@@ -29,7 +29,7 @@ return { energysolar = {
   },
 
   damageModifier                = 0.333,
-  energyMake                    = 2,
+  energyMake                    = 200000000000,
```
GoogleFrog: You've left this in. This PR will probably want a squash and merge based on the relatively small size and large number of commits, so don't worry about rebasing to remove this change.

GoogleFrog:
image

I prefer this, even though it means very large costs will end up under the health bar.

PR #5643, opened 2025-12-26
Add Colemak layout

allentiak:
Based on #5626


GoogleFrog:
I looks good except that I dislike that there are whitespace changes.

sprunk:
Just get rid of them.

PR #5647, opened 2025-12-31
Fix unit model rescale

XNTEABDSC:
make `UnitDef.customParams.model_rescale` applied to all `UnitModelRescale` so tech-k's UnitModelRescale works properly for ships.
make obj_artefact use `model_rescale` instead of `UnitModelRescale`
Im not sure whether `offset` of `UnitModelRescale` should be scaled because I didnt see any usage of `offset`

### DIFF
--- a/LuaRules/Gadgets/unit_model_rescale.lua
+++ b/LuaRules/Gadgets/unit_model_rescale.lua
@@ -26,7 +26,7 @@ local INLOS_ACCESS = {inlos = true}
 
 VFS.Include("LuaRules/Utilities/tablefunctions.lua")
 local suCopyTable = Spring.Utilities.CopyTable
-local spGetUnitDefID=Spring.GetUnitDefID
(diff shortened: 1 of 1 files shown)


GoogleFrog:
I don't like what `UnitModelRescale` now means. Previously it mean "set the unit model to this scale". Now it means "multiply the input by a custom param and set the unit to that scale". The former is pretty direct and the latter is a bit arbitrary. Also, if I see `UnitModelRescale(unitID, 1)` I expect it to be resetting scale, or a noop if there has been no scaling.

The PR happens to fix an interaction between two things that want to scale models multiplicatively (tech-k and custom params), but only because there are only though two things. When a third thing appears it will break again and the solution will be even more complex. The solution for this already exists though, suggesting model scale should just be rolled into the attributes gadget.

GoogleFrog:
That's a good point and idk where else we would want multiplicative scaling. The artefact fix is also big.

PR #5648, opened 2025-12-31
Typedefs

XNTEABDSC:
Added typedefs file and .luarc.json for vscode Lua Language Server. Now Lua Language Server can find some actual bugs, and provides comments.


XNTEABDSC:
Now Im not sure about it as well... we can have typedefs files elsewhere (we can set .luarc.json), 
but .luarc.json need to be at root directory (or should we gitignore?). 
workspace file is to make scripts/ a individual directory for its .luarc.json to work (as far as I know) (yes its name needs to be changed). 
Should we have all this in Zero-K, or have a tutorial about how to set it up?

PR #5659, opened 2026-01-15
Fix wrong GL DepthTest param

Helwor:
That may disturb other widgets (especially Eco Drawer).

### DIFF
--- a/LuaUI/Widgets/map_edge_extension2.lua
+++ b/LuaUI/Widgets/map_edge_extension2.lua
@@ -861,6 +861,7 @@ function widget:DrawGroundDeferred()
 	--gl.DepthTest(false)
 	--gl.DepthMask(false)
 	gl.Culling(GL.BACK)
+	gl.Culling(false)
 
 end
 
@@ -898,6 +899,7 @@ function widget:DrawWorldPreUnit()
 	gl.DepthTest(false)
 	gl.DepthMask(false)
 	gl.Culling(GL.BACK)
+	gl.Culling(false)
 end
 
 local lastSunChanged = -1
--- a/LuaUI/Widgets/map_edge_extension2.lua
+++ b/LuaUI/Widgets/map_edge_extension2.lua
@@ -894,7 +894,7 @@ function widget:DrawWorldPreUnit()
 	gl.Texture(1, false)
 	gl.Texture(2, false)
 
-	gl.DepthTest(GL.ALWAYS)
+	--gl.DepthTest(GL.ALWAYS)
 	gl.DepthTest(false)
 	gl.DepthMask(false)
 	gl.Culling(GL.BACK)


sprunk:
`widget:DrawGroundDeferred()` doesn't seem to be a real thing. Maybe it's BAR specific?

The full list of widgets using `widget:DrawWorldPreUnit()` after enabling this one:
```
Mex Placement Handler
Custom Markers
Lups
AllyCursors
Map Edge Extension
Commands FX
Map Edge Extension 2        <----
Showeco and Grid Drawer
Missile Impact Points
Defense Range Zero-K
Decloak Range
Antinuke Coverage
Selected Units GL4 2
API Unit Tracker GL4
```

It looks like lots of things below and above happens to rely on whatever state they inherit from the previous widget. I put [a little test on the `glDepthTest-debug` branch](https://github.com/ZeroK-RTS/Zero-K/tree/glDepthTest-debug) and at least defense ranges, antinuke ranges, mex spots, and OD pylon circles get broken in various ways if they start with the wrong `glDepthTest` value.

image
image

Helwor:
Yes I've tested it.
As a widget, when you says gl.DepthTest(true) you expect it will suppress anything below ground. 
For that it need the param to be set to GL.LEQUAL (or GL.LESS but it's not default). So having it as GL.ALWAYS is obviously problematic. It basically means that a widget doing gl.DepthTest(true) will draw as same as gl.DepthTest(false) aka draw everything. Widgets never expect that param to be changed. A widget that changes it should mandatorily reset it at the end.
Letting DepthTest ON or OFF, is a bit less problematic bc usually a widget will explicitly say if he want to draw with depth constraints or not. From what I understood: DrawWorldPreUnit have DepthTest false by default and DrawWorld have it true, not sure 100%.
Letting Culling ON is problematic bc a widget usually  don't expect the culling to be activated and don't bother set it to false before drawing.
Letting Culling Param is less problematic bc a widget using culling will most likely explicitely state which face he want to cull.
Eco Drawer use a shadow volume stencil trick to draw circles perfectly on the ground (glVolumes.lua). For that it need culling to be false and expect depth test to be GL.LEQUAL when activated.
Idk about other widgets, but if they are disturbed by that change (depth param GL.LEQUAL, culling false), they should adapt imo. For the reasons cited above...

Helwor:
I havent tested it yet, but widget:DrawGroundDeferred() is probably relying on Spring Settings
	AllowDrawMapPostDeferredEvents
	AllowDrawMapDeferredEvents
	AlwaysSendDrawGroundEvents

sprunk:
> Idk about other widgets, but if they are disturbed by that change (depth param GL.LEQUAL, culling false), they should adapt imo.

Yes, what GF was asking for was that part of this PR should be checking if this is the case for the default set of enabled widgets and adapting anything that gets disturbed, otherwise we will still have bad rendering and are just shuffling it around between different widgets.

GoogleFrog:
Exactly. I'm not saying "we must keep broken things". I'm saying that if fixing something that is technically broken causes other things to be predictably broken, then the PR with the fix should also fix the new practical brokenness.

Helwor:
I haven't tested thoroughly, I just never have this widget ON, and this widget is OFF by default, therefore if fixing it would break next widgets, next widgets would also be broken just by the fact this widget is OFF... simple logic. Context is that I had a user comlpaining about eco grid not showing properly and I found why.

GoogleFrog:
Well if the widget is off by default, then I'm a lot less worried.

Issue #5137, opened 2023-12-27
Migrate box configs to 1-indexed


sprunk:
See 22a619fed00e4b713568449be269ab62def0e6e3 for an example. This is how Lua tables normally work, and the explicit indexes are mistake prone for copy-paste reasons so this would make boxes slightly more modder-friendly and prevent the spread of bad standards. This is a trivial mechanical task anybody could do.

PR #5671, opened 2026-02-16
Convert startbox config to 1-based indexing

MexxPex:
Lua tables are naturally 1‑indexed, and explicit 0‑based keys are mistake‑prone.
This aligns the files with the standards described in [issue #5137](https://github.com/ZeroK-RTS/Zero-K/issues/5137)


sprunk:
> I could split the commit into multiple commits if I knew how to do it.

Generally you use `git reset --mixed HEAD^` and then add/commit a few files at a time, google it if needed.

> does this not need any code changes? As in can the gadget handle both types already?

It handles both:

https://github.com/ZeroK-RTS/Zero-K/blob/317077dfe2577f9ebe68ec0f5d068240935a2ca1/LuaRules/Gadgets/Include/startbox_utilities.lua#L158-L174

sprunk:
Reviewed, fixed (some places had some extra logic after the primary config that was missed), looks good. Thanks!

Issue #5673, opened 2026-02-19
HoverJam - The first sparrow of the match can hover in place.


Stuart1998:
https://zero-k.info/Battles/Detail/2366965?

The first sparrow produced of the match has unusual properties that only apply to it and not to any other sparrows from any other players. Specifically, this first sparrow is able to hover in place by setting its unit state to land rather than fly when idle and receiving wait commands immediately after takeoff. I tested all three possible ways of obtaining the first sparrow (morph, build from factory, spawn in directly with cheats) and regardless of method, so long as it was the first sparrow of any player, it would have this behavior. A related behavior is "space sparrow", which I was only able to reproduce with sparrows that could also hover and haven't figured out exact steps to reproduce but causes the sparrow to fly well above its normal height while in a landed state. Space sparrow is of minimal impact because a sparrow that's in the air while in landed state has no practical use, but HoverJam is deleterious for pvp play because it allows a sparrow to act as a stationary radar jammer—but only for the first player to make one in a given game. (Of course, this means that the other player can permanently stop the first player from using it by destroying the HoverJam sparrow, but the player using it may only need to have it active once for a short time to be able to get off a game winning stealth attack)

I suggest that any matches where a player knowingly takes advantage of this bug be no-elo'd until the bug is fixed.

PR #5677, opened 2026-03-08
Partially fix fixedwing takeoff funcs

sprunk:
They would all use the unitID of the first fixedwing aircraft.

This does NOT solve #5673 (first Sparrow being able to hover mid-air), now all Sparrows have this "ability".


sprunk:
I'm not sure how good from a gameplay PoV it is to spread the "ability" to hover, and the coding error doesn't seem to have been that bad in practice (maybe planes take off at weird heights while The First Plane is alive? idk) so this can wait if needed.

GoogleFrog:
It looks like you're replacing a NoOP with a bugged function. Presumably the bug has something to do with the signal being sent while the plane has its parameters set to an intermediate state.

GoogleFrog:
Ok well there is a ticket and I can reproduce the issue, so this is pullable.

PR #5686, opened 2026-04-06
Commander eggs

Licho1:
image

Buildable eggs that can spawn (morph to) empty comms if your comm is dead.
Egg is actually an immobile unit - can be transported by hercules or thrown with newtons. Can detonate on D-gun with shockley effect - for some extra fun.

Notes:
 - handles obscure cases like commshare or higher number of initial comms. Max number of comms per team is invariant and during commshare, comms that are hatched are allocated to a team with spare capacity.
 - comms or eggs can be given to allies, but comms belong to original owner for purpose of limits (so give egg if you want to help ally and not your comm)
 - captured comms still belong to original owner so capturing hurts an extra bit
 - commander corpses are not ressurectable

Credits go mostly to @Orlicek1

Should be squashed when merging


GoogleFrog:
The egg costs 500 metal and the morph is free, but creates a 1100 metal cost unit. So shouldn't I be able to reclaim my comm for 880 metal and make a profit? This costs energy, and removes the vanguard income.

image

The egg shows up in Orders tab, which is a bit front-and-centre for something that cannot do its primary purpose unless you lose your comm.

image

For engineer comm it shows up in the units tab alongside its blueprint unit, and no unit is shown in the Orders tab. This all happens because I made constructors with a single unit build option put it in the Orders tab, for blueprint.

The tabs on the command panel look pretty bad when six are squished together. This is part of why Athena went to the Strider hub. I think it would be preferable to put the egg in the special panel, because it is pretty special. Perhaps here:

image

Licho1:
@GoogleFrog 
updated the PR:
 - build cost is 600
 - living commander is not reclaimable 
 - egg is blocking units but can be pushed
 - egg can be transported by charon for consistency
 - comm price includes modules
 - egg is in special section of build menu
 - morph buttons are updated to match availability
 - logic moved mostly to extra file for better readability

PR #5701, opened 2026-04-22
Fix: Scylla stows it's tacnuke correctly

Stiofan-K:
Scylla sometimes ends up with it's missile sticking out when the doors are closed. It just needed turnspeed applied.

Slows Scylla firerate by just a little since it needs to turn it's missile upward again though.


sprunk:
Good find. Thanks!

PR #5703, opened 2026-04-29
Starlight description is wrong

luminaStuffphoton:
Craterpuncher is displayed as dealing 18k damage. The LUS sets craterpuncher to fire for 30 ticks every time it triggers, with the craterpuncher ticks replacing the cutter ticks. ZK weapon convention is that a unit can fire all of it's weapons at once, so we subtract the dps of cutter to get a damage of `(800 - 150)*30 = 19500` for craterpuncher

### DIFF
--- a/units/mahlazer.lua
+++ b/units/mahlazer.lua
@@ -133,7 +133,7 @@ return { mahlazer = {
       craterMult              = 0,
 
       customParams              = {
-        stats_damage = 18000,
+        stats_damage = 19500,
         stats_hide_shield_damage = 1,
         light_radius = 0,
         lups_noshockwave = [[1]],


sprunk:
821e52a4b05b59c9dea2c397efe41abc62982af4 says it was supposed to be 18k, perhaps lower the actual damage instead (doesn't make much difference but is a nicer number)? Also make sure that there aren't engine shenanigans going on that would make real damage lower than nominal.

GoogleFrog:
I'm not sure that ZK weapon convention is that units fire two weapons at once out of the same gun. Dante and Scorpion don't. Jugglenaut and Newton don't either. I can't think of any units that use one gun to shoot two weapons at the same time.

All this is to say that the damage listed for crater puncher should be the damage crater puncher does. I don't expect players to assume that the smaller laser is happening inside the larger one, and think that they will get the small laser damage during the big burst.

PR #5704, opened 2026-04-29
Fix default DepthTest and Culling states

Helwor:
Showeco and grid drawer (among others?) not working properly because of it.


sprunk:
Looks like another take at #5659 but that one has some unaddressed comments.

Helwor:
I've addressed the comments now

GoogleFrog:
Duplicate.

PR #5722, opened 2026-05-24
rml support

loveridge:
Adds the `RmlUi` engine table to the global context.

RmlUi is only initialized when a widget calls `RmlUi.GetContext` or `RmlUi.CreateContext`. At which point the fonts are loaded, the cursor aliases are set, and the `shared` context is created for widgets to use.


GoogleFrog:
Thanks. Fixed unrelated whitespace changes.

Issue #5726, opened 2026-05-25
Terraform MMB Behaviour


Birdulon:
Currently, terraforming has rather unusual behaviour surrounding middle mouse button, which I tend to use for panning while doing other commands. ~Not sure how much is engine vs widget.~ Ramp is currently in the best state.
EDIT: Looking at [gui_lasso_terraform](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Widgets/gui_lasso_terraform.lua) again I think it probably won't come down to engine.

## Expected Behaviour
Normal panning controls. Just like Move/AttackMove/Patrol/ForceFire/Guard/AreaMex/Repair/Reclaim/SetTarget pre-click phase (things get weird while you're LMB click-dragging those, but that should probably be another issue).

#### Ramp
Normal panning controls. Very convenient.

#### Level, Raise/Lower, Restore, and Smooth
Immediately cancels out of the command, as if RMB were pressed. Very frustrating.

#### Ramp
One click: Normal panning controls!
Drag click: While LMB held, seemingly nothing - but when you release LMB after having pressed MMB, it acts almost like you're in One click mode, but it doesn't allow changing the elevation at the "second" LMB click, and you can do various things to get it more broken from this state.

#### Level, Raise/Lower, Restore, and Smooth
On MMB press, it goes into straight line mode, similar to holding Ctrl. It doesn't stop when you release MMB. It doesn't stop when you release LMB. Whenever you (press and) release Ctrl in this state, it starts a new line segment. LMB must be released and pressed again to proceed to the elevation step for Level or Raise/Lower, or to execute the command for Restore or Smooth.

#### Ramp
Normal panning controls while setting width, though the width control fights the cursor movement in a janky way.

#### Level, Raise/Lower
Normal panning controls while setting elevation, though the elevation control fights the cursor movement in a way that makes either task very janky.

Birdulon:
Some more testing:
Adding `if button == 2 then return end` to the start of `widget:MousePress` and `widget:MouseRelease` prevents the cursor cancelling out on the Before Drawing step. Good start, but doesn't solve the steps that rely on captured mouse input. I experimented with naively trying to retain the mouse input with `return drawingLasso or setHeight or drawingRamp or simpleDrawingRamp or drawingRectangle or placingRectangle` and even `return true` but it didn't seem to actually do so. I'll have to play around more.

GoogleFrog:
[...]

PR #5727, opened 2026-05-26
Improve UI consistency of Terraform commands with respect to MMB

Birdulon:
Quick fix to the most annoying part of #5726, the rest of which will be more involved.


Birdulon:
Oops, this breaks normal RMB commands, time to add some more conditions

Birdulon:
Should be fine now

PR #5734, opened 2026-05-28
Commander egg description EMP -> disarm/slow

Stiofan-K:


sprunk:
Should probably say

> Morphs into a Commander. Disarm and slow explosion on death.

since that is what `units/commander_egg.lua` says.

Issue #5668, opened 2026-02-13
Reinstate FeatureDamaged etc


sprunk:
Useful for modding and AFAIK vanilla won't actually pay any perf cost if there is no gadget using it.

PR #5745, opened 2026-06-04
Nanoplague!

Stiofan-K:
Gadget that implements zombie revive as a onhit weapon custom parameter.

Nanoplague offeres another way to interact with or leverage wrecks, adding another layer to the usual importance wreckfields generate while also giving a combo option for wrecks generated by suicidal later raids.

Example applications here include Athena, Quake and DRP's green shot.

In testing in the refumble mod, Quake has found at least some enthusiasts with Athena being rare but still offering a unique tool you can follow up with.

Comes with some extra parameters to adjust the revive per weapon. Things like revive speed, additional delay, if multiple revives are possible. I think there is potential for more too, that could affect unit behavior.

Currently, previously zombiefied wrecks/neutral wrecks are blocked from zombiefication unless specified otherwise to avoid metal farming from wrecks. Alternatives could include zombies dying into fragments instead of wrecks.

Gaia/neutral team gains metal and E for each revived unit. 
The metal is experimental for neutral production should a factory revive, the E for neutrals with need for energy upkeep.

The zombie behavior is copied from the zombies gadget, but I wanted to split it to not mess with the existing stuff. Maybe there are some improvement ideas for how they should behave around too.

Generally theres alot of ways this could be interesting I think, so I'm curious about others thoughts on this.

### DIFF
--- a/LuaRules/Gadgets/unit_nanoplague.lua
+++ b/LuaRules/Gadgets/unit_nanoplague.lua
@@ -1,0 +1,469 @@
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+if not gadgetHandler:IsSyncedCode() then
+	return
+end
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
+local version = "0.0.1"
+
+function gadget:GetInfo()
+	return {
+		name        = "Nanoplague!",
+		desc        = "Reuse of Zombies as firable weapon/weaponCustomParam.", -- Unsure if I could reuse the zombie gadgets somehow... it is a good bit of duplication but on disabled zombies the gadget removed itself
+		author      = "Stiofan", -- adapted from Zombies! by Tom Fyuri, he credits banana_Ai and Anarchid. Most comments are from Tom.
+		license     = "GPL v2 or later",
+		layer       = math.huge,
+		enabled     = true
+	}
+end
+
+--SYNCED-------------------------------------------------------------------
+
+-- changelog
+-- 6 august 2014 - 0.1.3. Some magic which might fix crash. (At least it fixed the original zombie gadget)
+-- 7 april 2014 - 0.1.2. Added permaslow option. Default on. 50% is max slow for now.
+-- 5 april 2014 - 0.1.1. Sfx, gfx, factory orders added. Slow down upon reclaim added. Thanks Anarchid.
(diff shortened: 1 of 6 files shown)


sprunk:
The changes to individual units belong to a mod. The FeatureDamaged change in the gadget handler is good and could be a standalone PR, alongside FeaturePreDamaged (as per #5668). Will look at the actual nanoplague gadget when the two above are addressed, generally the capability sounds nice even if no vanilla ZK unit would immediately use it.

Stiofan-K:
Ah, so I should split off the Feature Damaged callin and/or add the PreDamage callin, and toss out the example plague applications?

sprunk:
Yes please.

Eventually having an example in vanilla ZK would probably be good (so that modders are aware of the tech existing, and also because it's cool) but that requires more thought than just slapping it onto existing stuff.

Stiofan-K:
DRP and Quake/Missile Silo felt like a natural fit since both are/have "one of each status" weapons/Missiles. Athena is then the risky reusable option, or that were the ideas.

I agree that there needs to be more discussion to get it right, there are alot of ways a zombiefying weapon param can go and Id like some guidance to make it as good as can be.

The current implementation works but Ive ended up with alot of questions that stray into design/preference more than implementation, that are difficult to answer well on my own.


Things to talk about:
1. Should zombies inherit nanoplague to their weapons and/or death explosion too?

2. Should nanoplague be a uniform reviving speed? Currently it is set flat per weapon, but could easily be something that scales with cost of affected units, or even both flat mutliplier on the scaling speed.

3. Should kills with a nanoplague weapon immediately start the process, or should it only have effect on already dead units?
 
4. Should zombies behave randomly, as per the normal zombie modoption/gadget, or do we sneak a brutal ai behind the controls? It would allow for zombies to use their metal with cons, and maybe lead to more persistent zombies.

5. Inversely to 3., should zombies be combat boosted but decay naturally instead?

6. The issue of farming reclaim off reviving wrecks. Its still possible via kill->reclaim down to 1% -> zombiffy-> capture -> kill.
An idea I had here was making the resurrection chance dependant on metal left in the wreck, but im unsure if that is smooth enough/fixes the problem, while it diminishes the frontline viability.


A few of those things are something that I can implememt as a sort of weapon param/modoption, so I'll see if I can test those in practice.

PR #5748, opened 2026-06-06
Better reactivity

Helwor:
On the demand of Godde


LuaUI/Widgets/gui_chili_gesture_menu.lua
```diff
@@ -15,7 +15,7 @@ function widget:GetInfo()
 		handler   = true,
 	}
 end
-
+Spring.Echo('test')
```
Licho1: echo?

Helwor: ops

PR #5752, opened 2026-06-11
FeatureDamaged and FeaturePreDamaged

Stiofan-K:
What it says on the tin!

FeatureDamage is tested in Refumble mod and seems to work as intended.
FeaturePreDamaged I am uncertain about, though it mimics the UnitPreDamaged Callin.

### DIFF
--- a/LuaRules/gadgets.lua
+++ b/LuaRules/gadgets.lua
@@ -2035,7 +2035,6 @@ end
 
 local FeaturePreDamaged_GadgetMap = {}
 local FeaturePreDamaged_first = true
-local allWeaponDefs = {}
 
 function gadgetHandler:FeaturePreDamaged(featureID, featureDefID, featureTeam,
 	damage, weaponDefID,
@@ -2045,7 +2044,7 @@ function gadgetHandler:FeaturePreDamaged(featureID, featureDefID, featureTeam,
 	if FeaturePreDamaged_first then
 		for _,g in r_ipairs(self.FeaturePreDamagedList) do
 			tracy.ZoneBeginN("G:FeaturePreDamaged_GetWantedWeaponDef :" .. g.ghInfo.name)
-			local weaponDefs = (g.FeaturePreDamaged_GetWantedWeaponDef and g:FeatureDamaged_GetWantedWeaponDef()) or allWeaponDefs
+			local weaponDefs = (g.FeaturePreDamaged_GetWantedWeaponDef and g:FeaturePreDamaged_GetWantedWeaponDef()) or allWeaponDefs
 			tracy.ZoneEnd()
 			for _,wdid in ipairs(weaponDefs) do
 				if FeaturePreDamaged_GadgetMap[wdid] then
--- a/LuaRules/gadgets.lua
+++ b/LuaRules/gadgets.lua
@@ -140,9 +140,6 @@ local callInLists = {
 	"FeatureDestroyed",
 	"FeaturePreDamaged",
 	"FeatureDamaged",
-	--[[ FeatureDamaged and FeaturePreDamaged missing on purpose. Basic damage control
-	     can be achieved via armordefs (use the "default" class, make sure to populate
-	     the others including "else" explicitly) so this way we avoid the perf cost. ]]
 
 	-- Projectile CallIns
 	"ProjectileCreated",
@@ -2040,14 +2037,8 @@ local FeaturePreDamaged_GadgetMap = {}
 local FeaturePreDamaged_first = true
 local allWeaponDefs = {}
 
-do
-	for i=-7,#WeaponDefs do
-		allWeaponDefs[#allWeaponDefs+1] = i
-	end
-end
-
 function gadgetHandler:FeaturePreDamaged(featureID, featureDefID, featureTeam,
-	damage, paralyzer, weaponDefID,
+	damage, weaponDefID,
 	projectileID, attackerID, attackerDefID, attackerTeam)
 	tracy.ZoneBeginN("G:FeaturePreDamaged")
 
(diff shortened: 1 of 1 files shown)


GoogleFrog:
What are they for? The comment you haven't removed says that they are missing on purpose to avoid paying the performance cost.

Stiofan-K:
Sprung had asked for it to be split off from the Nanoplague PR

Reasoning as per issue #5668

I had asked about how to do feature damaged detection the way the comment described but handnt gotten an answer/it was easier to implement the callin.

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2036,104 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
 	return
 end
 
+local FeaturePreDamaged_GadgetMap = {}
+local FeaturePreDamaged_first = true
+local allWeaponDefs = {}
+
+do
+	for i=-7,#WeaponDefs do
```
sprunk: why -7?

LuaRules/gadgets.lua
```diff
@@ -138,6 +138,8 @@ local callInLists = {
 	-- Feature CallIns
 	"FeatureCreated",
 	"FeatureDestroyed",
+	"FeaturePreDamaged",
+	"FeatureDamaged",
```
sprunk, line 142: the comment below was discovered to be untrue (no perf cost if callin not actually used) and no longer applies, remove it

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2036,104 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
...
+	projectileID, attackerID, attackerDefID, attackerTeam)
+	tracy.ZoneBeginN("G:FeaturePreDamaged")
+
+	if FeaturePreDamaged_first then
+		for _,g in r_ipairs(self.FeaturePreDamagedList) do
+			tracy.ZoneBeginN("G:FeaturePreDamaged_GetWantedWeaponDef :" .. g.ghInfo.name)
+			local weaponDefs = (g.FeaturePreDamaged_GetWantedWeaponDef and g:FeatureDamaged_GetWantedWeaponDef()) or allWeaponDefs
+			tracy.ZoneEnd()
+			for _,wdid in ipairs(weaponDefs) do
+				if FeatureDamaged_GadgetMap[wdid] then
```
sprunk: wrong var

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2036,104 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
...
+
+do
+	for i=-7,#WeaponDefs do
+		allWeaponDefs[#allWeaponDefs+1] = i
+	end
+end
+
+function gadgetHandler:FeaturePreDamaged(featureID, featureDefID, featureTeam,
+	damage, paralyzer, weaponDefID,
+	projectileID, attackerID, attackerDefID, attackerTeam)
```
sprunk, line 2041: this doesnt seem to be the actual signature of the function as passed by the engine

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2036,104 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
...
+	end
+
+	local rDam = damage
+	local rImp = 1.0
+
+	local gadgets = FeaturePreDamaged_GadgetMap[weaponDefID]
+	if gadgets then
+		if gadgetHandler.GG._AddUnitDamage_teamID then
+			attackerTeam = gadgetHandler.GG._AddUnitDamage_teamID
+		end
```
sprunk: probably remove this bit (unless an equivalent AddFeatureDamage hack for features is implemented)

sprunk (changes requested):

sprunk:
> I had asked about how to do feature damaged detection the way the comment described but handnt gotten an answer/it was easier to implement the callin.

Without the callin it's just not possible to identify non-lethal damage. The "use armordefs" comment is only about implementing basic "deals more/less/zero damage to features" kind of logic, which I think was the bulk of what people used to want to do.

LuaRules/gadgets.lua
Stiofan-K: This is an uncommeted offset taken from usage in UnitPreDamaged.
I'm unsure why it's there either, but assumed it was intentional. 

The UnitPreDamaged Callin is used a decent amount, also using the weapondef value within.
Its also reused as a fallback on line 2215 for the Explosion gadgetHandler

I'll remove the duplicate declaration I suppose, and maybe make an issue for the -7? (Line 1748)

Since it's the fallback option in all usecases, it may just never be used and has as such never failed yet?

LuaRules/gadgets.lua
Stiofan-K: Should do the same for FeatureDamaged at line 2109?

LuaRules/gadgets.lua
Stiofan-K: I missed the paralyze damage not being included in feature callins (Except add damage) when checking the recoil docs

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2033,95 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
 	return
 end
 
+local FeaturePreDamaged_GadgetMap = {}
+local FeaturePreDamaged_first = true
+local allWeaponDefs = {}
```
sprunk: this shadows the earlier var

sprunk (commented):
overall ok. i'll try to test during the week

LuaRules/gadgets.lua
```diff
@@ -2034,6 +2033,95 @@ function gadgetHandler:FeatureCreated(featureID, allyTeam)
...
+
+function gadgetHandler:FeaturePreDamaged(featureID, featureDefID, featureTeam,
+	damage, weaponDefID,
+	projectileID, attackerID, attackerDefID, attackerTeam)
+	tracy.ZoneBeginN("G:FeaturePreDamaged")
+
+	if FeaturePreDamaged_first then
+		for _,g in r_ipairs(self.FeaturePreDamagedList) do
+			tracy.ZoneBeginN("G:FeaturePreDamaged_GetWantedWeaponDef :" .. g.ghInfo.name)
+			local weaponDefs = (g.FeaturePreDamaged_GetWantedWeaponDef and g:FeatureDamaged_GetWantedWeaponDef()) or allWeaponDefs
```
sprunk:
```suggestion
			local weaponDefs = (g.FeaturePreDamaged_GetWantedWeaponDef and g:FeaturePreDamaged_GetWantedWeaponDef()) or allWeaponDefs
```

sprunk:
There were some issues but I fixed them. Thanks!

Stiofan-K:
Aye! Sorry for the trouble!

PR #5761, opened 2026-06-21
Likho plays with its ball more

SteelBlueZK:
fix for likho permanently hiding ball after first launch

### DIFF
--- a/scripts/bomberheavy.lua
+++ b/scripts/bomberheavy.lua
@@ -28,6 +28,8 @@ local SIG_TAKEOFF = 2
 local takeoffHeight = UnitDefNames["bomberheavy"].cruiseAltitude
 
 local function ShowBall()
+	Move(ball, y_axis, 0)
+	Move(ball, z_axis, 0)
 	Show(ball)
 	
 	Move(radiator_L, z_axis, 0, 2)


SteelBlueZK:
I think this could be animated "better" (move instead of teleport)

SteelBlueZK:
Actually, when giving the move command a speed, it breaks in another way. I don't nearly know enough about the engine systems, but my guess is that Spring.Unitscript.Move is not playing well with the custom Spring.SetUnitPieceMatrix calls. I cannot do more than the one commit.

sprunk:
Why does it move at all? Also what about this line?
https://github.com/ZeroK-RTS/Zero-K/blob/5eba6152fa2b86484159fe3a76626fd372b03d52/scripts/bomberheavy.lua#L62

> I think this could be animated "better" (move instead of teleport)

Originally it regrew from a 0 size, which the line above shrunk it to. I guess the regrowing part was removed when engine broke this so the line above is just a no-op? Anyway that is probably the appropriate animation.

GoogleFrog:
@sprunk was a dedicated scale function added to replace the engine change?

sprunk:
Yes, `Scale` and other ancillaries (`WaitForScale` etc)

SteelBlueZK:
Most use of Spring.SetUnitPieceMatrix is over my head. I do see that the hide function applies Spring.Unitscript.Move to the ball piece [here](https://github.com/SteelBlueZK/Zero-K/blob/master/scripts/bomberheavy.lua#L63). If you check the underside of the reloaded Likho, you can see the ball poking out the underside:

(stable, never fired likho↓)
image

(stable, once reloaded and armed likho↓)
image

Alright, I lied I do kinda understand Matrix manipulation. Thinking back on things, a translation matrix can do rotation and translation, which probably reset the piece to (0,0) relative back when it was working.

New information: Singu animation is broken. Reject this patch and make an engine ticket.

SteelBlueZK:
Not applicable change

PR #5764, opened 2026-06-27
Dgun prioritisation (for coms)

Stiofan-K:
A workaround attempt to avoid forcefire/settarget from conflicting with dgun commands.

### DIFF
--- a/scripts/dynassault.lua
+++ b/scripts/dynassault.lua
@@ -230,10 +230,17 @@ local function RestoreDGun()
 	Turn(lnanohand, x_axis, 0, ARM_SPEED_PITCH)
 end
 
+local dgunAim = false
+local function PrioritiseDgun()
+	dgunAim = true
+	Sleep(500)
+	dgunAim = false
+end
+
 function script.AimWeapon(num, heading, pitch)
 	local weaponNum = dyncomm.GetWeapon(num)
 
-	if weaponNum == 1 then
+	if weaponNum == 1 and not dgunAim then
 		Signal(SIG_LASER)
 		SetSignalMask(SIG_LASER)
 		isLasering = true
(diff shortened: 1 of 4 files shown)


Stiofan-K:
Issue is reproducable by setting forcefire/set target and then dgunning the opposite direction. Not really an issue for small or 0 aim deviations, but I assume it's the underlying cause of some dguns not firing.

Doesnt feel like a clean fix, though I used it for Dante too. PR is more to get the ball rolling and some insights.

From observation, it seems that it's set target and forcefire that conflict with dgun orders, while a commander given a stop command and as such has no target set, prioritises it's dgun without need for workaround even if it shoots other things inbetween.

I've tried to unset the target through Spring.SetUnitTarget and the Unitrules param "target_type" as the  unit_target_on_the_move gadget does it, but that didnt seem to do the same thing as a stop command, with some linger data that I assume causes the resulting jittering/aim conflict.

GoogleFrog:
A more robust fix may be to set some GG in script.AimWeapon for the num corresponding to manual fire. The GG would index by unitID, and have the frame it was set as a value. The set target gadget would then avoid setting targets if the unit attempted to dgun too recently (the GG could even be cleared on script.Shot, or whatever turns out to be approriate).

There is a related problem with set target where Attack commands fight for priority. This could be dealt with by setting the frame offset in the gadget equal to the SlowUpdate offset in the engine, but at some point SlowUpdates became staggered and the engine does not expose their offsets.

Stiofan-K:
I did note the set target and forcefire conflicts, but that is rarer to occure usually or I suppose less impactfull compared to a dgun not firing.

Whats does the GG stand for? :o

If the dgun command cleared target data like stopping does that should probably also fix the issue, since it only occures after a target was set, not on automatic/passive target aquisition. But I'm unsure what would need to be cleared there.

GoogleFrog:
GG is the gadget global table.

PR #5771, opened 2026-07-01
Missile orders

amnykon:
This pull request introduces a new 'Missiles' tab to the integral menu and adds a new widget to manage missile commands, stockpile counts, and target drawing.


LuaUI/Widgets/missile_command_center.lua
```diff
@@ -0,0 +1,717 @@
...
+  if changed or (emptySelection and not wasEmptySelection) then
+    Spring.ForceLayoutUpdate()
+  end
+  wasEmptySelection = emptySelection
+end
+
+function widget:CommandNotify(cmdID, cmdParams, cmdOptions)
+  for _, command in pairs(commands) do
+    if command:commandNotify(cmdID, cmdParams, cmdOptions) then return true end
+  end
```
GoogleFrog: Doing pairs to call a bunch of objects every time a command happens does not look good. The call is gated internally by `if cmdID == self.cmd`, so you should be able to make a simple table indexed by cmdID, reducing the cost and maintenance burden to a single table lookup.

LuaUI/Widgets/missile_command_center.lua
```diff
@@ -0,0 +1,717 @@
...
+end
+
+
+function widget:DrawWorld()
+  for _, command in pairs(commands) do
+    command:drawWorld()
+  end
```
GoogleFrog: See `widget:CommandNotify`. In this case the gate is the same, but behind `Spring.GetActiveCommand` (in the current version this is called redundantly multiple times.

LuaUI/Widgets/missile_command_center.lua
```diff
@@ -0,0 +1,717 @@
...
+  local x2, y2, z2 = Spring.GetUnitPosition(unit)
+
+  if not x1 or not x2 then return 0 end
+
+  if distance3(x1, y1, z1, x2, y2, z2) > 600 then return 0 end
+
+  return 1
+end
+
+local function EOS_controller_class()
```
GoogleFrog: The Bla_controller_classes have a lot of duplication so should have their essential parameters picked out and put into a config file. This is cleaner and lets modders find them.

LuaUI/Widgets/missile_command_center.lua
```diff
@@ -0,0 +1,717 @@
...
+      return coords[1], coords[2], coords[3], true
+    else
+      return Spring.GetFeaturePosition(mouseTarget)
+    end
+  else
+    return nil
+  end
+end
+
+local function distance3(x1, y1, z1, x2, y2, z2)
```
GoogleFrog: These exist (or should be added if missing) in Spring.Utilities.Vector

GoogleFrog:
image

When I click at A with this widget, the missile is targeted at the position on the ground behind the target unit.

LuaUI/Widgets/missile_command_center.lua
```diff
@@ -0,0 +1,717 @@
...
+
+local function DrawCircle(x, y, z, radius)
+  glPushMatrix()
+  glTranslate(x, y, z)
+  glScale(radius, radius, radius)
+  glBeginEnd(GL_LINE_LOOP, UnitCircleVertices)
+  glPopMatrix()
+end
+
+local function drawBlastRadius(tx, ty, tz, weaponDef)
```
GoogleFrog, line 85: Doesn't this exist in attack AoE?

GoogleFrog:
I want to see this sort of feature implemented, but I have misgivings about adding a global second-row tab.
- There is something quite satisfying in having a clear UI when nothing is selected.
- With Launch and nothing selected, pressing D and clicking launches a Shockley. This is quite bad as you might have, say, a Lobster selected, but it dies just as you press D and try to have it launch units somewhere. Or maybe you have a strider that gets napped (deselecting it) just as you try to use a D ability. Basically, having a global hotkey appear when your selection clears is a problem.
- The second row of the command menu looks a bit busy and janky. The main row of tabs are attached to the bottom while this one floats above.
- The logic of the launch menu means that, if you have units selected, the menu closes whenever you fire a missile. Often you'll want to fire multiple missiles, potentially of different types.
- I'm also a bit concerned about the maintainability of adding concepts such as global menu, two menu layers, and (especially) tab badges to integral menu.

image

I would like to see a version where the launch menu become accessible via a core select button, as I suspect this would work a lot better. 
- We could replace (only by default) the per-silo factory button with a single button that selects the launch screen.
- This button would contain the same information has the launch badge. The button could have a 2x2 grid of missiles in most cases, and switch to 3x3 if the player has at least five types of missiles..

Clicking the core selector button would open the launch menu of the PR, but I am not 100% sure how it would interact with it. Here are some ideas.
- Show the tabs appropriate to the selection, but highlight none of them, as the launch tab is selected.
- Clicking the launch button on core selector again should toggle the menu.
- Pressing Escape should close the menu (as happens now with construction tabs).
- There should probably be some other methods of closing the menu, which I would expect to become clear after some playing around with it.

The core selector approach makes the feature quite visible, but keeps it minimal for people who want to use the old system. Players are unlikely to accidentally launch missiles because all the functionality would sit behind pressing the button or a hotkey. We would have options to turn off the button and to show individual missile silos on the selector (as currently happens), which would probably be off by default. Some way to queue missiles in the launch menu would help make the individual selection buttons obsolete

Other issues.
- I feel like the integral config is almost autodetectable from customparams, which would help modding. Perhaps easy modding would just lose the ability to set the missile layout explicitly, which is an ok tradeoff.
- Similar to the above, the cmdIDs could probably be automatically determined as well.
- It looks like there is code reuse between attack AoE and the missile command widget. That is going to be a pain to maintain, and I'm not sure why missile command can't just do a simple invocation of attack AoE.
- Complicating the above is the fact that the closest silo is not always the best choice. Eg silo A is blocked by terrain while silo B is not.

image

Minor issues.
- You have written Eos in allcaps.
- The whitespace for the missile command widget is wrong.
- Where is Scylla?

amnykon:
@GoogleFrog i feel the critical issues is the ui. No to randomly floating buttons, no to 2nd row of tabs (this got worse when I added air), adding a launch or strike button in the core select menu is an interesting idea. Should this adde or replace the silo button? IMO it's add, as you can't build with the launcher menu. If we are not having a tab for it, I question the integration into the orders panel. Con upgrade shows buttons above the orders panel, I feel launch should show a single row of buttons for launch missiles.

Other comments are good and I will apply them (or point out how thay are already dealt with) after the ui for this has been defined.

amnykon:
3×3 badges on the selection panel will be way too small to show type + count + progress per missile. Instead I’m thinking a “show launchers” toggle in the top-left global commands panel — when on, the launch buttons appear in their own panel, probably just above the orders panel (or off to the side). Clean by default, full-size buttons with room for type/count/progress.
Same mechanism covers bombers (~8, incl. 2 for Odin) and scouts (swift, sparrow, oddly flea).
Different mechanism for two things, both further off, unrelated other then to show the progression I am planning:
	•	Global construction — use tab to switch global/direct modes; advanced widget, off by default. Currently tab shows the share screen, IMO it's a weak hotkey in a prime position for a hotkey.
	•	Striders — basically turns a nanotower into a strider hub plate; probably cleaner to just make it a proper plate than to have separate nanotower and strider hub side widgets.

GoogleFrog:
- I think >=5 missiles will be quite rare, and 2x2 will probably be sufficiently visible
- I said above, I would _by default_ replace the silo button, but there should just be independent settings for whether to show the launch button and whether to show silo buttons. The PR doesn't need to do anything special to the current silo button, as it is just handled like a factory.
- Opening integral menu as a hidden tab sounds fine to me. It makes sense for this type of UI to live there, and it uses the grid hotkeys. Clicking any of the tabs will close it.
- Con upgrade = comm upgrade? The comm upgrade UI isn't great, I don't want it to be copied for something that might be used as frequently as a missile UI (in bursts).
- I was considering something like "press Alt+A to queue an Eos". Or perhaps buttons on the third row to queue.

GoogleFrog:
I'm seeing a lot of this.

image

LuaRules/Configs/customcmds.lua
```diff
@@ -100,6 +100,17 @@ local commands = {
 	FIELD_FAC_UNIT_TYPE = 38694,
 	FIELD_FAC_QUEUELESS = 38695,
 
+	-- missile launcher (Missile Command Center widget; see Configs/missile_config.lua)
```
GoogleFrog, line 103: Why are these here and hardcoded?

LuaUI/Widgets/gui_attack_aoe.lua
```diff
@@ -215,6 +215,25 @@ end
 --initialization
 --------------------------------------------------------------------------------
 
+-- Vlaunch (starburst) flight parameters, derived purely from the weapon def.
+-- Used by the AoE preview here and exposed via WG for the missile launch preview,
+-- so both compute impact points from an identical trajectory model.
+local function BuildVlaunch(weaponDef)
```
GoogleFrog, line 221: Why vlaunch in particular? The next time someone wants a different type of weapondef are they going to make their own builder? Instead, getWeaponInfo should be split into the bits that need unitDef and the bits that don't (or have a fallback).

LuaUI/Widgets/gui_attack_aoe.lua
```diff
@@ -1027,10 +1095,12 @@ function widget:Initialize()
 		aoeDefInfo[unitDefID], dgunInfo[unitDefID], extraDrawRangeDefInfo[unitDefID] = SetupUnit(unitDef)
 	end
 	SetupDisplayLists()
+	WG.AttackAoE = {GetVlaunchImpact = GetVlaunchImpact, DrawAoEPreview = DrawAoEPreview}
```
GoogleFrog: This is not how these tables are usually constructed so will be hard to search. Use externalFunctions.

LuaUI/Widgets/gui_attack_aoe.lua
```diff
@@ -939,6 +980,33 @@ local function CalculateVlaunchImpact(info, fx, fy, fz, tx, ty, tz)
...
+--------------------------------------------------------------------------------
+-- Shared vlaunch impact query (used by the missile launch preview widget)
+--------------------------------------------------------------------------------
+
+local vlaunchInfoCache = {}
+
+-- Returns the terrain impact point (x, y, z) of a vlaunch (starburst) shot fired
+-- from (fx, fy, fz) at (tx, ty, tz), or nil if it reaches the target
+-- unobstructed or the weapon is not a vlaunch weapon.
+local function GetVlaunchImpact(weaponDefID, fx, fy, fz, tx, ty, tz)
```
GoogleFrog: This is duplicate and locally-dead code. Also in my tested the visuals were bugged.

LuaUI/Widgets/gui_chili_core_selector.lua
```diff
@@ -1747,10 +2067,27 @@ local function ClearData()
...
 	InitializeUnits()
-	
+
 	buttonList.SetImagesVisible(false)
 end
 
+-- Rebuild the button list when the silo-hiding option changes so existing silo
+-- buttons appear/disappear immediately (ClearData re-scans owned units; the old
+-- list is disposed next Update, same as on a team change).
+options.hideMissileSilos.OnChange = function()
```
GoogleFrog: Put the OnChange in the option.

GoogleFrog:
I don't like how the statefulness works in the integral code and UI. There are a lot of checks and extra bits of handling. Why does the missile menu close if I have a single Glaive selected, and that Glaive dies? Why does opening the menu cause me to have a missile attack command already selected?

The integral handling should be very simple. A special tab opened by a widget should only be closed when explicitly told to do so, either by the user pressing a tab button, or via the API. Bespoke cases should be handled by the external widget.

PR #5777, opened 2026-07-02
Overdrive cables: wiggle noise, twig rooting, and cliff-lip overshoot fixes

Anarchid:
Fixes three issues that have been plaguing the cables feature since release, plus a redo of the dead-weight cleanup that had drifted out of mergeability.

## Fixes

**Band-limited wiggle noise** (`a41d8fe`) — the wiggle offset was raw per-vertex white noise, so the pointiness-adaptive tessellation (which densifies vertices on cliffs) fed adjacent vertices fully-independent ±amplitude kicks: a tight full-amplitude zigzag exactly where the cable was already struggling. Replaced with cell-quantized, smoothstep-interpolated value noise (fixed 16-elmo feature wavelength), so denser sampling converges on the same smooth curve. Keyed on world XZ, so MST parent/child flips don't teleport the pattern.

**Twig roots welded to the rendered rail** (`beafd2a`) — twigs are emitted by separate GS invocations and must rebuild the exact trunk geometry to stay attached. The old root anchor offset from a belly-height spawn along a local-normal axis; on slopes it diverged from the trunk frame and dangled below the tube, and the noise change exposed the mismatch everywhere. The twig dispatch now rebuilds its target emitted vertex (and predecessor) and reconstructs the trunk's own rail frame — the root lands on the literal outer-rail vertex the tent invocation emits, with the same terrain clamp, plus a small Navg lift against z-fighting. The twig's growth basis is sign-aligned to the rail axis so it can't fold back across the trunk.

**Cliff-lip "inertia"** (`96f40b1`) — the centerline lift took the terrain max within ±0.85·(lenAB/numSeg) *along* the cable. A max-filter along the path is a morphological dilation: on any monotone slope the whole height profile shifts ~19 elmos downhill, so cables carried cliff-top height well past the lip before dropping — tracing the ground curvature perfectly but with a consistent horizontal push. The wide window predates the kink-adaptive tessellation and was doing anti-clip duty that clustering now covers. The profile scan now max-filters over one scan cell, and each emitted vertex takes the max over ~0.55 of its local gaps: same stay-above-terrain envelope, but the overshoot collapses to about one scan cell at the lip (where clustering is dense), and endpoints no longer sample terrain beyond the pylons.

## Cleanup redo (`68431aa`)

Re-derivation of the old `cable-cutting` cleanup against the post-tent-rework code (the original commit predates the tent geometry and could not be merged textually). No intended behaviour change: removes the never-read CPU-side bubble-phase integration (the FS advects purely from gameTime), shrinks the vertex payload 10→9 floats (`vertGrid` vec4→vec3 end-to-end; the ghost sentinel moves from `.w` to `.z`), and deletes `geomCache`, `EFF_EPSILON`/`lastSentEff`, dead CPU-router remnants, the twig `tipR`/`CONE_TIP_WIDTH` chain, and the unreachable `t > 0.90` FS discard. luacheck is down to the known gl/GL unsynced false-positives + pre-existing `edges` shadowing.

## Testing

Verified in a dev run: shaders compile on all three passes, cables hug cliff descents without the horizontal offset, wiggle stays smooth on oversampled slopes, and twig junctions stay seated. Worth a second pair of eyes on **ghost edges / enemy-LOS gating**, since the attribute repack moved the ghost sentinel component.

*(screenshot incoming)*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/LuaRules/Gadgets/Shaders/gfx_overdrive_cables.frag.glsl
+++ b/LuaRules/Gadgets/Shaders/gfx_overdrive_cables.frag.glsl
@@ -6,7 +6,6 @@
 uniform sampler2D infoTex;
 uniform sampler2D cableTex;
 uniform float gameTime;
-uniform float bakeTime;
 uniform float enableFlow;     // 1.0 = full bubble pass; 0.0 = static cables (no animation)
 uniform float ghostsEnabled;  // 1.0 = ghost branch active; 0.0 = enemy OOL discards immediately
 uniform sampler2DShadow shadowTex;  // engine shadow map ($shadow); sampled for shadow reception
@@ -26,7 +25,7 @@ in DataGS {
 	float width;
 	vec2 cableUV;
 	vec2 timeData;
-	vec3 gridData;          // was vec4; .z was unused, dropped to fund cableTangent (see GS BUDGET NOTE). Old .w (isOwnAlly / ghost flag) is now .z.
+	vec3 gridData;          // (gridEfficiency, flow, isOwnAlly / ghost flag); see GS BUDGET NOTE
 	vec3 cableTangent;      // smooth per-vertex cable along-direction; interpolated → drives the lit frame below
 	float spawnAlongMain;   // overloaded: main ribbon → lenPerSeg; twig → twig-along
 	flat int gsSlot;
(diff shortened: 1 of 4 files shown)


Anarchid:
image

GoogleFrog:
Oh, I might have wanted to take a look at this. But the large commit is meant to be a noop, and the other three commits look fairly reasonable, so I suppose it is fine.

PR #5782, opened 2026-07-07
Yet another missile silo panel widget

danfireman:
Competing with https://github.com/ZeroK-RTS/Zero-K/pull/5771:

image

**Only review if you're interested in this sort of approach/think it's a viable starting point/like the look, otherwise it can be closed.**

## Deliverable

A movable, factory-queue-styled strip for missile silos that aggregates all owned silos into one panel, with a button per missile type showing ready / queued / build progress and a shared slots meter.

- Left-click a type: arm fire-mode; click again, or when none are ready, to queue one instead.
- Middle-click: queue one at the best-placed silo. Right-click: cancel one (never destroys a finished missile).
- Hover or arm a type to preview each silo's range ring (solid/bright when that silo can fire it, dim/dashed otherwise) plus the AoE falloff at the cursor, mirroring the stock Attack-AoE indicator.

The panel auto-hides when you own no silos. Fire dispatches CMD.ATTACK to the specific ready missile of the chosen type, matching cmd_missile_silo's force-fire semantics (ground point, or unit-tracking for the homing Zeno).

## Questions unanswered

Unsure on how/whether to include Scylla/Reef (I definitely think Trinity should be handled differently, its more strategic and doesn't have the very fiddly UX).

Unsure on a good hide/expand silo panel UX (e.g. where the little tab should live)

### DIFF
--- a/LuaUI/Widgets/gui_attack_aoe.lua
+++ b/LuaUI/Widgets/gui_attack_aoe.lua
@@ -488,6 +488,41 @@ local function DrawAoE(tx, ty, tz, aoe, ee, alphaMult, offset, circleMode)
 end
 
 --------------------------------------------------------------------------------
+-- Shared blast-radius preview, exposed via WG so other widgets (e.g. the missile
+-- silo launch UI) draw their AoE footprint with THIS falloff code instead of
+-- duplicating it. Nested rings at (tx,ty,tz) whose alpha decays toward the edge by
+-- edgeEffectiveness `ee`. `color` (optional {r,g,b[,a]}) overrides the ring colour;
+-- `alphaMult` (optional) scales overall opacity. Line width is derived from the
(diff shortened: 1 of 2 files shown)


GoogleFrog:
I had a few comments but it looks good overall from the code. Here are some ideas to pursue.
- Register it with the top left area (like the cheat sheet button) once it becomes active.
- Put some settings in Settings/Interface/HUD Panels. There should be a tickbox that controls whether it is automatically enabled when a silo is built, and one that toggles whether it is visible (and all options should be hotkeyable by default).
- Attack AoE knows whether a missile is projected to impact the terrain. Use this as a filter on whether missile should be picked.

danfireman:
image
Suitable icon?

sprunk:
Perhaps the radar icon for missile silo?

danfireman:
Missile filtering working

PR #5784, opened 2026-07-09
Zombie api gadget

Stiofan-K:
Splitting off some the zombies code into a gadget to later be used by nanoplague


Stiofan-K:
I'm a bit on the fence on some things, specifically handling the zombies arrays and such. 

I could unify/port/standerdise the countdown to revive aswell I suppose, though then it would not be possible to differentiate between zombie or nanoplague zombies.

GoogleFrog:
What does each of the gadgets do? The descriptions of each should be updated to make this clear, because api_ and unit_ don't mean a whole lot.

Stiofan-K:
aaaaaye.
The api gadget are essentially just helper functions to be used by anything that wants to do zombie esque things, with the base/unit zombies being the global passive wreck reviving.

Ideally I want to put as much into the helper gadget to avoid duplication, but I'm unsure if I can put more there without it becoming inflexible.

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,207 @@
...
+
+function gadget:Initialize()
+  
+	mapWidth = Game.mapSizeX
+	mapHeight = Game.mapSizeZ
+
+  GG.Zombies = {
+    TurnFeatureIntoUnit     = TurnFeatureIntoUnit,
+    SetZombieSpeedMult      = SetZombieSpeedMult,
+    SetZombieBehavior       = SetZombieBehavior,
```
sprunk: I'd get rid of the "set behaviour" and just let gadgets ask for the wander directly. Units are essentially "controlled by luaAI" by default.
```suggestion
    SetZombieWander         = GiveZombiesRandomOrders,
```

sprunk (commented):
looks ok overall, i'll see if i can find some time to test during the week

Stiofan-K:
aye thanks, I'll change it around! 

Thoughts on if theres sense in trying to put the to be rezed arrays in the api gadget?
Same with unslowing/speeding zombies, since some may not want their zombies to reset speed.

And is LuaAi different from amove behavior? since thats essentially what wander applies, and I'd be curious if more complex behavior can be coaxed out for nanoplague zombies.

sprunk:
> Thoughts on if theres sense in trying to put the to be rezed arrays in the api gadget?
> Same with unslowing/speeding zombies

IMO slowing is ok as-is, rezzed arrays may be ok to eventually put in the api but I'll have to think of how to do it cleanly.

> And is LuaAi different from amove behavior? since thats essentially what wander applies, and I'd be curious if more complex behavior can be coaxed out for nanoplague zombies.

LuaAI can make units do anything, it's just a matter of writing an implementation. This includes arbitrarily powerful "legit" behaviour (e.g. pyros use jump; glaives perfectly dodge projectiles and avoid LLTs; units congregate into hivemind armies and strategically raid mexes...) but also beyond that (e.g. radars call down orbital tacnukes; detriment roars, whistles, and even plays music).

Stiofan-K:
For rezzed arrays I think the issue would be referencing back to the gadgets that put the features "in queue", so that you can differentiate between different zombie types.

And :O Detriment with all the bells and whistles. Is there a good example for LuaAI floating about? Wandering works generally for whats intended, but cons building structures and units being smarter or 'biased' toward the team that made them could be interesting to give Nanoplagued zombies a smarter feel over normal ones, being a less degraded/feral version of the same tech essentially.

sprunk:
> For rezzed arrays I think the issue would be referencing back to the gadgets that put the features "in queue", so that you can differentiate between different zombie types.

Yea I'm thinking there could just be a callback like so
```
GG.ApplyZombieBuildpower(featureID, buildpower, function (unitID, featureID)
  -- bla
end)
```

> Is there a good example for LuaAI floating about?

Not really, what I listed is mostly just what's possible. In theory CAI is a LuaAI and tries to be smart but it may be hard to make a "detached" instance that won't try to control the rest of non-zombies owned by Gaia.

Stiofan-K:
*poke*
Should I try to add such a thing or do we stick to this for now? o:

sprunk:
Sounds fine either way. I was/am sort of busy but I'll try to prioritize looking at this deeper soon.

Stiofan-K:
Aye, just wanted to get some confirmation. No rush!

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+
+local function TurnFeatureIntoUnit(featureID,teamID,reclaimPercentHealth)
+  
+  local featureDefName,facing = GetFeatureResurrectData(featureID)
+  local x, y, z = Spring.GetFeaturePosition(featureID)
+  
+  if reclaimPercentHealth then
+    local currentMetal, maxMetal = Spring.GetFeatureResources(featureID)
+    if currentMetal and maxMetal and (maxMetal > 0) then
+      partialReclaim = currentMetal/maxMetal
```
sprunk, line 65: When the code does not enter these `if`s then the `partialReclaim` var (which is a global btw) will be nil,

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+  if (unitID) then
+    gadgetHandler:NotifyUnitCreatedByMechanic(unitID, false, "zombies")
+    local size = UnitDefNames[featureDefName].xsize
+    Spring.SpawnCEG("resurrect", x, y, z, 0, 0, 0, size)
+    Spring.GiveOrderToUnit(unitID, CMD.FIRE_STATE, 2, 0)
+    GG.PlayFogHiddenSound(REZ_SOUND, 12, x, y, z)
+    if partialReclaim ~= 1 then
+      local health = Spring.GetUnitHealth(unitID)
+      if health then
+        Spring.SetUnitHealth(unitID, health*partialReclaim)
```
sprunk, line 81: and since `nil ~= 1` it will enter there and crash on `health * partialReclaim` due to multiplication by nil.

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+      end
+    end
+  end
+  
+  return unitID
+end
+
+-- Works on non zombie units too.
+local function SetZombieSpeedMult(unitID,speedMult)
+	Spring.SetUnitRulesParam(unitID, "zombieSpeedMult", speedMult, LOS_ACCESS)
```
sprunk, line 91: Now that it's an API it needs a bit more care around argument correctness, perhaps add something like
```suggestion
	if type(speedMult) ~= 'number' or speedMult < 0 then
		error("SetZombieSpeedMult: mult must be number >= 0")
	end
	Spring.SetUnitRulesParam(unitID, "zombieSpeedMult", speedMult, LOS_ACCESS)
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
+--------------------------------------------------------------------------------
+--------------------------------------------------------------------------------
+
```
sprunk, line 3: needs a synced code check, same as zombies

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+function gadget:GetInfo()
+	return {
+		name        = "Zombie helper api",
+		desc        = "The place to handle your zombie esque needs!",
+		author      = "Stiofan",
+		date        = "June 2026",
```
sprunk, line 9: largely just shuffling of existing code
```suggestion
		author      = "TomFyuri, Stiofan",
		date        = "Mar 2014",
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+	local buildopts = UnitDefs[unitDefID].buildOptions
+	if (not buildopts) or #buildopts <= 0 then
+		return
+	end
+	local orders = {}
+	for i = 1, math.random(10, 30) do
+		orders[#orders + 1] = {-buildopts[math.random(1, #buildopts)], 0, 0 }
+	end
+	if (#orders > 0) then
+		if not Spring.GetUnitIsDead(unitID) then
```
sprunk, line 108: death checks could be done at the top, before rolling dice 30 times. the orders check is redundant since there's always at least 10 orders.

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+			Spring.GiveOrderArrayToUnitArray({unitID}, orders)
```
sprunk, line 109: the array always has just 1 unit, so could be just `Spring.GiveOrderArrayToUnit`

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+		end
+	end
+end
+
+
+local function disSQ(x1, y1, x2, y2)
+	return (x1 - x2)^2 + (y1 - y2)^2
+end
```
sprunk, line 117: could use `math.diag` for distance

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+  
+  if reclaimPercentHealth then
+    local currentMetal, maxMetal = Spring.GetFeatureResources(featureID)
+    if currentMetal and maxMetal and (maxMetal > 0) then
+      partialReclaim = currentMetal/maxMetal
+    end
+  end
+  
+  Spring.DestroyFeature(featureID)
+  local unitID = Spring.CreateUnit(featureDefName, x, y, z, facing, teamID)
```
sprunk, line 70: these could be swapped, would keep a brief window where they exist together to allow transferring some traits from the feature to the unit in the future, if needed

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,191 @@
...
+
+local mapWidth
+local mapHeight
+
+local REZ_SOUND = "sounds/misc/resurrect.wav"
+
+--unused, may be used depending on how things shake out
+local ZOMBIE_SOUNDS = {
+	"sounds/misc/zombie_1.wav",
+	"sounds/misc/zombie_2.wav",
```
sprunk, line 34: sounds easy to copypaste if/when that happens

sprunk (commented):
I left a bunch of comments, but couldn't resist immediately addressing them all myself. I pushed the resulting commit to https://github.com/ZeroK-RTS/Zero-K/tree/zombie-api-cleaned-up.

Your choice on whether you want to have your share of fun and learning and do them on your own, or if you want me to just merge that cleaned up commit so you can move on to nanoplague etc.

LuaRules/Gadgets/api_zombie_helper.lua
Stiofan-K: so I'll just leave it in for now?

LuaRules/Gadgets/api_zombie_helper.lua
Stiofan-K: Hope I used it correcly, i seems to function anyhow

Stiofan-K:
Thanks for giving me the option to fix it myself :P 
I do understand the desire to just "fix it now!!!"

Depending on how things go I may try myself at the callback function shenanigans later today, but that could also be a different PR if this would be good to go.

LuaRules/Gadgets/api_zombie_helper.lua
sprunk: Nah, remove because it's unused. If it becomes needed it's easy to copy-paste from the old zombie gadget since it's still there

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+	local best_dist
+	local x, y, z = Spring.GetUnitPosition(unitID)
+	local units = Spring.GetUnitsInCylinder(x, z, range)
+	for i = 1, #units do
+		local allyID = units[i]
+		local allyTeam = Spring.GetUnitTeam(allyID)
+		local allyDefID = Spring.GetUnitDefID(allyID)
+		if (allyID ~= unitID) and (allyTeam == GaiaTeamID) and (Spring.Utilities.getMovetype(UnitDefs[allyDefID]) ~= false) then
+			local ox, oy, oz = Spring.GetUnitPosition(allyID)
+			local dist = math.diag(x, z, ox ,oz)
```
sprunk: it accepts distances in each dimension:

```suggestion
			local dist = math.diag(x - ox, z - oz)
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+	end
+	Spring.SetUnitRulesParam(unitID, "zombieSpeedMult", speedMult, LOS_ACCESS)
+	GG.UpdateUnitAttributes(unitID)
+end
+
+
+-- Zombie commands
+
+local function RandomFactoryOrders(unitID, unitDefID) -- give factory something to do
+	if not Spring.GetUnitIsDead(unitID) then
```
sprunk: Rather than
```lua
if condition then
  -- the entirety of the function
end
```
prefer
```lua
if not condition then
  return
end
-- the rest of the function
```

So far the function is small so it doesn't matter that much, but the principle avoids it turning into huge pyramids like this

Image

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+		local buildopts = UnitDefs[unitDefID].buildOptions
+		if (not buildopts) or #buildopts <= 0 then
+			return
+		end
+		local orders = {}
+		for i = 1, math.random(10, 30) do
+			orders[#orders + 1] = {-buildopts[math.random(1, #buildopts)], 0, 0 }
+			Spring.GiveOrderArrayToUnit(unitID, orders)
+		end
```
sprunk: GiveOrder needs to be outside the loop that builds the array, as-is it would give an increasing series of commands like `{A}, {A, B}, {A, B, C}, ...`
```suggestion
		for i = 1, math.random(10, 30) do
			orders[#orders + 1] = {-buildopts[math.random(1, #buildopts)], 0, 0 }
		end
		Spring.GiveOrderArrayToUnit(unitID, orders)
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+  
+	mapWidth = Game.mapSizeX
+	mapHeight = Game.mapSizeZ
+
+  GG.Zombies = {
+    TurnFeatureIntoUnit     = TurnFeatureIntoUnit,
+    SetZombieSpeedMult      = SetZombieSpeedMult,
+    SetZombieBehavior       = GiveZombiesRandomOrders,
+    GetFeatureResurrectData = GetFeatureResurrectData
+  }
```
sprunk: whitespace: indentation should be tabs
```suggestion
	GG.Zombies = {
		TurnFeatureIntoUnit     = TurnFeatureIntoUnit,
		SetZombieSpeedMult      = SetZombieSpeedMult,
		SetZombieBehavior       = GiveZombiesRandomOrders,
		GetFeatureResurrectData = GetFeatureResurrectData
	}
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+		ry = Spring.GetGroundHeight(rx,rz)
+		if IsTargetReallyReachable(unitID, rx, ry, rz, x, y, z) then
+			orders[#orders+1] = {CMD.FIGHT, {rx, ry, rz}, CMD.OPT_SHIFT}
+		end
+	end
+	if (#orders > 0) then
+		if not Spring.GetUnitIsDead(unitID) then
+			Spring.GiveOrderArrayToUnitArray({unitID},orders)
+		end
+	end
```
sprunk: * IsDead is redundant, it's already checked in the first line of the function.
* the `# > 0` check also doesn't do anything, since in the rare case where all orders happen to get dropped the 0-length order array is still fine.
* the unit array always contains just 1 unit, so could use the non-array version.

```suggestion
	Spring.GiveOrderArrayToUnit(unitID, orders)
```

LuaRules/Gadgets/api_zombie_helper.lua
```diff
@@ -0,0 +1,194 @@
...
+	local x,y,z = Spring.GetUnitPosition(unitID)
+	if (near_ally) and  math.random(0, 5) < 4 then -- 60% chance to guard nearest ally
+		orders[#orders + 1] = {CMD.GUARD, {near_ally}, 0}
+	end
+	for i = 1, math.random(10, 30) do
+		rx = math.random(0, mapWidth)
+		rz = math.random(0, mapHeight)
+		ry = Spring.GetGroundHeight(rx,rz)
```
sprunk: These aren't used anywhere else so it's better to make them locals here and delete line 149:

```suggestion
		local rx = math.random(0, mapWidth)
		local rz = math.random(0, mapHeight)
		local ry = Spring.GetGroundHeight(rx,rz)
```

This is because a more limited var scope is good (if somebody uses it beyond the intended scope it's likely a mistake) and for Lua specifically it's also minimally faster.

sprunk (commented):
Looks decent, more comments

LuaRules/Gadgets/api_zombie_helper.lua
Stiofan-K: mh I hadnt seen any documentation on math.diag, but then also forgot to check for other examples in code >.>

Stiofan-K:
More comments addressed! :>

sprunk:
Thanks!

Issue #5772, opened 2026-07-01
turretaaflak becomes inoperable when gun is in allied hitbox


SteelBlueZK:
<img width="944" height="744" alt="Image" src="https://github.com/user-attachments/assets/7f20698e-b43f-4fb9-9dc1-bc60c35b0111" />

Similar to https://github.com/ZeroK-RTS/Zero-K/issues/2386
Cannon weapons are more complicated than just the aimpoint. Similarly to a big bertha sticking it's nose in the ground/in an allied unit. I'll submit something later.

When the muzzle is inside allied colvolume, it never aims cause it never has clear shot.

GoogleFrog:
The fix probably involves creating some fake higher set of queried points that are used when the turret is stowed.

PR #5785, opened 2026-07-09
Thresher unsticker

SteelBlueZK:
featuring 6 second unstick

https://github.com/ZeroK-RTS/Zero-K/issues/5772

### DIFF
--- a/scripts/turretaaflak.lua
+++ b/scripts/turretaaflak.lua
@@ -8,6 +8,7 @@ local floatbase = piece 'floatbase'
 local trueaim = piece 'trueaim'
 
 local gun_to_use = 1
+local targetlock = false
 
 local SIG_AIM = 1
 
@@ -29,6 +30,7 @@ end
 local function RestoreAfterDelay()
 	SetSignalMask(SIG_AIM)
 	Sleep(6000)
+	targetlock = false
 	Turn (a1, x_axis, 0, math.rad(20))
 	Turn (a2, x_axis, 0, math.rad(60))
 	Turn (a3, x_axis, 0, math.rad(50))
@@ -68,6 +70,7 @@ function script.AimWeapon(num, heading, pitch)
 
 	Signal (SIG_AIM)
 	SetSignalMask (SIG_AIM)
+	targetlock = true
 
 	while disarmed do
 		Sleep(34)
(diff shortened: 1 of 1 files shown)


sprunk:
* isn't the actual problem that the turret just doesn't have room to maneuver? Clipping is also a visual problem. Perhaps flak should just become 4x4, I don't think anything hinges on it being 3x3 and it's just the usual reasons of being unwieldy and possibly awkward with how thin it is?
 * what if the first AimWeapon call doesn't put the barrels away from clipping (e.g. maybe the flak is on a hill and a plane arrives exactly from the front)? Won't it still fail to shoot for 6s?
 * I am unsure but I feel like the query point aim check shouldn't prevent AimWeapon from being called and rotating the weapon. What is AimFrom even for then? Maybe this is a subtle engine regression?

SteelBlueZK:
-point 1, room to maneuver

If we're going to increase the footprint, we need to go to 5x5 so that the closest possible box of built threshers don't stick turrets in the next colvol. I hadn't considered touching the footprint. 5x5 might be too large, zk doesn't have a lot of large footrprints and often puts it on the player to block/not block base path finding and 5x5 is a lot of space. I don't think the engine supports pathfinding through the construction grid (natively) the way supcom does, but I feel thresher would appreciate it here.

-point 2, 6 seconds

I just put the toggle on the same delay as restoreafterdelay's anim delay. I figure getting stuck is infrequent, occlusion is more common. Moving the queryweapon point so jaringly(subframe) can cause a frienldyfire problem, but it only does 2 dps to ground units in the worst case scenario, not a big issue.

-point 3, engine

I think there's a problem with the logic of the engine on at least cannon weapons, but also there's little in the way of constraints. Mods are allowed to use large models, freely animate any model on any callin, so these somewhat frequent "Snail Eyes" sort of guns like thresher, bulkhead just run the risk of putting the end of their gun someplace cramped that don't shine. I actually don't know what all the pieces are for, AimFrom by convention is the turret base, QueryWeapon is the muzzle.

sprunk:
> I don't think the engine supports pathfinding through the construction grid (natively) the way supcom does, but I feel thresher would appreciate it here.

Pathing-only footprint exists, it's `u` in the yardmap.

GoogleFrog:
I don't want to touch the footprint. Thresher looks like it should be able to be placed on a relatively small spire, which puts it at 3x3.

PR #5788, opened 2026-07-12
Singularity Generator rattles when hit

SteelBlueZK:
Imagine a baby playing with the most powerful physics.

The orb in the middle of singu will now respond to enemy attacks, bouncing away from the impact and springing back towards the center. When it gets pushed close enough to the arms, the ball shrinks to avoid egregious model clipping. The y-axis jiggle is randomly generated, because the engine does not pass the vertical component of the impulse vector. As the Singu scales from 100% to 20% HP, the spring dampening, or resistance to permanently jiggling, decreases, so a near-death Singu has a much more noticeable jiggle for longer.

I've tested this with about 12% of the unit roster, and the animation seems sane and fun about 99% of the time.

Disclaimer: PD boilerplate code written by chatgpt, but fine tuning and other formula are all me.

### DIFF
--- a/scripts/energysingu.lua
+++ b/scripts/energysingu.lua
@@ -18,10 +18,8 @@ local spGetUnitHealth = Spring.GetUnitHealth
 local sin = math.sin
 local max = math.max
 local min = math.min
-local abs = math.abs
-local rand = math.random
 
-local SPRING = 0.18    -- Pull toward center
+local CENTER_FORCE = 0.18    -- Pull toward center
 local DAMPING_BASE = 0.4    -- Removes oscillation
 local DAMPING_MIN = 0.03
 local RATTLE_THRESHOLD = 26  -- elmo range threshold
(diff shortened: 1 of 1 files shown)


scripts/energysingu.lua
```diff
@@ -13,18 +13,78 @@ local arm3 = piece "arm3"
...
 
+local spSetUnitRulesParam = Spring.SetUnitRulesParam
+local spGetUnitHealth = Spring.GetUnitHealth
+local sin = math.sin
+local max = math.max
+local min = math.min
+local abs = math.abs
+local rand = math.random
+
+local SPRING = 0.18    -- Pull toward center
```
sprunk: this will bite us in the ass during some mass string replace action in 15 years

SteelBlueZK: I can see this happening with SPRING

Issue #5587, opened 2025-09-29
Odin shows its hugeass technical range


sprunk:
With "show selected unit range", bomberassault shows a third circle with huge radius which confuses people.

<img width="1013" height="686" alt="Image" src="https://github.com/user-attachments/assets/c73c9026-ec2f-4379-8d86-7146765b3aad" />

Its weapon seems to have two ranges defined, one "true" and one technical:
https://github.com/ZeroK-RTS/Zero-K/blob/caaa315f2e880906ecb39a72523eaf23dcf6f1a8/units/bomberassault.lua#L126
https://github.com/ZeroK-RTS/Zero-K/blob/caaa315f2e880906ecb39a72523eaf23dcf6f1a8/units/bomberassault.lua#L155

Figure out if the technical range can be reduced to match "true", otherwise make the ranges widget not render non-"true" ranges.

GoogleFrog:
Range has to be huge to stop the engine killing the noExplode projectile too soon (as distance timeout is the only way noExplodes can be removed). The widget should draw truerange.

SteelBlueZK:
Could the range be extended on the frame the projectile is launched like Gauss hax?
https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/turretgauss.lua#L162
https://github.com/ZeroK-RTS/Zero-K/blob/master/scripts/hoverraid.lua#L91

Gauss and Dagger both have GG.Attributes.SetRangeUpdater(unitID, RangeUpdate) and a local RangeUpdate function defined. Detriment no longer has guass range hax fix, but that's a separate issue.

PR #5794, opened 2026-07-20
Hide odin technical range in LUS

SteelBlueZK:
fix for https://github.com/ZeroK-RTS/Zero-K/issues/5587


scripts/bomberassault.lua
```diff
@@ -129,10 +142,23 @@ function script.FireWeapon(num)
 end
 
 function script.BlockShot(num, targetID)
```
SteelBlueZK, line 144: Check for style in this function

sprunk:
Why not make a customparam that adjusts any projectile range?

SteelBlueZK:
9001 is a meme number, previously it was 3000. I don't think the number really matters, just that it is larger than any potential travel distance. I did manage to make an unrealistically high mountain and drop a bomb in an unreasonable pit and watch it expire before hitting it's target, so I just bumped the number. It could be a custom parameter, but I don't think it's any useful information.

sprunk:
The advantage of customparam is that a modder would be able to put thermite on some other bomber. Even if the numerical value is arbitrary in this case, accepting a number is useful for decoupling targeting range from expiration range to deliberately introduce overshoot for some "regular" weapons (at least I remember @GoogleFrog made a mantis ticket about 10 years ago with the request).

sprunk:
With customparam, true range should be a gadget that handles units generally, rather than in any specific unit's LUS. The fake range probably stays in LUS because I assume it's some bomber-specific nonsense. LUS FireWeapon and EndBurst are available to gadgets as ScriptFireWeapon and ScriptEndBurst.

So something like:
```lua
local rangeDefs = {} -- [unitDefID][weaponNum] = {trueRange, nominalWeapondefRange}
for unitDefID, unitDef in pairs(UnitDefs) do
  for weaponNum = 1, #unitDef.weapons do
    local weaponDef = WeaponDefs[unitDef.weapons[i].weaponDef]
    local trueRange = weaponDef.customParams.truerange
    if trueRange then
      rangeDefs[unitDefID] = rangeDefs[unitDefID] or {}
      rangeDefs[unitDefID][weaponNum] = {truerange, weaponDef.range}
    end
  end
end

function gadget:ScriptFireWeapon(unitID, unitDefID, weaponNum)
  local rangeDef = rangeDefs[unitDefID]
  if not rangeDef then
    return
  end

  local weaponRanges = rangeDef[weaponNum]
  if not weaponRanges then
    return
  end

  Spring.SetUnitWeaponState(unitID, weaponNum, "range", weaponRanges[1])
end

function gadget:ScriptEndBurst(unitID, unitDefID, weaponNum)
  local rangeDef = rangeDefs[unitDefID]
  if not rangeDef then
    return
  end

  local weaponRanges = rangeDef[weaponNum]
  if not weaponRanges then
    return
  end

  Spring.SetUnitWeaponState(unitID, weaponNum, "range", weaponRanges[2])
end
```

GoogleFrog:
I don't like how the PR constantly modifies game mechanical values purely to draw a circle at a different range. Modifying ranges always comes with some jank (even if you do it via attributes, which this does not do) and I don't think the circle is worth it.

SteelBlueZK:
If changing weapon range is undesireable, would it be better to just patch the widget that draws weapon ranges to ignore the odin fake range?

GoogleFrog:
Oh I thought this PR was about the engine range drawer. It being a widget makes this easy. https://github.com/ZeroK-RTS/Zero-K/commit/0f6058b16a056b0d91aaea22f5ad73a6e466b858

PR #5797, opened 2026-07-22
Assorted patches

SteelBlueZK:
LuaUI/Widgets/gui_reclaiminfo.lua
- Initilized 2 variables instead of 1 for clarity

effects/roach.lua
- Table defined 2 properties "groundflash", got rid of the redundant one. The redundant one was getting overwritten by the latter one.

lups/lups.lua
- global scope variable pulled back to local scope

LuaUI/Widgets/unit_smart_nanos.lua
- uhhh.... Errr... My linter informed me there was a reference to a global variable, which was probably meant to reference a local variable defined below it, but then I noticed many other things that are messy with this file.... This isn't even the default caretaker controller, and maybe it should be deleted. The logic is actually all over the place, Histidine added it to the repo "so someone else can figure it out" and noone seems to have done so.

### DIFF
--- a/LuaUI/Widgets/gui_reclaiminfo.lua
+++ b/LuaUI/Widgets/gui_reclaiminfo.lua
@@ -36,7 +36,7 @@ local rangeend = {}  --counting radius end point
 local b1was = false  -- cursor was outside the map?
 local vsx, vsy = widgetHandler:GetViewSizes()
 local form = 12 --text format depends on screen size
-local xstart,ystart = 0
+local xstart,ystart = 0, 0
 local cmd,xend,yend,x,y,b1,b2
 local inMinimap = false --mouse cursor in minimap
 
--- a/effects/roach.lua
+++ b/effects/roach.lua
@@ -40,16 +40,6 @@ return {
     },
   },
   blastwing = {
-    groundflash = {
-      flashalpha         = 1,
-      flashsize          = 108,
-      ttl                = 75,
-      color = {
-        [1]  = 0.7,
-        [2]  = 0.3,
-        [3]  = 0.1,
-      },
-    },
     redploom = {
       air                = true,
       class              = [[CExpGenSpawner]],
--- a/lups/lups.lua
+++ b/lups/lups.lua
@@ -921,7 +921,7 @@ local function GameFrame(_,n)
 	CleanInvalidUnitFX()
 
 	--// update FXs
-	framesToUpdate = thisGameFrame - lastGameFrame
+	local framesToUpdate = thisGameFrame - lastGameFrame
 	for _,partFx in pairs(particles) do
 		if (n>=partFx.dieGameFrame) then
 			--// lifetime ended


LuaUI/Widgets/unit_smart_nanos.lua
```diff
@@ -26,23 +26,24 @@ end
...
+local GetUnitCommandCount     = Spring.GetUnitCommandCount
+local GetFeatureDefID         = Spring.GetFeatureDefID
+local GetFeatureResources     = Spring.GetFeatureResources
+local AreTeamsAllied          = Spring.AreTeamsAllied
+local GetFeaturePosition      = Spring.GetFeaturePosition
+local GetGameSeconds          = Spring.GetGameSeconds
+local GetSelectedUnits        = Spring.GetSelectedUnits
+local GetUnitTeam             = Spring.GetUnitTeam
+local GetTeamResources        = Spring.GetTeamResources
+local spGetFeaturesInCylinder = Spring.GetFeaturesInCylinder
```
sprunk: The project-wide convention is `spFoo = Spring.Foo`, but the existing lines are just `Foo = Spring.Foo` so now there's inconsistent style. If we're linting the file perhaps bring the others up to the `spFoo` standard.

LuaUI/Widgets/unit_smart_nanos.lua
```diff
@@ -358,16 +348,13 @@ function widget:Update(deltaTime)
 					local ordered = false
 					
 					local nearUnits = GetUnitsInCylinder(unitDefs.posX,unitDefs.posZ,unitDefs.buildDistance)
-					local nearFeatures = Spring.GetFeaturesInRectangle(
-						unitDefs.posX - (unitDefs.buildDistance+75), unitDefs.posZ - (unitDefs.buildDistance+75),
-						unitDefs.posX + (unitDefs.buildDistance+75), unitDefs.posZ + (unitDefs.buildDistance+75)
-					)
+					local nearFeatures = spGetFeaturesInCylinder(unitDefs.posX, unitDefs.posZ, unitDefs.buildDistance)
```
sprunk: What happened to the +75?

sprunk (commented):
Changes in the 3 smaller files look good, some nits around smart nanos though.

LuaUI/Widgets/unit_smart_nanos.lua
SteelBlueZK: Whoops I didn't see the parenthesis right, I thought this was just an offset problem and removed it. I guess I did change the logic.

I looked over the rest of the code, and this distance is later clamped to buildDistance+featureRadius depending on the features returned by the engine. I'm not sure if nanos can even operate if the target center is outside the buildRange like that.

SteelBlueZK:
What do you say I remove changes to unit_smart_nanos and just push the 3 tiny changes?

my linter pointed to [this line](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaUI/Widgets/unit_smart_nanos.lua#L370) and said "global variable" and indeed "bestUnit" is global in this scope. It gets defined in local scopes down the function in 2 separate if statements.

sprunk:
If you'd like to do more with smart nanos then a standalone PR for the 3 tiny changes to get them out of the way sounds ok.

But could also remove the `sp` from `spGetFeaturesInCylinder` to be consistent with what exists (rather than plow through the entire file to add `sp` to everything else) and I think that would also be immediately mergeable.

Issue #5801, opened 2026-07-25
Mod caused nil error in zk utility function


SteelBlueZK:
`[t=00:07:46.006454][f=0030033] Error: [LuaRules::RunCallInTraceback] error=2 (LUA_ERRRUN) callin=GameFrame trace=[Internal Lua error: Call failure] [string "LuaRules/Utilities/unittypechecker.lua"]:63: attempt to index local 'ud' (a nil value)`
This error seems to happen every frame once it starts
[source link](https://github.com/ZeroK-RTS/Zero-K/blob/master/LuaRules/Utilities/unitTypeChecker.lua#L63)

This error comes out of this replay:
https://zero-k.info/Battles/Detail/2472912

running this mod: https://zero-k.info/Mods/Detail/31

The mod changes stuff about commanders, and is most likely an interaction with code from the commander egg update.

Probably just needs a change to return false when it's given an improper unit def ID and let errors scream elsewhere.

sprunk:
34e42f82b871ebd549c9bbe8941b126a641533a5

PR #5802, opened 2026-07-25
Patch 03

SteelBlueZK:
Potential fix for https://github.com/ZeroK-RTS/Zero-K/issues/5801

This essentially silent fails, "If this Unitdef is not a unitdef, then it is not a commander"


sprunk (approved):
Personally I think the error is good (since the mod is passing an incorrect value and should just stop doing that) so perhaps the ideal improvement would have been to make it more helpful:
```lua
if not ud then
  error("Spring.Utilities.isComm() called with invalid arg: " .. tostring(ud))
end
```

But as-is the PR is consistent with how everything is polluted with a billion nil checks.

GoogleFrog:
I suppose it is the case that a unitDefID that fails to correspond to a unit is also not a unitDefID that corresponds to a commander.

PR #5818, opened 2026-08-24
Fixes #5811: Unit Auto Reclaim/Heal/Assist no longer leaves ghost idle cons

### LINKED ISSUE #5811: Auto Reclaim/Heal/Assist widget causes idle con count to not decrease if con dies
This concerns widgets `gui_chili_core_selector.lua` and `unit_auto_reclaim_heal_assist.lua`.

### Steps to reproduce:

0. Launch a fresh ZK install / have default settings
1. Start skirmish, **enable Auto Reclaim/Heal/Assist widget**
2. place commander, plop factory
3. Ctrl-K commander

### Expected result:
number of idle cons is 0

### Actual Result:
number of idle cons ticks to 0 for a moment and immediately goes back to 1

## Likely cause:
The auto reclaim widget has a `ConController` object which keeps track of a given con, and deletes itself when the con dies. The deletion function is `ConController.unset`, triggered, among other things, by `widget:UnitDestroyed`:
```lua
	unset = function(self)
		-- Echo("IdleConAssist removed: " .. self.unitID)
		GiveOrderToUnit(self.unitID,CMD.STOP, {}, {""},1)
		return nil
	end,
```
Following liberal use of Spring.Echo, I found that Spring seemingly lets the widget issue an order even to a dead unit, which then triggers the UnitIdle callin (after the order "completes") in the core selector widget, after UnitDestroyed already fired. This creates a ghost idle constructor that will never go away.

## Proposed solutions:
a) Make core selector somehow check that the newly idle unit is still alive. (possibly by `Spring.GetUnitIsDead`?)
b) Stop auto reclaim from issuing orders if unset was called after the unit was destroyed.
c) Engine change to completely block commands to dead units (granted, I'm guessing with this, no idea what really goes on in the engine)

Of these, c) might be cleanest, but i have no idea how difficult that would be, a) is probably more future-proof, and b) seems simplest.

As a) or b) sound like they'd be very simple changes of a few lines, I could do it myself and make a PR, depending on what the preferred solution is.

Petronous:
~~Changed the widget to not issue a stop command if the unit just died (proposed solution b in my issue).~~

~~Frankly I am not sure if we ever want it to issue a stop command.~~

The core selector widget now checks whether the unit in UnitIdle is really alive before setting its IdleCons entry to true.

I tested the change and it actually does fix the issue, idle con count stays correct. I have not done exhaustive testing with other widgets, but with such a small change it should hopefully not be necessary.

### DIFF
--- a/LuaUI/Widgets/gui_chili_core_selector.lua
+++ b/LuaUI/Widgets/gui_chili_core_selector.lua
@@ -1922,7 +1922,7 @@ function widget:UnitIdle(unitID, unitDefID, unitTeam)
 		wantUpdateCons = true
 	end
 	local ud = UnitDefs[unitDefID]
-	if CanBeAnIdleCons(ud) and IsConNotCarriedByEnemyTransport(unitID) then
+	if CanBeAnIdleCons(ud) and IsConNotCarriedByEnemyTransport(unitID) and Spring.GetUnitIsDead(unitID) == false then
 		idleCons[unitID] = true
 		wantUpdateCons = true
 	end
--- a/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
+++ b/LuaUI/Widgets/unit_auto_reclaim_heal_assist.lua
@@ -33,9 +33,9 @@ local ConController = {
 		return self
 	end,
 
-	unset = function(self)
+	unset = function(self, isDying)
 		-- Echo("IdleConAssist removed: " .. self.unitID)
-		GiveOrderToUnit(self.unitID,CMD.STOP, {}, {""},1)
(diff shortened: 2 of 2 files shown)


sprunk:
What if another widget gives a stop command, or even a player for a con with longer death animation? Core selector should probably be the one to notice that it's a dead unit receiving an order and not count it

Petronous:
Yeah, it's probably the better option. I suppose I was reluctant to jump straight into editing a core widget.
 
Is checking `Spring.GetUnitIsDead(...)` enough?

sprunk:
Sounds like it should be.

Petronous:
I could also put the same check into `UnitCreated` and `UnitLoaded`, which also set `idleCons[...] = true`, but I don't know if there is even a possibility that those could fire for a dead unit.

PR #5824, opened 2026-09-01
GUI for retreat order, havens player colored

SteelBlueZK:
inspired greatly by Helwor's Hel-K widget suite
https://github.com/Helwor/New-Hel-K

--------------------

Summary: Changed the way Retreat Zones (Havens) are drawn to be team-colored, left default color and behavior as is in stable right now, but will now show all player retreat zones when spectating by default to increase discoverability/awareness of the feature.

Overview:
options revamp
- bool "Only Show My Zones" has been replaced with "Display Zone options" a 5 option radio
- added option to draw team color playernames over all retreat zones
- added options to change the default red color of Havens

Whitespace changes

Spring functions localized

Various variable cleanup (reduced to lower scopes, moved around some logic)

### DIFF
--- a/LuaUI/Widgets/gui_havens.lua
+++ b/LuaUI/Widgets/gui_havens.lua
@@ -98,13 +98,13 @@ local RADIUS = 0
 
 ----------------------------------------------------------------------------------------
 -- functions
-function GetTeamName(teamID)
+local function GetTeamName(teamID)
 	local _, leaderID = spGetTeamInfo(teamID, false)
 	local playerName = spGetPlayerInfo(leaderID, true)
 	return playerName
 end
 
-function GetTeamHavens(teamID)
+local function GetTeamHavens(teamID)
 	local teamHavenCount = spGetTeamRulesParam(teamID, "haven_count")
 	if teamHavenCount then
 		local teamLeaderName = GetTeamName(teamID) or "???"
@@ -124,7 +124,7 @@ function GetTeamHavens(teamID)
 	end
 end
 
-function HavenUpdate()
+local function HavenUpdate()
 	havens = {}
 	local teams = spGetTeamList()
 	for i = 0, #teams-1 do
--- a/LuaUI/Widgets/gui_havens.lua
+++ b/LuaUI/Widgets/gui_havens.lua
@@ -2,16 +2,16 @@
 -------------------------------------------------------------------------------------
 
 function widget:GetInfo()
-  return {
-    name      = "Haven Handler",
-    desc      = "Haven Handler for Retreat Gadget",
-    author    = "CarRepairer",
-    date      = "2014-04-10",
-    license   = "GNU GPL, v2 or later",
-	handler   = true,
-    layer     = 0,
-    enabled   = true,
-  }
+	return {
+		name      = "Haven Handler",
+		desc      = "Haven Handler for Retreat Gadget",
+		author    = "CarRepairer",
+		date      = "2014-04-10",
+		license   = "GNU GPL, v2 or later",
+		handler   = true,
+		layer     = 0,
+		enabled   = true,
+	}
 end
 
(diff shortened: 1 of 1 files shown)


LuaUI/Widgets/gui_havens.lua
```diff
@@ -61,54 +63,52 @@ local glBillboard      = gl.Billboard
...
 local abs 	= math.abs
 
 local havens = {}
-local havenCount = 0
 local RADIUS = 0
+local DEFAULT_HAVEN_COLOR = {1, 0.1, 0.1, 0.8}
 
 ----------------------------------------------------------------------------------------
 -- functions
+function GetTeamName(teamID)
```
SteelBlueZK: Is this function achieved anywhere else in the code? If I can just call it instead of copying it that would be good.

LuaUI/Widgets/gui_havens.lua
```diff
@@ -20,19 +20,22 @@ VFS.Include("LuaRules/Configs/customcmds.h.lua")
...
 
 options = {
-	onlyShowMyZones = {
-		name = 'Only Show My Zones',
-		desc = 'With this enabled you will only see your retreat zones.',
+	drawAllyZones = {
+		name = 'Show All Player Retreat Zones',
+		desc = 'Displays retreat zones of other players',
 		type = 'bool',
 		value = true,
```
SteelBlueZK: review re-wording on old option, and setting "Label Player Names". What should default be?

LuaUI/Widgets/gui_havens.lua
```diff
@@ -159,37 +159,55 @@ local function DrawWorldFunc()
...
-		gl.LineWidth(4)
-		glColor(1, 1, 1, 0.5)
-		gl.DrawGroundCircle(x, y, z, RADIUS, 32)
-
-		gl.LineWidth(2)
-		glColor(1, 0.1, 0.1, 0.8)
-		gl.DrawGroundCircle(x, y, z, RADIUS, 32)
-
+		-- TODO spGetLocalPLayerID interacts poorly when local player is cheating between teams and spectator.
+		-- Change comparison logic to work better in this edge case.
```
SteelBlueZK: In regards to "TODO" comment:

The line below and some other places have "havenPosition.team == spGetLocalPlayerID()". I only tested this locally, I don't know how this interacts in multiplayer. The intention of this comparison is if we are drawing all retreat zones or not, options.drawAllyZones. If we're not drawing ally zones, we only want to draw the zones of the local player. Is this the best comparison to achieve the goal? Is there a different spring function that does better?

LuaUI/Widgets/gui_havens.lua
SteelBlueZK: also this needs to be made local

SteelBlueZK:
image

LuaUI/Widgets/gui_havens.lua
```diff
@@ -159,37 +159,55 @@ local function DrawWorldFunc()
...
+		if options.drawAllyZones.value or havenPosition.team == spGetLocalPlayerID() then
```
sprunk: should probably be GetLocalTeamID rather than GetLocalPlayerID

LuaUI/Widgets/gui_havens.lua
SteelBlueZK: It's local now

LuaUI/Widgets/gui_havens.lua
SteelBlueZK: "Resolved" thanks sprunk

SteelBlueZK:
Default settings:
image

SteelBlueZK:
I have 0 idea if this updates properly in multiplayer. It would probably be good to test that, if I'm a player and a different player updates their Havens, do I reflect that? I've been cheating between teams so local team ID is always the player being updated.

PR #5826, opened 2026-09-04
Magpie bone

SteelBlueZK:
If it already doesn't collide with allies, might as well give it targeting when all the magpies are colliding with each other in one easy to burst ball. And let it forcefire at wrecks.


sprunk:
`collide=false, avoid=true` is often just protection against grazing against ally hitboxes and having expensive, impactful shots cause teamkill instead (which feels terrible) while still not causing suspension of disbelief by outright shooting through the middle of the ally. Although for air units it's not necessarily noticeable even then.

GoogleFrog:
> collide=false, avoid=true is often just protection against grazing against ally hitboxes and having expensive, impactful shots cause teamkill instead (which feels terrible) while still not causing suspension of disbelief by outright shooting through the middle of the ally.

Yes

sprunk:
For shooting against wrecks there could be `force_ignore_ground = [[1]]`, if GF doesn't mind. That way explicit attack orders can shoot wrecks but the unit will still avoid wasting regular shots on wrecks unlike plain `avoidFeature=true`.

GoogleFrog:
> For shooting against wrecks there could be `force_ignore_ground = [[1]]`, if GF doesn't mind. That way explicit attack orders can shoot wrecks but the unit will still avoid wasting regular shots on wrecks unlike plain `avoidFeature=true`.

I am confused by this comment as I added this behaviour ages ago.

sprunk:
Yes, but not specifically to Magpie. This PR contained an `avoidFeature = false` change which earlier comments didn't really address. That specific change is probably also bad (in that it just leads to wasting shots on wrecks), but since the goal was force-firing at wrecks I brought up the `force_ignore_ground` customparam which sounds ok to apply.

PR #5831, opened 2026-09-07
icon_generator Config update.

Rallehop:
This version of the config generates the appropriate icon for Phantom, Fencer, Support Commander and Recon Commander.

Does *not* have fixes for Emissary: I used a scuffed manual pose method by setting the wait value to a high number and timing it. I cannot get it to pose, no matter what values i give.
Does *not* contain anything about Dominatrix, because that seems to crash the game.

Lines 303 to 310 might be without purpose. What actually fixed the generated icons was adding `unfold = true` to `dynrecon1` and `dynsupport1`. If you manually input `commrecon1` or `commsupport1` it will generate the correct icon, and save them without the 1. I don't know if this is desirable or not, but they are not inBatch so they won't generate unless you manually request them.

If someone wants to look this over before merging that'd be cool and at the very least i seem unable to fix Emissary's icon outside of a scuffed method.


sprunk:
Looks ok but it would be nice if it wasn't 27 commits.

